PR #68, opened 2014-04-22
Use an environment variable to set the location of the configuration directory

number33:
Updated code in _include.php to optionally set an alternate configuration directory locaion.  Updated the Memcache Sync, Log clearner and analyzer to accept two optional command line arguments for setting an alternate configuration direcotry location.  The code now will look for the environment variable SIMPLESAMLPHP_CONFIG_DIR for an alternate configuration location.

### DIFF
--- a/www/_include.php
+++ b/www/_include.php
@@ -96,13 +96,23 @@ class SimpleSAML_IncPrefixWarn {
 /* Set the $SIMPLESAML_INCPREFIX to a reference to the class. */
 $SIMPLESAML_INCPREFIX = new SimpleSAML_IncPrefixWarn();
 
+/* Get environment variable that specifies an alernate location for the configuration files. */
+$configdir = getenv("SIMPLESAMLPHP_CONFIG_DIR");
 
-$configdir = dirname(dirname(__FILE__)) . '/config';
-if (!file_exists($configdir . '/config.php')) {
-	header('Content-Type: text/plain');
-	echo("You have not yet created the simpleSAMLphp configuration files.\n");
-	echo("See: http://rnd.feide.no/content/installing-simplesamlphp#id434777\n");
-	exit(1);
+/* Check if environment variable is set and if the directory exists.
+ * If neither condition is met use the default configuration directory.
+ */
+if (empty($configdir) || !file_exists($configdir)) {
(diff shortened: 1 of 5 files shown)


www/_include.php
```diff
@@ -96,13 +96,23 @@ function __toString() {
...
+ */
+if (empty($configdir) || !file_exists($configdir)) {
+	$configdir = dirname(dirname(__FILE__)) . '/config';
+	if (!file_exists($configdir . '/config.php')) {
+		header('Content-Type: text/plain');
+		echo("You have not yet created the simpleSAMLphp configuration files.\n");
+		echo("See: http://rnd.feide.no/content/installing-simplesamlphp#id434777\n");
+		exit(1);
+	}
+} else {
```
jaimeperez: Since you used `file_exists()`, here you are basically having the following condition:

```php
if (!empty($configdir) && file_exists($configdir)) {
```

Note that in that case, if _SIMPLESAMLPHP_CONFIG_DIR_ is set, for instance, to _/etc/passwd_, you would end up here, causing an error since that's not a directory. `is_dir()` should be used to make sure the path specified is a directory.

www/_include.php
```diff
@@ -96,13 +96,23 @@ function __toString() {
...
+	/* Set the configuration directory to the new location. */
+	SimpleSAML_Configuration::setConfigDir($configdir);
```
jaimeperez: And here you are assuming that if someone specified a directory (well, or a file actually), it will be a valid configuration directory containing the _config.php_ file. We should not assume that and explicitly check it in the same way we do when we use a pre-fixed location.

jaimeperez:
@tombooj I agree with you in that this is an interesting feature. However, there's a few things wrong with it that should be fixed before merging.

Regarding the _real feedback_, I don't know what you mean.

cooperaj:
I'd like to see this implemented as well. It's not possible to use it as a vendor library if you then have to change it to configure it. We've had to implement a halfway house solution where we commit our changes to just this project within vendor/ and then have removed it from our projects composer.json.

jaimeperez:
Hi,

I'm closing this PR due to lack of response, and favoring #221 instead.

PR #75, opened 2014-05-02
Change metadata-converter output to textarea

mkrogh:
I found it much more pleasing to work with the metadata-converter when the output is put into a textarea instead of a pre. 

It might not be a tool you use every day (especially not if you have setup the metarefresh module), but it really helps newcomers such as me.

### DIFF
--- a/templates/metadata-converter.php
+++ b/templates/metadata-converter.php
@@ -32,7 +32,7 @@ if($output !== NULL) {
 		}
 
 		echo('<h3>' . htmlspecialchars($type) . '</h3>' . "\n");
-		echo('<pre class="metadatabox">' . htmlspecialchars($text) . '</pre>' . "\n");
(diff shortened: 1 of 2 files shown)


tdiscuit:
An alternative to this might be to keep it as a pre, but use javascript to select all the text in the pre when you click on the pre or even a button. That would give you the advantages of going to a textarea while not making it editable client-side. Users would still have to manually copy, but at least you wouldn't have to select all the text by dragging your mouse around.

mkrogh:
Another way these days is to use [clipboard.js](https://zenorocha.github.io/clipboard.js/).

tdiscuit:
I just added it to a test instance of SimpleSAMLphp I have and clipbard.js works quite well. Doesn't support Safari 100%, but it at least selects the text.

Didn't take very long to get working either which was nice and works with the existing pre.

jaimeperez:
I've added support for clipboard.js in all the metadata related pages, hope that helps! :smiley:

PR #81, opened 2014-06-25
Treat objectguid as a binary LDAP attribute

harrisony:
Hi!

When integrating Office365, their sync tools use the LDAP objectGUID attribute (a binary attribute) as the SAML NameID by default.

This pull request includes a commit to base64 the attribute so that the NameID is returned correctly in the SAML response.

### DIFF
--- a/lib/SimpleSAML/Auth/LDAP.php
+++ b/lib/SimpleSAML/Auth/LDAP.php
@@ -541,7 +541,7 @@ class SimpleSAML_Auth_LDAP {
 					continue;
 				}
 
-				// Base64 encode jpegPhoto.
+				// Base64 encode binary attributes.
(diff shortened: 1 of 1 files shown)


PatelParas:
might be incase of encoded uid as guid we need following conversion for decoding :
```php
public static function base64ToobjectGUID($guid_b64) {
            $guid_bin = base64_decode($guid_b64);
            return join('-', array(
                bin2hex(strrev(substr($guid_bin, 0, 4))),
                bin2hex(strrev(substr($guid_bin, 4, 2))),
                bin2hex(strrev(substr($guid_bin, 6, 2))),
                bin2hex(substr($guid_bin, 8, 2)),
                bin2hex(substr($guid_bin, 10, 6))
            ));
 }
```

PR #103, opened 2014-09-04
Add NameID to attributes status page.

thijskh:
This page allows to view nearly everything of the active session,
except for the NameID. Display this as well.


jaimeperez:
Hi @thijskh!

Sorry that this took so long... I had some trouble merging since there was conflicts with some latest commits, but hopefully everything should be in place now.

PR #105, opened 2014-09-05
Enhance the metadata import to preserve the "required" flag of requested attributes

gollmann:
To allow for more focused attribute release it would be useful if the metadata import would preserve which of the requested attributes are required. This patch collects the required attributes into an “attributes.required” element.


jaimeperez:
This is now merged. Thanks a lot @gollmann!

PR #108, opened 2014-09-10
Bug fix to include the subject nameid in the data returned from a SAML 1.0 SSO authentication.

nickamon:
Fixed a bug where the subject nameid was not being returned back to the SP after successful authentication.


jaimeperez:
Thanks a lot @nickamon! And **sorry so much** for the delay this took to be merged... :cold_sweat:

PR #116, opened 2014-10-12
Add privacyIDEA module

cornelinux:
to enable One Time Password (OTP) authentication with
several different OTP tokens (HOTP, TOTP, Yubikey,
Google Authenticator, SMS, Email...) supported by
privacyIDEA.

The idea behind it is described in this howto: http://www.howtoforge.com/how-to-setup-single-sign-on-with-otp-using-simplesaml-php-and-privacyidea

### DIFF
--- a/config-templates/authsources.php
+++ b/config-templates/authsources.php
@@ -384,5 +384,29 @@ $config = array(
 
     ),
     */
+    /*
+    // example for an authentication against privacyIDEA (http://www.privacyidea.org).
+    // to enable One Time Password authentication.
+    // The credentials are sent to privacyIDEA via the web API and validated if
+    // these are valid OTP values for this user.
+    // attributemap can contain a mapping of the returned attributes.
+    'example-privacyidea' => array(
+        'privacyidea:privacyidea',
+        'privacyideaserver' => 'https://privacyidea',
+        'sslverifyhost' => False,
+        'sslverifypeer' => False,
+        'realm' => '',
+        'attributemap' => array('username' => 'samlLoginName',
+                                'surname' => 'surName',
+                                'givenname' => 'givenName',
+                                'email' => 'emailAddress',
+                                'phone' => 'telePhone',
+                                'mobile' => 'mobilePhone',
+                                'realm' => 'memberOf',
+                                'resolver' => 'resolver'
+                                ),
+    ),
+
+     */
 
 );
--- a/modules/privacyidea/privacyidea/docs/privacyidea.txt
+++ b/modules/privacyidea/privacyidea/docs/privacyidea.txt
@@ -1,0 +1,75 @@
+privacyidea module
+==================
+
+This module provides an authentication module for simpleSAMLphp to talk
+to the privacyIDEA authentication server.
+
+`privacyidea:privacyidea`
+: Authenticate a user against a privacyidea server.
+
+The module contacts the privacyIDEA server via the API
+
+  https://privacyidea/validate/samlcheck
+ 
+and authenticates the user according to the token assigned to the user.
+
+The response can also contain some attributes.
+
+To enable this module, create a file 'enable' in the
+directorpy 'modules/privacyidea/'.
+Then you need to add the authentication source 'privacyidea:privacyidea' to
+'config/authsources.php'. Do it like this:
+
+    'example-privacyidea' => array(
+        'privacyidea:privacyidea',
+
+        /* 
+         * The URI (including protocol and port) of the privacyidea server
+         * Required.
+         */
+        'privacyideaserver' => 'https://your.server.com',
+
+        /*
+         * Check if the hostname matches the name in the certificate
+         * Optional.
+         */
+        'sslverifyhost' => False,
+
+        /*
+         * Check if the certificate is valid, signed by a trusted CA
+         * Optional.
+         */
+        'sslverifypeer' => False,
+        
+        /*
+         * The realm where the user is located in.
+         * Optional.
+         */
+        'realm' => '',
+        
+        /*
+         * This is the translation from privacyIDEA attribute names to 
+         * SAML attribute names.
+         * Optional.
+         */
+         'attributemap' => array('username' => 'samlLoginName',
+         			 'surname' => 'surName',
+         			 'givenname' => 'givenName',
+         			 'email' => 'emailAddress',
+         			 'phone' => 'telePhone',
+         			 'mobile' => 'mobilePhone'),
+    ),
+
+
+User attributes
+---------------
+At the moment privacyIDEA will know and return the following attributes, 
+that can be mapped to SAML attributes:
+
+username:	The login name
+surname:	The real world name of the user as it is retrieved from the user source
+givenname:	The real world name of the user as it is retrieved from the user source
+mobile:		The mobile phone number of the user as it is retrieved from the user source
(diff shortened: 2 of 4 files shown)


jaimeperez:
Hi @cornelinux!

Did you find time to take a look into this?

cornelinux:
Hi,
honestly I was quite busy migrating the OTP backend to (privacyidea) to another framework (pylons->flask). Thus I did not find enough time to do so... :-/

jaimeperez:
No worries, just checking!

Just update the ticket whenever you find time to spend on it :smiley:

jaimeperez:
Hi @cornelinux!

Any update on this? I'm pushing towards 1.14 and your module would be a great addition to the release, IMHO :smile:

cornelinux:
Hi @jaimeperez 

I did failed with packagist. So I wonder if a I can add the module in the composer.json even if it is a subdirectory in a git project like
https://github.com/privacyidea/privacyidea/tree/master/authmodules/simpleSAMLphp

(I am absoultely unfamiliar with composer ;-)

jaimeperez:
Hi!

No, unfortunately you cannot do that. Modules must have their own repository and even follow a specific naming convention. In your case, the module should sit in a repository with base URL `https://github.com/privacyidea/simplesamlphp-module-privacyidea`. Then you can add the composer.json and enable packagist support. Take a look at [the instructions of the module installer](https://github.com/simplesamlphp/composer-module-installer) for specifics. You can also take a look at other modules to see how they look like and use that as a base.

Please tell me if I can help with anything!

cornelinux:
Hi,
I added privacyIDEA module to packagist and added composer.json. But I seem to lack a version number.
Do you know, where to add this version number?

php composer.phar require privacyidea/simplesamlephp-module-privacyidea

or

php composer.phar require privacyidea/simplesamlephp-module-privacyidea 1.*

does not work out.

jaimeperez:
Hi @cornelinux, thanks a lot for the update!

The version number is not strictly necessary if you don't want it. You can always point to a specific commit or install the latest version by using *dev-master* as version number. In any case, it's not a bad idea to have a version number. You can do that by tagging a specific commit with a version number, like `v1.0`. Then you can create a new release with that tag.

A couple more comments:

* I see your module is inside a "privacyidea" directory. That's not going to work. You need all the files in that directory to sit in the root of your repository, as that's also the root of your module.
* You also have it disabled by default. Since you need to explicitly install the module, it is a bit weird that after explicitly installing it is disabled, and you need to enable it. The approach we are following lately is to enable by default all modules that have their own repo. In order to do that, you just need to rename `default-disable` to `default-enable`.

Thanks again!

cornelinux:
...obviously I have to run
php composer.phar require privacyidea/simplesamlphp-module-privacyidea:dev-master

cornelinux:
Thanks for the info. I will do so.

jaimeperez:
Hi @cornelinux!

What's the current status of the module? Did you manage to get it to install through composer?

cornelinux:
I ran it. once.

But I totally does not get it.
I installed simplesamlphp on an ubuntu 14.04 machine. 

By which command am I supposed to run this composer thingy again.

Something is missing. There is no "composer.phar".
I installed simplesaml 1.13.2.

jaimeperez:
Hi!

That's the normal way to use composer. You always need to install it on your own, for several reasons:

* It could be already installed in your system.
* You could prefer to install it using your favourite package manager (apt-get, yum, whatever).
* The project using composer is not the authoritative source where to get it from. We shouldn't include our own version of composer, and if we want to keep up to date, we should update it every time there's a new release.

Imagine there's a security issue with composer. Should we release a new version of SimpleSAMLphp with no changes at all, just because we need to update composer? Should our users update the entire SimpleSAMLphp installation because of that, while they could just run `composer.phar self-update`?

cornelinux:
All true. I think I was just a bit disappointed, that composer was not available in the distributions default repo.
I will upgrade the source at packagist and test it again, since there were some majer enhancements last week. 
Now it also supports Challenge Response with OTP and also U2F devices.
I will return here, after having checked it and updated at packagist.

cornelinux:
Hi, ok, I was able to install the privacyIDEA module via composer.
It installed to:

./vendor/privacyidea/simplesamlphp-module-privacyidea/

my authentication class is called "sspmod_privacyidea_Auth_Source_privacyidea".

How do I activate it (I touched an "enable" file)? When I installed it to ./modules/privacyidea I used this entry in the authsources:

    'example-privacyidea' => array(
        'privacyidea:privacyidea',
        'privacyideaserver' => 'https://172.16.200.138/pi',
        'sslverifyhost' => False,
        'sslverifypeer' => False,
        'realm' => '',
         'attributemap' => array('username' => 'samlLoginName',
                                 'surname' => 'surName',
                                 'givenname' => 'givenName',
                                 'email' => 'emailAddress',
                                 'phone' => 'telePhone',
                                 'mobile' => 'mobilePhone'),
    ),

But now, with the code under "vendor" I get this exception:

Backtrace:
0 /var/simplesamlphp/www/module.php:179 (N/A)
Caused by: Exception: Could not resolve 'privacyidea:privacyidea': No class named 'sspmod_privacyidea_Auth_Source_privacyidea'.
Backtrace:
6 /var/simplesamlphp/lib/SimpleSAML/Module.php:134 (SimpleSAML_Module::resolveClass)
5 /var/simplesamlphp/lib/SimpleSAML/Auth/Source.php:207 (SimpleSAML_Auth_Source::parseAuthSource)
4 /var/simplesamlphp/lib/SimpleSAML/Auth/Source.php:247 (SimpleSAML_Auth_Source::getById)
3 /var/simplesamlphp/lib/SimpleSAML/Auth/Default.php:59 (SimpleSAML_Auth_Default::initLogin)
2 /var/simplesamlphp/lib/SimpleSAML/Auth/Simple.php:136 (SimpleSAML_Auth_Simple::login)
1 /var/simplesamlphp/modules/core/www/authenticate.php:45 (require)
0 /var/simplesamlphp/www/module.php:134 (N/A)

jaimeperez:
Thanks a lot!

The code should not be under `vendor`, but in `modules` as it always has been. I've been taking a look and the problem is related to a recent change in composer that broke our module installer plugin. It has now been fixed, so it should just work fine.

Can you try it?

cornelinux:
Marvelous!

Everything went fine, now:

root@privacyidea:/var/simplesamlphp# php composer.phar require simplesamlphp/composer-module-installer
Using version ^1.1 for simplesamlphp/composer-module-installer
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Removing simplesamlphp/composer-module-installer (v1.1.2)
  - Installing simplesamlphp/composer-module-installer (v1.1.3)
    Downloading: 100%         

Writing lock file
Generating autoload files
root@privacyidea:/var/simplesamlphp# php composer.phar require privacyidea/simplesamlphp-module-privacyidea:v1.2
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Installing privacyidea/simplesamlphp-module-privacyidea (v1.2)
    Downloading: 100%         

Generating autoload files


So the module installs fine and I can use it for authentication with the same authsource!

jaimeperez:
That's great news!

I've just added your module to [the list of modules in the website](https://simplesamlphp.org/modules) :-)

PR #131, opened 2014-11-22
Add coveralls support

br00k:
This might need a few revisions before it gets incorporated. ...but worth a try.


thijskh:
Second attempt has been merged including these changes, although I'm not fully sure it works yet.

PR #135, opened 2014-12-08
Patch to have ownCloud/user_saml work

deltafunction:
when using simplesamlphp with the default PHP session store

### DIFF
--- a/lib/SimpleSAML/Auth/State.php
+++ b/lib/SimpleSAML/Auth/State.php
@@ -162,6 +162,18 @@ class SimpleSAML_Auth_State {
 		$session = SimpleSAML_Session::getSessionFromRequest();
 		$session->setData('SimpleSAML_Auth_State', $id, $serializedState, self::getStateTimeout());
 
+		// This is for ownCloud/user_saml to work (using the default PHP session store)
+		$sh = SimpleSAML_SessionHandler::getSessionHandler();
+		try {
+				$sh->saveSession($session);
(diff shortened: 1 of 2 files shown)


deltafunction:
Any progress on this?

jaimeperez:
Hi @deltafunction!

Thanks a lot for your patch and my apologies for the delay. Could you elaborate a little bit on what's exactly the problem you are solving here, and why your patch is needed?

thijskh:
@deltafunction Could you clarify the intention? I must say that it's also not evident to my why this change is needed.

thijskh:
Could it be that this problem is the same as the one in #260, @deltafunction?

deltafunction:
It looks related yes. Let me try out the patch or, preferably, a new
release with the patch, and get back to you.

thijskh:
We'd prefer to know if the patch solves your problem before we create a release :)

jaimeperez:
Hi @deltafunction!

Were you able to test this out?

deltafunction:
Hi,

I tried out the 1.14.0-rc1 tarball and it does not fix my problem. I get the usual "Session information lost" (or something along those lines).

thijskh:
I'm still a bit in doubt about this, because we don't have a clear understanding about what ownCloud specifically does that causes the problem (and is not a problem with other SPs), and therefore cannot judge whether the proposed solution is the right way to go.

PR #137, opened 2014-12-15
Change attribute representation in authmemcookie

jornane:
Allow for different representations in the config file. The current
representation gives issues when handling multi-value attributes
containing colons, like entitlements.

### DIFF
--- a/config-templates/authmemcookie.php
+++ b/config-templates/authmemcookie.php
@@ -81,6 +81,33 @@ $config = array(
 	 */
 	'memcache.port' => 11211,
 
+	/*
+	 * This option specifies how the attributes are represented in the memcache store.
+	 * Up until SimpleSamlPhp 1.13, the format was "implode".
+	 * In this format, the representation consists of multiple lines, one attribute-name per line.
+	 * Multi-value attributes would be concatinated with colons in between.
+	 * For backwards compatibility, this is still the standard behaviour.
+	 * 
+	 * The possible values are:
+	 * # implode
+	 *   Classic behaviour, memcache will contain one line per attribute name,
+	 *   if the attribute is multi-value, the values are concatinated with a colon (:) in between.
+	 *   Lines are separated with Windows-style newlines (\r\n)
+	 * 
+	 * # multi
+	 *   Looks like implode, except that there will not be one line per attribute name,
+	 *   but per attribute value. Multi-value attributes will be represented with multiple lines
+	 *   with identical keys.
+	 *   This option is recommended if you want to handle multi-value attributes,
+	 *   but you have to be compatible with the old format as well.
+	 *
+	 * # json
+	 *   The attributes are serialised as JSON. The root is a dictionary with the attribute name as key,
+	 *   and an array of attribute values as value.
+	 */
+	'format' => 'implode', // backwards-compatible
+	//'format' => 'json', // recommended
+
 );
 
 ?>
--- a/www/authmemcookie.php
+++ b/www/authmemcookie.php
@@ -91,12 +91,25 @@ try {
 	/* Store the authentication data in the memcache server. */
 
 	$data = '';
+	$amcConfig = SimpleSAML_Configuration::getConfig('authmemcookie.php');
+	$format = strtolower($amcConfig->getBoolean('format', 'implode'));
+	assert('in_array($format, array("implode", "multi", "json"))');
 	foreach($authData as $n => $v) {
-		if(is_array($v)) {
-			$v = implode(':', $v);
+		if(!is_array($v)) {
+			$v = array((string)$v);
+		}
+		if ($format === 'multi') {
+			$v = array(implode(':', $v));
 		}
 
-		$data .= $n . '=' . $v . "\r\n";
+		if (in_array($format, array('implode', 'multi'))) {
+			foreach($v as $v2) {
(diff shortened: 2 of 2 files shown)


config-templates/authmemcookie.php
```diff
@@ -81,6 +81,33 @@
 	 */
 	'memcache.port' => 11211,
 
+	/*
+	 * This option specifies how the attributes are represented in the memcache store.
+	 * Up until SimpleSamlPhp 1.13, the format was "implode".
```
jaimeperez: s/SimpleSamlPhp/SimpleSAMLphp/

Just trying to keep coherence with the format we chose to use for the name :smiley:

config-templates/authmemcookie.php
```diff
@@ -81,6 +81,33 @@
...
+	 * In this format, the representation consists of multiple lines, one attribute-name per line.
+	 * Multi-value attributes would be concatinated with colons in between.
```
jaimeperez: s/concatinated/concatenated/

config-templates/authmemcookie.php
```diff
@@ -81,6 +81,33 @@
...
+	 * For backwards compatibility, this is still the standard behaviour.
+	 * 
+	 * The possible values are:
+	 * # implode
+	 *   Classic behaviour, memcache will contain one line per attribute name,
+	 *   if the attribute is multi-value, the values are concatinated with a colon (:) in between.
```
jaimeperez: s/concatinated/concatenated/

config-templates/authmemcookie.php
```diff
@@ -81,6 +81,33 @@
...
+	 * # multi
+	 *   Looks like implode, except that there will not be one line per attribute name,
+	 *   but per attribute value. Multi-value attributes will be represented with multiple lines
+	 *   with identical keys.
+	 *   This option is recommended if you want to handle multi-value attributes,
+	 *   but you have to be compatible with the old format as well.
+	 *
+	 * # json
+	 *   The attributes are serialised as JSON. The root is a dictionary with the attribute name as key,
+	 *   and an array of attribute values as value.
```
jaimeperez: Maybe it would be useful to have a simple example for each format?

www/authmemcookie.php
```diff
@@ -91,12 +91,25 @@
 	/* Store the authentication data in the memcache server. */
 
 	$data = '';
+	$amcConfig = SimpleSAML_Configuration::getConfig('authmemcookie.php');
+	$format = strtolower($amcConfig->getBoolean('format', 'implode'));
```
jaimeperez: This should be using `getString()`, as the option is a `string`, not a `boolean`.

www/authmemcookie.php
```diff
@@ -91,12 +91,25 @@
...
+	assert('in_array($format, array("implode", "multi", "json"))');
```
jaimeperez: I don't think we should use an `assert()` here. Asserts are intended for debugging, and we are already abusing them all over the place. If the assertion fails, it would be due to a configuration mistake, not a development mistake, so we should throw an exception.

www/authmemcookie.php
```diff
@@ -91,12 +91,25 @@
...
-		$data .= $n . '=' . $v . "\r\n";
+		if (in_array($format, array('implode', 'multi'))) {
+			foreach($v as $v2) {
+				$data .= $n . '=' . str_replace("\r\n", "\n", $v2) . "\r\n";
+			}
+		}
+	}
+	if ($format === 'json') {
+		$data = json_encode($authData);
 	}
```
jaimeperez: This should be an else clause, and the foreach loop should be inside. There's no point on iterating over the set of authentication data if we are just JSON-encoding it anyway.

www/authmemcookie.php
```diff
@@ -91,12 +91,25 @@
...
+	$amcConfig = SimpleSAML_Configuration::getConfig('authmemcookie.php');
+	$format = strtolower($amcConfig->getBoolean('format', 'implode'));
+	assert('in_array($format, array("implode", "multi", "json"))');
 	foreach($authData as $n => $v) {
-		if(is_array($v)) {
-			$v = implode(':', $v);
+		if(!is_array($v)) {
+			$v = array((string)$v);
+		}
+		if ($format === 'multi') {
```
jaimeperez: This sounds like a bug. If I understood the comments correctly, _implode_ will yield something like this:

```ini
attribute=value1:value2:value3
```

while _multi_ will yield:

```ini
attribute=value1
attribute=value2
attribute=value3
```

But here we are imploding all the different values with a colon in between in case format is _multi_, which is incoherent with the documentation. So I understand this should be:

```php
if ($format === 'implode') {
```

www/authmemcookie.php
```diff
@@ -91,12 +91,25 @@
...
+			$v = array(implode(':', $v));
 		}
 
-		$data .= $n . '=' . $v . "\r\n";
+		if (in_array($format, array('implode', 'multi'))) {
```
jaimeperez: Since there's only a couple of options that we are comparing to, I suggest making the condition more readable (and slightly more efficient):

```php
if ($format === 'implode' || $format === 'multi') {
```

jaimeperez:
Thanks @yorn! (and sorry for the delay)

I've made some comments directly on the code, take a look at them whenever you have time :smiley:

tvdijen:
This will not be adressed in the codebase for simplesamlphp, since it moved to a module

PR #139, opened 2014-12-19
Make sure metadata-converter.php requires an admin user

relaxnow:
Depending on server configuration this may be used in a Denial Of Service attack by tying up all webserver workers with large POST bodies.

### DIFF
--- a/www/admin/metadata-converter.php
+++ b/www/admin/metadata-converter.php
@@ -2,6 +2,9 @@
 
 require_once('../_include.php');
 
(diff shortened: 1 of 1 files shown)


jaimeperez:
Thanks for submitting this pull request @relaxnow!

I don't think nobody is relying on this in a way that could break if now they are required to authenticate as admin, and the perils of leaving it like it is now are not negligible, as you have identified, so I'm merging it now :+1:

PR #140, opened 2015-01-02
protectmetadata was not enforced for SP metadata.

thijskh:
This check was not ported to the saml2 module.


jaimeperez:
Thanks @thijskh!

Issue #141, opened 2015-01-02
Support for SQL Server as Data Store for Sessions


tashbaugh:
Inside of config.php I have parameters setup like this:

'store.type'                    => 'sql',
'store.sql.dsn'                => 'sqlsrv:Server=DBServer-1,1433;Database=SAMLDB',
'store.sql.username'     => 'dbuser',
'store.sql.password'      => 'dbpassword',
'store.sql.prefix'            => 'SAML',

I have other PHP applications on the same server as SimpleSAMLPHP that are communicating with this specific SQL Server database and do not have any issues. 

With this configuration, when I utlize the Test Authentication Sources and click on my named SP I am appropriately redirected to the IDP. When I login to the IDP I am then redirected back to the SP but am presented with a State Information Lost page. 

When I inspect the database no new tables have been added with a prefix of SAML. No PHP errors are logged and nothing is in the logs directory of SimpleSAMLPHP. 

If I switch the store type back to phpsession then I the redirect back to the SP works and I have a valid session. 

Upon reviewing the Windows Event Log I found this error occurring multiple times when simplesamlphp attempts to store session information in the database:

"SQLSTATE[42000]: [Microsoft][SQL Server Native Client 11.0][SQL Server]Implicit conversion from data type nvarchar to timestamp is not allowed. Use the CONVERT function to run this query."

kpugazh:
I have the same issue with MySQL
I did configured the SAML to store session in DB with teh following configuration in config/config.php

'store.type' => 'sql',
'store.sql.dsn' => 'mysql:host=localhost;dbname=samldb',
'store.sql.username' => 'root',
'store.sql.password' => 'dbpassword',
'store.sql.prefix' => 'simpleSAMLphp',

Yet I am getting the same error. I did not create any table in this db since the link 
https://simplesamlphp.org/docs/1.10/simplesamlphp-maintenance#section_2
but I see not tables are created

I am able to create tabels from PhP code through mysql PDO.

jaimeperez:
Hi @kpugazh,

Are you getting a Microsoft's SQL Server specific error while using a MySQL backend? :open_mouth:

tvdijen:
@close since no response in 3yrs

PR #142, opened 2015-01-06
Adding support for SQLSRV driver to SQL.php

tashbaugh:
Issue #141 Create kvstore _expire column as a datetime data type when using the sqlsrv driver. Added case 23000 for insert and update statements for kvstore table when using the sqlsrv driver. Once these changes were made in my install of simplesamlphp I was able to successfully use a SQL Server 2008 database as my store for session data.


thijskh:
Thanks. Can you perhaps clarify why you need to ignore that error even after you've changed the column type? Why does that error occur?

tashbaugh:
After I changed the column type, I continually encountered the following error:

simplesamlphp[25152]
4 [29783bcf35] PDOException: SQLSTATE[23000]: [Microsoft][SQL Server Native Client 11.0][SQL Server]Violation of PRIMARY KEY constraint 'PK__SAML_kvs__63DAC062505BE5AD'. Cannot insert duplicate key in object 'dbo.SAML_kvstore'. The duplicate key value is (69bca63ac3f3eb24d3929d125e2a5528, session).

Once I added the switch case for error code 23000 then the above error no longer occurred and I was able to use SimpleSAMLPHP with SQL Server for session management.

Additionally, I found another problem on logout with data type of TIMESTAMP when it needed to be DATETIME for SQL Server. I will be making a separate PR for that one.

dustinmoris:
Hi, thanks to your pull request I have submitted a very similar change, but instead of catching an exception and trying to update the row I used the MS SQL server MERGE command which is the equivalent of the MySQL REPLACE INTO command.

My change can be found here:

thijskh:
I don't think we'll be merging this in this form.  #238 can be a starting point for anyone wanting to fix this issue (and has access to an actual MS SQL server).

PR #146, opened 2015-01-13
Adding MDX metadata handler

sitya:
With this MDX metadata handler simpleSAMLphp can load metadata dynamically from MDX servers (like [pyFF]).

More about Metadata Query Protocol: https://github.com/iay/md-query

[pyff]: https://github.com/leifj/pyFF

### DIFF
--- a/lib/SimpleSAML/Metadata/MetaDataStorageHandlerMDX.php
+++ b/lib/SimpleSAML/Metadata/MetaDataStorageHandlerMDX.php
@@ -201,7 +201,8 @@ class SimpleSAML_Metadata_MetaDataStorageHandlerMDX extends SimpleSAML_Metadata_
 		case 'shib13-sp-remote':
 			return $entity->getMetadata1xSP();
 		case 'attributeauthority-remote':
-			return $entity->getAttributeAuthorities();
+			$ret = $entity->getAttributeAuthorities();
+			return $ret[0];
 
 		default:
 			SimpleSAML_Logger::warning('MetaData - Handler.MDX: Unknown metadata set: ' . $set);
@@ -262,9 +263,7 @@ class SimpleSAML_Metadata_MetaDataStorageHandlerMDX extends SimpleSAML_Metadata_
 		}
 
 		$entity = SimpleSAML_Metadata_SAMLParser::parseString($xmldata);
-		SimpleSAML_Logger::debug('MetaData - Handler.MDX: Completed parsing of [' .
-			$mdx_url . ']'  );
-
+		SimpleSAML_Logger::debug('MetaData - Handler.MDX: Completed parsing of [' .	$mdx_url . ']'  );
 
 		if( $this->validateFingerprint !== NULL) {
 			if(!$entity->validateFingerprint($this->validateFingerprint)) {
@@ -274,10 +273,8 @@ class SimpleSAML_Metadata_MetaDataStorageHandlerMDX extends SimpleSAML_Metadata_
 
 		$data = self::getParsedSet($entity, $set);
 		if ($data === NULL) {
-			throw new Exception('No metadata for set "' . $set .
-				'" available from "' . $index . '".');
+			throw new Exception('No metadata for set "' . $set . '" available from "' . $index . '".');
 		}
-		if (is_array($data[0])) $data = $data[0];
 
 		$this->writeToCache($set, $index, $data);
 
--- a/lib/SimpleSAML/Metadata/MetaDataStorageHandlerMDX.php
+++ b/lib/SimpleSAML/Metadata/MetaDataStorageHandlerMDX.php
@@ -1,0 +1,290 @@
+<?php
+
+/**
+ * This class implements SAML Metadata Exchange Protocol
+ *
+ * @author Andreas Åkre Solberg, UNINETT AS.
+ * @author Olav Morken, UNINETT AS.
+ * @author Tamas Frank, NIIFI
+ * @package simpleSAMLphp
+ */
+class SimpleSAML_Metadata_MetaDataStorageHandlerMDX extends SimpleSAML_Metadata_MetaDataStorageSource {
+
+    /**
+     * The URL of MDX server (url:port)
+     */
+    private $server;
+
+    /**
+     * The fingerprint of the certificate used to sign the metadata. You don't need this option if you don't want to validate the signature on the metadata.
+     */
+    private $validateFingerprint;
+
+	/**
+	 * The cache directory, or NULL if no cache directory is configured.
+	 */
+	private $cacheDir;
+
+
+	/**
+	 * The maximum cache length, in seconds.
+	 */
+	private $cacheLength;
+
+
+	/**
+	 * This function initializes the dynamic XML metadata source.
+	 *
+	 * Options:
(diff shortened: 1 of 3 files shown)


jaimeperez:
Hi @sitya! Thanks a lot for your pull request!

I'll try to set this up asap and verify everything works fine. I'll try also to give the code a brief look and give some comments back.

lib/SimpleSAML/Metadata/MetaDataStorageHandlerMDX.php
```diff
@@ -0,0 +1,292 @@
...
+	 * @param $index  The entityId or metaindex we are looking up.
+	 * @param $set  The set we are looking for metadata in.
+	 * @return An associative array with metadata for the given entity, or NULL if we are unable to
+	 *         locate the entity.
+	 */
+	public function getMetaData($index, $set) {
+		assert('is_string($index)');
+		assert('is_string($set)');
+
+		if (!preg_match('@(https?://([-\w\.]+)+(:\d+)?(/([\w/_\.]*(\?\S+)?)?)?)@', $index)) {
```
jaimeperez: Why are you enforcing the entity ID to be a URL? The SAML spec doesn't mandate it to be a URL, so it could virtually be anything. In fact it's quite common to have URNs as entity IDs...

sitya: Yes, I forgot to remove this section. I was rewriting MetaDataStorageHandlerDynamicXML.php and it remained in it. I have commited this now.

trscavo:
If you want to further test this code, you can do so against the metadata query server at mdq-beta.incommon.org. See the [Per-Entity Metadata Pilot](https://spaces.internet2.edu/x/3w7kAg) wiki page for details.

For reference, there are configuration examples for the Shibboleth SP and IdP at the bottom of the above wiki page.

jaimeperez:
Thanks @trscavo, very useful indeed!

lib/SimpleSAML/Metadata/MetaDataStorageHandlerMDX.php
```diff
@@ -0,0 +1,287 @@
...
+				throw new Exception ('Error, could not verify signature for entity: ' . $index . '".');
+			}
+		}
+
+		$data = self::getParsedSet($entity, $set);
+		if ($data === NULL) {
+			throw new Exception('No metadata for set "' . $set .
+				'" available from "' . $index . '".');
+		}
+		if (is_array($data[0])) $data = $data[0];
```
jaimeperez: I understand this is to handle the case when multiple entities are returned from the MDX server, which I don't know if it's even possible. In any case, I assume the most common case would be the MDX server will return exactly one entity when queried for an `entityID`, and in that case this code generates a warning in the log file, since the index 0 won't exist in the parsed metadata array. So if we want to see if the result from `getParsedSet()` returns a bunch of entities, we should do something like this:

```php
        $multiple_entities = true;
        foreach (array_keys($data) as $key) {
            if (!is_numeric($key)) {
                $multiple_entities = false;
                break;
            }
            if (!is_array($data[$key])) {
                $multiple_entities = false;
                break;
            }
        }
        if ($multiple_entities) {
            $data = reset($data);
        }
```

Basically, we assume we have multiple entities, and then we iterate over the keys of the array. As soon as we find one key that's not numeric or is not an array, we know this is not an array with multiple entities. If the loop finishes correctly, then we can fetch the first entity because it's probable that we got a bunch of them.

Could you please fix this in the pull request? As soon as this is fixed I think the code can be merged :+1:

jaimeperez:
Hi @sitya,

Apart from this, which is a nice functionality, have you considered implementing MDX support for the `metarefresh` and `aggregator2` modules? The `metarefresh` case is a bit more difficult, since we would need to configure it on beforehand with the entity IDs we want it to fetch, but it could be interesting in any case.

@trscavo, are there any plans to allow an MDX server to return a list of entities registered when queried for it? It would be very useful, IMHO.

trscavo:
I believe you're asking me about the Metadata Query Protocol (or MDQ Protocol). An overview page with links to the authoritative spec documents will be found in the Spaces wiki: https://spaces.internet2.edu/x/2w7kAg

The short answer is yes: a Metadata Query Server serves arbitrary collections of entity descriptors. The spec elaborates on per-entity metadata (as you suspected) and shows how to retrieve an all-entities aggregate as well. Anything in between is possible (but not specified).

That said, a Metadata Query Server is first and foremost a per-entity metadata server. I would concentrate on that functionality. SimpleSAMLphp already has a metarefresh module that can refresh aggregates so I'm not sure why you would want to reinvent that wheel.

lib/SimpleSAML/Metadata/MetaDataStorageHandlerMDX.php
sitya: Ok, I fixed it.

The code `if (is_array($data[0])) $data = $data[0];` was needed because if `getParsedSet()` got `attributeauthority-remote`, then the metadata of the AttributeAuthority entity was under a '0' indexed subarray. If you look at `SAMLParser.php`, there is no function like `getMetadata20IdP` only `getAttributeAuthorities` which returns an array with subarrays. I didn't want to edit `SAMLParser.php`, so I handled this case at `getParsedSet()`.

I have tested, and it worked fine, without any warning or error message.

jaimeperez:
I've just merged this, thanks a lot for your work @sitya, and sorry for the delay!

@trscavo, basically what I'm wondering is how to make this more usable. Currently, the first request arriving the IdP/SP will have a significant delay corresponding to the time the IdP/SP takes to fetch the metadata from the MDX server, process and cache it. I'd like to be able to _prefetch_ metadata periodically, that meaning:

1. Run a cronjob that fetches a list of entities from our trusted MDX server, and download metadata for each of them.
2. Cache the downloaded metadata, so that it's available immediately.
3. Reduce the delays only to those entities that are so fresh that the cronjob didn't download metadata for them yet. I believe this is unlikely to happen, as users shouldn't be trying to access an entity as soon as it's been published (not as a rule, at least). Maybe those testing that the entity is working would suffer the delay, but in that case it's not a big deal, I assume.

So of course to implement such a feature we need a way to fetch the list of entities that are provided by the MDX server. One could argue that for that we already have metarefresh and standard metadata feeds, but I think MDX provides some value that we don't have with metarefresh, like conditionally fetching entities that have been changed or are new, or still having the possibility to fetch them on the fly, eliminating the gap between publication in the feed and the time an entity is finally available.

trscavo:
@sitya, thank you for this important contribution.

@jaimeperez, I have thoughts along these lines but I'm a fedop, not an implementer, so I'll resist the urge. May I suggest you post these questions to the [metadata-support](https://lists.incommon.org/sympa/info/metadata-support) mailing list? This is where we had hoped just this kind of discussion would take place. Others involved in the [Per-Entity Metadata Pilot](https://spaces.internet2.edu/x/3w7kAg) are subscribed to that mailing list. Would you mind?

Related to this, are you still testing against mdq-beta.incommon.org? The reason I ask is: I'm compiling a list of current uses of that server to support a request to extend the pilot.

sitya:
great, thanks :)

jaimeperez:
@trscavo, no problem. Thanks for the suggestion!

PR #171, opened 2015-01-16
RPM specfile for simpleSAMLphp, tested on CentOS 7

jschlyter:


extra/simplesamlphp.spec
```diff
@@ -0,0 +1,94 @@
...
+
+Name:      %{name}
+Summary:   %{summary}
+Version:   %{version}
+Release:   %{release}
+License:   %{license}
+Group:     %{group}
+Source0:   %{source}
+BuildArch: noarch
+Requires:  httpd, mod_ssl, mod_php, php-ldap, php-mcrypt, php-pecl-krb5, php-xml
```
jaimeperez: I don't think you should require php-pecl-krb5. I doubt it's used for anything unless you are using the negotiate module, which is just a marginal usage of SSP.

extra/simplesamlphp.spec
```diff
@@ -0,0 +1,94 @@
+%define name      simplesamlphp
+%define summary   SAML IDP/SP written i PHP
+%define version   1.13.2
+%define release   10
+%define license   LGPL 2.1
+%define group     Networking/WWW
+%define source    %{name}-%version.tar.gz
+%define url       https://simplesamlphp.org
+%define vendor    Feide RND
```
jaimeperez: The vendor here should be UNINETT AS (at least for now).

extra/simplesamlphp.spec
```diff
@@ -0,0 +1,94 @@
+%define name      simplesamlphp
+%define summary   SAML IDP/SP written i PHP
```
jaimeperez: Typo? _written i**n** PHP_?

extra/simplesamlphp.spec
```diff
@@ -0,0 +1,94 @@
...
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/config-login-feide.php
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/authmemcookie.php
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/translation.php
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/authsources.php
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/acl.php
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/ldapmulti.php
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/cas-ldap.php
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/config-login-auto.php
+%config(noreplace) %attr(0640, root,apache) /var/lib/simplesamlphp/config/ldap.php
+%dir %attr(0750, root,apache) /var/lib/simplesamlphp/metadata
```
jaimeperez: I don't know how necessary would be to protect the metadata directory, in general there should be no secrets or private information there, but I don't think it makes any harm anyway. What I do miss here is protection for the *cert* directory, which holds the certificates used by the IdP (including their private keys) and therefore it's critical that the directory is not accessible by anyone except the owner of the web server process.

jschlyter:
minor errors fixed now.

extra/simplesamlphp.spec
```diff
@@ -0,0 +1,94 @@
...
+#    semanage fcontext -a -t httpd_sys_rw_content_t '/var/lib/simplesamlphp/data(/.*)?'
+#    semanage fcontext -a -t httpd_sys_rw_content_t '/var/lib/simplesamlphp/log(/.*)?'
+
+%preun
+
+
+%files
+%defattr(-,root,root)
+/var/lib/simplesamlphp/
+%dir %attr(0750, root,apache) /var/lib/simplesamlphp/config
```
jaimeperez: Most of the files under _config/_ are not commonly used. The most important here are _config.php_ and _authsources.php_, which must **always** be present.

jaimeperez:
Hi @jschlyter, thanks a lot for your contribution!

Apart from the comments inline (you already fixed most of them), I'm wondering whether this would be usable for other linux distributions using RPM for package management. I'm asking because it's nice to have the spec for CentOS, but of course for other distributions too, and in case this is not generic enough to work for all of them, I think we should mark it as something specific for CentOS then...

extra/simplesamlphp.spec
jschlyter: I've fixed cert directory as well, although I personally would prefer /etc/pki/saml on a CentOS/RedHat system.

jschlyter:
This specfile _should_ work on other RPM-based distributions as well, but we need more feedback on that.

extra/simplesamlphp.spec
jaimeperez: Yeah, I'd be fine with that too, but in that case the _config.php_ file should be configured automatically to point to that directory. Given my complete ignorance on RPM package specs, would that be something that we could do automatically?

It could also be interesting to have the possibility to ask the user for the base URL and configure the `baseurlpath` parameter in _config.php_ according to that, but I'm thinking out loud here...

jschlyter: I believe we can leave the cert directory as default, as well as baseurlpath, for now. This - as well as logging options - depends on deployment.

jschlyter:
Maybe I should note that CentOS/RedHat/Fedora should all work with this spec.

thijskh:
The simplesaml download page currently contains an note in **bold** face recommending not to use the distribution packaged versions of SSP. Besides from it in my opinion being untrue I also find it rather unfriendly against those putting effort into making SSP work well within the distribution packaging systems.

I'd gladly help supply more information about distribution packaged software and why it's not necessarily harmful to use.

jschlyter:
We package SSP internally for production systems and having a specfile as part of the source may be helpful to other.

thijskh:
My comment was not directed at you but at the SSP webmaster :)

jaimeperez:
@thijskh, that comment has been there for a while, mainly because of the packaged version shipped in debian based systems. That package had been unmaintained for long time (I see now they've updated to the latest version), and that means it ended up being a very old version with several security vulnerabilities and bugs, which is why we were not recommending to use it.

jaimeperez:
Now that I look more carefully into it... Testing and unstable are in 1.13.1, which is quite good. Stable is on 1.9.2, which is 2 years and a half old now. Not the best situation, IMHO.

jschlyter:
This is exactly why we should ship packing specs for deb/rpm with the source so people easily can rebuild their own packages.

jaimeperez:
Well, most of the time I think it's the distro's policy that prevents using a reasonably up-to-date version, but in any case I agree that it's a good idea to have packaging specs, and your last commit looks quite good as a starting point, so I'm merging :smile: 

Thanks a lot for the good work!

thijskh:
That the version number is old doesn't mean it's unmaintained. Distros do not upgrade to new major upstream releases in their stable branches, but just apply the security fixes. This is a feature. If you have a working installation based on 1.9, there's no need to upgrade if there's no security fixes in there.

We're using distro versions of Apache and PHP for similar reasons: when there's an issue we do not want to upgrade to a new codebase with all kinds of changes, but just run apt-get upgrade and get a patched Apache that just fixes the problems. I don't see why SSP would be different.

jaimeperez:
I totally agree @thijskh. The main problem as I see it, is that if you end up using very old versions it's difficult to apply security updates, sometimes even impossible. But that's just a personal opinion :wink:

thijskh:
I also found this: https://bugzilla.redhat.com/show_bug.cgi?id=800867. Maybe submit it also to there?

jschlyter:
Sure, please do :-)

PR #175, opened 2015-01-30
Extract RSA or DSA PEM data from cert. 

ottenhoff:
Google treats RSA and DSA equally in their documentation (https://developers.google.com/google-apps/help/articles/sso-keygen)

From the OpenSSL docs https://www.openssl.org/docs/apps/dsa.html:

The PEM private key format uses the header and footer lines:

 -----BEGIN DSA PRIVATE KEY-----
 -----END DSA PRIVATE KEY-----
The PEM public key format uses the header and footer lines:

 -----BEGIN PUBLIC KEY-----
 -----END PUBLIC KEY-----

### DIFF
--- a/lib/SimpleSAML/Configuration.php
+++ b/lib/SimpleSAML/Configuration.php
@@ -1125,10 +1125,14 @@ class SimpleSAML_Configuration {
 				throw new Exception($this->location . ': Unable to load certificate/public key from file "' . $file . '".');
 			}
 
-			/* Extract certificate data (if this is a certificate). */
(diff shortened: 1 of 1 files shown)


ottenhoff:
Even with this modification, I can't get Google to accept a SAML assertion signed with a DSA key.

jaimeperez:
@ottenhoff, were you able to make it work?

ottenhoff:
No, I was never able to get a DSA key working

jaimeperez:
That's bad :disappointed: 

I'm closing this for the moment then, feel free to reopen if you finally make it work.

PR #176, opened 2015-01-30
CAS Server module should store its tickets in the central store

ottenhoff:
This means that if an institution is load-balancing across two nodes, the CAS ticket
can be properly stored in the multiple memcache nodes. The previous behavior
stored the ticket information in a file. Storing in a file means that load-balancing
is very difficult and storing in a file is much slower than using memcache.

### DIFF
--- a/modules/casserver/config-templates/module_casserver.php
+++ b/modules/casserver/config-templates/module_casserver.php
@@ -13,8 +13,6 @@ $config = array (
 	// Legal values: saml2, shib13
 	'auth' => 'saml2',
 	
-	'ticketcache' => 'ticketcache',
-
 	'attrname' => 'mail', // 'eduPersonPrincipalName',
 	#'attributes' => TRUE, // enable transfer of attributes
 	
--- a/modules/casserver/www/login.php
+++ b/modules/casserver/www/login.php
@@ -39,10 +39,8 @@ if (!$as->isAuthenticated()) {
 
 $attributes = $as->getAttributes();
 
-$path = $casconfig->resolvePath($casconfig->getValue('ticketcache', '/tmp'));
-
 $ticket = str_replace( '_', 'ST-', SimpleSAML_Utilities::generateID() );
-storeTicket($ticket, $path, array('service' => $service,
+storeTicket($ticket, array('service' => $service,
 	'forceAuthn' => $forceAuthn,
 	'attributes' => $attributes,
 	'proxies' => array(),
--- a/modules/casserver/www/proxy.php
+++ b/modules/casserver/www/proxy.php
@@ -22,12 +22,10 @@ $legal_service_urls = $casconfig->getValue('legal_service_urls');
 if (!checkServiceURL($targetService, $legal_service_urls))
 	throw new Exception('Service parameter provided to CAS server is not listed as a legal service: [service] = ' . $service);
 
-$path = $casconfig->resolvePath($casconfig->getValue('ticketcache', 'ticketcache'));
-
-$ticket = retrieveTicket($pgt, $path, false);
+$ticket = retrieveTicket($pgt, false);
 if ($ticket['validbefore'] > time()) {
 	$pt = str_replace( '_', 'PT-', SimpleSAML_Utilities::generateID() );
-	storeTicket($pt, $path, array(
+	storeTicket($pt, array(
 		'service' => $targetService,
 		'forceAuthn' => false,
 		'attributes' => $ticket['attributes'],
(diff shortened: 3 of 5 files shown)


shoaibali:
Very nice! I also ran in to dead-end with this, but if you are stuck with only having to use CAS then this would be very useful to have. I ended up going down authmemcookie path and using this apache2 module instead for solving high availability scenario https://github.com/piersharding/apache2-mod_auth_memcookie

shamus13:
I have been maintaining a fork of the CAS module at https://github.com/shamus13/simpleSAMLphp-sbcasserver for years, where i have addressed this and several other issues.

I have been meaning to have a little chat with Jaime about the CAS support in simplesamlphp, but never quite gotten around to it.

ottenhoff:
Yes, the shamus13 fork is quite a bit better than the casserver functionality in simpleSAMLPHP 1.13.2.

@jaimeperez do you want individual pull requests for each piece of enhanced functionality?  Or is there a plan to bring the entire CAS server fork back into core?

jaimeperez:
Hi guys! First of all, thanks a lot for your valuable input and contributions!

Currently we don't have any plans regarding the CAS module. We have many modules in our agenda to be extracted out of the standard distribution and make them installable through composer. However, CAS is initially not included in that list. Considering the amount of work that could be for everybody to pull everything back into core, maybe it's better to make the fork installable with composer and just remove the original module. I see two clear advantages with that:

* people would be able to use a CAS module more up to date, with more functionalities and better in general, and
* there would be someone maintaining the module with knowledge of CAS and the module itself.

What do you all think?

ottenhoff:
My guess is that the core CAS module needs 3-4 pull requests for it to be in good shape.

I believe CAS remains an important legacy protocol for higher education.  I believe it's an important part of the value of simpleSamlPHP. 

I would be happy to install any non-core functionality via composer. I'm skeptical of maintaining personal CAS forks instead of helping to maintain one central CAS module. @shamus13 what do you think?

shamus13:
I sort of agree with both of you. I am not crazy about idea of merging the fork back into the main project only to split it out again later, but i will have a look at it, maybe it wont be so hard after all.

I am not in favor of personal forks either, so, if a merge proves impractical, i would suggest creating a simplesamlphp/simplesamlphp-module-casserver project, where we can get things sorted out, before removing the casserver module from core and abandoning my fork.

shamus13:
I seem to have managed the merge to some extent at least. We lost a bit of history locally, when the project was initially forked and later when switching version control system, so i ended up merging two repositories without a shared history.

I ended up using 'git filter-branch --tree-filter' to move files about and rename the module before pulling. The result is at https://github.com/shamus13/simplesamlphp. Any other ideas on how to do this are most welcome.

At any rate ill have to give the module a workout before creating a pull request.

jaimeperez:
Hi guys, sorry for the late response to this.

The problem with the CAS module (as with many others) is that there's no one here with knowledge enough to maintain it, so if we have someone who can take care of possible issues who actually knows what (s)he's doing, that would be the perfect scenario for us.

Now, CAS is quite used, but it's definitely not the main purpose for SimpleSAMLphp to integrate with it. I would say the vast majority of SSP users aren't using CAS now, nor they will in the future. So that's why I see it more or less reasonable to extract the module out of the main distribution, to be installed as an external module. In that case, it would be easier to update / fix / maintain the module, as it won't interfere with SSP development (nor will it depend on our current lack of resources).

I agree with both of you though on avoiding personal forks, specially for a module like this. Actually what I was thinking of was not to use directly @shamus13's module, but creating a new repo under simplesamlphp/ and allowing those with knowledge to manage that specific repository. I would be more than happy to have both of you as committers in there, if you agree :smile:

shamus13:
Hi Jaime. I would gladly step up as (co)maintainer, especially since i really doubt, it will require much work:-)  :smile:

jaimeperez:
That's great, thanks @shamus13! @ottenhoff, what do you think? Does my plan sound reasonable?

ottenhoff:
Yes, I agree with your plan.  I am happy to be a maintainer.

jaimeperez:
Hi @ottenhoff & @shamus13!

I've just added you both as collaborators in the [simplesamlphp/simplesamlphp-module-casserver](https://github.com/simplesamlphp/simplesamlphp-module-casserver) repository. You should now have write access there, so that you can contribute your work directly.

I'm closing this PR for now, since the *casserver* module is not part of the codebase any longer.

Thanks again for stepping up and your help!

jaimeperez:
Actually it still is part of SimpleSAMLphp, which reminds me I should remove it :blush:

PR #177, opened 2015-02-01
Fix Exception type and set timeout when fetching metadata

girgen:
If the AuthState is missing, usually because the users has saved a bookmark or in some other way is following a link that is half way in to the authentication process. Reading the code, it seems obvious that the author had the NoState exception in mind for this, but it was not used. Also, add some friendly error information.

The timeout in MetaLoader.php is important as well:

### DIFF
--- a/modules/core/dictionaries/no_state.translation.json
+++ b/modules/core/dictionaries/no_state.translation.json
@@ -83,6 +83,9 @@
 		"cs": "N\u00e1vrhy pro vy\u0159e\u0161en\u00ed tohoto probl\u00e9mu:",
 		"eu": "Arazo hau konpontzeko iradokizunak:"
 	},
+	"suggestion_badlink": {
+		"sv": "Kontrollera att länken du använde för att komma åt webplatsen är korrekt."
+	},
 	"suggestion_goback": {
 		"no": "G\u00e5 tilbake til forrige side og pr\u00f8v p\u00e5 nytt.",
 		"nn": "G\u00e5 tilbake til forrige side og pr\u00f8v p\u00e5 nytt.",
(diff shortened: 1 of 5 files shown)


mcouillard:
These look like excellent additions. We use FastCGI and timeouts do cause mayhem.

thijskh:
It's a bit confusing that this PR combines two issues in one. Re the metarefresh one: I think in SSP 1.13 metarefresh was changed to not drop the old metadata if refreshing it fails. So I'm not sure that "all metadata is wiped" should still be a problem with the current release. Can you confirm?

thijskh:
I was also wondering if 5 sec read timeout would be a good value to hardcode. On slower connections, 5 sec may not be enough to retrieve large sets like e.g. eduGAIN. Isn't it better to just defer to the PHP ini setting `default_socket_timeout`? If you require a different timeout you can already configure it there..

girgen:
Sorry about the combination of the two commits is my bad, it was not intentional. They should have been two different pull requests really.

About the timeout, yes, five seconds should of course not be hardcoded. default_socket_timeout is probably already used, that's why it is failing. I need to set it lower for this specific request, since all metadata fetching must be done within the timeout of one single fastcgi request.

I did not see the fix in 1.13, maybe that makes it better, at least not wiping all metadata anymore?

modules/metarefresh/lib/MetaLoader.php
```diff
@@ -159,7 +159,7 @@ private function createContext($source) {
 			}
 		}
 
-		return array('http' => array('header' => $rawheader));
+		return array('http' => array('header' => $rawheader, 'timeout' => 5));
```
jaimeperez: Should we hardcode this value here, use a constant, or allow configuring it?

modules/core/dictionaries/no_state.translation.json
```diff
@@ -83,6 +83,9 @@
 		"cs": "N\u00e1vrhy pro vy\u0159e\u0161en\u00ed tohoto probl\u00e9mu:",
 		"eu": "Arazo hau konpontzeko iradokizunak:"
 	},
+	"suggestion_badlink": {
```
jaimeperez: Translations are handled through the [translation portal](https://translation.rnd.feide.no/), not directly in the code, so you don't need to provide them here. We just have to update the strings in the portal so that you can translate them there, and afterwards we'll be able to generate updated translations to be shipped with SimpleSAMLphp.

jaimeperez:
Thanks a lot @girgen!

Would you be so kind to split this pull request into two, each of them addressing one individual problem? I've also commented a couple of things in the code. Please take a look at them.

Thanks again!

PR #178, opened 2015-02-10
Add documentation to sp-remote for signature.certificate

dtsosie:
Adds documentation on behavior of signature.certificate when signature.privatekey is specified for a service provider.

### DIFF
--- a/docs/simplesamlphp-reference-sp-remote.txt
+++ b/docs/simplesamlphp-reference-sp-remote.txt
@@ -257,6 +257,10 @@ The following SAML 2.0 options are available:
 :   Passphrase for the private key. Leave this option out if the private key is unencrypted.
 :   Note that this option only is used if `signature.privatekey` is present.
 
+`signature.certificate`
+:   Certificate file included by IdP for KeyInfo within the signature for the SP, in PEM format. The filename is relative to the cert/-directory.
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @dtsosie!

I might be missing something, but I don't think such an option exists on SimpleSAMLphp...

dtsosie:
modules/saml/lib/Message.php function addSign has this effect if the spMeta has signature.privatekey set. Finally got a service provider that cared if KeyInfo was returned and noticed this setting wasn't in spremote docs.

jaimeperez:
Oh, I see... It's a bit strange that this option has been there undocumented up until now. Before merging, I'm going to check why and see if we should evaluate the existence of that option, so that maybe we don't even need to document it.

Of course, if you are using it and find it useful, feedback is welcome :+1:

dtsosie:
Signature.privatekey Has its best purpose for staged idp key rollover.I use ssp as an idp for many mission critical systems that can't go down. Ssp didn't allow per service provider key changes prior. This setting allows a scheduling and phasing to occur per sp to minimize simultaneous disruption . Granted privatekey use, certificate is needed to compliment it for when x509 key info is required.

jaimeperez:
Ah, that's true, now I recall! Yep, I agree, this should be properly documented :+1:

PR #179, opened 2015-02-16
Add metadata signature verification by using pub key from certificate

s-hal:
Added the ability to verify the metadata signature with a certificate file. I added the functionality to both the cli and the automatic metarefresh module. It is possible to use more than one certificate to handle key rollover.

### DIFF
--- a/docs/simplesamlphp-automated_metadata.txt
+++ b/docs/simplesamlphp-automated_metadata.txt
@@ -75,7 +75,10 @@ Here's an example of a possible configuration for both the Kalmar Federation and
 				'sources'	=> array(
 					array(
 						'src' => 'https://kalmar.feide.no/simplesaml/module.php/aggregator/?id=kalmarcentral&mimetype=text/plain&exclude=norway',
-						'validateFingerprint' => '591d4b4670463eeda91fcc816dc0af2a092aa801',
+						'certificates' => array(
+							'current.crt',
+							'rollover.crt',
+						),
 						'template' => array(
 							'tags'	=> array('kalmar'),
 							'authproc' => array(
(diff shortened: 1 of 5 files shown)


jaimeperez:
Hi @s-hal, thanks for your contribution!

A couple of questions:

* I understand there's no validation of the certificate itself. So provided that the same public key is used to sign metadata, it doesn't really matter if the certificate does not validate or is expired, right?

* Should the two options, validate by fingerprint or by certificate, be exclusive? I don't know if it makes much sense to have both at the same time, so maybe certificate-based validation should take precedence?

s-hal:
•	That is correct, the certificate are not validated nor is the expiration date checked

•	I am pro that certificate-based verification should take precedence. I will make the changes.

jaimeperez:
Thanks a lot @s-hal!

I just noticed some detail I overlooked before, sorry. The module does not have its own documentation (and that's why I didn't notice before), but it's documented in `docs/simplesamlphp-automated_metadata.txt`. We should have the new option documented there. Would you mind adding a line or two explaining the new `certificates` option and that it takes precedence over the `validateFingerprint` one? Maybe we should even mark the latter as deprecated...

s-hal:
I have added the text about the new certificates option. Should validateFingerprint realy be deprecated at this stage? Why not keep both option?

trscavo:
This is an important new feature. Thanks for contributing this work.
Jumping in with a quick comment. Deprecating the fingerprint approach makes sense since it clearly communicates the superiority of the other approach. In the InCommon Federation, we've had to wrap our signing key with a new cert twice in the last five years (don't ask). Our Shibboleth deployments were guaranteed not break but every other deployment (including SSP deployments) either configured the new cert in a timely fashion or broke.

jaimeperez:
@s-hal, a huge thanks for your work with this!

I agree with Tom, even though I think configuring a certificate that we are later on ignoring (except for the public key) is sending the wrong message and creates confusion. In any case, if this is useful for some of our users, this is very much welcome. Using fingerprints is definitely more complicated than just dropping a cert in a directory, so I think we should favour the new option (we are already doing by giving it precedence), and if that's the case, deprecating fingerprints sends a clear message: _do not use fingerprints any longer, use certificates instead_.

Thanks a lot again!

s-hal:
I agree that it sends a clear messages, and since fingerprints are only deprecated not removed, it is probably the best solution. In the last commit, fingerprints where not deprecated. Should I create a new pull request?

I am only contributing, it is you guys doing the real work.

thijskh:
@s-hal As an aside: your commit timestamps are far into the future. You may want to check your system clock :)

s-hal:
Thank you for pointing that out. It was a VM that I used to test expired certificate. The VM has already been recycled.

PR #184, opened 2015-03-27
Disable login button with JS after first click.

thijskh:
Impatient users will press the button twice which leads (esp in Chrome it seems) to NOSTATE errors.

### DIFF
--- a/dictionaries/login.definition.json
+++ b/dictionaries/login.definition.json
@@ -11,6 +11,9 @@
 	"login_button": {
 		"en": "Login"
 	},
+	"processing": {
+		"en": "Processing..."
+	},
 	"username": {
 		"en": "Username"
 	},
--- a/modules/core/templates/loginuserpass.php
+++ b/modules/core/templates/loginuserpass.php
@@ -67,7 +67,7 @@ if ($this->data['rememberUsernameEnabled'] || $this->data['rememberMeEnabled']) 
 } else {
 	$text = $this->t('{login:login_button}');
 	echo str_repeat("\t", 4);
-	echo "<input type=\"submit\" tabindex=\"4\" id=\"regularsubmit\" value=\"{$text}\" />";
+	echo "<input onclick=\"this.value='" . $this->t('{login:processing}') . "';this.disabled=true;this.form.submit();return true;\" type=\"submit\" tabindex=\"4\" id=\"regularsubmit\" value=\"{$text}\" />";
 }
 ?>
 			</td>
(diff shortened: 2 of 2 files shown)


thijskh:
I'm told this is not only "impatient" users but also Windows victims who do not know when to click or when to double click.

jaimeperez:
Hi Thijs!

I certainly see the reason for this, but I'm a bit hesitant on the solution, specially if the problem is people double clicking on the login button. What if the hit the button faster than javascript is able to execute and disable the button? Shouldn't we try to find a way to fail more gracefully on the IdP?

thijskh:
I doubt that it is practically possible to click the second time before the JS has disabled the button.

I'm not sure about your idea to handle it more gracefully. The nostate error may be presented by the remote SP. Can you reliably prevent sending a duplicated request to them?

jaimeperez:
That's what I was trying to figure out...

Detecting a duplicated login should be relatively simple. After all, you already have a full session by then. But in that case, you could skip sending a SAML response to the SP. The problem in that case is that whatever you do, the user will be stuck in the IdP, since the browser will keep the response to the second click and disregard the first one. I could imagine some dirty trick like switching to an IdP-initiated login then, but sounds like a bit overkilling.

By the way, I'm trying to reproduce the problem, and I'm failing miserably both with Firefox and Chrome. Have you been able to reproduce it?

thijskh:
It's harder to reproduce if you run your IdP locally or on a dedicated system since it may then be too fast and the request is already sent out before you get the chance to click twice. What happens is when I do get this reproduced in Chrome is that it represents the login form, but entering the details again there will after submit trigger a nostate error in the remote SP. It may help to add a sleep to your loginuserpass.php.

I like my fix since it tackles the problem at the source: it prevents duplicate submissions wholesale (the browser will definitely process the JS event before it processes the next click event so it's guaranteed to solve the issue). This in contrast of trying to work it around by detecting this behaviour and then devising a strategy to somehow reconcile these requests...

jaimeperez:
Sorry for the delay @thijskh!

I've been considering it further and I think you are right. After looking into the patch more deeply, it's pretty clear that it won't be possible to click the submit button twice now. And this is probably the cleanest and simplest solution.

**But** I have one concern. What happens with browsers without javascript, or with javascript disabled? Will they submit anyway, ignoring the *onclick*? Should we have an alternative submit button without the *onclick* event inside a `noscript`?

thijskh:
If there's no JavaScript the form continues to work as it did before. JS is needed to achieve the disabling the button, if there's no JS, the button will not be disabled. The onclick attribute is plainly ignored.

I've verified that this is indeed the case.

jaimeperez:
That was what I was expecting, but just wanted to confirm. I'm merging then!

PR #187, opened 2015-04-01
LDAP connection improvements

mschwager:
Hey guys!

I've found it useful to be able to specify the LDAP port I'm connecting to from configuration. It's possible to listen for connections on a non-standard port (389).

I also found a bug in the way you're handling encrypted LDAP connections. It appears you're trying to make STARTTLS connections, which it's worth noting that this is different from LDAPS (https://en.wikipedia.org/wiki/Lightweight_Directory_Access_Protocol#StartTLS). The problem occurs when you check if `ldaps` is in the hostname. When `ldaps` is in the hostname the `ldap_connect` call uses the standard LDAPS port of 686. The problem is that STARTTLS uses the standard `ldap` scheme and defaults to port 389. So the code is checking for `ldaps`, implying the LDAPS protocol (port 686), and then trying to perform a STARTTLS connection (port 389).

Overall, I think it's better to explicitly specify the port instead of relying on `ldap_connect` to choose a port for us based on the existence of a substring in the hostname.

Let me know if you have any feedback!


mschwager:
I think I initially misread the code, and totally missed the `!` in front of the `preg_match` call. So that check was correct after all. Sorry about that. However, I do think switching the call to `strpos` will offer speed and readability enhancements: https://php.net/manual/en/function.preg-match.php#refsect1-function.preg-match-notes.

thijskh:
Thanks for the PR! The port thing looks good.
About the STARTTLS change, the regular expression was case instensitive, I'm not sure how many people will actually depend on being able to specify `LDAPS:` but to be sure you could change it to a `stripos`. It may make sense to rebase the PR so we end up with two commits (one for the port, one for the starttls check).

mschwager:
Ok, this PR is getting increasingly ridiculous as I try to rebase it.

PR #190, opened 2015-04-09
add attributes from skolfederation.se/SIS

jschlyter:
three attributes used by skolfederation.se (defined by TK 450 AG04: Attribut).
perhaps they should be defined in more files, please let me know.


jaimeperez:
Tack @jschlyter!

In general, I would say the definitions should be at least in the following attribute maps:

* `name2oid`
* `name2urn`
* `oid2name`
* `oid2urn`
* `urn2name`
* `urn2oid`

Also, note that even though attributes are sorted alphabetically (or by number, if the key of the map is an OID), attributes are grouped by their origin. So the *schac* attributes are all together, the same as *eduPerson* or *norEdu*.

jaimeperez:
Merged, tack så mycket! :smile:

PR #191, opened 2015-04-22
fix SimpleSAML\Utils\getBaseURL return value

enricocavalli:
getServerPort does not return “:portnumber” which is expected by
getBaseURL to construct a correct redirection url

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -444,7 +444,7 @@ class HTTP
             $protocol .= '://';
 
             $hostname = self::getServerHost();
-            $port = self::getServerPort();
+            $port = ':'.self::getServerPort();
             $path = '/'.$globalConfig->getBaseURL();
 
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @enricocavalli, thanks for your pull request!

The bug is actually different. You cannot unconditionally prepend a colon to what getServerPort() returns for two reasons:

* If it returns a port number, it will be already prepended by a colon, and
* if the port number is the standard one for the protocol in use (that being, HTTP or HTTPS), then it will return an empty string.

So your patch would render URLs like `http://www.example.com::123/path` or `https://www.example.com:/path`.

The problem here is that the rewritten function was returning '80' and '443' (without the colons) for the new standard ports, where it should return an empty string. Commit 8078b2094d780ae309b331bc0dec7c8b68b7f5c8 should address that issue.

enricocavalli:
thank you ! I did not think in general terms ....

jaimeperez:
Don't worry. You raised my attention on a bug that would have been unnoticed otherwise :smile:

Issue #193, opened 2015-04-23
`cert` directory mentioned in documentation does not exist in default installation


j13k:
The `cert` directory referred to in [section 1.1](https://simplesamlphp.org/docs/stable/simplesamlphp-sp#section_1_1) of the documentation does not exist in the default source tree, and it's not clear how to proceed:
* should the user create the directory themselves?
* where should the directory be located?

It would be a good idea to create this directory in the correct place of the default distribution to save confusion.

PR #194, opened 2015-04-23
Created default 'cert' directory as suggested in issue #193

j13k:
A `cert` directory placeholder file to help avoid confusion described in issue #193.

### DIFF
--- a/cert/.gitignore
+++ b/cert/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore


jaimeperez:
Hi @j13k! Thanks for the PR!

While I understand the issue here, I see little value on having the `cert` directory in the repository with just a placeholder file. In the end, it's exactly the same as with the `config` and `metadata` directories.

I think the proper solution to this would be to modify the `bin/build-release.sh` so that the `cert` directory is always created when we are creating a new release of the software.

PR #196, opened 2015-04-28
remove misleading and unused config option

dancsa:
this admins array is not used anywhere,
but it may be misleading to administrators who wants to restrict
the access to the metadata registry.

Please consider removing this configure options.


jaimeperez:
Thanks for your PR @dancsa!

As far as I can see, it's not used anywhere, so you are right, we should remove it. I'm merging right away :smile:

PR #198, opened 2015-04-30
Support Doctrine/DBAL store

sgomez:
This is a new Store to support Doctrine/DBAL support. I did it to improve my oauth2 module.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "php": "~5.3",
         "simplesamlphp/saml2": "~0.3",
         "simplesamlphp/xmlseclibs": "~1.3.2",
-        "whitehat101/apr1-md5": "~1.0"
+        "whitehat101/apr1-md5": "~1.0",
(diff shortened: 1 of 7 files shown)


thijskh:
Otherwise it looks like a worthwhile addition.

sgomez:
Well, IMHO the are some reasons:

* Using DBAL is more safe when you are building sql sentences. The QueryBuilder API is more intuitive and safe than write the sql directly.
* DBAL have automated type conversions.
* The actual SQL Store has a "TableVersionTable" than I guess is used to update the schema on the future. The problem with that is you need to write all the sql sentences. With DBAL is easier migrate the schema to a new configuration.

Anyway, the problem is that SimpleSAML_Store configure the stores with a switch-case, so I can't change the store with other implementation. So maybe the alternative could be to improve it to let inject new Store classes and then i could have this code in an external module.

sgomez:
Sorry. I just saw I can create a module for this. I am working on it.

Issue #50, opened 2014-02-27
Implement SAML2 sender-vouches subject confirmation method (followup to #333) (was #600)


sspghost:
_From [fefefan...@googlemail.com](https://code.google.com/u/109263125316745138061/) on November 27, 2013 09:53:46_

What steps will reproduce the problem? 

1. Configure an IdP which uses the 'urn:oasis:names:tc:SAML:2.0:cm:sender-vouches' SubjectConfirmation-Method

2. Use the Authentication test page

I would expect to be authenticated, but get an exception:

>Backtrace:
>0 /var/simplesamlphp/www/module.php:180 (N/A)
>Caused by: SimpleSAML_Error_Exception: Error validating SubjectConfirmation in Assertion: Invalid Method on SubjectConfirmation: 'urn:oasis:names:tc:SAML:2.0:cm:sender-vouches'
>Backtrace:
>3 /var/simplesamlphp/modules/saml/lib/Message.php:694 (sspmod_saml_Message::processAssertion)
>2 /var/simplesamlphp/modules/saml/lib/Message.php:528 (sspmod_saml_Message::processResponse)
>1 /var/simplesamlphp/modules/saml/www/sp/saml2-acs.php:81 (require)
>0 /var/simplesamlphp/www/module.php:135 (N/A) What version of the product are you using? On what >operating system? 1.11.0 on Ubuntu 13.10.

This is a follow-up for Issue #333 .

_Original issue: http://code.google.com/p/simplesamlphp/issues/detail?id=600_

brianV:
I've taken a stab at this. See the two pull requests referenced above (https://github.com/simplesamlphp/saml2/issues/41 and https://github.com/simplesamlphp/simplesamlphp/issues/200)

jaimeperez:
Thanks @brianV for your pull requests! :smile:

PR #200, opened 2015-05-11
Add support for sender-vouches

brianV:
This patch adds support for the sender-vouches SubjectConfirmation Method, requested in issue #50. 

It also depends on the constant added in https://github.com/simplesamlphp/saml2/pull/41

### DIFF
--- a/modules/saml/lib/Message.php
+++ b/modules/saml/lib/Message.php
@@ -587,8 +587,9 @@ class sspmod_saml_Message {
 
 		$found = FALSE;
 		$lastError = 'No SubjectConfirmation element in Subject.';
+		$validSCMethods = array(SAML2_Const::CM_BEARER, SAML2_Const::CM_HOK, SAML2_Const::CM_VOUCHES);
 		foreach ($assertion->getSubjectConfirmation() as $sc) {
-			if ($sc->Method !== SAML2_Const::CM_BEARER && $sc->Method !== SAML2_Const::CM_HOK && $sc->Method !== SAML2_Const::CM_VOUCHES) {
+		    if (!in_array($sc->Method, $validSCMethods)) {
 				$lastError = 'Invalid Method on SubjectConfirmation: ' . var_export($sc->Method, TRUE);
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


thijskh:
Looks good to me

relaxnow:
```!in_array()``` might've been more readable but :+1: for the PR as it is.
@jaimeperez what do you think?

jaimeperez:
I agree with you, both that it looks good and that it would be more readable using `in_array()`.

There's one thing that worries me though. The new constant is already in place in the SAML2 library. However, we need to bump the version of the SAML2 library we are using in SimpleSAMLphp to avoid PHP errors when the interpreter evaluates the condition and finds a constant that's not defined. Currently, SimpleSAMLphp requires version `0.3` of the SAML2 library, which is more than a year old now, and of course a lot has changed there in the meantime.

Are we comfortable with the current status of the SAML2 library? Is it tested enough to start using the latest version in SimpleSAMLphp?

brianV:
Let's make it unanimous - it is more readable using `!in_array()`.

I've added a commit that should do it.

relaxnow:
@jaimeperez at SURF we've been using it for lots of purposes (including 2 production SAML2 gateways) and it's perfectly stable. In fact to better match the semver spec I bumped the version numbers to 1.x (we are currently on v1.5.3, did you create that tag?).
You've seen and :+1:ed every change that happened and no bwc breaks occurred. Only awesome new stuff :smile:.

relaxnow:
@brianV thanks!

jaimeperez:
Wow, that was fast @brianV! Thanks a lot!

@relaxnow I'm guessing that it must be pretty stable, but I need to be the paranoid one here :blush: Oh, and I just created 1.5.3, yep! @thijskh, what you think? Should we bump the version we are using in SSP? In that case, I would first do that, so that we ensure the constant is in place, then merge this PR.

thijskh:
I see no problem at all in bumping the version in SSP, for the reasons Boy outlined.

jaimeperez:
Good! I trust you guys, so I just bumped the version and merged this PR. Thanks again @brianV!

Issue #49, opened 2014-02-27
no support for attribute filter rules for entity-category or registrationauthority (was #591)


sspghost:
_From [schofi...@terena.org](https://code.google.com/u/103727203147080158829/) on November 06, 2013 10:28:00_

What steps will reproduce the problem? 1. use of core:AttributeLimit can only match requested attributes - doesn't support required (isRequire="true") or the flexibility of entity-category.
2. attached patches get us a lot closer to that mark :-)
3. clearly looking forward to a range of feedback on this patch
4. this is not meant to be feature complete with implementations in parallel products but does cover many of the general purpose cases. What is the expected output? What do you see instead? the attached patches allow you to specify filter rules within config-attributelimit.php that match on EntityID, RegistrationAuthority and/or EntityCategory. This does NOT support arbitrary matching on EntityAttributes. What version of the product are you using? On what operating system? r3288 Please provide any additional information below. The patches relate to:
- modules/core/lib/Auth/Process/AttributeLimit.php
- config-templates/config-attributelimit.php
  an example config-templates file is attached (config-attributelimit.php)
- lib/SimpleSAML/Metadata/SAMLParser.php
- modules/saml/www/sp/metadata.php

_Original issue: http://code.google.com/p/simplesamlphp/issues/detail?id=591_

jaimeperez:
Hi!

Just a quick update on this. A [new module](https://github.com/simplesamlphp/simplesamlphp-module-entitycategories) adding support for entity categories in general and *Research & Scholarship* and eduGAIN's *Code of Conduct* in particular is on the way 😉

thijskh:
@jaimeperez @br00k Can you make explicit in this issue what needs to be done in SimpleSAMLphp core now that there is a new module? Or perhaps it makes sense to close this long winding issue and open (a) new one(s) for concrete remaining action items?

jaimeperez:
I think it makes more sense to just close this issue and move all development efforts to the new module, @thijskh. Thanks for bringing this up!

PR #206, opened 2015-05-26
Improve the local generation of metadata to support registrationAutho…

br00k:
…rity

Useful in the leadup to entity category support to publish
registrationAuthority (which might be overridden by a federation) and
attributes.required.

This was half of #49 and should be unbundled from that request.

### DIFF
--- a/lib/SimpleSAML/Metadata/SAMLParser.php
+++ b/lib/SimpleSAML/Metadata/SAMLParser.php
@@ -478,6 +478,9 @@ class SimpleSAML_Metadata_SAMLParser {
 		if (array_key_exists('attributes', $spd)) {
 			$ret['attributes'] = $spd['attributes'];
 		}
+		if (array_key_exists('attributes.required', $spd)) {
+			$ret['attributes.required'] = $spd['attributes.required'];
+		}
 		if (array_key_exists('attributes.NameFormat', $spd)) {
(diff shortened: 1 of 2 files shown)


lib/SimpleSAML/Metadata/SAMLParser.php
```diff
@@ -478,6 +484,12 @@ public function getMetadata1xSP() {
 		if (array_key_exists('attributes', $spd)) {
 			$ret['attributes'] = $spd['attributes'];
 		}
+		if (array_key_exists('attributes.required', $spd)) {
+			$ret['attributes.required'] = $spd['attributes.required'];
+		}
+		if (array_key_exists('attributes.required', $spd)) {
+			$ret['attributes.required'] = $spd['attributes.required'];
+		}
```
thijskh: This block is duplicated?

jaimeperez:
Just for the record, so that Brook can complain about me denying his pull requests (:wink:) , support for `registrationAuthority` (and *RegistrationInfo* in general as an extension) is already in place.

thijskh:
Given the back-and-forths it may be useful to squash the commits before merging.

thijskh:
:+1:

jaimeperez:
Merged! Just because it was you, Brook :wink:

PR #207, opened 2015-05-26
A-Select: always add uid+organization to attributes

praseodym:
uid and organization attributes are not always in `$creds['attributes']`, so we add an option for that.


thijskh:
It has been refactored a bit after discussion on IRC between @praseodym, @wsldankers and me. Looks fine now, thanks!

PR #208, opened 2015-05-27
extend authX509 module with authentication processing filter for generating certificate expiry warnings

joostd:
In some simplesamlphp deployments there is a chicken&egg situation where users authenticate at their IDP using a X509 certificate, and apply for certificates at an SP using federated login. When a certificate expires, they are immediately locked out of their account, and are unable to fix this situation. This filter will warn users if their certificate is about to expire, so they can renew their certificate in time.

### DIFF
--- a/modules/authX509/dictionaries/X509warning.translation.json
+++ b/modules/authX509/dictionaries/X509warning.translation.json
@@ -1,0 +1,11 @@
+{
+	"warning": {
+		"nl": "Je certificaat verloopt over %days% dagen. Vervang tijdig je certificaat."
+	},
+	"warning_header": {
+		"nl": "Je certificaat verloopt binnenkort."
+	},
+	"proceed": {
+		"nl": "Verder"
+	}
+}
--- a/modules/authX509/dictionaries/X509warning.translation.json
+++ b/modules/authX509/dictionaries/X509warning.translation.json
@@ -1,11 +1,32 @@
 {
-	"warning": {
-		"nl": "Je certificaat verloopt over %days% dagen. Vervang tijdig je certificaat."
-	},
-	"warning_header": {
-		"nl": "Je certificaat verloopt binnenkort."
-	},
-	"proceed": {
-		"nl": "Verder"
-	}
+  "warning": {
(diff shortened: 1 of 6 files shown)


modules/authX509/dictionaries/X509warning.translation.json
```diff
@@ -0,0 +1,32 @@
...
+    "es": "Su certificado está a punto de caducar."
+  },
+  "renew": {
+    "nl": "Vervang tijdig je certificaat.",
+    "no": "Vennligst forny sertifikatet ditt før det utløper.",
+    "da": "Forny venligst dit certifikat i tide.",
+    "es": "Por favor, renueve su certificado a tiempo."
+  },
+  "renew_url": {
+    "nl": "<a href='%renewurl%'>Vernieuw<\/a> tijdig je certificate.",
```
thijskh: `certificate` has to be `certificaat` & it says "Vernieuw" while it says "Vervang" above.

modules/authX509/lib/Auth/Process/ExpiryWarning.php
```diff
@@ -0,0 +1,93 @@
...
+        }
+        $validTo = $client_cert_data['validTo_time_t'];
+        $now = time();
+        $daysleft = (int)(($validTo - $now) / (24*60*60));
+        if ($daysleft > $this->warndaysbefore) {
+            /* We have a certificate that will be valid for some time. Skip the warning. */
+            return;
+        }
+
+        SimpleSAML_Logger::warning('authX509: user certificate expires in ' . $daysleft . ' days');
```
thijskh: Maybe not log this at level 'warning', since to the administrator of the IdP it should not be a warning but in fact normal behaviour. Info-level seems more appropriate.

joostd: Not sure about the Info-level - technically, this doesn't clarify as a warning, but the warning is only logged when the certificate is about to expire. Sysadmins could anticipate expiry and urge users to renew (without having to lower the logging threshold to INFO)

PR #215, opened 2015-06-03
Corrected logic in checkSessionCookie

enricocavalli:
Raise InvalidArgumentException if $retryURL is not a string AND is not null


jaimeperez:
Merged! Thanks Enrico!

PR #217, opened 2015-06-04
do not report success if the metaedit save was unsuccessful

dancsa:
hi,

Metaedit reports "Metadata successfully saved" in all case even if errors occurred during the save.

The saving goes through MetaDataStorageHandlerSerialize.php.  If the directory does not exist yet, the handler tries to create it first. if something goes wrong, logs error and returns with FALSE.
The caller in modules/metaedt/www/edit.php doesn't check the return value from saveMetadata(), then reports successful saving.

### DIFF
--- a/modules/metaedit/www/edit.php
+++ b/modules/metaedit/www/edit.php
@@ -66,7 +66,9 @@ if (isset($_POST['submit'])) {
 	} catch(Exception $e) {}
 	if ($testmetadata) requireOwnership($testmetadata, $userid);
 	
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @dancsa! Thanks a lot for your PR!

The metaedit module was moved to [its own repository](https://github.com/simplesamlphp/simplesamlphp-module-metaedit), so that it can be installed by those who really need it. Would you be so kind to apply your PR there?

Thanks!

PR #219, opened 2015-06-09
Use $metadata array for example IDP and SP metadata

peter-:
As per https://groups.google.com/forum/#!topic/simplesamlphp/nFXerlX839E


thijskh:
Agreed, probably they were intended to be 'snippets' to be put into an existing `$metadata = array(`... block, but the changed notation matches what is done in the template configuration files.

PR #221, opened 2015-06-10
Moved logic pertaining to finding configdir to central location, added ability to set via environment variable

tldev:
There is a similar pull request for this feature (I even borrowed the environment variable), but I wanted to try a different approach that makes this a little easier to understand.  The logic for finding the config directory is centralized.

I can make the necessary changes for the manual, but I wanted to put in the pull request and have someone look at it first before I finish it off.  If it looks good I'll finish it off.

For the bin files, the environment variable can be passed via cli like so:

```
SIMPLESAMLPHP_CONFIG_DIR=/home/user/samlconfig php modules/statistics/bin/loganalyzer.php
```

### DIFF
--- a/lib/SimpleSAML/Utils/Config.php
+++ b/lib/SimpleSAML/Utils/Config.php
@@ -55,4 +55,32 @@ class Config
 
         return $secretSalt;
     }
+
+    /**
+     * Returns the path to the config dir
+     *
+     * If the SIMPLSAMLPHP_CONFIG_DIR environment variable has been set, it takes precedence
+     * over the default $simplesamldir/config directory.
+     *
+     * @return string
+     */
+    public static function getConfigDir()
+    {
+        $configDir    = dirname(dirname(dirname(__DIR__))) . '/config';
+        $configDirEnv = getenv('SIMPLESAMLPHP_CONFIG_DIR');
+        if ($configDirEnv !== false) {
+            if (!is_dir($configDirEnv)) {
+                throw new \InvalidArgumentException(
+                    sprintf(
+                        'Config directory specified by environment variable SIMPLESAMLPHP_CONFIG_DIR is not a ' .
+                        'directory.  Given: "%s"',
+                        $configDirEnv
+                    )
+                );
+            }
+            $configDir = $configDirEnv;
+        }
+
+        return $configDir;
+    }
 }
--- a/bin/memcacheSync.php
+++ b/bin/memcacheSync.php
@@ -24,7 +24,7 @@ $baseDir = dirname(dirname(__FILE__));
 require_once($baseDir . '/lib/_autoload.php');
 
 /* Initialize the configuration. */
-$configdir = $baseDir . '/config';
+$configdir = SimpleSAML\Utils\Config::getConfigDir();
 SimpleSAML_Configuration::setConfigDir($configdir);
 
 /* Things we should warn the user about. */
--- a/lib/SimpleSAML/Configuration.php
+++ b/lib/SimpleSAML/Configuration.php
@@ -144,7 +144,7 @@ class SimpleSAML_Configuration {
 			if ($configSet !== 'simplesaml') {
 				throw new Exception('Configuration set \'' . $configSet . '\' not initialized.');
 			} else {
-				self::$configDirs['simplesaml'] = dirname(dirname(dirname(__FILE__))) . '/config';
+				self::$configDirs['simplesaml'] = SimpleSAML\Utils\Config::getConfigDir();
 			}
 		}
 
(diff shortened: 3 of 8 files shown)


lib/SimpleSAML/Utils/Config.php
```diff
@@ -55,4 +55,21 @@ public static function getSecretSalt()
...
+     * Returns the path to the config dir
+     * If the SIMPLSAMLPHP_CONFIG_DIR environment variable has been set, it takes precedence
+     * over the default $simplesamldir/config directory.
+     *
+     * @return string
+     */
+    public static function getConfigDir()
+    {
+        $configDir = getenv('SIMPLESAMLPHP_CONFIG_DIR');
+        if ($configDir === false || !is_dir($configDir)) {
```
thijskh: If `SIMPLESAMLPHP_CONFIG_DIR` is set to some path which does not exist, it will fall back to another config dir silently. This seems like a source of big confusion. Maybe better that if someone sets that env and it's invalid, to throw an error?

thijskh:
I think it's a nice and simple approach. What do you think, @jaimeperez?

Besides my comment above, it would be nice if you could also include a test for the new method. We're trying to (at the very least) not reduce test coverage in merging pull requests. Indeed some documentation would be good; and as a nitpick: you could leave out all the whitespace changes from your commit (e.g. by using `git add -p` to add only the relevant chunks.)

tom-1bg:
@thijskh I agreed with your comment on throwing an exception and made the necessary change.  I also removed my whitespace changes.  (I'm thinking I'll clean up this whole repo's whitespace)  The project itself could use a linter for pull requests.

tldev:
@thijskh I agreed with your comment on throwing an exception and made the necessary change. I also removed my whitespace changes. (I'm thinking I'll clean up this whole repo's whitespace) The project itself could use a linter for pull requests.

I will wait on tests until I hear this is a merge-able solution.

jaimeperez:
Hi @tldev!

Thanks a lot for your PR, and sorry so much for the long time it took me to get my hands on this.

I've been taking a look into the code and it looks pretty good, and it's definitely a nice feature that many people is demanding. Since this would be such a critical code, I think tests in this case are a must.

Apart from that, I'm absolutely positive about merging this :+1:

tldev:
Hey @jaimeperez 

We should be good to go.  Let me know if you need anything else.

Thanks.

jaimeperez:
Thanks a lot @tldev!

I've just pulled your branch to test it out, and I noticed that when the path specified in the environment variable is not a directory, an exception is thrown. This makes sense. However, since nothing is capturing it, it ends up displayed in the browser. At that point we don't have any configuration, so we cannot offer a nice page to display the exception, so it is a very ugly message difficult to read. Wouldn't it be better to display a simple, illustrative message, and abort execution?

Also, would it be too much to ask you to document this in the installation instructions? This is a feature many people is willing to use, but it won't be really helpful if we don't document it properly.

dialogik:
How do I use `SimpleSAML_Configuration::setConfigDir()` (also asking @tldev)? I tried like this

    SimpleSAML_Configuration::setConfigDir('my-config');
    $auth = new SimpleSAML_Auth_Simple('my-auth');

and I also tried using `putenv()`

    putenv('SIMPLESAMLPHP_CONFIG_DIR=my-config');
    $auth = new SimpleSAML_Auth_Simple('my-auth');

but get this error

> You have not yet created the simpleSAMLphp configuration files.

while `$configdir` in `www/_include.php` is pointing to the `config/` in `/vendor/simplesamlphp/simplesamlphp/`, but should be `/config/`.

thijskh:
@dialogik Just to verify: you did apply this PR to your local installation?

dialogik:
Yes, I did. This is my `composer.json`

    ...
    "repositories":
    [
            {
                "type": "vcs",
                "url": "https://github.com/tldev/simplesamlphp"
            }
    ],
    "require": {
        ...
        "simplesamlphp/simplesamlphp": "dev-master"
    }

and this is my `Config::getConfigDir()`

    public static function getConfigDir()
    {
        $configDir    = dirname(dirname(dirname(__DIR__))) . '/config';
        $configDirEnv = getenv('SIMPLESAMLPHP_CONFIG_DIR');
        if ($configDirEnv !== false) {
            if (!is_dir($configDirEnv)) {
                throw new \InvalidArgumentException(
                    sprintf(
                        'Config directory specified by environment variable SIMPLESAMLPHP_CONFIG_DIR is not a ' .
                        'directory.  Given: "%s"',
                        $configDirEnv
                    )
                );
            }
            $configDir = $configDirEnv;
        }

        return $configDir;
    }

dialogik:
Setting the config directory via `SimpleSAML_Configuration::setConfigDir()` is not possible at all because it is simply not handled in `SimpleSAML\Utils\Config::getConfigDir()`. When implementing, one should be aware of the priority in case it is set via both, the `setConfigDir()` and `putenv()` variants.

And the problem with `putenv()` seems to be as follows: I declare the environment variable in my `<root>/index.php` but the problem is, that –if auth is required– I'm being redirected to `simplesamlphp/www/module.php/saml/disco.php?entityID=<entity.php>` and as soon as `module.php` is processed, my `putenv()` setting is gone.

jaimeperez:
Hi,

I took the liberty of fixing the conflicts and documenting the use of this feature in the installation guide. It is now merged and available in master.

Thanks for your work @tldev!

dialogik:
Great! Thanks to you both, @jaimeperez and @tldev.

PR #223, opened 2015-06-29
Some gitignore goodness

mapsi:


jaimeperez:
Thanks @mapsi!

I've merged your PR with some modifications to include some more directories that should be ignored, and to remove the `.gitignore` file from the `log` directory, which is now also gone.

PR #224, opened 2015-07-03
Updated newSessionId function to allow for just the session id to be …

peternolland:
…regenerated and not destroy the whole session.

session_regenerate_id(TRUE) is avaliable in PHP 4 >= 4.3.2 and PHP 5, this way the session is not destroyed when the id is regenerated, while still aiding in the prevention of session fixation and session hijacking attacks. 

This allows our system to set some session variables before the login attempt that are needed for when the user is redirected back.

### DIFF
--- a/lib/SimpleSAML/SessionHandlerPHP.php
+++ b/lib/SimpleSAML/SessionHandlerPHP.php
@@ -77,18 +77,22 @@ class SimpleSAML_SessionHandlerPHP extends SimpleSAML_SessionHandler {
 		}
 
 		/* Generate new (secure) session id. */
-		$sessionId = SimpleSAML_Utilities::stringToHex(SimpleSAML_Utilities::generateRandomBytes(16));
-		SimpleSAML_Session::createSession($sessionId);
 
-		if (session_id() !== '') {
-			/* Session already started, close it. */
-			session_write_close();
-		}
+		/**
+		*
+		* Regenerate the session id instead of destroying the current session.
+		* This will keep the current session data intact once the new session id
+		* has been generated.
+		*
+		*/
+		$old_session_id = session_id();
+		
+		// regenerates the session id, set to TRUE to delete the old associated files (set to false to keep)
+        session_regenerate_id(TRUE);
+        $new_session_id = session_id();
 
-		session_id($sessionId);
-		session_start();
-
-		return session_id();
+        // returns the new session
+		return $new_session_id;
 	}
 
 


jaimeperez:
Hi @peternolland!

Thanks a lot for your PR.

I don't see any difference, beyond what you are suggesting to be able to retain data stored in `$_SESSION` after regenerating the session identifier. I'm also unable to see what your use case is to need to store anything in `$_SESSION`, so I'm a bit hesitant on this. Could you please elaborate what you need to store in there, and why you need it?

Thanks again!

peternolland:
Hi @jaimeperez 

Thanks for your response,

In the system I'm working on we store information about the user, for instance the entry asset, entry url and visited assets to allow us to not only track the user from the beginning of their session but also once they've logged in be able to matchup what they've done before with what they are doing after. 

Our asset that interfaces with simplesamlphp also has functionality to use these session variables to redirect the user after the login has occurred for instance back the the entry asset for example, the user entered on the homepage, then clicked to another page which redirected to the login, then controlling the redirect after the authentication has passed the user back to say the entry asset of the session.

Hope this helps please contact me if you need anything else

jaimeperez:
Hi!

I'm afraid I'm a bit confused here. Keeping track of what the user was doing does only make sense in the SP, since in the IdP there's only... well, the IdP itself. Different domains, different machines probably, so I'm assuming you are talking about establishing a session in the SP.

But SimpleSAMLphp already takes care of all that for you. See [the documentation on the SP API](https://simplesamlphp.org/docs/stable/simplesamlphp-sp-api). The user will return automatically to the current page, or you can tell SSP to return to a different place if you like. Even if authentication is triggered while processing a POST request, you can ask SimpleSAMLphp to keep that request for you and recreate it after authentication.

In any case, if you really need something more complex than just having the user land in the original page that triggered authentication (bear in mind that changing this behaviour could be a nightmare in terms of usability), that's the responsibility of the end application, not SSP. If you are using SSP's session to store application data, that's definitely not the way you should do it.

peternolland:
Sorry my wording isn't the best at times.

So basically we make a cms that is using simplesamlphp as a SP, when we initiate the library in our system the session started by simplesamlphp wipes our existing session completely, i can see reasons for wiping the session, but the same result can be achieved with a lot less of an impact to the SP if the sessionid is just regenerated.

We are only storing basic info in the session,
- Entry asset
- Entry URL
- Visited assets

It is these that our application processes in a lot of functions that are common to a lot of the asset types we have.

jaimeperez:
Hi!

What you are describing is not the normal behaviour and should never happen. SimpleSAMLphp should not destroy the session of the application. If it does, that's probably due to a misconfiguration. I assume you are using the `phpsession` session handler. Make sure that the option 'session.phpsession.cookiename` is set in your `config/config.php`, and its value is different to whatever you are using in your CMS.

Just to be clear, I'm not against your patch, but I think you should solve your misconfiguration first. If you solve it by applying this patch it might work for you, but it could also create undefined behaviour for other apps.

peternolland:
Hi sorry for the delayed response, I can confirm that both session.phpsession.cookiename set in my config/config.php, is different to the value in my CMS.

jaimeperez:
Hi @peternolland! Sorry so much for the delay!

I think you are basically hitting a bug, as described in #349. Sessions existing **before** SimpleSAMLphp session is initialized/retrieved should be untouched, but the `phpsession` session handler is apparently wiping them, as you were describing. I added a patch in #349 that should fix the bug, enabling you to retrieve your session untouched **after** using SimpleSAMLphp.

Would you mind applying it and telling me if it solves your issue?

Thanks!

jaimeperez:
Hi again! I've just committed 8dc545b87cc1e6ba6f987b2141cc458f8616105e, closing this issue. There were a few bugs in SimpleSAMLphp's PHP session handler that should now be fixed. In any case, due to the nature of PHP's session handling, it's impossible to manage two sessions at the same time, so a method has been added to `SimpleSAML_Session` in order to restore a session existing previously to invoking SimpleSAMLphp. Refer to the commit for more information.

jaimeperez:
Just for completeness, here's a simple PHP script that can be used to verify that this fix works (as well as illustrate how to restore the previous session).

```php
<?php

// configure according to your needs
$path = '/path/to/your/script.php';
$host = 'your.machine.com';
$secure = true || false;
$session_name = 'name_of_the_cookie',
$ssproot = '/path/to/simplesamlphp';

// create our own session
session_set_cookie_params(0, $path, $host, $secure);
session_name($session_name);
session_start();
$id = session_id();

// if our session was not initialized or there's no attributes, require auth
if (!array_key_exists('foo', $_SESSION) || !array_key_exists('attributes', $_SESSION)) {
    $_SESSION['foo'] = 'bar';
    include $ssproot.'www/_include.php';
    $auth = new SimpleSAML_Auth_Simple('default-sp');
    $auth->requireAuth();

    // restore our own session
    $session = SimpleSAML_Session::getSessionFromRequest();
    $session->cleanup();

    // save the attributes to the session
    $_SESSION['attributes'] = $attributes;

    // test that SSP's session is still working
    $session->doLogout("default-sp");
}

// print details about *our* session
header("Content-Type: text/plain; charset=utf-8");
echo "Session Name: ".session_name()."\n\n";
echo "Session ID: ".session_id()."\n";
echo "\nSession contents: \n".var_export($_SESSION, true);
```

PR #233, opened 2015-07-28
ECP support to simplesamlphp

sitya:
This It is a reorganized and developed version of (abandoned?) https://github.com/bkryza/simplesamlphp/commits/ecp repository. It requires https://github.com/simplesamlphp/saml2/pull/48 PR to be part of simplesamlphp/saml2. 

I have tested with different ECP clients and it was working properly with all of them.


jaimeperez:
Hi @sitya!

Finally I got my hands on this, and I've been able to get it working with your ECP client and an IdP instance that I have deployed locally. I still need to go through the specification and make sure that your code makes us comply with all the requirements needed to claim support for the ECP profile.

There's also quite a few problems with the code, unfortunately. I'll comment directly in the code, if you don't mind.

docs/simplesamlphp-ecp-idp.txt
```diff
@@ -0,0 +1,66 @@
...
+
+Enabling ECP Profile on the IdP
+-----------------------------------
+
+To enable the IdP to send ECP assertions you must add the `saml20.ecp` option to the `saml20-idp-hosted` metadata file:
+
+    $metadata['__DYNAMIC:1__'] = array(
+        [....]
+        'auth' => 'example-userpass',
+        'saml20.ecp' => TRUE,
```
jaimeperez: This is a general comment regarding coding style. We try to avoid using uppercase for booleans and other constants as `null`, not only for coherence with how PHP dumps them into strings, but also to comply with PSR-2.

Refer to [our contribution guidelines](https://github.com/simplesamlphp/simplesamlphp/blob/master/CONTRIBUTE.md).

docs/simplesamlphp-ecp-idp.txt
```diff
@@ -0,0 +1,66 @@
...
+	    array (
+	      'Binding' => 'urn:oasis:names:tc:SAML:2.0:bindings:PAOS',
+	      'Location' => 'https://sp.example.org/ECP',
+	      'index' => 2,
+	    ),
+	  ),
+
+For this SPs you have to enable to recieve authnrequests via SOAP, so put this code to the SP config:
+
+	'saml20.soap.auth.allow' => true,
```
jaimeperez: I think the name of the configuration option is a little bit excessive, but I can live with that if we cannot find anything better.

In any case, if we want to support ECP in SSP as a Service Provider, this option should be included in the metadata exported by the SP (in PHP format). The other way around, if an IdP is parsing metadata (in XML format) from a Service Provider that's ECP-enabled, this option should be included in the resulting metadata in PHP format. The idea is that copying and pasting from the metadata converter tool should always yield a working setup that does not need manual edition of the converted metadata.

docs/simplesamlphp-reference-idp-hosted.txt
```diff
@@ -224,6 +224,10 @@ The following SAML 2.0 options are available:
 
 :   Note that this requires a configured memcache server.
 
+`saml20.ecp`
+:   Set to `TRUE` to enable the IdP to recieve authnrequests and send responses according the Enhanced Client or Proxy (ECP) Profile.
+    Defaults to `FALSE`.
+
 `saml20.hok.assertion`
 :   Set to `TRUE` to enable the IdP to send responses according the [Holder-of-Key Web Browser SSO Profile](./simplesamlphp-hok-idp).
     Defaults to `FALSE`.
```
jaimeperez: I think we need some comments in the IdP documentation about ECP. Basically, we need to explicitly tell that any **authentication filters that require interaction with the user will not work with ECP**, and therefore they must be disabled. The *consent* module is a good example of this. If it is enabled, authentication will not work unless the user gave consent for the SP on beforehand.

lib/SimpleSAML/Utilities.php
```diff
@@ -230,6 +230,10 @@ private static function _doRedirect($url, $parameters = array())
             SimpleSAML_Logger::warning('Redirecting to a URL longer than 2048 bytes.');
         }
 
+        if (isset($GLOBALS['SimpleSAML.debugDisableRedirect'])) {
+                throw new SimpleSAML_Error_Exception('Redirect was attempted with redirecting disabled: ' . var_export($url, TRUE));
+        }
+
```
jaimeperez: The `SimpleSAML_Utilities` class is deprecated and scheduled for removal. This check, in case it must be here inevitably, should be in `SimpleSAML_Utils_HTTP::redirect()` instead.

modules/core/lib/Auth/UserPassBase.php
```diff
@@ -169,6 +169,22 @@ public function isRememberMeChecked() {
 	public function authenticate(&$state) {
 		assert('is_array($state)');
 
+
+		if (isset($state['core:auth:username']) && isset($state['core:auth:password'])) {
+			/* It looks like we are doing HTTP Basic authentication. */
```
jaimeperez: Another comment on coding style. Multi-line comment delimiters should be used for comments that are truly multi-line. One-line comments like this should use their corresponding syntax `//`.

modules/core/lib/Auth/UserPassBase.php
```diff
@@ -169,6 +169,22 @@ public function isRememberMeChecked() {
...
+			$username = $state['core:auth:username'];
+			$password = $state['core:auth:password'];
+
+			if ($this->forcedUsername !== NULL && $this->forcedUsername !== $username) {
+				throw new SimpleSAML_Error_Error('WRONGUSERPASS');
+			}
+
+			$attributes = $this->login($username, $password);
```
jaimeperez: We should probably surround this with a `try { ... } catch` statement and log the error.

jaimeperez: After testing with a wrong username/password, I've noticed that the ECP-enabled IdP doesn't even reply with an error message. I think the problem is that this exception is not caught, so it escalates up and the code that should handle the error doesn't catch it because the exception is "unknown" to it. But I'm just speculating here...

modules/saml/lib/IdP/SAML2.php
```diff
@@ -87,10 +90,21 @@ public static function sendResponse(array $state) {
 	 */
 	public static function handleAuthError(SimpleSAML_Error_Exception $exception, array $state) {
 		assert('isset($state["SPMetadata"])');
-		assert('isset($state["saml:ConsumerURL"])');
+		assert('array_key_exists("saml:ConsumerURL", $state)'); // Can be NULL.
 		assert('array_key_exists("saml:RequestId", $state)'); // Can be NULL.
 		assert('array_key_exists("saml:RelayState", $state)'); // Can be NULL.
 
+ 		if ($exception instanceof SimpleSAML_Error_Error && $exception->getErrorCode() === 'WRONGUSERPASS') {
```
jaimeperez: This is what I was talking about in `UserPassBase.php`. Unless the error thrown by the call to `login()` in there throws a `SimpleSAML_Error_Error`, and there is a bug here, so it should read:

```php
if ($exception instanceof SimpleSAML_Error_Error || $exception->getErrorCode() === 'WRONGUSERPASS') {
```

Also, I think we should have something specific to ECP as the trigger for this particular code, because *WRONGUSERPASS* is something generic enough that it could be understood as a generic failure with password authentication, probably leading to mistakes in the future.

modules/saml/lib/IdP/SAML2.php
```diff
@@ -87,10 +90,21 @@ public static function sendResponse(array $state) {
...
+ 			/*
+ 			 * This should only happen during SOAP authentication, and indicates
+ 			 * that we do not have a valid username or password.
+ 			 */
+ 			header('WWW-Authenticate: Basic realm="IDPAUTH"', TRUE, 401);
+ 			header('Content-Type: text/plain');
+ 			echo('SOAP authentication requires basic auth.');
```
jaimeperez: Can the ECP client understand this error coming from the IdP and act accordingly? I'm just asking because I don't remember seeing anything in the spec...

modules/saml/lib/IdP/SAML2.php
```diff
@@ -87,10 +90,21 @@ public static function sendResponse(array $state) {
...
+ 			exit(0);
+ 		}
+
```
jaimeperez: Looking more into it, I think this whole if clause is intended to deal with missing authentication headers, but not with errors. That reaffirms my thoughts on the need for a specific code for the exception, different than *WRONGUSERPASS*.

I also think we need another if clause for `SimpleSAML_Error_Error` exceptions with a different code (say *ECP_AUTH_FAILURE*, for example), and act accordingly by sending back a SAML response to the ECP client informing about the error.

modules/saml/lib/IdP/SAML2.php
```diff
@@ -218,6 +232,41 @@ private static function getAssertionConsumerService(array $supportedBindings, Si
...
+ 	/**
+	 * Check a SOAP AuthnRequest.
+	 *
+	 * @param SimpleSAML_Configuration $idpMetadata  The metadata for the IdP.
+	 * @param SimpleSAML_Configuration $spMetadata  The metadata for the SP.
+	 * @param array &$state  The state array.
+	 */
+	private static function processSOAPAuthnRequest(SimpleSAML_Configuration $idpMetadata, SimpleSAML_Configuration $spMetadata, SAML2_AuthnRequest $request, array &$state) {
+
+		$GLOBALS['SimpleSAML.debugDisableRedirect'] = TRUE;
```
jaimeperez: I understand this is a mechanism to make the implementation of the ECP client easier, so that it doesn't need to follow HTTP redirections. I don't know if the spec mandates it, but I think it's a good idea anyway.

The problem though is the use of globals. That's bad practice and I don't think we should accept it like that. The state array could be used instead to store a flag with the same purpose. However, `SimpleSAML_Utils_HTTP::redirect()`, where this variable should be checked instead of `SimpleSAML_Utilities::_doRedirect()`, does not receive the state array as a parameter, nor it should, IMHO, as its function should be pretty clear and simple: perform an HTTP redirection.

Is it possible to check the places that might be calling for a redirection, and check the state array there, avoiding the redirection if the flag is set?

sitya:
Thanks for the reviewing, I will do some work on the code, I hope I can push the changes in this week.

jaimeperez:
Thanks a lot @sitya! I still couldn't go through all the code, it's hard to find time to focus on just one thing for a while, but it's ok. I'll try to review the rest of it tomorrow.

Oh, and by the way, if you don't agree with any of my comments or you think I got something wrong, please tell me!

sitya:
Hi, I hope I built in most of your suggestions. A did not touch the question about `saml20.soap.auth.allow`, but I was thinking about it, and I would say it might be unnecessary at all. I will check it again in the documentation of the profile.

lobbin:
Is this patch standalone, or it does still require pull request 48?

sitya:
> On 2015. dec. 10., at 13:19, Robin Helgelin <notifications@github.com> wrote:
> 
> Is this patch standalone, or it does still require pull request 48?

It requires PR#48.

lobbin:
Is this support to work with Office365? I tried to get a discussion going on the mailinglist, but no luck :)

sitya:
> On 2015. dec. 21., at 12:04, Robin Helgelin <notifications@github.com> wrote:
> 
> Is this support to work with Office365? 

I hope so :)

kgaofficial:
Quick question:
Will this enable any SP to communicate over SOAP to the SimpleSAMLPHP IdP? i.e. is the end-point a SOAP endpoint of POAS endpoint?

Sorry to ask this if it is obvious. Am a bit new to ECP.

sitya:
> @kgaofficial wrote:
> Will this enable any SP to communicate over SOAP to the SimpleSAMLPHP IdP?

Yes.

jaimeperez:
Hi!

I don't think we can say that so easily. It would enable those service providers using HTTP basic auth to authenticate to the IdP, assuming they are ok with this particular, limited implementation. Since this is not a complete implementation of the ECP specification, there might be SPs that won't interoperate with SSP even when this code is merged.

kgaofficial:
>enable those service providers using HTTP basic auth to authenticate to the IdP, assuming they are ok with this particular, limited implementation. 

I think, I am okay with that. ECP (the non 2.0 variety) would require only two such authentication schemes. One would be HTTP Basic and another would be Certificate based.

If the above code is merged, however limited, SSP will be the rarest of the open source IDPs that have any true ECP support. I have tried a couple of other famous IDPs, and despite claiming ECP, they don't really provide anything usable.

pgh70:
I would like (and need) the functionality this PR brings... But it will not merge with the current state of the master.... I did some rework and crunched all the work in a single commit to make it possible to rebase it on master again... My effort is in https://github.com/pgh70/simplesamlphp/tree/ecp

pculka:
Can somebody merge this? I will be probbably implementing ECP for one project, so this might aswell be handy and I could give a helping hand if needed

modules/saml/lib/IdP/SAML2.php
```diff
@@ -238,6 +238,31 @@ private static function getAssertionConsumerService(
...
     }
 
+    /**
+     * Check a SOAP AuthnRequest.
+     *
+     * @param SimpleSAML_Configuration $idpMetadata The metadata for the IdP.
+     * @param SimpleSAML_Configuration $spMetadata  The metadata for the SP.
+     * @param array                    &$state      The state array.
+     */
+    private static function processSOAPAuthnRequest(SimpleSAML_Configuration $idpMetadata, SimpleSAML_Configuration $spMetadata, SAML2_AuthnRequest $request, array &$state)
```
JohnMaguire: `SAML2_AuthnRequest` is now namespaced as `SAML2\AuthnRequest`

modules/saml/lib/IdP/SAML2.php
```diff
@@ -376,6 +401,12 @@ public static function receiveAuthnRequest(SimpleSAML_IdP $idp)
             'idpInit'     => $idpInit,
         ));
 
+        if ($binding instanceof SAML2_SOAP) {
```
JohnMaguire: `SAML2_SOAP` is now namespaced as `SAML2\SOAP`

modules/saml/lib/IdP/SAML2.php
```diff
@@ -428,6 +459,14 @@ public static function receiveAuthnRequest(SimpleSAML_IdP $idp)
             'saml:RequestedAuthnContext'  => $authnContext,
         );
 
+        if ($binding instanceof SAML2_SOAP) {
```
JohnMaguire: `SAML2_SOAP` is now namespaced as `SAML2\SOAP`

modules/saml/lib/IdP/SAML2.php
```diff
@@ -238,6 +238,31 @@ private static function getAssertionConsumerService(
...
+     * @param array                    &$state      The state array.
+     */
+    private static function processSOAPAuthnRequest(SimpleSAML_Configuration $idpMetadata, SimpleSAML_Configuration $spMetadata, SAML2_AuthnRequest $request, array &$state)
+    {
+
+        // Send the response via SOAP.
+        $state['saml:Binding'] = \SAML2\Constants::BINDING_SOAP;
+
+        if (!sspmod_saml_Message::checkSign($spMetadata, $request)) {
+            throw new SimpleSAML_Error_Exception('SOAP authentication request not signed.');
```
JohnMaguire: From reading over [the ECP spec](https://docs.oasis-open.org/security/saml/Post2.0/saml-ecp/v2.0/saml-ecp-v2.0.html) it doesn't seem like an `AuthnRequest` must be signed except in certain conditions...

Particularly, section 2.3.1 states:

> Request for a signed SAML request indicated by the option value "urn:oasis:names:tc:SAML:2.0:profiles:SSO:ecp:2.0:WantAuthnRequestsSigned"

This is stating that the SP should return a signed `AuthnRequest` to the ECP though, not that the IdP wants `AuthnRequest` to be signed. This is further stated in section 2.3.2, which also states:

> In parallel, the service provider MUST include a corresponding `<cb:ChannelBindings>` element in the `<samlp:Extensions>` element of its `<samlp:AuthnRequest>` message for each SOAP header block it attaches, containing channel bindings of a particular type. Within each extension element, the `Type` attribute MAY be set to the channel binding type (if known), and the raw channel binding data MUST be base64-encoded and the result used as the content of the element (per the "default" encoding specified in [ChanBind]). When channel bindings are included, the `<samlp:AuthnRequest>` message MUST be signed via [XMLSig].

However, this is only necessary when channel bindings are used, and is still the responsibility of the SP, not a requirement of the IdP.

In section 2.3.4.1, '"Holder of Key" Subject Confirmation', it is stated that the client must demonstrate possession of the key by signing the entire SOAP message, when applicable, but I think this is a different sort of check than is being performed here.

In section 2.3.6, there are some requirements on the IdP for scenarios where the client is authenticated via a digital signature, but again, that shouldn't require the `AuthnRequest` to be signed.

During testing with O365, Exchange Online _does not_ sign their `AuthnRequest`, and you cannot auth with this check present.

JohnMaguire:
When discovering the issue around [`AuthnRequest` signing](https://github.com/simplesamlphp/simplesamlphp/pull/233#discussion_r111827193), I also found an error in the error handling around `processSOAPAuthnRequest()`:

```
Apr 17 16:53:26 simplesamlphp INFO [d6b1e67a4c] SAML2.0 - IdP.SSOService: incoming authentication request: 'urn:federation:MicrosoftOnline'
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] SimpleSAML_Error_Exception: Error 8 - Undefined index: core:IdP
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] Backtrace:
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] 4 /var/www/html/simplesamlphp/www/_include.php:86 (SimpleSAML_error_handler)
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] 3 /var/www/html/simplesamlphp/lib/SimpleSAML/IdP.php:156 (SimpleSAML_IdP::getByState)
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] 2 /var/www/html/simplesamlphp/modules/saml/lib/IdP/SAML2.php:115 (sspmod_saml_IdP_SAML2::handleAuthError)
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] 1 /var/www/html/simplesamlphp/modules/saml/lib/IdP/SAML2.php:466 (sspmod_saml_IdP_SAML2::receiveAuthnRequest)
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] 0 /var/www/html/simplesamlphp/www/saml2/idp/SSOService.php:19 (N/A)
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] SimpleSAML_Error_Exception: Error 1 - Call to a member function getString() on null
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] Backtrace:
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] 2 /var/www/html/simplesamlphp/www/_include.php:86 (SimpleSAML_error_handler)
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] 1 /var/www/html/simplesamlphp/www/_include.php:54 (SimpleSAML_exception_handler)
Apr 17 16:53:26 simplesamlphp ERROR [d6b1e67a4c] 0 [builtin] (N/A)
```

modules/saml/lib/IdP/SAML2.php
JohnMaguire: I haven't tested it yet, but it looks like replacing this block with:

```
sspmod_saml_Message::validateMessage($idpMetadata, $spMetadata, $request);
```

Should do the trick... but it looks like this is already performed on line 359 of `receiveAuthnRequest()`. I need to verify that's true before removing this.

JohnMaguire: I also think we could probably set `$state['saml:Binding']` properly in that `receiveAuthnRequest()` function, and then all this function really does is process HTTP Basic Authentication.

modules/saml/lib/IdP/SAML2.php
```diff
@@ -376,6 +401,12 @@ public static function receiveAuthnRequest(SimpleSAML_IdP $idp)
             'idpInit'     => $idpInit,
         ));
 
+        if ($binding instanceof SAML2_SOAP) {
+            $supportedBindings = array(
+                \SAML2\Constants::BINDING_PAOS,
+            );
```
JohnMaguire: I think this could cause a problem if we're expected to respond on a different binding. We can simply append this to `supportedBindings` if `saml20.ecp` is enabled, up where we set the others.

thijskh:
Replaced by #626.

PR #238, opened 2015-08-17
Fixed bugs for SQL Server support

dustinmoris:
I fixed a few things to support SQL Server as well.

The main change was in SQL.php where I re-factored the insertOrUpdate method to support the MERGE statement for SQL Server 2008 and higher.

I moved the various db specific implementations into smaller private methods, so that the code can be re-factored to a more strategy based approach in the future.

### DIFF
--- a/lib/SimpleSAML/Store/SQL.php
+++ b/lib/SimpleSAML/Store/SQL.php
@@ -145,31 +145,89 @@ class SimpleSAML_Store_SQL extends SimpleSAML_Store {
 	 * we abstract it away here.
 	 *
 	 * @param string $table  The table we should update.
-	 * @param array $key  The key columns.
+	 * @param array $keys  The key columns.
 	 * @param array $data  Associative array with columns.
 	 */
 	public function insertOrUpdate($table, array $keys, array $data) {
-		assert('is_string($table)');
+    switch ($this->driver) {
+		case 'mysql':
+			$this->_insertOrUpdateMySQL($table, $keys, $data);
+			return;
+		case 'sqlite':
+			$this->_insertOrUpdateSQLLite($table, $keys, $data);
+			return;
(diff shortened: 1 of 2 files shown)


dustinmoris:
What is the status on this? Is there any hope to get this pull request accepted or at least an answer?

jaimeperez:
Hi @dustinmoris!

First of all, my apologies for the delay. We are quite short on resources so it takes time to address all issues, specially when we don't use the affected functionalities ourselves.

In this particular case, it's quite difficult for us to test this as we don't have any SQL server instance at hand that we could use. It would really help us out if someone else could confirm that your fixes work. Additionally, you might have seen that we have a new `SimpleSAML_Database` class and we are planning to migrate all classes making use of databases to this one, so that the main interface is always the same and things like the store need only to use it instead of implementing it again and again. In that sense, it would be nice if we could solve this issue in a way compatible with the new class.

dustinmoris:
I am not working with simpleSAMLphp anymore so it is not easy to make your suggested change now. If no one else was able to confirm my changes by now then clearly no one is using SQL Server with simpleSAMLphp. I will close it now, because I don't think anything is going to happen on this one.

rjimenezhsc:
Howdy!

We are starting to use SimpleSAMLphp in our project, however SQL Server support is highly desirable.

Is there anything I can do to help make this a reality?

I could fork this repo, apply the changes and use it instead.
However, I would prefer not to do that if there is no need.

Thank you for your time and hard work!

thijskh:
If you can confirm that the changes indeed work, I believe we can go forward with them. Creating a new PR is likely the best way to do that.

As Jaime said, the ideal solution would be something compatible with the new Database class, although I do not know from the top of my heaf what its status is.

PR #239, opened 2015-08-18
Isolate persistent state in SimpleSAML_Auth_Source::loginCompleted

astoltz:
When SimpleSAMLPHP runs an IDP with multiauth as the authentication source, upon successful login, the redirect back to the multiauth plugin fails with:

SimpleSAML_Error_Assertion: Assertion failed: 'isset($state["core:IdP"])'

I tracked this down to a recent refactoring of SimpleSAML_Auth_Source::loginCompleted(). $state is filtered which destroys much of the state data passed to "call_user_func($return, $state);" within the same function.

This change simply isolates the filter so the callback has full state information.


jaimeperez:
Hi,

This was indeed a bug that has been recently resolved in 5128993. Thanks for the PR!

PR #240, opened 2015-08-18
Fix assertion scope

astoltz:
The authState assertion in SimpleSAML_Auth_Source::initLogin fails because of incorrectly accessing the variable it's checking.


jaimeperez:
Hi @astoltz! Thanks for your PR!

This bug was introduced in 28a732d and fixed afterwards in c6cf761.

PR #242, opened 2015-08-19
Make SQLPersistentNameID more flexible

ghalse:
SQLPersistentNameID is very conservative about adding entries into the data store, and will only do so in situations where it knows they're explicitly wanted. This is a problem if there are situations where you want it to behave more like other NameID generation filters -- for instance, to create a persistent NameID even if the SP did not explicitly ask for it.

This patch makes it more flexible, whilst still preserving the existing behaviour as the default. It allows you to  override three of the existing checks by means of config booleans.

### DIFF
--- a/modules/saml/docs/nameid.txt
+++ b/modules/saml/docs/nameid.txt
@@ -69,12 +69,24 @@ This filter generates and stores a persistent NameID in a SQL datastore.
 To use this filter, simpleSAMLphp must be configured to use a SQL datastore.
 See the `store.type` configuration option in `config.php`.
 
-This filter will only create new NameIDs when the SP specifies `AllowCreate="true"` in the authentication request.
-
 ### Options
 
 `attribute`
 :   The name of the attribute we should use as the unique user ID.
+
+`allowUnspecified`
+:   Whether a persistent NameID should be created if the SP does not specify any NameID format in the request.
+    The default is `FALSE`.
+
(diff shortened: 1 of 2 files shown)


ghalse:
My use case for this is migrating from a Shibboleth IdP to simpleSAMLphp. We used to use Shibboleth's StoreID connector to generate and store persistent ids, and so had a SQL database with all existing persistent ids. SQLPersistentNameID seemed like the obvious way to preserve those when we changed the new IdP -- we simply munged the StoreID database schema into SQLPersistentNameID's.

This worked well, except in the case where an SP didn't explicitly request a persistent NameID.  Both Shibboleth's StoreID connector and simpleSAMLphp's PersistentNameID filter still create a persistent NameID in this situation; SQLPersistentNameID would not. By allowing this be be overridden, we've managed to preserve existing behaviour and successfully migrate without having to fix individual SPs.

Having done this, we had a single SP that sent a request with AllowCreate=false. This worked in Shibboleth, and would work with PersistentNameID, but didn't in SQLPersistentNameID. We did not have access to change this SP, hence a more complete set of override booleans.

The code in this request is currently running on our IdP without problems.

jaimeperez:
Hi,

That sounds quite reasonable to me, specially if this way the different _*PersistentNameID_ filter behave the same. I'm mergin the PR, thanks @ghalse!

PR #248, opened 2015-08-27
Added missing mappings for schacHomeOrganization[Type]

grawity:
Source: https://wiki.surfnet.nl/display/surfconextdev/Attributes+in+SURFconext


jaimeperez:
Hi @grawity!

Thanks a lot for your PR. There's a small issue with it, though. Some time ago, schac got its own namespace, so [all the URNs in the *terena.org* namespace are now deprecated](https://wiki.refeds.org/display/STAN/SCHAC+Registry+Information). The main issue is that the [new URN attribute definitions](https://wiki.refeds.org/display/STAN/SCHAC+URN+Registry) are explicitly forbidden for their use with SAML 2, but we could leave that to the user. Attribute mappings are protocol independent, anyway. In these two cases:

* `urn:schac:attribute-def:schacHomeOrganization`
* `urn:schac:attribute-def:schacHomeOrganizationType`

Could you please update that in your PR? Thanks!

grawity:
Hmm, well, there still are several SPs in eduGAIN which list requested attributes by their URN, and they all use the `urn:mace:terena.org:` namespace. (Although most of them list both URNs and OIDs, but I do see [one SP with URNs only][1].) So changing the mappings to `urn:schac:` would kind of defeat the point, no?

(I doubt these SPs will _ever_ use `urn:schac:`, especially given the old namespace is merely 'deprecated', but the new one is outright 'forbidden'… More likely that they'll migrate directly to OIDs.)

[1]: https://engine.surfconext.nl/authentication/sp/metadata

jaimeperez:
That's strange, because that's contrary to the SAML 2.0 LDAP/X500 Attribute Profile. OIDs should be used with SAML 2.0. In their case, the `urn:schac:attribute-def` namespace is definitely a no go, but they should switch to OIDs. For SPs using SAML 1.1 is more important, because in that case, URNs should be used (and the new namespace can be used then).

It's probably ok to keep support for the old namespace for a while. But at least we should also add support for the new one. Which leads me to the next thought. Shouldn't the mappings be between profiles instead of formats? Doesn't it make more sense to have *ldap2saml2* or *ldap2saml1* mappings, rather than *name2oid* or *name2urn*? That way we are explicitly fostering the use of the right attribute profiles, instead of allowing users to decide (and break the existing common rules).

As a compromise solution, maybe we could migrate to the new namespace all schac attributes, and provide a couple new mappings that translate from the old namespace to the new one, and vice versa?

grawity:
Well I'm not really qualified enough, but yeah, that might make more sense. Maybe that should be a separate PR though.

jaimeperez:
Ok, let's do this then. Since we always try to avoid breaking backwards compatibility between minor versions, I'm merging this PR as is, then adding mappings for the missing schac attributes, and providing mappings between the old and new namespaces.

Thanks for your contribution!

PR #257, opened 2015-09-21
Fix session problems

piotrbelina:


jaimeperez:
Hi @piotrbelina, thanks for your PR!

Could you please give us some information? Like why is this PR needed and what problems are you trying to address? Also, why are you storing the SAML responses like that? That's not very reasonable, since if we start storing the responses the backend store could grow really fast and big in very short time in case of SimpleSAMLphp instances with lots of authentications.

Besides, I'm afraid that we cannot merge pull requests that have conflicts or that duplicate commits that already exist. You would need to rebase your PR first.

piotrbelina:
Hi @jaimeperez,

I submitted this pull request by accident and I can not deleted myself. Could you do it please?

Thanks,
Piotr

jaimeperez:
Sure, no problem! We can't delete pull requests but we can just close it :smile:

PR #258, opened 2015-09-25
Added 'default' parameter to AttributeAlter. 

gtkrug:
The default parameter only works when %replace and 'target' are specified, and instead of creating the attribute with no value when the match fails, it creates the attribute with the default value.

The way it was implemented should be fully backwards compatible.  The functionality should remain exactly as is if this parameter is not present.


gtkrug:
This was a bad idea...  There is a simpler fix.

PR #259, opened 2015-09-25
Changed how %replace is handled when 'target' is specified

gtkrug:
Currently when %replace is specified when there is a 'target', instead of replacing the targets value with the new value, the new value is appended to the array of values for the target.  Semantically I think %replace should mean replace any value in this attribute with the new value if the match hits, so this change wipes out any existing value and replaces it with the new value if the match hits.

### DIFF
--- a/modules/core/docs/authproc_attributealter.txt
+++ b/modules/core/docs/authproc_attributealter.txt
@@ -90,6 +90,17 @@ Get the domain of the email and put it in a separate attribute:
         '%replace',
     ),
 
+Defaulting an attribute to one value (Add it with the default before altering) unless another attribute meets a condition: 
+
+    10 => array ('class' => 'core:AttributeAdd',
+                 'myAttribute' => 'default-value'),
+    11 => array ('class' => 'core:AttributeAlter',
+        'subject' => 'entitlement',
+        'pattern' => '/faculty/',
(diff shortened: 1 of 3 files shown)


gtkrug:
Initially I solved my issue by adding a new parameter to AttributeAlter called default, but then I decided the problem was really a bug in how %replace was handled when 'target' was specified, so I removed the changes I made to add the 'default' parameter and instead patched the behavior for the case of %replace and 'target' both being specified.

thijskh:
Hi gtkrug,

Thanks for your PR. Some remarks:
* Could you squash the two commits into one? Now the PR touches quite some lines while it's in essence just a one line change.
* Before changing the behaviour I think it would be good to have test coverage. It should probably not be hard to add a testcase for this filter, given those already present in `tests/modules/core/lib/Auth/Process`.
* Maybe the documentation should also be amended to make it explicit how this case is handled.

gtkrug:
I don't know how to actually run the tests...  There are no tests for alter at all right now, and the potential scope of all tests for alter is rather insane;I don't mind creating a few simple first tests, but I don't have time to try and create a thorough testing of that module.

thijskh:
Thanks! Your test base was very useful and I've added some more on top to reach 100% coverage for the filter, and merged the whole lot.

PR #260, opened 2015-09-25
Bug/session not saved till after redirect

BrettMerrick:
Occasionally a request that was the result of a redirect would attempt to load the session state before it had been saved by the previous request.

This caused either an unnecessary repeat of the redirect, or a SimpleSAML_Error_NoState: NOSTATE error.

The following log excerpt with additional debug messages reveals the sequencing issue.

###### Here the session state is read by the subsequent request after it is set which _is_ ok
```
Sep 25 11:30:30 dev01 simplesamlphp[18535]: [2015-09-25 11:30:30.291000] Redirect header sent
Sep 25 11:30:30 dev01 simplesamlphp[18535]: [2015-09-25 11:30:30.701500] Key set simpleSAMLphp.session.e9e35f8009185e9e8dbcd5c917c99048
-- 
Sep 25 11:30:31 dev01 simplesamlphp[13758]: [2015-09-25 11:30:31.963500] Key read simpleSAMLphp.session.e9e35f8009185e9e8dbcd5c917c99048
Sep 25 11:30:31 dev01 simplesamlphp[13758]: 7 [3c790b3d3a] Received message:
Sep 25 11:30:31 dev01 simplesamlphp[13758]: 7 [3c790b3d3a] <samlp:Response 
--
Sep 25 11:30:32 dev01 simplesamlphp[13758]: 7 [3c790b3d3a] Session: doLogin("1")
Sep 25 11:30:32 dev01 simplesamlphp[13758]: [2015-09-25 11:30:32.040200] Redirect header sent
```
###### Later the session state is set after it is read by the subsequent request which _is not_ ok
```
Sep 25 11:30:32 dev01 simplesamlphp[17980]: [2015-09-25 11:30:32.395400] Key read simpleSAMLphp.session.e9e35f8009185e9e8dbcd5c917c99048
Sep 25 11:30:32 dev01 simplesamlphp[17980]: 7 [3c790b3d3a] Session: '1' not valid because we are not authenticated.
Sep 25 11:30:32 dev01 simplesamlphp[13758]: [2015-09-25 11:30:32.429000] Key set simpleSAMLphp.session.e9e35f8009185e9e8dbcd5c917c99048
Sep 25 11:30:32 dev01 simplesamlphp[17980]: 7 [3c790b3d3a] Saved state: '_71671ba4009ce28926a6042542e7af1c844ad078ec'
```

This pull request separates the save method from the session destructor, and ensures that the session is saved with an explicit call immediately prior to sending the redirect header.

### DIFF
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -290,6 +290,15 @@ class SimpleSAML_Session
      */
     public function __destruct()
     {
+        $this->save();
+    }
+    
+    /**
+     * Saves the session to the session handler if the session has been
+     * marked as dirty. Do nothing otherwise.
+     */
+    public function save()
+    {
         if (!$this->dirty) {
             // session hasn't changed - don't bother saving it
             return;
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -119,6 +119,8 @@ class HTTP
      * HTTP version is 1.1. Otherwise, a "HTTP 302 Found" redirection will be used.
      *
      * The function will also generate a simple web page with a clickable link to the target page.
+     * 
+     * The function requests that the session is saved if dirty to ensure it is available after the redirect.
      *
      * @param string   $url The URL we should redirect to. This URL may include query parameters. If this URL is a
      *     relative URL (starting with '/'), then it will be turned into an absolute URL by prefixing it with the
@@ -134,6 +136,7 @@ class HTTP
      * @author Olav Morken, UNINETT AS <olav.morken@uninett.no>
      * @author Mads Freek Petersen
      * @author Jaime Perez, UNINETT AS <jaime.perez@uninett.no>
+     * @author Brett Merrick, Our School Ltd <brett.merrick@ourschool.co.nz>
      */
     private static function redirect($url, $parameters = array())
     {
@@ -159,6 +162,10 @@ class HTTP
         if (strlen($url) > 2048) {
             \SimpleSAML_Logger::warning('Redirecting to a URL longer than 2048 bytes.');
         }
+
+        // ensure session is saved and available after redirect
+        $session = \SimpleSAML_Session::getSessionFromRequest();
+        $session->save();
 
         // set the location header
         header('Location: '.$url, true, $code);


jaimeperez:
Hi @BrettMerrick!

Thanks a lot for your PR and your debugging.

I'm trying to reproduce the issue and I have to confess I'm failing miserably. In order to keep things simple, I've created a very simple script that uses a class with a `__destruct()` method that delays execution with a call to `sleep()`. Here it is:

```php
<?php
error_log("start of script: ".time());

class Test
{
    public function __construct()
    {
        error_log("constructor: ".time());
    }

    public function __destruct()
    {
        error_log("destructor called: ".time());
        sleep(5);
        error_log("destructor completed: ".time());
    }
}

if (!isset($_REQUEST['stop'])) {
    $t = new Test();
    sleep(2);
    header("Location: /tools/headers.php?stop=true");
    echo "redirecting...";
    error_log("header sent: ".time());
    exit();
}

?>
<html>
<body>
<a style="display: block; text-align: center; width: 100%; height: 100%;
    background: <?php echo sprintf('#%06X', mt_rand(0, 0xFFFFFF)); ?>;" href="/tools/headers.php">
    <h2 style="color: <?php echo sprintf('#%06X', mt_rand(0, 0xFFFFFF)); ?>;">Try again!</h2>
</a>
</body>
</html>

<?php
error_log("end of script: ".time());
```

Then I'm opening it through a web server with several browsers, while monitoring the browser (*live HTTP headers* in firefox) and even the network with *wireshark*. During all the tests I've made, the behaviour is coherent: the browser does not get the response from the server until all PHP code has been run, including the code in the destructor. Here's an example log of the execution:

```
[Tue Sep 29 13:13:41.553789 2015] [:error] [pid 1244] [client 192.168.56.1:63570] start of script: 1443525221
[Tue Sep 29 13:13:41.553822 2015] [:error] [pid 1244] [client 192.168.56.1:63570] constructor: 1443525221
[Tue Sep 29 13:13:43.554198 2015] [:error] [pid 1244] [client 192.168.56.1:63570] header sent: 1443525223
[Tue Sep 29 13:13:43.554269 2015] [:error] [pid 1244] [client 192.168.56.1:63570] destructor called: 1443525223
[Tue Sep 29 13:13:48.556247 2015] [:error] [pid 1244] [client 192.168.56.1:63570] destructor completed: 1443525228
[Tue Sep 29 13:13:48.560518 2015] [:error] [pid 1244] [client 192.168.56.1:63570] start of script: 1443525228
[Tue Sep 29 13:13:48.560584 2015] [:error] [pid 1244] [client 192.168.56.1:63570] end of script: 1443525228
```

This log also verifies what I've seen through the browser and *wireshark*. This behaviour given, it's impossible that the browser asks for a page it's been redirected to before the redirecting script finishes its execution, as it won't get the response at all and therefore it won't know where to go (not even what to do, actually).

I'm therefore assuming there's a problem somewhere else if you experience a different behaviour. What webserver are you running? What PHP version? Could you also run this same test and tell me if you get any different results?

Also, please note that you are mixing two different things in your test: the session and the state array (referenced in the log by the message *Saved state*). The log lines in your example are normal (except for your own added lines, of course), and they reflect what happens when a user tries to authenticate with SimpleSAMLphp with an invalid session. If a session exists from previous requests (sessions are created right away during the first request to any SimpleSAMLphp script), it will fetch it. After determining the session is invalid (the user never authenticated himself/herself), it will initiate authentication, and the session (as well as the state array) will be saved with relevant information about the current status for the user. This is coherent with your logs: you first read from the storage, then save after updating the session.

So I'm thinking that if you have repeated redirections (that's quite strange, can you reproduce that somehow and send a trace?) or `NOSTATE` exceptions, it must be due to some other issue. Maybe [this document](https://simplesamlphp.org/docs/development/simplesamlphp-nostate) can help finding the source of your trouble?

BrettMerrick:
Hi @jaimeperez,

Thanks for your comprehensive response!

Unfortunately the test code provided only works the way it does because it too relies on a quirk of output buffering.

This slight modification demonstrates this in a LAMP environment with fairly default settings, although since there are several system tuning parameters which can influence the result you may need to increase the length of the filler to trigger a flush of all the buffers.

```php
if (!isset($_REQUEST['stop'])) {
    $t = new Test();
    sleep(2);
    header("Location: /tools/headers.php?stop=true");

    $filler = str_pad('', 8192, 'b');  // 8049 is the tipping point for my testing.
//    error_log("Pre Filler" . var_export(ob_get_status(true), true));
    echo $filler;
//    error_log("Post Filler" . var_export(ob_get_status(true), true));

    error_log("header sent: ".time());
    exit();
}
```
Uncommenting the additional logging may provide a hint as to the value required on your system, however higher than default buffers downstream could prevent a flush at this level reaching the browser.
```log
Oct  1 08:21:32 dev01 start of script: 1443640892
Oct  1 08:21:32 dev01 constructor: 1443640892
Oct  1 08:21:34 dev01 header sent: 1443640894
Oct  1 08:21:34 dev01 destructor called: 1443640894
Oct  1 08:21:34 dev01 start of script: 1443640894
Oct  1 08:21:34 dev01 end of script: 1443640894
Oct  1 08:21:39 dev01 destructor completed: 1443640899
```
In our case I suspect the default output buffer flush was triggered occasionally by a nested output buffer callback function introduced by a PHP APM extension we use which occasionally injects javascript into the output to help it generate browser side traces.  But it could have been something as simple as a very long redirect url included in the output - max 2083 chars included 3 times, the location header, the a tag and the href.

Given the vulnerability of the current design to external factors including but not limited to:

* TCP buffer tuning
* Apache buffer tuning
* Apache modules
* PHP buffer tuning
* PHP extensions
* SimpleSAMLphp modules

it seems the only predictable, reliable solution is to ensure that the session is saved before the location header is sent.

BrettMerrick:
Hi @jaimeperez,

Do I need to open a new pull request to have this reconsidered now that it's been closed, or are you able to re-open this one?

jaimeperez:
Hi @BrettMerrick, thanks for the further details!

I need to take a closer look into this. I'll come back to you as soon as I can.

jaimeperez:
Hi again!

I've been digging a bit into this, and you are absolutely right. It turns out that this is a race condition issue with output chunking. I should have seen that coming. Intuitively, if PHP determines the output is going to be big and could take a while to be fetched, it will send it out chunked. In that case, headers should of course be sent before the rest of the content, and the script will keep running even though there's output sent already to the client. If a web browser is eager to make the experience faster for the user, which probably all of them do nowadays, then it will most likely ignore the output received in a redirect (who cares, anyway?) and just proceed with the next request while the first one is still ongoing.

My main headache here is that fixing this properly is difficult. I would really prefer to keep the session autonomous. Your PR addresses the particular problem with redirections, but this problem can happen in other places where there's no redirection at all:

* normal output + automated client that fetches the next URL in a flow without waiting for the previous one to finish completely (e.g. monitoring scripts).
* normal output + an extremely slow backend + a user clicking on links *too fast*.

I tried registering a shutdown function. However, the problem is exactly the same as with destructors. I also tried another alternative, which is registering a handler that executes **before** the output buffer is flushed to the client. In that case, we could make sure that the session is saved before any output is sent. Unfortunately, the way to do that is by calling `ob_start()` with the callback as first parameter, which also enables output buffering. That would be an ugly hack, apart from affecting performance real bad, since every request would need to wait until execution is completely over.

The only proper solution I can see is to use [`header_register_callback`](http://php.net/manual/en/function.header-register-callback.php). The problem with that one is that it was introduced in PHP 5.4, while we are still supporting PHP 5.3 in SimpleSAMLphp. I don't think it would be extremely bad to bump the minimum PHP version, but on the other hand I know there's still quite a lot stable distributions out there running PHP 5.3 and backporting security fixes.

So my proposal here would be to use that function (it's exactly what we need) in case it is available, fixing your problem *and* any other potential problems due to this. It will only work for those using PHP 5.4, though, so the solution for those running still 5.3 would be to upgrade. How does that sound?

Could you also apply this patch and tell me if it resolves the issue for you?

```patch
diff --git a/lib/SimpleSAML/Session.php b/lib/SimpleSAML/Session.php
index b13a64d..71326dd 100644
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -75,6 +75,14 @@ class SimpleSAML_Session


     /**
+     * Tells the session object that the save callback has been registered and there's no need to register it again.
+     *
+     * @var bool
+     */
+    private $callback_registered = false;
+
+
+    /**
      * This is an array of objects which will expire automatically after a set time. It is used
      * where one needs to store some information - for example a logout request, but doesn't
      * want it to be stored forever.
@@ -139,7 +147,7 @@ class SimpleSAML_Session

         $this->trackid = bin2hex(openssl_random_pseudo_bytes(5));

-        $this->dirty = true;
+        $this->markDirty();

         // initialize data for session check function if defined
         $globalConfig = SimpleSAML_Configuration::getInstance();
@@ -285,17 +293,22 @@ class SimpleSAML_Session
     }

     /**
-     * Destructor for this class. It will save the session to the session handler
-     * in case the session has been marked as dirty. Do nothing otherwise.
+     * Save the session to the store.
+     *
+     * This method saves the session to the session handler in case it has been marked as dirty.
+     *
+     * WARNING: please do not use this method directly unless you really need to and know what you are doing. Use
+     * markDirty() instead.
      */
-    public function __destruct()
+    public function save()
     {
         if (!$this->dirty) {
-            // session hasn't changed - don't bother saving it
+            // session hasn't changed, don't bother saving it
             return;
         }

         $this->dirty = false;
+        $this->callback_registered = false;

         $sh = SimpleSAML_SessionHandler::getSessionHandler();

@@ -311,6 +324,39 @@ class SimpleSAML_Session
     }

     /**
+     * Mark this session as dirty.
+     *
+     * This method will register a callback to save the session right before any output is sent to the browser.
+     */
+    public function markDirty()
+    {
+        $this->dirty = true;
+
+        if (!function_exists('header_register_callback')) {
+            // PHP version < 5.4, can't register the callback
+            return;
+        }
+
+        if ($this->callback_registered) {
+            // we already have a shutdown callback registered for this object, no need to add another one
+            return;
+        }
+        $this->callback_registered = header_register_callback(array($this, 'save'));
+    }
+
+
+    /**
+     * Destroy the session.
+     *
+     * Destructor for this class. It will save the session to the session handler
+     * in case the session has been marked as dirty. Do nothing otherwise.
+     */
+    public function __destruct()
+    {
+        $this->save();
+    }
+
+    /**
      * Retrieve the session ID of this session.
      *
      * @return string|null  The session ID, or null if this is a transient session.
@@ -385,7 +431,7 @@ class SimpleSAML_Session

         SimpleSAML_Logger::debug('Session: doLogin("'.$authority.'")');

-        $this->dirty = true;
+        $this->markDirty();

         if (isset($this->authData[$authority])) {
             // we are already logged in, log the user out first
@@ -443,7 +489,7 @@ class SimpleSAML_Session
             return;
         }

-        $this->dirty = true;
+        $this->markDirty();

         $this->callLogoutHandlers($authority);
         unset($this->authData[$authority]);
@@ -554,7 +600,7 @@ class SimpleSAML_Session
         assert('isset($this->authData[$authority])');
         assert('is_int($expire) || is_null($expire)');

-        $this->dirty = true;
+        $this->markDirty();

         if ($expire === null) {
             $globalConfig = SimpleSAML_Configuration::getInstance();
@@ -587,7 +633,7 @@ class SimpleSAML_Session
         }

         $this->authData[$authority]['LogoutHandlers'][] = $logout_handler;
-        $this->dirty = true;
+        $this->markDirty();
     }

     /**
@@ -612,7 +658,7 @@ class SimpleSAML_Session
         }

         unset($this->dataStore[$type][$id]);
-        $this->dirty = true;
+        $this->markDirty();
     }

     /**
@@ -677,7 +723,7 @@ class SimpleSAML_Session

         $this->dataStore[$type][$id] = $dataInfo;

-        $this->dirty = true;
+        $this->markDirty();
     }

     /**
@@ -836,7 +882,7 @@ class SimpleSAML_Session

         $this->associations[$idp][$association['id']] = $association;

-        $this->dirty = true;
+        $this->markDirty();
     }


@@ -899,7 +945,7 @@ class SimpleSAML_Session

         unset($this->associations[$idp][$associationId]);

-        $this->dirty = true;
+        $this->markDirty();
     }
```

Thanks!

BrettMerrick:
Hi,

I'd thought about the ob_start() callback, but agree it seemed like a bit of a sledgehammer approach.

I was just starting to look into cases where something external to the browser - such as a monitoring script, or speedy user - was causing the error, and considering how best to broaden the catchment of the solution in the PR.

I hadn't spotted the header_register_callback in 5.4, but it seems like a much better option. And only registering with it when the session is dirty is a nice solution. Nice work!

I'll run some tests and let you know how it looks.

BrettMerrick:
Hi @jaimeperez,

This patch has resolved the issues we were seeing. Thanks!

jaimeperez:
Thanks Brett!

Glad to hear it resolved the issue. I've just merged the patch into master.

PR #261, opened 2015-09-28
Remove setting ReturnURL for default SimpleSAML_Auth states

madmatt:
[RFC BEFORE MERGE]

I'm not sure if this is the correct way to do this, but the SAML2 spec states
that the relayState param must be 80 bytes or less. If ReturnURL is set here, it
is set to `https://my-site.gov/simplesaml/module.php/core/postredirect.php?RedirId=_54b431cf9a27821d0eeeb17072e7ee5648075f6e81`,
which is always >80 bytes when adding the domain.

It seems there may have been a way to override this `relayState` in previous
versions of SimpleSAMLphp, as I've found references in `authsources.php` to
setting an array parameter of `'RelayState' => null`, but that doesn't appear to
work anymore. Is that a better solution? I'm happy to re-work my pull request if
that's the case.

### DIFF
--- a/lib/SimpleSAML/Auth/Source.php
+++ b/lib/SimpleSAML/Auth/Source.php
@@ -179,12 +179,7 @@ abstract class SimpleSAML_Auth_Source
                 'SimpleSAML_Auth_Source.logoutSource' => $this->authId,
             ),
         ));
-
-        if (is_string($return)) {
-            $state['SimpleSAML_Auth_Default.ReturnURL'] = $return; // TODO: remove in 2.0
-            $state['SimpleSAML_Auth_Source.ReturnURL'] = $return;
-        }
-
+		
         if ($errorURL !== null) {
             $state[SimpleSAML_Auth_State::EXCEPTION_HANDLER_URL] = $errorURL;
         }


jaimeperez:
Hi @madmatt!

This is a part of the spec that we are basically ignoring on purpose, since it's much more convenient to use the RelayState parameter to carry a URL where you can redirect the user after authentication. You can set that parameter in the URL of an IdP-first flow in order to get redirected to a particular place in your SP after authentication, or you can set it in `config/authsources.php` as you noticed.

At some point we should stop using RelayState like this. However, that means a lot of work (definitely not just what you did in your PR) with little benefit, as we haven't noticed any interoperability issue so far due to *abusing* the spec like this...

madmatt:
Hey @jaimeperez,

Thanks for your comments! Unfortunately I've run into a case where the IdP does in fact enforce a 80-byte limit on the `RelayState` param. It's the New Zealand government's single login system called RealMe.

If I made another PR that allowed you to set the base path to `RelayState` within `authsources.php` for example, do you think that would be acceptable? This PR blindly deletes it which obviously isn't acceptable, but in some cases we don't need `RelayState`, and could remove it.

I think I might do that, and then we can continue discussion on the forthcoming PR, rather than here :)

Thanks again!

tractorcow:
Just a question; This behaviour could be left in, but could we not implement a new option for "enforce strict relaystate limit"? It could be an option that's off by default, but allows integration with systems such as the above by setting an option from 0 to 1. ;)

@madmatt do you think this would be a feasible config option?

madmatt:
@tractorcow I'm not sure such an option would make sense - if you set that to `1` and the `RelayState` is longer than 80 bytes, would you truncate it to 80 bytes or just force it `null`? Either way, it seems like a sub-optimal solution, as I'm not sure that either would be expected from a system. I'll implement it as I described in my previous comment, meaning that we can work around this for RealMe and any other IdP implementations that require a strict 80 byte length, and don't otherwise use the `RelayState` value for anything.

tractorcow:
Or maybe it could just be "don't add returnurl"?

PR #262, opened 2015-09-28
[NOT FOR MERGE] Add common headers that reverse proxies set

madmatt:
These proxy headers are often set by reverse proxies, or web servers like nginx
that are capable of proxying requests to a backend application server that runs
SimpleSAMLphp.

I've marked this as 'not for merge' because theoretically anyone could set these
headers, and we want to make sure that they're only set by trusted proxies. I
propose using an environment variable to enumerate a list of trusted IP
addresses, and setting a define() with a boolean true/false for a quick check
point, e.g. `if(TRUSTED_PROXY && isset(...`, however I don't know how this
should be handled in the SimpleSAMLphp codebase.

Are there any thoughts on how this could be handled better? Ideally we wouldn't
hard-code things in config.php, because these would likely vary per-environment
(development environments may not have them, and trusted proxies are likely
different IP addresses in staging/pre-prod environments and production).

Any thoughts on how to solve this properly?

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -73,7 +73,23 @@ class HTTP
      */
     private static function getServerHTTPS()
     {
-        if (!array_key_exists('HTTPS', $_SERVER)) {
+        // Convention for (non-standard) proxy signaling a HTTPS forward, see
+        // https://en.wikipedia.org/wiki/List_of_HTTP_header_fields
+		if(isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) == 'https') {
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @madmatt!

Thanks a lot for your PR!

This has been discussed several times in the mailing list, I believe. The issue here is that there's nothing you can do in the server (meaning in SimpleSAMLphp) to ensure those headers are ok and you can trust them. Anyone can forge them and send you a header that matches your trust rules. The only way to make sure the headers you receive are "secure" is to make sure that the load balancers or reverse proxies you have in front of your servers are blindly removing the headers coming in the original request and setting their own ones.

In any case, that's irrelevant and not even necessary. You don't need to use these headers to guess what your address is. Just set the `baseurlpath` configuration option in `config/config.php` to the canonical URL of your service.

PR #263, opened 2015-09-28
Removed new call before makeException.

mschwager:
The `makeException` function returns an instantiated object, so the
caller shouldn't also call `new`.


thijskh:
Thanks!

Issue #265, opened 2015-10-06
Security Report: Remote File Include Alert


dialogik:
I've been reported about a security issue by our data security officers (who ran some kind of penetration tool on my service) that refers to SimpleSAMLphp. See the attached screenshot for full information.

![RFI suspicion](https://cloud.githubusercontent.com/assets/1260548/10302736/f46d4416-6c0e-11e5-9140-b647ab3cf495.png)

Long story short, that report says that passing a URL to `disco.php`s `return` parameter or to `discoresp.php`s `AuthID` parameter would invoke a [Remote File Inclusion](https://en.wikipedia.org/wiki/File_inclusion_vulnerability). I checked both and what happens is, that in both scenarios the user is *redirected*  to those URLs but no file inclusion takes place whatsoever. I also took a look into the sourcecode which confirmed this assumption.

Now two questions remaining:

1. I am no security expert, so I'd like to ask the community: **Is there a way to inject remote code anyhow by this?** (I'd be very surprised btw as SimpleSAMLphp is a security relevant library and I assume will be well audited.)
1. **Why is it necessary to pass those values via the URI, why can't they be read from the config files?** Wouldn't that be more secure? Because, while I don't see a RFI problem, at least this might be used in order to perform some kind of phishing or clickjacking attack.

thijskh:
Unlikely that code injection is possible. Open redirects are considered a security risk though because it can be used in phishing to obscure the real URL the user is sent to (see OWASP).

My first question would be whether you've set `trusted.url.domains` in config.php? See the bottom of that file (if you copied the template from config-templates/) for the setting and explanation.

dialogik:
No, I haven't yet. Thanks for pointing me there. Have set it to an empty array as recommended in order to disallow all redirections).

But still: Wouldn't it be better to configure any redirection path via configuration instead of passing them via the URI?

gfaust-qb:
Wouldn't it be better to generally set it to secure defaults?

jaimeperez:
Hi!

You are quite right, there's no *remote file inclusion* in there. I wonder how the auditors ended up identifying an open redirect with an RFI...

[...]

PR #266, opened 2015-10-06
Set trusted.url.domains default to array()

dialogik:
See #265.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -834,6 +834,6 @@ $config = array(
      * Example:
      *   'trusted.url.domains' => array('sp.example.com', 'app.example.com'),
      */
-    'trusted.url.domains' => null,
+    'trusted.url.domains' => array(),
 
 );
--- a/docs/simplesamlphp-upgrade-notes-1.14.txt
+++ b/docs/simplesamlphp-upgrade-notes-1.14.txt
@@ -179,3 +179,5 @@ The following modules will no longer be shipped with the next version of SimpleS
     * `openidProvider`
     * `saml2debug`
     * `themefeidernd`
+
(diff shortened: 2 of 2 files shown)


jaimeperez:
Thanks a lot for this @dialogik!

Just one thing. Since this change could break some current installations, would you mind to add a note to the upgrade notes telling about this change and what SimpleSAMLphp users would need to do if they rely on redirections to third-party sites?

Thanks again!

thijskh:
This changes the default in the template config (a good start). But I think we'd also want to change the code in Utils\HTTP to the different default, so that if someone does not have the setting in their config at all, they will get the 'secure' behaviour?

jaimeperez:
That's probably the best approach. In that case it's even more important to update the upgrade notes, as well as the comment in the config template, warning about this option being enforced if not specified.

dialogik:
The added the information to the [docs/simplesamlphp-upgrade-notes-1.14.txt](https://github.com/dialogik/simplesamlphp/blob/e5daae771ffdd780aa2285276232c41e8587eba1/docs/simplesamlphp-upgrade-notes-1.14.txt).

The `config.php` template – imho – doesn't need to be adapted as all required information are already contained (`null`, empty `array()` and an example `array()` with third party sites).

I'll take a look at `Utils/HTTP.php` later (maybe not until the weekend).

jaimeperez:
I've just merged this PR and modified `SimpleSAML_Utils_HTTP::checkURLAllowed()` to change the default behaviour to forbid all redirections that are not to a trusted site.

dialogik:
Thanks for accomplishing, didn't have the time.

Issue #268, opened 2015-10-07
Negotiate-module issue with LDAP-referrals


tvdijen:
I ran into an issue with the negotiate-module that I'd like to share. I have attached a piece of the log to this issue.

When doing a tcpdump, I could see that the LDAP-search was actually performing fine without any errors! So I'm a bit clueless on where the 'Operations error' is coming from.
Anyway, I found out that the problem was lying in the 'referrals' parameter.. It had to be turned off for the Active Directory I'm connecting to.

In /simplesamlphp-1.13.2/modules/negotiate/lib/Auth/Source/Negotiate.php, line 103, the LDAP-module is being called:
>> new SimpleSAML_Auth_LDAP($this->hostname, $this->enableTLS, $this->debugLDAP, $this->timeout);
This call lacks the arguments 'port' and 'referrals'.

I would like to suggest that these settings are made customizable in the authsources.php. They are for the LDAP-part of the config, but not for the Negotiate-part...

tvdijen:
Will post pull-request soon

tvdijen:
Log-file, because uploading a text-file tot Github is not possible at the moment:

>> Oct 07 08:20:52 simplesamlphp DEBUG [f93eca0527] Session: 'AD-KERBEROS' not valid because we are not authenticated.
>> Oct 07 08:20:52 simplesamlphp DEBUG [f93eca0527] Negotiate - SP disable flag not found
>> Oct 07 08:20:52 simplesamlphp DEBUG [f93eca0527] Negotiate - authenticate(): looking for Negotate
>> Oct 07 08:20:52 simplesamlphp DEBUG [f93eca0527] Negotiate - authenticate(): Negotate found
>> Oct 07 08:20:52 simplesamlphp DEBUG [f93eca0527] Library - LDAP __construct(): Setup LDAP with host='somewhere@example.nl', tls=false, debug=false, timeout=30, referrals=true
>> Oct 07 08:20:52 simplesamlphp INFO [f93eca0527] Negotiate - authenticate(): Snoa_TA_ssc98@EXAMPLE.NL authenticated.
>> Oct 07 08:20:52 simplesamlphp DEBUG [f93eca0527] Negotiate - authenticate(): Binding as system user 'CN=sa-gdi-idpldp-a,OU=ServiceAccounts,DC=example,DC=nl'
[...]

PR #269, opened 2015-10-09
Adresses issue #268 and makes the negotiate module a bit more configu…

tvdijen:
…rable, like you would configure ldap

### DIFF
--- a/tests/modules/core/lib/Auth/Process/AttributeAlterTest.php
+++ b/tests/modules/core/lib/Auth/Process/AttributeAlterTest.php
@@ -33,20 +33,92 @@ class Test_Core_Auth_Process_AttributeAlter extends PHPUnit_Framework_TestCase
 
         $request = array(
             'Attributes' => array(
-                 'test' => array('wrong'),
+                 'test' => array('somethingiswrong'),
              ),
         );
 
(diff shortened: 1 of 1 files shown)


tvdijen:
It would probably be even better if the negotiate-module would use the ldap-settings from the ldap-module, but that would require more changes... I'll leave that decision up to the developers..

thijskh:
Hi Tim,
Thanks for your work. But a first comment is that you appear to have changed all the newlines in the file. This leads to the diff becoming huge while in fact just a few lines have been changed. Can you push the commit again with the whitespace changes undone?

tvdijen:
I tried, used different text-editors, but it keeps doing that... It looks like git is messing it up

thijskh:
Maybe these links are helpful:
http://wiki.secondlife.com/wiki/How_to_avoid_DOS_line_endings_in_Windows_tools
http://stackoverflow.com/questions/8195839/choose-newline-character-in-notepad

tvdijen:
I've tried it all, it's not working.. Different platforms, different editors..

tvdijen:
I've managed to fix it using the online editor.. Did a new pull request

thijskh:
Thanks, diff is much clearer now (in #270)!

PR #272, opened 2015-10-12
Update LDAP.php

tvdijen:
Fixed to follow PSR-2 coding guidelines, proper indentation and a minor fix in equality operand (== vs ===)

### DIFF
--- a/lib/SimpleSAML/Auth/LDAP.php
+++ b/lib/SimpleSAML/Auth/LDAP.php
@@ -12,7 +12,7 @@ define('ERR_AS_ATTRIBUTE', 6);
 
 // not defined in earlier PHP versions
 if (!defined('LDAP_OPT_DIAGNOSTIC_MESSAGE')) {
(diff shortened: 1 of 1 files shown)


thijskh:
Thanks!

tvdijen:
You're welcome! This was a file I accidentally ran into... Are you aware of any other files that need reformatting?

thijskh:
I'm not aware of a concrete list, but maybe @jaimeperez is?

jaimeperez:
Hi!

First of all, thanks a lot for this @tvdijen!

I don't have a list of specific files that should be revamped to conform with PSR-2. My usual approach is to reformat every file I open for other reasons and that I see is still wrongly formatted. Since I'm using an IDE with visual warnings for everything out of the formatting rules, it's quite easy to spot them.

When I'm reformatting a file, I usually let the IDE to do that for me, and then fix some things that might be missing. For instance, I tend to use `//` for one-line comments and leave `/* */` for multi-line, use lowercase for constants like `true`, `false` or `null`, and also try to make the phpdoc comments as verbose and complete as possible.

PR #276, opened 2015-10-24
Avoid data-collision with memcache

tvdijen:
Avoid data-collision when running multiple instances of SSP on the same host that are all configured to use memcache


tvdijen:
Forgot something.. Will send in new PR

PR #277, opened 2015-10-24
Avoid memcache data collision

tvdijen:
Avoid memcache data collision when running multiple SSP instances on the same host

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -763,6 +763,18 @@ $config = array(
         ),
     ),
 
+    /*
+     * This value is the name under which the data which the data should be
+     * stored in memcache. It defaults to 'simpleSAMLphp', which is fine
+     * in most cases.
+     *
+     * Change this value when you have multiple instances of SimpleSAMLphp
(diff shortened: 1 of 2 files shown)


thijskh:
Hi Tim,

Thanks, this seems like a useful change to have.

I'm wondering if it wouldn't be nicer to fetch the config value in `SimpleSAML_Store_Memcache`'s constructor and use it in each of the methods?

Also I'm not so sure about the four commits, which all do not have a really descriptive commit message. Maybe you can rebase them into two (one for the purely functional change, one for the coding style changes)?

tvdijen:
Hi Thijs, I'll do that and see if I can put it in the constructor as you said.. I'll make it a new PR then

thijskh:
You need not make a new PR; you can instead force-push to the branch you used to make the PR.

tvdijen:
Hehe, well, I guess it's too late for that now.. I should probably read a bit about version control, because sometimes it makes no sense to me at all...

PR #278, opened 2015-10-24
Avoid data collision in memcache and PSR-2 reformatting

tvdijen:
This is an update to Memcache.php to prevent data collision when running multiple instances of SSP on one host and more than one of them is configured to use memcache.
It introduces a new setting in config.php that allows you to come up with a unique prefix per instance.
Also PSR-2 reformatting of the Memcache.php file is done.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -765,6 +765,17 @@ $config = array(
 
 
     /*
+     * This value allows you to set a prefix for memcache-keys. The default
+     * for this value is 'simpleSAMLphp', which is fine in most cases.
+     *
+     * When running multiple instances of SSP on the same host, and more
+     * than one instance is using memcache, you probably want to assign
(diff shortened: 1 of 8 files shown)


thijskh:
Much better, thanks :-)

I've rebased it to remove the 2 merge commits and squashed the config one, but that's more a matter of taste I guess.

tvdijen:
You're welcome... But, I've made a really stupid mistake.. The code is referencing the wrong property in the config file, because I was in doubt whether I should call it 'name' or 'prefix'... Does that require a new PR? I've fixed it in my fork...

thijskh:
I don't see the fix in your fork?

tvdijen:
https://github.com/tvdijen/simplesamlphp/commit/a9cdecf5fb01638ba8390aa3e99bc04c62138a99

thijskh:
I've merged that diff now. Thanks again for your work!

PR #279, opened 2015-10-26
Corrected typo in core:show_metadata template

tdiscuit:
There is a typo in `show_metadata.tpl.php` referencing the key "blacklink" when it is actually "backlink".


thijskh:
Thanks!

PR #284, opened 2015-11-05
Make SHA-256 the default signature algorithm.

thijskh:
SHA-1 is still supported but needs to be configured explicitly if you need it.

SHA-1 has been marked as deprecated in SSP 1.12 (2014-03-24). SSP 2.0 seems like a good moment to change the default.


jaimeperez:
Thanks Thijs! I quite agree with you, so I'm marking this to be merged for 2.0.

jaimeperez:
Thijs, did I miss something, or closing this issue was a some kind of mistake?

thijskh:
I needed to put it on a different branch. Have also rebased it so it applies cleanly to current master again. See #441.

PR #286, opened 2015-11-09
add OIDs from SIS (Swedish Standards Institute), TK 450 AG04

jschlyter:
reference: https://www.skolfederation.se/wordpress/wp-content/uploads/2015/06/Skolfederation_Attributprofil_3_1.pdf


jschlyter:
Could we get this merged and flushed out? (or dropped)

jaimeperez:
Tack så mycket @jschlyter och ursäkta!

I know I held this for a reason, though I can't remember it now, so probably it wasn't that important. In any case, I think it's better to just merge it.

PR #288, opened 2015-11-17
Filtering function in MetaLoader

pradtke:
Testable with:

phpunit --configuration tools/phpunit tests/modules/metarefresh/lib/MetaLoaderTest.php

Change is add ability to filter entities by a callback function (or a currying function).  The main use case is filtering metadata that has eduGain entities by the registration authority.

'source' can now take new keys:

'filterCallback' is the name of a function that accepts a 'SimpleSAML_Metadata_SAMLParser' and returns true/false

'filterFactory' is the name of a function that returns a function that meets the requirements of 'filterCallback'.   'filterFactoryArgs' is an array of arguments passed to the factory function.

---

One thing I could not figure out was how to enable the metarefresh module for unit test.  Jamie described a technique in https://groups.google.com/d/msg/simplesamlphp-dev/Sp0Hhe86KTQ/pzY6LubrAAAJ however (because of lack of php and simplesaml experience) I couldn't get to work.  For short term I added an ENV variable to load all modules and that ENV gets set by phpunit. I will need some help on this part if we want to use a different technique.

If we have some agreement on the new attribute names and the functionality then I can work on updating the docs in a subsequent PR

Lastly, my co-worker pointed out that the filtering is happening post XML processing/conversion to entities which is not memory efficient. It might be desirable to move filtering into the SimpleSaml SimpleSAML_Metadata_SAMLParser::parseDescriptorsElement() call or to have that call return a generator so that only one entity is converted/filtered at a time.

### DIFF
--- a/lib/_autoload_modules.php
+++ b/lib/_autoload_modules.php
@@ -29,6 +29,7 @@ function temporaryLoader($class)
 
     if (!strstr($class, 'SimpleSAML_')) {
         return; // not a valid class name for old classes
+
     }
     $original = $class;
 
--- a/modules/metarefresh/lib/CommonFilters.php
+++ b/modules/metarefresh/lib/CommonFilters.php
@@ -5,22 +5,37 @@
 class sspmod_metarefresh_CommonFilters {
 
 
+    /**
+     * @param $authority The registration authority the entity should match
+     * @return Closure A closure that will return true if called with an entity registered by $authority
+     */
     public static function registeredAuthorityFilterFactory($authority) {
         return function(SimpleSAML_Metadata_SAMLParser $entityDesc) use ($authority) {
-            $metaData = self::getMetadata($entityDesc);
+            $metaData = sspmod_metarefresh_CommonFilters::getMetadata($entityDesc);
             return isset($metaData['RegistrationInfo']['registrationAuthority']) && $metaData['RegistrationInfo']['registrationAuthority'] === $authority;
         };
     }
 
+    /**
+     * @param $name The name of the entity attribute to check
+     * @param $value The value that the entity attribute should contain
+     * @return Closure A closure that will return true if called with an entity that has an attribute with name $name
+     *  value $value.
+     */
     public static function entityAttributeFactory($name, $value) {
         return function(SimpleSAML_Metadata_SAMLParser $entityDesc) use ($name, $value) {
-            $metaData = self::getMetadata($entityDesc);
+            $metaData = sspmod_metarefresh_CommonFilters::getMetadata($entityDesc);
             return isset($metaData['EntityAttributes'][$name]) && in_array($value,$metaData['EntityAttributes'][$name], true);
         };
     }
 
-
-    private static function getMetadata(SimpleSAML_Metadata_SAMLParser $entity) {
+    /**
+     * An internal helper function. Limitations in php 5.3 prevent referencing this helper function from anonymous methods
+     * unless it is public.
+     * @param SimpleSAML_Metadata_SAMLParser $entity
+     * @return array
+     */
+    public static function getMetadata(SimpleSAML_Metadata_SAMLParser $entity) {
         $metaData = $entity->getMetadata20SP();
         if (!isset($metaData)) {
             $metaData = $entity->getMetadata20IdP();
--- a/modules/metarefresh/lib/CommonFilters.php
+++ b/modules/metarefresh/lib/CommonFilters.php
@@ -1,31 +1,38 @@
 <?php
+
 /*
  * Filter callbacks and callback factories that are useful to most SSP users
  */
-class sspmod_metarefresh_CommonFilters {
+
+class sspmod_metarefresh_CommonFilters
+{
 
 
     /**
-     * @param $authority The registration authority the entity should match
-     * @return Closure A closure that will return true if called with an entity registered by $authority
+     * Creates a filter that checks the <mdrpi:RegistrationInfo/> element of an entity.
+     * @param $authority The registration authority the entity should match.
+     * @return Closure A closure that will return true if called with an entity registered by $authority.
      */
-    public static function registeredAuthorityFilterFactory($authority) {
-        return function(SimpleSAML_Metadata_SAMLParser $entityDesc) use ($authority) {
+    public static function registeredAuthorityFilterFactory($authority)
+    {
+        return function (SimpleSAML_Metadata_SAMLParser $entityDesc) use ($authority) {
             $metaData = sspmod_metarefresh_CommonFilters::getMetadata($entityDesc);
             return isset($metaData['RegistrationInfo']['registrationAuthority']) && $metaData['RegistrationInfo']['registrationAuthority'] === $authority;
         };
     }
 
     /**
-     * @param $name The name of the entity attribute to check
-     * @param $value The value that the entity attribute should contain
+     * Creates a filter that checks the <mdattr:EntityAttributes> element of an entity.
+     * @param $name The name of the attribute to check.
+     * @param $value A value that the attribute should contain.
      * @return Closure A closure that will return true if called with an entity that has an attribute with name $name
-     *  value $value.
+     *  and value $value.
      */
-    public static function entityAttributeFactory($name, $value) {
-        return function(SimpleSAML_Metadata_SAMLParser $entityDesc) use ($name, $value) {
+    public static function entityAttributeFactory($name, $value)
+    {
+        return function (SimpleSAML_Metadata_SAMLParser $entityDesc) use ($name, $value) {
             $metaData = sspmod_metarefresh_CommonFilters::getMetadata($entityDesc);
-            return isset($metaData['EntityAttributes'][$name]) && in_array($value,$metaData['EntityAttributes'][$name], true);
+            return isset($metaData['EntityAttributes'][$name]) && in_array($value, $metaData['EntityAttributes'][$name], true);
         };
     }
 
@@ -35,7 +42,8 @@ class sspmod_metarefresh_CommonFilters {
      * @param SimpleSAML_Metadata_SAMLParser $entity
      * @return array
      */
-    public static function getMetadata(SimpleSAML_Metadata_SAMLParser $entity) {
+    public static function getMetadata(SimpleSAML_Metadata_SAMLParser $entity)
+    {
         $metaData = $entity->getMetadata20SP();
         if (!isset($metaData)) {
             $metaData = $entity->getMetadata20IdP();
--- a/modules/metarefresh/lib/CommonFilters.php
+++ b/modules/metarefresh/lib/CommonFilters.php
@@ -40,6 +40,12 @@ class sspmod_metarefresh_CommonFilters {
         if (!isset($metaData)) {
             $metaData = $entity->getMetadata20IdP();
         }
+        if (!isset($metaData)) {
+            $metaData = $entity->getMetadata1xSP();
+        }
+        if (!isset($metaData)) {
+            $metaData = $entity->getMetadata1xIdP();
+        }
         return $metaData;
     }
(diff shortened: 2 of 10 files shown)


pradtke:
I realized one of the files I changed (lib/SimpleSAML/Metadata/SAMLParser.php) will instead need to be applied to the SimpleSAML repo.

CONTRIBUTE.md
```diff
@@ -64,6 +64,7 @@ deploy, and therefore we try to avoid it.
 * Add **unit tests** to verify that your code not only works but also keeps working over time. When adding tests, keep
 the same directory structure used for regular classes. Try to cover **all your code** with tests. The bigger the test
 coverage, the more reliable and better our library is.
+    * `phpunit --configuration tools/phpunit tests/path/to/your/test.php`
```
jaimeperez: It would be nice to elaborate a bit more.

On the other hand, as long as the test is inside the `tests` directory, there's no need to pass it in the command line, as phpunit will run all tests it finds in there.

lib/SimpleSAML/Metadata/SAMLParser.php
```diff
@@ -124,6 +124,11 @@ class SimpleSAML_Metadata_SAMLParser
      */
     private $entityAttributes;
 
+    /**
+     * An associative array of attributes from the RegistrationInfo element.
+     * @var array
+     */
+    private $registrationInfo;
```
jaimeperez: I think this addition to `SimpleSAML_Metadata_SAMLParser` makes a lot of sense by itself, but I feel like it should have its own PR. It is just a prerequisite for the functionality implemented by this PR, but others might find this useful too.

That way we can also make this PR smaller and easier to manage.

lib/_autoload_modules.php
```diff
@@ -24,7 +24,7 @@ function SimpleSAML_autoload($className)
     $module = substr($className, $modulePrefixLength, $modNameEnd - $modulePrefixLength);
     $moduleClass = substr($className, $modNameEnd + 1);
 
-    if (!SimpleSAML_Module::isModuleEnabled($module)) {
+    if (getenv('SIMPLESAMLPHP_ENABLE_ALL_MODULES') == false && !SimpleSAML_Module::isModuleEnabled($module)) {
```
jaimeperez: This is a bit confusing. This change has nothing to do directly with the PR itself, and honestly I don't think you need it or even that it is a good idea.

Also, strict comparison operator should be used in a case like this.

tools/phpunit/phpunit.xml
```diff
@@ -9,6 +9,10 @@
          stopOnFailure="false"
          syntaxCheck="false"
          bootstrap="./../../vendor/autoload.php">
+    <php>
+        <!-- During testing we want all modules enabled -->
```
jaimeperez: Actually, we don't :smile: 

How could we test then that `SimpleSAML\Module::isModuleEnabled()` fails appropriately if all modules are enabled?

I think we need an alternative way to do this, using the `module.enable` configuration option that we have already.

tests/modules/metarefresh/lib/MetaLoaderTest.php
```diff
@@ -0,0 +1,221 @@
...
+        $this->verifyEntityPresent(true, $entities, 'https://auth.cs.serialssolutions.com/auth/Metadata/Shib');
+        $this->verifyEntityPresent(false, $entities, 'https://idp.nuim.ie/idp/shibboleth');
+
+    }
+
+    /**
+     * Assert that the metadata has or doesn't have the entityId
+     * @param $present
+     * @param $metadata
+     * @param $entityId
```
jaimeperez: I know this is just a private method, but would you mind completing the phpdoc block with the correct types for the parameters, as well as a description?

tests/modules/metarefresh/lib/MetaLoaderTest.php
```diff
@@ -0,0 +1,221 @@
...
+
+        // This Idp doesn't match the filter
+        $this->verifyEntityPresent(false, $entities, 'urn:mace:incommon:osu.edu');
+        // Only Wikis match the filter
+        $this->verifyEntityPresent(true, $entities, 'https://carmenwiki.osu.edu/shibboleth');
+        $this->verifyEntityPresent(true, $entities, 'https://wiki.cac.washington.edu/');
+        $this->verifyEntityPresent(false, $entities, 'https://beta.projecteuclid.org/shibboleth-sp');
+    }
+
+    public static function entityIdIsWiki(SimpleSAML_Metadata_SAMLParser $entityDesc) {
```
jaimeperez: This should also have a phpdoc block.

modules/metarefresh/lib/MetaLoader.php
```diff
@@ -124,6 +124,29 @@ public function loadSource($source) {
 			return;
 		}
 
+        $filterFunction = null;
+        if (isset($source['filterCallback'])) {
+            if (!is_callable($source['filterCallback'])) {
+                SimpleSAML_Logger::debug('Invalid filter callback ' . $source['filterCallback'] . ' - attempting to re-use cached metadata');
```
jaimeperez: I'm not sure this should be `debug`. In the end, this would be a configuration error, right? So in that case you want the user to notice about it, even though in most deployments people won't use *debug* as log level.

modules/metarefresh/lib/MetaLoader.php
```diff
@@ -124,6 +124,29 @@ public function loadSource($source) {
...
+            }
+        }
+
+        if(isset($source['filterFactory'])) {
+            if (isset($filterFunction)) {
+                SimpleSAML_Logger::warning("Both 'filterCallback' and 'filterFactory' defined. Ignoring 'filterFactory");
+            } elseif (is_callable($source['filterFactory'])) {
+                $filterFunction = call_user_func_array($source['filterFactory'], $source['filterFactoryArgs']);
+            } else {
+                SimpleSAML_Logger::debug('Invalid filter factory ' . $source['filterFactory'] . ' - attempting to re-use cached metadata');
```
jaimeperez: Same here.

modules/metarefresh/lib/CommonFilters.php
```diff
@@ -0,0 +1,59 @@
...
+            return isset($metaData['EntityAttributes'][$name]) && in_array($value, $metaData['EntityAttributes'][$name], true);
+        };
+    }
+
+    /**
+     * An internal helper function. Limitations in php 5.3 prevent referencing this helper function from anonymous methods
+     * unless it is public.
+     * @param SimpleSAML_Metadata_SAMLParser $entity
+     * @return array
+     */
```
jaimeperez: Could you please complete the phpdoc block? :smile:

modules/metarefresh/lib/CommonFilters.php
```diff
@@ -0,0 +1,59 @@
...
+        $metaData = $entity->getMetadata20SP();
+        if (!isset($metaData)) {
+            $metaData = $entity->getMetadata20IdP();
+        }
+        if (!isset($metaData)) {
+            $metaData = $entity->getMetadata1xSP();
+        }
+        if (!isset($metaData)) {
+            $metaData = $entity->getMetadata1xIdP();
+        }
```
jaimeperez: I might be missing something, but why this particular order in the `if` statements?

jaimeperez:
Thanks a lot for this PR @pradtke! Outstanding job!

I've commented on a few things in the code itself. I think the main things we need to address in order to merge this PR are resolving the conflicts, and splitting the PR into 2 (maybe more), so that things are self-contained. Other than that, it's a great work :smile:

CONTRIBUTE.md
pradtke: I added more info.  My goal was for someone not well versed in php to be able to create/run tests.

lib/SimpleSAML/Metadata/SAMLParser.php
pradtke: Ported this to
https://github.com/simplesamlphp/simplesamlphp/pull/335

If it conflicts with this PR, then I can remove the change from this PR once 335 is approved.

lib/_autoload_modules.php
pradtke: See my comment below

tools/phpunit/phpunit.xml
pradtke: Both this, and the hack in _autoload_modules were because of my inability to enable the module any other way.  You had mentioned to me previously about creating a temp dir with a minimal config.php and overriding the default configuration location - but that is beyond my current ability with ssp + php.   I agree we need a better way to do it, I'm just not sure I can create it :)

tests/modules/metarefresh/lib/MetaLoaderTest.php
pradtke: done

tests/modules/metarefresh/lib/MetaLoaderTest.php
pradtke: done

modules/metarefresh/lib/MetaLoader.php
pradtke: agree. changed to `warning`

modules/metarefresh/lib/MetaLoader.php
pradtke: done.

modules/metarefresh/lib/CommonFilters.php
pradtke: done.

modules/metarefresh/lib/CommonFilters.php
pradtke: you didn't miss anything :)  The RegistrationInfo is the metadata, so it just looks up different metadatas until one has data. I updated the comment to reflect this.

pradtke:
Thanks @jaimeperez I've run through the fixes. I broke out 1 PR (and can remove that code from this, once its approved).  As mentioned in the comments I'm unable to get an alternate way to enable a module for testing to work. I agree my way is wrong :) It was mostly just to get the tests to run.

jtgasper3:
Hey all,

I'm looking for this functionality to filter out entities so that my SSP instance only has InCommon Research and Scholarship tagged IdPs. Is the only outstanding issue the merge conflicts?

Thanks,
John

pradtke:
I still need to get the tests working in a better fashion as a prerequisite for merging. I'll go ahead and fix the PR to avoid conflicts.

jtgasper3:
Thanks for the update, Patrick. Sounds good. I'll be patient... or if I get
bored, I'll see if I can add anything. :)

pradtke:
@jtgasper3 I've done an update. So it will apply cleanly to master, and should apply to 1.14.x branches if you limit the patch files to `modules/metarefresh/lib/CommonFilters.php` and `modules/metarefresh/lib/MetaLoader.php` - all the other files are test or documentation related.

I think the major change between 1.14 and 1.13 is that I had pulled out the RegistrationAuthority changes into their own patch that is part of 1.14.

To preserve 1.13 compatibility I created a copy of the branch at https://github.com/pradtke/simplesamlphp/tree/feature/callback-metadata-filter-1_13 and there is also a gist https://gist.github.com/pradtke/94422f5534c69a02bae7.

In regards to tests and getting this PR approved, I started this mailing list discussion https://groups.google.com/forum/#!topic/simplesamlphp-dev/OKfZtXrc3P0 to figure out how the project wants to handle global state spill over between tests.

tvdijen:
@pradtke I think the recently merged https://github.com/simplesamlphp/simplesamlphp/pull/819 will solve your unit testing issues

tvdijen:
@pradtke This is so outdated and the module has been moved to it's own repository in the meantime.
If this is still desirable functionality, would you mind creating a new PR in the metarefresh-repo?

PR #291, opened 2015-11-20
Added 'secure' session cookie option

mschwager:
Along with the `httponly` cookie option, the `secure` option is another good one to use.

https://www.owasp.org/index.php/SecureFlag

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -366,6 +366,7 @@ $config = array(
     'session.phpsession.cookiename' => null,
     'session.phpsession.savepath' => null,
     'session.phpsession.httponly' => true,
+    'session.phpsession.secure' => true,
 
     /*
      * Option to override the default settings for the auth token cookie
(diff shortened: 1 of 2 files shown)


thijskh:
There is already the `session.cookie.secure` option for that, I think?

mschwager:
I believe `session.cookie.secure` refers to the authentication token cookie (typically `SimpleSAMLAuthToken`), whereas `session.phpsession.secure` refers to the PHP session cookie (typically `PHPSESSID`).

olavmrk:
Actually, the `secure` flag for PHP sessions is set from `session.cookie.secure`. The same goes for some of the other cookie parameters.

Note that `SimpleSAML_SessionHandlerPHP::getCookieParams()` [calls `SimpleSAML_SessionHandler::getCookieParams()`](https://github.com/simplesamlphp/simplesamlphp/blob/v1.14.0-rc1/lib/SimpleSAML/SessionHandlerPHP.php#L223). That function [sets the `secure` flag](https://github.com/simplesamlphp/simplesamlphp/blob/v1.14.0-rc1/lib/SimpleSAML/SessionHandler.php#L148).

PR #295, opened 2015-12-03
Updates openid/php-openid version dependency

### LINKED ISSUE #294: Open ID version update for Composer Install
OpenId version 2.2.2 on dev-master is no longer available or has been updated to 2.3.0, can we please update the dependency requirement.

mathewpeterson:
This should fix #294

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php": "~5.3",
         "simplesamlphp/saml2": "~0.3",
         "simplesamlphp/xmlseclibs": "~1.3.2",
-        "openid/php-openid": "dev-master#ee669c6a9d4d95b58ecd9b6945627276807694fb as 2.2.2"
+        "openid/php-openid": "^2.3"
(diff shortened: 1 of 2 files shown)


jaimeperez:
Hi @mathewpeterson! Thanks for your PR!

This is fixed in master (actually, in master there's no longer such dependency), and there won't be any more 1.13.X releases, so it doesn't make sense to merge this.

mathewpeterson:
I appreciate your feedback and understand that there is no longer any effort to maintain that branch.

The problem for us is that you still haven't made a new release and our best practice forbids us to use pre stable branches. Since you are removing the openid/php-openid requirement I would assume this would be a new major release (per semver.org) anyways. 

I hope you will reconsider releasing a patch for 1.13 with this fix so that people can continue using 1.13 without much effort. If not, we will keep our fork with this fix so we can continue working.

Thank you for your time!

PR #297, opened 2015-12-07
Fix multiauth selected source timeout

astoltz:
By default, multiauth only stores the selected authentication source for the datastore length, which is 4 hours. The default session length is 8 hours. If you try logging out between the 4 and 8 hour marks, you will get an exception saying "Invalid authentication source during logout".

My fix is to store the auth source until session expiration when it won't be needed anymore.


jaimeperez:
Thanks a lot @astoltz!

PR #300, opened 2015-12-16
gitignore for custom modules

clinta:
This gitignore ignores anything in the modules directory while explicitly including modules that ship with simplesamlphp. This allows users to create and maintain custom modules while tracking upstream.


jaimeperez:
Thanks a lot @clinta!

PR #305, opened 2015-12-31
Added option for custom AssertionConsumerService sent in SSO request

freezy-sk:


jaimeperez:
Thanks for the PR @freezy-sk!

Could you please provide some background info on why you would need this?

freezy-sk:
Hi @jaimeperez 
I'm using this as vendor library outside public part of project and when I don't want to expose whole handling mechanism defined in www, I need this option to set custom URL of my Assertion Consumer because it's checked with Identity Provider's configuration

thijskh:
I'm not too enthusiastic about this. First of course because SSP works in a certain way as an SP. You can limit access to the www/ parts with various settings, and could even use Apache settings if you have even more stringent requirements. Redirecting to a completely external ACS URL does not seem like a logical solution that we want to support to me.

tvdijen:
Similar functionality was added in https://github.com/simplesamlphp/simplesamlphp/pull/988

PR #308, opened 2016-01-15
support Basic Authorization to web proxy

restena-sw:
as discussed on the mailing list

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -823,7 +823,14 @@ $config = array(
      *   'proxy' => 'tcp://proxy.example.com:5100'
      */
     'proxy' => null,
-
+    
+    /*
+     * Username/password authentication to proxy (Proxy-Authorization: Basic)
+     * Example:
+     *   'proxy.auth' = 'myuser:password'
+     */
+    'proxy.auth' => null
+    
     /*
      * Array of domains that are allowed when generating links or redirections
      * to URLs. SimpleSAMLphp will use this option to determine whether to
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -336,7 +336,8 @@ class HTTP
 
 
     /**
-     * Helper function to retrieve a file or URL with proxy support.
+     * Helper function to retrieve a file or URL with proxy support, also 
+     * supporting proxy basic authorization.
      *
      * An exception will be thrown if we are unable to retrieve the data.
      *
@@ -361,9 +362,13 @@ class HTTP
         $config = \SimpleSAML_Configuration::getInstance();
 
         $proxy = $config->getString('proxy', null);
+        $proxy_auth = $config->getString('proxy.auth', null);
         if ($proxy !== null) {
             if (!isset($context['http']['proxy'])) {
                 $context['http']['proxy'] = $proxy;
+            }
+            if (!empty($proxy_auth) ) {
+                $context['http']['header'] = "Proxy-Authorization: Basic".base64_encode($proxy_auth);
             }
             if (!isset($context['http']['request_fulluri'])) {
                 $context['http']['request_fulluri'] = true;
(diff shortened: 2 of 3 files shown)


jaimeperez:
Hi Stefan! Thanks a lot for the PR!

There's a couple of things we would need to fix, though. First, you are submitting two different changes as part of the same PR. It would be better to have them split as that way we can check and merge them independently, so that one does not affect or delay the other. You can do that by creating branches (for example, `feature/proxy-auth` and `feature/conditional-endpoints`), and then creating a PR in github for each of them.

Secondly, I see you used the `simplesamlphp-1.14` branch as the source for your changes. This has two undesired consequences:

* Merging the PR is difficult because 1.14 is now closed for new features. As soon as 1.14.0 is released, we can add minor features that are backwards-compatible
* Continuous Integration with Travis is apparently broken in 1.14, due to a change in one of our dependencies (*php-coveralls*) which has been introducing a lot of trouble lately. This means we cannot make sure that your changes build properly right now (I'll try to fix this ASAP).

Therefore, the way to proceed would be:

1. Checkout the `master` branch.
2. Create a `feature/proxy-auth` branch (or something else, the name is just an example, of course).
3. Apply the related patch.
4. Submit a PR here.
5. Create a `feature/conditional-endpoints` branch (ditto).
6. Apply the related patch.
7. Submit the new PR.

There's a couple of things I'd like to comment on too, but I'll do that in the code where they apply.

Thanks again!

config-templates/config.php
```diff
@@ -823,7 +823,14 @@
...
      */
     'proxy' => null,
-
+    
+    /*
+     * Username/password authentication to proxy (Proxy-Authorization: Basic)
+     * Example:
+     *   'proxy.auth' = 'myuser:password'
+     */
+    'proxy.auth' => null
```
jaimeperez: I think we should document the proxy feature in general. Probably the best place to do that is a new section in the [advanced features page](https://simplesamlphp.org/docs/stable/simplesamlphp-advancedfeatures).

We tend to add options to the configuration templates without properly documenting them (and I'm the first to claim responsibility of that :hand: ), and that's causing our users to miss a lot of important information, I think.

Needless to say this won't be a show stopper.

lib/SimpleSAML/Utils/HTTP.php
```diff
@@ -361,10 +362,14 @@ public static function fetch($url, $context = array(), $getHeaders = false)
         $config = \SimpleSAML_Configuration::getInstance();
 
         $proxy = $config->getString('proxy', null);
+        $proxy_auth = $config->getString('proxy.auth', null);
         if ($proxy !== null) {
             if (!isset($context['http']['proxy'])) {
                 $context['http']['proxy'] = $proxy;
             }
+            if (!empty($proxy_auth) ) {
```
jaimeperez: The comparison here should be:

```php
if ($proxy_auth !== null) {
```

Setting `proxy.auth` to `0` or `""` (empty string) would avoid triggering authorization. Actually, it's probably better to default it to false and compare this way:

```php
if ($proxy_auth !== false) {
```

Then we should state clearly in the documentation that setting `proxy.auth` to `false` will disable it.

modules/saml/www/sp/metadata.php
```diff
@@ -58,6 +58,7 @@
 
 $index = 0;
 $eps = array();
+$supported_protocols = array();
```
jaimeperez: I think this is the way to go and I really want to see this merged. However, there's a problem with backwards-compatibility. Our rule of thumb is that we try to keep backwards-compatibility always between minor versions (1.14 in this case). Metadata is a specially critical case where backwards-compatibility is essential for interoperability, so we always try to be extra careful with it.

In this case, this change would mean different metadata, so I think we need a configuration option to enable this, and set it to keep the current behaviour by default. Then you can switch that option to remove unused protocols depending on your configuration.

Later on, when we release 2.0, I'd get rid of that option and make this the default behaviour. I wouldn't spend much time on documenting the configuration option therefore.

How does that sound?

PR #311, opened 2016-01-17
Better handling of multi-line logs (e.g., with backtrace)

bradjones1:
Hopefully this PR isn't too broad, but I've attempted to clean up some of the logging logic particularly when it comes to handling multi-line messages. Reading through the existing code, most messages destined for the log appear as $string, and a string is assumed, though the Logger's log() function includes an implode() if it's indeed passed an array.

I ship my logs off to an offsite log aggregation service, using rsyslog as an intermediary. Furthermore I like to package log messages as JSON, either in the application or with an rsyslog template. This PR is backwards-compatible with existing configurations but will result in only one "main" error message being logged, as opposed to the foreach() that was running which resulted in one log message per line. That can become a bit unwieldy with the backtrace.

If the logging service is syslog and the format is set to "json," the message is maintained as an array and is then JSON-encoded.

Also this downgrades the "Error report generated" message to a notice, since it is not itself the error, but simply a note about the logged error.

Includes some abstraction in Exception.php as the result of the removal of those foreach() loops.

### DIFF
--- a/lib/SimpleSAML/Error/Error.php
+++ b/lib/SimpleSAML/Error/Error.php
@@ -211,7 +211,7 @@ class SimpleSAML_Error_Error extends SimpleSAML_Error_Exception
         $etrace = implode("\n", $data);
 
         $reportId = bin2hex(openssl_random_pseudo_bytes(4));
-        SimpleSAML_Logger::error('Error report with id '.$reportId.' generated.');
+        SimpleSAML_Logger::notice('Error report with id '.$reportId.' generated.');
 
         $config = SimpleSAML_Configuration::getInstance();
         $session = SimpleSAML_Session::getSessionFromRequest();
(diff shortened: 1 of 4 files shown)


jaimeperez:
Thanks for the PR @bradjones1!

Could you provide some example outputs for both the default, backwards-compatible logging, and the new format with JSON?

jaimeperez:
Hi again @bradjones1!

I've recently enhanced a bit the handling of logs, including those with backtraces. Backtraces are now logged as debug, so they can be ignored easily if you want.

Regarding the formatting in JSON, wouldn't that be possible if you provide a JSON-encoded string in the `logging.format` configuration option? In case it's not possible, or you need something more elaborate, I think it might make sense to have a separate log handler for this. That's another improvement I introduced recently. Now it's possible to use custom log handlers just by implementing the interface `\SimpleSAML\Logger\LoggingHandlerInterface`, making the class available for the autoloader, and referencing it in the `logging.handler` configuration option.

PR #312, opened 2016-01-17
Include .editorconfig file

bradjones1:
Include an .editorconfig file, per http://editorconfig.org/

Lifted from Drupal; I think the settings are correct for this project, though. Main item being 4-space indents.


olavmrk:
Hi, could you squash the two commits together in a single commit, and give it a bit more of a description? E.g. the text you included in the pull request?

bradjones1:
@olavmrk This passes with the exception of the coveralls test... though since it's just an editor config file it wouldn't need feature testing with the rest of the codebase.

jaimeperez:
Thanks @bradjones1!

Since this is intended for those developing SimpleSAMLphp and not for users, I think we should remove the file when shipping a new code release. Could you please modify the `bin/build-release.sh` script to do that too? Then we have everything together in this PR.

I'll have to remove other dot files too, but that can be done separately.

bradjones1:
@jaimeperez Branch updated per your feedback and up to date with master. :-)

jaimeperez:
Thanks a lot @bradjones1!

PR #313, opened 2016-01-19
Added port to the LDAP base filter configuration

justinsg:
The base class for the LDAP auth process filters doesn't fetch the LDAP port from the auth source configuration. This causes a "Can't connect to LDAP server" type error when using an LDAP server (or tunnel) with a port other than the default 389.

### DIFF
--- a/modules/ldap/lib/Auth/Process/BaseFilter.php
+++ b/modules/ldap/lib/Auth/Process/BaseFilter.php
@@ -137,6 +137,7 @@ abstract class sspmod_ldap_Auth_Process_BaseFilter extends SimpleSAML_Auth_Proce
 			$authconfig = array();
 			$authconfig['ldap.hostname']   = @$authsource['hostname'];
 			$authconfig['ldap.enable_tls'] = @$authsource['enable_tls'];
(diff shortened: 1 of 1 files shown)


jaimeperez:
Thanks a lot for this PR @justinsg!

Could I ask you to provide also some documentation for this option on the relevant configuration file templates and / or documentation?

Thanks!

justinsg:
Hi, I'm not sure this option requires any extra documentation - [section 2.2 of the LDAP AuthFilter](https://simplesamlphp.org/docs/stable/ldap:ldap#section_2_2) docs already states:

> It will look at the authsource for all LDAP connection information

All I have done in this PR is to ensure this actually happens - previously the `port` option was missed out.

LDAP options defined here: 
https://github.com/justinsg/simplesamlphp/blob/9941f6f3d66b7d095c3d6c4a9730510527364bc2/modules/ldap/lib/ConfigHelper.php#L136-L143

The LDAP AuthFilter base class pulls those options in here:
https://github.com/justinsg/simplesamlphp/blob/9941f6f3d66b7d095c3d6c4a9730510527364bc2/modules/ldap/lib/Auth/Process/BaseFilter.php#L138-L147

PR #314, opened 2016-01-19
Use saml2 v2 and xmlseclibs v2

relaxnow:
Use the new PSR2 classes from simplesaml/saml2 and robrichards/xmlseclibs.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,8 +23,8 @@
     },
     "require": {
         "php": ">=5.3",
-        "simplesamlphp/saml2": "~1.7",
-        "robrichards/xmlseclibs": "~1.4.1",
+        "simplesamlphp/saml2": "~2.0",
+        "robrichards/xmlseclibs": "~2.0",
         "whitehat101/apr1-md5": "~1.0"
     },
     "require-dev": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "79273d493e7d184e2726c3e6b13c0d05",
-    "content-hash": "675e663d2d81f0fb3ab32dbb80d2892d",
+    "hash": "75e524197b53d4ae052bb0c519f9fbb5",
+    "content-hash": "6e843f26cd794daae89dc8c11ab3a9ab",
     "packages": [
         {
             "name": "psr/log",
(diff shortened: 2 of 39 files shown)


DRvanR:
As discussed, `class_alias`es for all existing SAML2 classes should be added. They should be loaded always. Furthermore the autoloader should be expanded to check if classes are loaded the old way and trigger a deprectated notice (`trigger_error('some meaningful message', E_USER_DEPRECATED)`).

This provides users and module maintainers with a BC way to keep using SAML2 while allowing SSP to move forward and adopt SAML2 2.x, which in turn makes maintenance and development of SAML2 easier.

jaimeperez:
Hi guys!

I've just added an autoloader to the SAML2 library, creating aliases on demand for all old-style class names. It also considers the case of the only class that has been renamed (`SAML2_Const`, for obvious reasons).

It does not log any warning yet, as I wanted to discuss that with you two. I'd consider `error_log()` out of the question, and `trigger_error()` might be a bit excessive if it ends up dumping stack traces to your logs, depending on your configuration. Are we logging anything in the SAML2 library, via the SSP compat layer? What would you suggest to use?

BTW, we should probably log warnings too for all the classes deprecated in SimpleSAMLphp that will disappear or be renamed in 2.0.

DRvanR:
Cheers, much appreciated!

> trigger_error() might be a bit excessive if it ends up dumping stack traces to your logs, depending on your configuration

afaik warnings do not trigger the dumping of a strack trace but this would just be lines like:
`Deprecated: <message> in <path> on line <lineno>` which to me would be perfectly acceptable. Easy to grep and helps locate code to fix. 

Logging anything through the SSP compat layer doesn't seem to be a good option to me as this project is used outside of SSP as well.

Personally I'd opt for the `trigger_error($msg, E_USER_DEPRECATED)` - nice sideeffect would be that you could suppress the warnings if you'd want to :wink:

jaimeperez:
I've seen the traces myself while using `trigger_error()`, that's why I was asking. The message logged is also a bit different than that, but I guess it could be the particular configuration in my development VM.

Let's proceed with `trigger_error()` in any case :+1:

jaimeperez:
By the way, I submitted a PR to XMLSecLibs (robrichards/xmlseclibs#93) to provide the exact same functionality. Let's see if Rob accepts it, it would simplify things a lot.

jaimeperez:
Hi all!

This was finally solved in c23722a47dc196428f9ab5d68fd25ad74a3c6cdf.

In the end, it was more complicated than expected, even though we had the class aliasing in the SAML2 library. Basically, since class aliases create new virtual classes, those classes lose all inheritance relationships, so it's no longer possible to test an object and see if it inherits from some abstract class or interface.

In any case, all the code in master is finally using the 2.x branch. Sorry that I couldn't merge your work directly, Boy!

PR #315, opened 2016-01-19
Add GenerateAffiliation Authentication Processing Filter and unittest.

mrvanes:
GenerateAffiliation generates an attribute with fixed values based on (multiple) values found in a source attribute.
In short: if member of group "A" create target attribute with value "student".


modules/core/lib/Auth/Process/GenerateAffiliation.php
```diff
@@ -0,0 +1,110 @@
...
+        parent::__construct($config, $reserved);
+
+        assert('is_array($config)');
+
+        /* Validate configuration. */
+        foreach ($config as $name => $value) {
+            if (is_int($name)) {
+                // check if this is an option
+                if ($value === '%replace') {
+                        $this->replace = TRUE;
```
jaimeperez: [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#25-keywords-and-truefalsenull) :wink:

thijskh:
Hi Martin, thanks, I think this would be useful to have, as I don't think there's something there currently that does this. AttributeAlter is geared specifically towards regexp-like transformations, which could in theory do this but is unwieldy in the case of this kind of replacements, AttributeMap targets mapping attribute names, not values.

I do think that the name could be more generic as it doesn't do anything specific to Affiliation, but could be used to map any kind of values to other values. Maybe AttributeValueMap?

@jaimeperez, what do you think?

jaimeperez:
I fully agree :+1: 

I think this would be a useful addition as a more general filter that allows crafting attributes out of the values in others. But for that, the source and target attribute should be mandatory configuration options, and no defaults would be allowed.

I have also one doubt: what happens if the target attribute already exists?

mrvanes:
I thought long about the name and came up with this because there were a lot of collisions with existing filters (that I could come up with). Thijs' suggestion looks fine with me. I do realise the overwrite case is not covered. I could introduce an %overwrite flag to force overwriting existing attribute and append otherwise? In the other filters %replace is used for this, do you prefer consistency and create a different flag for not-keeping the source attribute? Have a a suggestion in that case?

modules/core/docs/authproc_generateaffiliation.txt
```diff
@@ -0,0 +1,61 @@
+`core:GenerateAffiliation`
+===================
+
+Filter that generate an attribute to the user based on value(s) in another attribute.
```
thijskh: *generates   *for the user

modules/core/lib/Auth/Process/GenerateAffiliation.php
```diff
@@ -0,0 +1,110 @@
...
+                unset($attributes[$this->memberattribute]);
+            }
+
+        } else {
+            SimpleSAML_Logger::warning('GenerateAffiliation - memberattribute does not exist: ' . $this->memberattribute);            
+        }
+    }
+}
+
+?>
```
thijskh: Please omit closing `?>` at EOF

PR #317, opened 2016-01-20
sqlite3 tests require the ext-sqlite3 module

br00k:
add sqlite3 as a dev dependency. you might not notice this when running
tests locally otherwise.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "whitehat101/apr1-md5": "~1.0"
     },
     "require-dev": {
+        "ext-sqlite3": "*",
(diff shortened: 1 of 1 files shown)


thijskh:
I've done some tests and it seems that what's actually required is the `pdo_sqlite` extension. This is contained in the same `php5-sqlite` Debian/Ubuntu package as `sqlite3` so nearly always either both or none are installed, but naming the actually required extension `pdo_sqlite` in `composer.json` is more correct.

thijskh:
Great, thanks!

PR #318, opened 2016-01-21
testing of core:AttributeLimit

br00k:
test core:AttributeLimit - since array can only contain int and string
I can’t raise the non covered constructor exception.


thijskh:
Great work!

PR #319, opened 2016-01-21
remove src_dir from .coveralls.yml

br00k:
this was added to fix a bug - now we need to remove it :-)


br00k:
travis-ci results now published back to coveralls https://coveralls.io/builds/4802382

PR #323, opened 2016-01-28
Azure does not support some Scoping Elements

OlivierP-VPG:
This patch aims to bypass the creation of the Element <Scoping> that is not fully supported by Azure Active Directory SAML Authentication Protocol (https://msdn.microsoft.com/en-US/library/azure/dn195589.aspx)

Scoping Element in AuthnRequest

The Scoping element, which includes a list of identity providers, is optional in AuthnRequest elements sent to Azure Active Directory.

Do not include the ProxyCount attribute, IDPList option, or the RequesterID element, in the Scoping element of an AuthnRequest sent to Azure Active Directory. They are not supported.

### DIFF
--- a/lib/SimpleSAML/Metadata/MetaDataStorageHandlerFlatFile.php
+++ b/lib/SimpleSAML/Metadata/MetaDataStorageHandlerFlatFile.php
@@ -113,7 +113,7 @@ class SimpleSAML_Metadata_MetaDataStorageHandlerFlatFile extends SimpleSAML_Meta
         foreach ($metadataSet as $entityId => &$entry) {
             if (preg_match('/__DYNAMIC(:[0-9]+)?__/', $entityId)) {
                 $entry['entityid'] = $this->generateDynamicHostedEntityID($set);
-            } else {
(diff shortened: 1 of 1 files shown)


thijskh:
Hi Olivier, I'm having trouble relating your proposed change with the problem described. Are you sure you added the right commit? It seems to me that Azure's lack of support for scoping should not relate to a specific metadata storage handler?

OlivierP-VPG:
Hi Thijs,
You're right, no direct relation with the metadata storage handler. Didn't have time to propose a better solution yet. Let me explain a bit more what I encountered:
When building an Authnrequest, the <Scoping> element is conditioned here [https://github.com/simplesamlphp/saml2/blob/master/src/SAML2/AuthnRequest.php#L735](url).
As an Idp hosted, the metadata  handler determine an entityid from the key of the array of metedata configuration. and this entityid is used as a "RequesterID".

Any idea to find a better solution would be much appreciated.

Thanks!

OlivierP-VPG:
Not a good approach

PR #324, opened 2016-01-28
detect SSL through proxy servers or load balancers

akinspe:
This pull request allows detecting SSL when running through a load balancer or proxy server.

It uses the standard X-Forwarded-Proto header to determine SSL termination.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -73,6 +73,11 @@ class HTTP
      */
     private static function getServerHTTPS()
     {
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @akinspe, thanks for your contribution!

However, we don't need such a feature, since it is already possible to tell SimpleSAMLphp it should build HTTPS URLs even though it's not currently running over it. You just need to specify the full URL (using `https://`, of course) in the `baseurlpath` configuration option. This is much more reliable, as it will always work regardless of what you have in front of your server(s).

PR #333, opened 2016-02-18
Fix misplaced </tr> (2x)

tvdijen:
The login-page would not verify as valid XHTML 1.0 Transitional


jaimeperez:
Thanks @tvdijen!

PR #334, opened 2016-02-22
Add AttributeValueMap Authentication Processing Filter and unittest. 

mrvanes:
Rewrite of previous pull request to cater for name change. Also, squashed a few bugs.


modules/core/lib/Auth/Process/AttributeValueMap.php
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+
+        if (!$this->keep) {
+            unset($attributes[$this->sourceattribute]);
+        }
+    }
+}
+
+?>
```
thijskh: Please omit closing `?>` before EOF.

mrvanes: Why is this bad practice?

thijskh: Procedural argument: because our coding standard says so:
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#22-files
More explanation:
http://php.net/manual/en/language.basic-syntax.phptags.php

jaimeperez: Leaving the closing `?>` makes it possible to add contents after it, even some that could go unnoticed, like newlines, spaces or tabs. Those characters will be printed to standard output, forcing PHP to send the headers. And once the headers have been sent, you can't send more headers.

So if something is printed by the end of a file and you want to make a redirection after importing that file, you won't be able. Imagine you have a trailing newline in a processing filter like this, and the result of the filter should be to redirect to another page. It won't work because headers were sent when you included the file.

Same applies to setting cookies and anything else that needs to send an HTTP header to the browser.

mrvanes: Very clear, won't happen again! :)

jaimeperez:
Thanks a lot @mrvanes! :smile: 

What do you think @thijskh? Ready to merge?

thijskh:
:+1:

jaimeperez:
Merged then! Thanks a lot again @mrvanes!

PR #335, opened 2016-02-23
Expose RegistrationInfo in parsed metadata

pradtke:
- Add test case
- Fix .gitignore since it ignored all metadata folders

Based on https://github.com/simplesamlphp/simplesamlphp/pull/288#discussion-diff-53787274

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 .gitignore
 log
 config
-metadata
+./metadata
 cert
 
 !config/.gitkeep
--- a/lib/SimpleSAML/Metadata/SAMLParser.php
+++ b/lib/SimpleSAML/Metadata/SAMLParser.php
@@ -124,6 +124,11 @@ class SimpleSAML_Metadata_SAMLParser
      */
     private $entityAttributes;
 
+    /**
+     * An associative array of attributes from the RegistrationInfo element.
+     * @var array
+     */
+    private $registrationInfo;
 
     /**
      * @var array
@@ -180,6 +185,7 @@ class SimpleSAML_Metadata_SAMLParser
         $this->scopes = $ext['scope'];
         $this->tags = $ext['tags'];
         $this->entityAttributes = $ext['EntityAttributes'];
+        $this->registrationInfo = $ext['RegistrationInfo'];
 
         // look over the RoleDescriptors
         foreach ($entityElement->RoleDescriptor as $child) {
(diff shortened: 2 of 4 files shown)


jaimeperez:
Great! Thanks @pradtke!

jaimeperez:
Uhm, maybe I was a bit too fast merging this.

What happens if the `<mdrpi:RegistrationInfo>` element is an extension to an `EntitiesDescriptor` element instead of an `EntityDescriptor`? For federation feeds it actually makes much more sense like that, so I think we must cover that case.

trscavo:
> What happens if the <mdrpi:RegistrationInfo> element is an extension to an EntitiesDescriptor element instead of an EntityDescriptor?

That is logically equivalent to having the same <mdrpi:RegistrationInfo> element on each EntityDescriptor.

> For federation feeds it actually makes much more sense like that, so I think we must cover that case.

No, every entity in eduGAIN metadata has its own <mdrpi:RegistrationInfo> element. As we move to per-entity metadata, that will solidify.

jaimeperez:
Hi Tom!

> That is logically equivalent to having the same mdrpi:RegistrationInfo element on each EntityDescriptor.

Exactly. And that's what we should do here, but my impression is that the PR does not deal with that use case.

> No, every entity in eduGAIN metadata has its own mdrpi:RegistrationInfo element. As we move to per-entity metadata, that will solidify.

Actually, we are talking about different things here. eduGAIN is not a federation, but a **confederation**. Since it joins many federations together by fetching their metadata and joining all the entities in one single, huge feed, there's no other way to use `RegistrationInfo` there. There's no separation between entities from different federations, so you have to mark them with a `RegistrationInfo` element inside every single `EntityDescriptor`.

For a metadata feed produced by a federation, that's unnecessary and a huge waste of space and resources. Every typical feed would be one `EntitiesDescriptor` element with many `EntityDescriptor` elements inside. As per your first comment, `RegistrationInfo` inside the `EntitiesDescriptor` element will apply then to all `EntityDescriptor` elements, which is reflecting reality exactly as it is: one single federation operator is both the one joining and publishing all those entities, and the one registering them, acting as an authoritative source.

My point is: regardless of whether your feed contains `mdrpi:RegistrationInfo` elements on `EntitiesDescriptor` or `EntityDescriptor` elements, SimpleSAMLphp should deal with both cases equally. And in the first case, the information should be "inherited" by all entities in the same feed.

pradtke:
Good catch. I'll take a look a making a new PR that handles inheritance from EntitiesDescriptor.

PR #336, opened 2016-02-24
RegistrationInfo can be inherited from EntitiesDescriptor

pradtke:
Allow RegistrationInfo to be inherited from parent EntitiesDescriptors.
Based on discussion https://github.com/simplesamlphp/simplesamlphp/pull/335#issuecomment-188178369

### DIFF
--- a/lib/SimpleSAML/Metadata/SAMLParser.php
+++ b/lib/SimpleSAML/Metadata/SAMLParser.php
@@ -159,11 +159,13 @@ class SimpleSAML_Metadata_SAMLParser
      * @param int|NULL                      $maxExpireTime The unix timestamp for when this entity should expire, or
      *     NULL if unknown.
      * @param array                         $validators An array of parent elements that may validate this element.
+     * @param array                         $parentExtensions An optional array of extensions from the parent element.
      */
     private function __construct(
         SAML2_XML_md_EntityDescriptor $entityElement,
         $maxExpireTime,
-        array $validators = array()
+        array $validators = array(),
+        array $parentExtensions = null
     ) {
         assert('is_null($maxExpireTime) || is_int($maxExpireTime)');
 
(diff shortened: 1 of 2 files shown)


jaimeperez:
Awesome @pradtke! Thanks so much! :smile: 

Just one more thing to make sure. Reading the MDRPI spec, I see this:

>36  The `<mdrpi:RegistrationInfo>` element MUST NOT appear more than once within the `<md:Ex-`
>37  `tensions>` element of a given `<md:EntitiesDescriptor>` or `<md:EntityDescriptor>` element.
>38  When a `<mdrpi:RegistrationInfo>` element appears in the `<md:Extensions>` element of a
>39  `<md:EntitiesDescriptor>` element it applies to all descendant `<md:EntitiesDescriptor>` and
>40  `<md:EntityDescriptor>` elements. That is to say, this is equivalent to putting an identical
>41  `<mdrpi:RegistrationInfo>` on every descendant `<md:EntityDescriptor>`. **When used in this**
>42  **manner, descendant `<md:EntitiesDescriptor>` and `<md:EntityDescriptor>` elements MUST**
>43  **NOT contain a `<mdrpi:RegistrationInfo>` element in their `<md:Extensions>` element.**

This is dealt with in the PR, right? I had a brief look at the patch and it looks like that, but I just wanted to confirm with you :grinning:

trscavo:
I was going to quote that passage yesterday but decided not to :-)

Note that the spec you quoted is not an implementation spec. An implementer might first check if there is an `<mdrpi:RegistrationInfo>` element on the `<md:Extensions>` element of an `<md:EntitiesDescriptor>` element, and if so, blindly apply it to each descendant `<md:EntityDescriptor>` element. I wouldn't call that wrong. A more robust implementation would check if the parent `<mdrpi:RegistrationInfo>` element matched the descendant `<mdrpi:RegistrationInfo>` element, and if not, log a warning. I would not raise an error condition if the two `<mdrpi:RegistrationInfo>` elements did not match. Just my two cents.

jaimeperez:
Actually I fully agree with you Tom, I think that should definitely trigger a warning, but not an error. Triggering an error would be very picky, and in the end that's some kind of error we can recover from (by just logging and ignoring), so yes, I take your two cents :wink:

pradtke:
The code (and tests) handles the override case gracefully (e.g. the resulting parsed data ignores the fact that you attempted an override).  I adjusted the log level to warn if the xml attempts an override that changes the registrationAuthority. If the XML just overrides with the same RegistrationAuthority value then no warning is logged.

jaimeperez:
Sounds great then. I see no reason not to merge it right away, so that's what I'm doing :smile: 

Thanks a lot again!

PR #339, opened 2016-03-01
SimpleSAML namespace usage in config.php

szabogyula:
Minor changes to use SimpleSAML namespace to Logger logging level definitions.


jaimeperez:
Thanks a lot @szabogyula!

PR #340, opened 2016-03-01
Several updates to the Radius module

thijskh:
* Code formatting
* Improve error messages
* Add new option `realm` that will be postfixed to the entered username


jaimeperez:
Bedankt Thijs!

Issue #337, opened 2016-02-29
Problem with port in method \SimpleSaml\Utils\HTTP::checkURLAllowed


andres-blanco:
Hey, I'am running an SP on http://localhost:4003. When I try to logout I get an exception that looks like this:

```
SimpleSAML_Error_Error: UNHANDLEDEXCEPTION
Backtrace:
0 /var/local/huarpe-core/vendor/simplesamlphp/simplesamlphp/www/module.php:170 (N/A)
Caused by: SimpleSAML_Error_Exception: URL not allowed: http://localhost:4003/
Backtrace:
2 /var/local/huarpe-core/vendor/simplesamlphp/simplesamlphp/lib/SimpleSAML/Utils/HTTP.php:332 (SimpleSAML\Utils\HTTP::checkURLAllowed)
1 /var/local/huarpe-core/vendor/simplesamlphp/simplesamlphp/modules/core/www/as_logout.php:18 (require)
0 /var/local/huarpe-core/vendor/simplesamlphp/simplesamlphp/www/module.php:127 (N/A)
```
After looking through the code we found that the method ``` \SimpleSaml\Utils\HTTP::getSelfHost``` is not returning the port, even tough in the function docs says this: ```* @return string The current host (with non-default ports included).```

Is this a bug?

jaimeperez:
Hola @andresote!

What value do you have in your `baseurlpath` configuration option? (`config/config.php`)

Also, can you verify the environment of your PHP script? Add a call to `error_log()` and monitor your apache error log to see what value does SimpleSAMLphp get from the environment, something like:

```php
private static function getServerPort()
    {
        $port = (isset($_SERVER['SERVER_PORT'])) ? $_SERVER['SERVER_PORT'] : '80';

        error_log("the port is: ".$port);

        if (self::getServerHTTPS()) {
            if ($port !== '443') {
                return ':'.$port;
            }
        } else {
            if ($port !== '80') {
                return ':'.$port;
            }
        }
        return '';
    }
```

andresote:
Hola @jaimeperez !

In the ```baseurlpath``` I have the default value: ```simplesaml/```.

The log in the ```getServerPort``` method returns: ```string(4) "4003"```

I think the problem is in the ```getSelfHost``` method:
```
    public static function getSelfHost()
    {
        $url = self::getBaseURL();

        $start = strpos($url, '://') + 3;
        // HERE. Why is the character ':' in the second parameter of strcspn?
[...]

PR #341, opened 2016-03-02
Fixes issue 337

andres-blanco:
This fixes the issue 337.
I added a function getSelfHostWithout port to retain the original behaviour. 
I've also added some unit tests.

Cheers

### DIFF
--- a/lib/SimpleSAML/Metadata/MetaDataStorageHandler.php
+++ b/lib/SimpleSAML/Metadata/MetaDataStorageHandler.php
@@ -207,11 +207,7 @@ class SimpleSAML_Metadata_MetaDataStorageHandler
         }
 
         // then we look for the hostname
-        $currenthost = \SimpleSAML\Utils\HTTP::getSelfHost(); // sp.example.org
-        if (strpos($currenthost, ":") !== false) {
-            $currenthostdecomposed = explode(":", $currenthost);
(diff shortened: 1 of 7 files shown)


thijskh:
I think this is a good approach to fix #337.

I'm wondering if it would not be marginally better to have getSelfHostWithoutPort call getSelfHost and chop off any port.

andres-blanco:
I think it's a matter of taste. For me, the normal behaviour is that the port is part of the host. 
But I think that you would be a better judge for the decision since you deal with hosts a lot more than me!
If you want, I'll make that change

jaimeperez:
¡Muchas gracias Andrés!

Strictly speaking, the host is the DNS name or IP address where you can reach the server, and the port is more related to the TCP endpoint, not the host itself.

In any case, I think the most important thing to keep in mind is coherence, and also try to be explicit. We already have a method called `getSelfHostWithPath()`, so it makes sense to have a `getSelfHostWithPort()`, better than a variant **without**. Additionally, the method will not display standard ports (which I think makes sense), so I'd better be explicit about that, renaming it to `getSelfHostWithNonStandardPort()`. Quite long, but no room for misunderstandings :smile: 

I've done this changes myself, adapted (and completed) the test, and merged everything into master. Please confirm that the final code fixes issue #337.

andres-blanco:
Everything works fine! 

Cheers!

PR #345, opened 2016-03-08
Changes for PHP 7.

brentboghosian:


thijskh:
Hi Brent, thanks! About the change for set_magic_quotes_runtime(), I'm not sure that's right. According to the documentation, in PHP 5.3.x this function is still relevant, so I don't think it should only be invoked when the PHP version < 5.3. Also, in PHP 7 get_magic_quotes_runtime is still there but will always return false, so why is the extra wrap with version_compare necessary?

brentboghosian:
On Tue, Mar 8, 2016 at 3:11 PM, Thijs Kinkhorst <notifications@github.com>
wrote:

> Hi Brent, thanks! About the change for set_magic_quotes_runtime(), I'm not
> sure that's right. According to the documentation, in PHP 5.3.x this
> function is still relevant, so I don't think it should only be invoked when
> the PHP version < 5.3. Also, in PHP 7 get_magic_quotes_runtime is still
> there but will always return false, so why is the extra wrap with
> version_compare necessary?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/simplesamlphp/simplesamlphp/pull/345#issuecomment-193948259>
> .
>
Very good point.
Would you like me to remove that change and resubmit the pull request,
without it?

Cheers!
-- 

Brent Boghosian           brent.boghosian@remote-learner.net
Lead Software Architect   http://www.remote-learner.net/
519-772-3910 x176         Open Source * Moodle Hosting/Development

thijskh:
You can remove that locally and force-push to your fork, so this PR will be updated.

brentboghosian:
On Wed, Mar 9, 2016 at 1:46 AM, Thijs Kinkhorst <notifications@github.com>
wrote:

> You can remove that locally and force-push to your fork, so this PR will
> be updated.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/simplesamlphp/simplesamlphp/pull/345#issuecomment-194141478>
> .
>
Ok - it should be done.


-- 

Brent Boghosian           brent.boghosian@remote-learner.net
Lead Software Architect   http://www.remote-learner.net/
519-772-3910 x176         Open Source * Moodle Hosting/Development

thijskh:
Thanks! You may want to consider to send your change to the Yubico-class also in a PR to the official "upstream" for this file: https://github.com/Yubico/php-yubico/blob/master/Yubico.php

PR #347, opened 2016-03-15
AllowCreate is hardcoded

AngelAlvarado:
This PR allows you to set AllowCreatePolicy (values TRUE or FALSE) in a Service Provider configured in the authsources.php. For example: 

> 'taxnotes-dev' => array(
        'saml:SP',
        'certificate' => 'D...',
        'entityID' => 'X...',
        'AllowCreatePolicy' => FALSE,
        'NameIDPolicy' => 'urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress',
    ),

Which will look like: 

>  < samlp:NameIDPolicy Format="urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress" AllowCreate="false" />

This is the second step in allowing Service Providers to configure the AllowCreate propriety in the NameIdPolicy in order to meet some IdP's specifications.

In order for this patch to work, this PR has to be accepted:  [SAML2](https://github.com/simplesamlphp/saml2/pull/63). The [SAML2 library](https://github.com/simplesamlphp/saml2/blob/master/src/SAML2/AuthnRequest.php) makes a validation for the AllowCreate property which does not accept anything but 'true'.

This is a use case I faced while implementing SSO for the [SPNEGO](https://en.wikipedia.org/wiki/SPNEGO) Identity Provider. 

Due the SPNEGO's specification you cannot logout a SPNEGO user gracefully **and** when trying to login again(after the user logs out) the user will get an exception. Therefore, it is necesary to send “allowcreate=false” in the SAML request token in order to let the SPNEGO IdP know that it is 'OK' to not refresh/try to authenticate the user in every SAML insertion.

Note: I'm creating this PR against the 1.13 version since I have only fully tested this version. 1.14/master seems to be a candidate for this 'enhancement'.

### DIFF
--- a/modules/saml/lib/Message.php
+++ b/modules/saml/lib/Message.php
@@ -402,10 +402,16 @@ class sspmod_saml_Message {
 			$nameIdPolicy = $spMetadata->getString('NameIDFormat', SAML2_Const::NAMEID_TRANSIENT);
 		}
 
+		if ($spMetadata->hasValue('AllowCreatePolicy')) {
+			$allowCreatePolicy = $spMetadata->getBoolean('AllowCreatePolicy', TRUE);
+		} else {
+			$allowCreatePolicy = TRUE;
+		}
+
 		if ($nameIdPolicy !== NULL) {
 			$ar->setNameIdPolicy(array(
 				'Format' => $nameIdPolicy,
-				'AllowCreate' => TRUE,
+				'AllowCreate' => $allowCreatePolicy,
 			));
 		}
 


AngelAlvarado:
Since simpleSAMLphp requires saml2 ~1.9. This is the [PR](https://github.com/simplesamlphp/saml2/pull/64) needed on that side in order for this patch to work. I confirm that the branch master works as expected with this patch. It should be shipped for 1.14 too. Ok to create two more PRs?

jaimeperez:
¡Hola de nuevo!

Thanks a lot for the PR!

Apart from what I told already in simplesamlphp/saml2#63, I think we need to start re-thinking the entire configuration files. We are adding more and more options for everything and it's starting to get messy. Not to talk about the fact that we are using a configuration option named *NameIDPolicy* to define just the format, while the `NameIDPolicy` element in SAML authentication requests carries more stuff like the `AllowCreate` you are mentioning or the `SPNameQualifier`.

So I'm thinking this *NameIDPolicy* option should be an array, specifying each of those elements if needed, with defaults equal to the ones we have today. Then we would avoid creating a new configuration option `AllowCreatePolicy`, and we could even deprecate and get rid of `SPNameQualifier` in IdP remote metadata (which is also inconsistent, since you cannot define that in the SP local configuration). The resulting configuration would be something like this:

```php
    ...
    'NameIDPolicy => array(
        'Format' => SAML2_Const::NAMEID_TRANSIENT,
        'AllowCreate' => false,
        'SPNameQualifier' => 'something',
    ),
    ...
```

I think that makes much more sense. Do you agree? If so, could you please apply this patch on top of master and tell me if it works for you?

```patch
diff --git a/modules/saml/lib/Message.php b/modules/saml/lib/Message.php
index 04318e9..685be4f 100644
--- a/modules/saml/lib/Message.php
+++ b/modules/saml/lib/Message.php
@@ -403,18 +403,29 @@ class sspmod_saml_Message {

 		$ar = new SAML2_AuthnRequest();

-		if ($spMetadata->hasValue('NameIDPolicy')) {
-			$nameIdPolicy = $spMetadata->getString('NameIDPolicy', NULL);
-		} else {
-			$nameIdPolicy = $spMetadata->getString('NameIDFormat', SAML2_Const::NAMEID_TRANSIENT);
+		// get the NameIDPolicy to apply. IdP metadata has precedence.
+		$nameIdPolicy = array();
+		if ($idpMetadata->hasValue('NameIDPolicy')) {
+			$nameIdPolicy = $idpMetadata->getValue('NameIDPolicy');
+		} elseif ($spMetadata->hasValue('NameIDPolicy')) {
+			$nameIdPolicy = $spMetadata->getValue('NameIDPolicy');
+		}
+
+		if (!is_array($nameIdPolicy)) {
+			// handle old configurations where 'NameIDPolicy' was used to specify just the format
+			$nameIdPolicy = array('Format' => $nameIdPolicy);
 		}

-		if ($nameIdPolicy !== NULL) {
-			$ar->setNameIdPolicy(array(
-				'Format' => $nameIdPolicy,
-				'AllowCreate' => TRUE,
-			));
+		$nameIdPolicy_cf = SimpleSAML_Configuration::loadFromArray($nameIdPolicy);
+		$policy = array(
+			'Format' => $nameIdPolicy_cf->getString('Format', SAML2_Const::NAMEID_TRANSIENT),
+			'AllowCreate' => $nameIdPolicy_cf->getBoolean('AllowCreate', true),
+		);
+		$spNameQualifier = $nameIdPolicy_cf->getString('SPNameQualifier', false);
+		if ($spNameQualifier !== false) {
+			$policy['SPNameQualifier'] = $spNameQualifier;
 		}
+		$ar->setNameIdPolicy($policy);

 		$ar->setForceAuthn($spMetadata->getBoolean('ForceAuthn', FALSE));
 		$ar->setIsPassive($spMetadata->getBoolean('IsPassive', FALSE));
diff --git a/modules/saml/lib/Message.php b/modules/saml/lib/Message.php
index 685be4f..8c8b821 100644
--- a/modules/saml/lib/Message.php
+++ b/modules/saml/lib/Message.php
@@ -700,7 +700,8 @@ class sspmod_saml_Message {
 			break;
 		}
 		if (!$found) {
-			throw new SimpleSAML_Error_Exception('Error validating SubjectConfirmation in Assertion: ' . $lastError);
+			SimpleSAML\Logger::warning('No SubjectConfirmation elements were found in Subject, or could not validate them.');
+			//throw new SimpleSAML_Error_Exception('Error validating SubjectConfirmation in Assertion: ' . $lastError);
 		}

 		/* As far as we can tell, the assertion is valid. */
```

AngelAlvarado:
¡Hola!

Indeed, the entire configuration files should be refactored. It does make a lot of sense to change the whole NameIDPolicy into an array. I applied your patch.
Using:

```
'NameIDPolicy => array(
        'Format' => SAML2_Const::NAMEID_TRANSIENT,
        'AllowCreate' => false,
        'SPNameQualifier' => 'something',
    ),
```
This is what the SAML token looks like:

```
<samlp:NameIDPolicy Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient"
                        SPNameQualifier="something"
                        />
```
I guess [saml2/pull/63](https://github.com/simplesamlphp/saml2/pull/63) can help here. Or, my SP+IdP are misconfigured? [These are my configuration files btw](https://github.com/AngelAlvarado/continuous-delivery-saml-sso/tree/master/cnf/SAML)

jaimeperez:
Hola de nuevo,

That's what I was expecting indeed. When you send an authentication request without `AllowCreate` in the `NameIDPolicy` element, does it fix your issue with SPNEGO?

In case not, is by any chance the IdP implemented with SimpleSAMLphp + the _negotiate_ module?

AngelAlvarado:
[SPNEGO](https://msdn.microsoft.com/en-us/library/ms995331.aspx)-GSS-API/Kerberos is only implemented by Microsoft.  I'll check if the absence of `AllowCreate` solves this issue.

jaimeperez:
Great, thanks! :smile:

jaimeperez:
Ángel, did it work?

AngelAlvarado:
Hey Jaime! 

Apologies for the late response, I got sidetracked. Indeed the IdP works as expected. It is not necessary to set the AllowCreate=false explicitly.

jaimeperez:
That's great, thanks for the feedback! I'll merge the patch asap :smile:

PR #349, opened 2016-03-18
transfer-session-variables-when session-already-exists

gvanimurthy:
Since SAML is creating a new session to authenticate user, we lose the session variables we have created before authentication.

This allows us to retrieve the session variables from unauthenticated session to use them in authenticated session.

Our use case: We are dealing with various forms and storing them in session object in order to access them across pages. This information is crucial and is wiped off if user chooses to authenticate himself after completing forms.

### DIFF
--- a/lib/SimpleSAML/SessionHandlerPHP.php
+++ b/lib/SimpleSAML/SessionHandlerPHP.php
@@ -87,17 +87,42 @@ class SimpleSAML_SessionHandlerPHP extends SimpleSAML_SessionHandler
         SimpleSAML_Session::createSession($sessionId);
 
         if (session_id() !== '') {
-            // session already started, close it
-            session_write_close();
+            // session already started, transfer it
+            self::update_session($sessionId);
         }
-
-        session_id($sessionId);
-        session_start();
+        else {
+            session_id($sessionId);
+            session_start();
+        }
 
         return session_id();
     }
 
+    public function update_session($newsessid = '') {
 
+        // Backup the current session
+        $session_backup = $_SESSION;
+
+        // Set current session to expire in 1 minute
+        $_SESSION['OBSOLETE'] = true;
+        $_SESSION['EXPIRES'] = time() + 60;
+
+        // Close the current session
+        session_write_close();
+
+        // Set a new session id and start the session
+        $newSession = session_id($newsessid);
+        session_start();
+
+        // Restore the previous session backup
+        $_SESSION = $session_backup;
+
+        // Clean up
+        unset($session_backup);
+        unset($_SESSION['OBSOLETE']);
+        unset($_SESSION['EXPIRES']);
+    }
+    
     /**
      * Retrieve the session ID saved in the session cookie, if there's one.
      *


jaimeperez:
Thanks @gvanimurthy!

I'm trying to reproduce the issue with master, by setting some value inside `$_SESSION` in some page, then authenticating, and checking for the same value inside `$_SESSION` afterwards. The value survives authentication as expected.

What version of SimpleSAMLphp are you using, and how are you experiencing this issue?

gvanimurthy:
We are using version 2.0.

Session value survives if we load saml library and let it create & handle the session.

Our case, we arent loading saml unless user chooses to login. Hence session_start() in our code is creating an initial session. When user clicks on Login, we load saml library. 
That case according to the following code in SessionHandlerPHP.php ,our session is being closed and saml session generated by `$sessionId = SimpleSAML_Utilities::stringToHex(SimpleSAML_Utilities::generateRandomBytes(16));` is started.

```
if (session_id() !== '') {
			/* Session already started, close it. */
			session_write_close();
} 
```
Hence at this point, if we transferred the session data, our values survives

thijskh:
> We are using version 2.0.

Are we talking about simpleSAMLphp here, or the saml2 library? You pull request is against the former, and that has no version 2.0 yet (the latest is 1.14.2).

gvanimurthy:
apologies...we are using simplesamlphp version 1.13.2

jaimeperez:
Hi again @gvanimurthy! Sorry for the delay.

I've been looking a bit more into this, and actually the problem is a bug in our PHP session handler. Instead of managing its own session, it is assuming there is no session already initialised. When it discovers there's already a session, it tries to load it and fails (obviously), and ends up creating a new session with the same identifier of the existing one. The session gets overwritten, and its contents lost.

I really don't think it's a good idea to allow applications to mess with the same session SimpleSAMLphp is using. Therefore, the proper solution is not to _copy_ the data over to the new session, IMHO. The proper solution would be to fix the bug in the PHP session handler, making it close any existing sessions and create or load its own session with an independent name and identifier. It won't be transparent for application developers, but I think it's a much safer approach.

Could you please apply the following patch (it's from **master**, though I think it might work too with 1.13.2) and tell me if it works for you?

```patch
diff --git a/lib/SimpleSAML/SessionHandlerPHP.php b/lib/SimpleSAML/SessionHandlerPHP.php
index da031eb..34b03ec 100644
--- a/lib/SimpleSAML/SessionHandlerPHP.php
+++ b/lib/SimpleSAML/SessionHandlerPHP.php
@@ -29,35 +29,33 @@ class SimpleSAML_SessionHandlerPHP extends SimpleSAML_SessionHandler
         // call the parent constructor in case it should become necessary in the future
         parent::__construct();

-        /* Initialize the php session handling.
-         *
-         * If session_id() returns a blank string, then we need to call session start. Otherwise the session is already
-         * started, and we should avoid calling session_start().
-         */
-        if (session_id() === '') {
-            $config = SimpleSAML_Configuration::getInstance();
-
-            $params = $this->getCookieParams();
-
-            session_set_cookie_params(
-                $params['lifetime'],
-                $params['path'],
-                $params['domain'],
-                $params['secure'],
-                $params['httponly']
-            );
+        if (session_status() === PHP_SESSION_ACTIVE) {
+            // we shouldn't have a session at this point, so it might be an application session. Write & close.
+            session_write_close();
+        }

-            $this->cookie_name = $config->getString('session.phpsession.cookiename', null);
-            if (!empty($this->cookie_name)) {
-                session_name($this->cookie_name);
-            } else {
-                $this->cookie_name = session_name();
-            }
+        $config = SimpleSAML_Configuration::getInstance();

-            $savepath = $config->getString('session.phpsession.savepath', null);
-            if (!empty($savepath)) {
-                session_save_path($savepath);
-            }
+        $params = $this->getCookieParams();
+
+        session_set_cookie_params(
+            $params['lifetime'],
+            $params['path'],
+            $params['domain'],
+            $params['secure'],
+            $params['httponly']
+        );
+
+        $this->cookie_name = $config->getString('session.phpsession.cookiename', null);
+        if (!empty($this->cookie_name)) {
+            session_name($this->cookie_name);
+        } else {
+            $this->cookie_name = session_name();
+        }
+
+        $savepath = $config->getString('session.phpsession.savepath', null);
+        if (!empty($savepath)) {
+            session_save_path($savepath);
         }
     }

@@ -107,37 +105,38 @@ class SimpleSAML_SessionHandlerPHP extends SimpleSAML_SessionHandler
      */
     public function getCookieSessionId()
     {
-        if (session_id() === '') {
-            if (!self::hasSessionCookie()) {
-                return null;
-            }
+        if (!self::hasSessionCookie()) {
+            return null; // there's no session cookie, can't return ID
+        }

-            $session_cookie_params = session_get_cookie_params();
+        // do not rely on session_id() as it can return the ID of a previous session. Get it from the cookie instead.
+        session_id($_COOKIE[$this->cookie_name]);

-            if ($session_cookie_params['secure'] && !\SimpleSAML\Utils\HTTP::isHTTPS()) {
-                throw new SimpleSAML_Error_Exception('Session start with secure cookie not allowed on http.');
-            }
+        $session_cookie_params = session_get_cookie_params();

-            $cacheLimiter = session_cache_limiter();
-            if (headers_sent()) {
-                /*
-                 * session_start() tries to send HTTP headers depending on the configuration, according to the
-                 * documentation:
-                 *
-                 *      http://php.net/manual/en/function.session-start.php
-                 *
-                 * If headers have been already sent, it will then trigger an error since no more headers can be sent.
-                 * Being unable to send headers does not mean we cannot recover the session by calling session_start(),
-                 * so we still want to call it. In this case, though, we want to avoid session_start() to send any
-                 * headers at all so that no error is generated, so we clear the cache limiter temporarily (no headers
-                 * sent then) and restore it after successfully starting the session.
-                 */
-                session_cache_limiter('');
-            }
-            session_start();
-            session_cache_limiter($cacheLimiter);
+        if ($session_cookie_params['secure'] && !\SimpleSAML\Utils\HTTP::isHTTPS()) {
+            throw new SimpleSAML_Error_Exception('Session start with secure cookie not allowed on http.');
         }

+        $cacheLimiter = session_cache_limiter();
+        if (headers_sent()) {
+            /*
+             * session_start() tries to send HTTP headers depending on the configuration, according to the
+             * documentation:
+             *
+             *      http://php.net/manual/en/function.session-start.php
+             *
+             * If headers have been already sent, it will then trigger an error since no more headers can be sent.
+             * Being unable to send headers does not mean we cannot recover the session by calling session_start(),
+             * so we still want to call it. In this case, though, we want to avoid session_start() to send any
+             * headers at all so that no error is generated, so we clear the cache limiter temporarily (no headers
+             * sent then) and restore it after successfully starting the session.
+             */
+            session_cache_limiter('');
+        }
+        session_start();
+        session_cache_limiter($cacheLimiter);
+
         return session_id();
     }
```

Please also note that **after calling SimpleSAMLphp**, you will need to close the existing session, and reload yours by calling `session_id()` and `session_start()`:

```php
<?php

session_name('mysession');
session_start();
$id = session_id();

if (!isset($_SESSION['foo'])) {
    $_SESSION['foo'] = 'bar';
}

include '../_include.php';
$auth = new SimpleSAML_Auth_Simple('default-sp');
$auth->requireAuth();

header("Content-Type: text/plain");
session_commit();
session_name('mysession');
session_id($id);
session_start();
echo "Session ID: ".session_id()."\n";
echo "\nSession contents: \n".var_export($_SESSION, true);
```

jaimeperez:
Did you try the patch, @gvanimurthy?

xwan510:
my application was affected by this bug as well, so im excited when i saw this report finally gets progress.

I have tested your patch and it's all good except there is a minor problem, that i have to work around it.

My application's session cookie extensively uses all  parameters: lifetime, path, domain etc.
However, in your code, when it is about to create a new session, it just overrides the my application's session cookie by using the same cookie name, and it calls getCookieParams() which ONLY grabs pre-defined params in config file. But i don't have any thing customised in the config file, so eventually the cookie was incorrectly set with a wrong "path",  that breaks my application.


Can you please change your code a bit, so, if it ever tries to override the session cookie,  it should grab the params from the current session, not the config file. 
If there is a reason that we must enforce the settings in config file, at least we should not override the current session cookie, and use a new name.

so it probably should go along like below
`
  
        $this->cookie_name = $config->getString('session.phpsession.cookiename', null);
        $params = null;
        if (!empty($this->cookie_name)) {
            session_name($this->cookie_name);
           $params = $this->getCookieParams();
        } else {
            $this->cookie_name = session_name();
           // if we are overriding the cookie, use the same params
           $params = session_get_cookie_params();
        }

      session_set_cookie_params(
            $params['lifetime'],
            $params['path'],
            $params['domain'],
            $params['secure'],
            $params['httponly']
        );


`

jaimeperez:
Thanks for the feedback @xwan510!

Actually, SimpleSAMLphp was not overriding the session cookie (as per the patch), but loading its own session. **And it should definitely not use the cookie parameters from the existing session**. If we do that, the settings in the configuration are useless, and an application could break SimpleSAMLphp by setting cookie parameters that are valid for it, but not for SSP.

Unfortunately, there's no easy, transparent way to fix this. Due to the nature of PHP sessions, it's not possible to have two sessions loaded at the same time. Therefore, the only way to go is to save all the session details of the current session before loading or creating one in the context of SimpleSAMLphp. If you need to access your session again **after** invoking SSP, then you will need to restore it manually. For convenience, a `cleanup()` method has been added to `SimpleSAMLphp_Session`, so that the previous session is restored after ensuring SSP's session is properly saved. Please refer to the commit to see how to use this new method. The documentation will be automatically refreshed in a few hours with this information too.

jaimeperez:
Just for completeness, here's a simple PHP script that can be used to verify that this fix works (as well as illustrate how to restore the previous session).

```php
<?php

// configure according to your needs
$path = '/path/to/your/script.php';
$host = 'your.machine.com';
$secure = true || false;
$session_name = 'name_of_the_cookie',
$ssproot = '/path/to/simplesamlphp';

// create our own session
session_set_cookie_params(0, $path, $host, $secure);
session_name($session_name);
session_start();
$id = session_id();

// if our session was not initialized or there's no attributes, require auth
if (!array_key_exists('foo', $_SESSION) || !array_key_exists('attributes', $_SESSION)) {
    $_SESSION['foo'] = 'bar';
    include $ssproot.'www/_include.php';
    $auth = new SimpleSAML_Auth_Simple('default-sp');
    $auth->requireAuth();

    // restore our own session
    $session = SimpleSAML_Session::getSessionFromRequest();
    $session->cleanup();

    // save the attributes to the session
    $_SESSION['attributes'] = $attributes;

    // test that SSP's session is still working
    $session->doLogout("default-sp");
}

// print details about *our* session
header("Content-Type: text/plain; charset=utf-8");
echo "Session Name: ".session_name()."\n\n";
echo "Session ID: ".session_id()."\n";
echo "\nSession contents: \n".var_export($_SESSION, true);
```

PR #350, opened 2016-03-18
Bugfix for empty attributes array

tvdijen:
An empty array shouldn't lead to an aborted logon.
Example when we try to get the 'mail' attribute:
UserA has a value set in the mail-attribute and passes the test -> Negotiate logon succeeds
UserB has an empty mail-attribute and fails the test -> Negotiate logon fails and the user has to log on manually (fallback to LDAP is initiated)

### DIFF
--- a/modules/negotiate/lib/Auth/Source/Negotiate.php
+++ b/modules/negotiate/lib/Auth/Source/Negotiate.php
@@ -151,7 +151,7 @@ class sspmod_negotiate_Auth_Source_Negotiate extends SimpleSAML_Auth_Source
                 $user = $auth->getAuthenticatedUser();
                 SimpleSAML\Logger::info('Negotiate - authenticate(): '.$user.' authenticated.');
                 $lookup = $this->lookupUserData($user);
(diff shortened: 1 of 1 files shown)


thijskh:
Because lookupUserData() either returns "something with attributes" or `null`, would it be clearer to make the check `$lookup !==null`?

tvdijen:
Yeah, either that or is_array($lookup)...
I'm not sure what I was thinking when I came up with that :')
Which one would you preffer?

thijskh:
I would say `!== null` because `null` is exactly that what lookupUserData returns when the function did not succeed.

tvdijen:
Alright, I'll close this PR and make a new one

thijskh:
You can fix your commit locally and "force push" it to your branch, that will update the PR.

tvdijen:
I'm not working locally.. I make my changes on the github.com site...
By the way, should I make a separate PR for the typo's? That accidentally ended up in this PR too...

thijskh:
No, just include those.

PR #352, opened 2016-03-21
Crude translation

thijsbraem:
"Vette pech" (fat chance) appears to offend some of our users. The translations for these seem to vary wildly:
* dr: "pas de chance" (no luck)
* de: "Tut uns leid" (we are sorry)
* it, da,... : this part is left out. This is the best option in my opinion.


thijskh:
A very reasonable request. The translation seems to stem from the original English string "Too bad!". Not sure if that adds much there as well.

jaimeperez:
Thanks for fixing this guys.

At some point we really need to go through the strings and review them thoroughly. Most of them are way too informal, and in many cases not even useful at all.

PR #354, opened 2016-03-25
Fix missing curly bracket

tvdijen:
Fixes: Mar 25 11:06:53 simplesamlphp INFO [0015a148dc] Template: Looking up [{login:processing]: not translated at all.


thijskh:
Good catch, thanks!

tvdijen:
You're welcome.. I have noticed another thing, that's not really worth making an issue for.
When debug logging is on, the line "Valid session found with '<authsource>'" is printed twice when a login occurs.. From what I can see, this is because the function that prints it, isValid, is called twice during the login proces. The second call is done by getAttributes(). I think the logging-line should be moved to the end of the doLogin function. What do you think about that?

PR #357, opened 2016-04-05
Added SIR attribute mappings

lonoak:
These mappings are used by former RedIRIS PAPI IdPs, and most have an equivalent to existing names, with the only exception of iMMA and IMAA (irisMailMainAddress and irisMailAlternateAddress), which are part of the iriseduPerson, defined here:

http://wiki.rediris.es/gtschema/Iriseduperson


thijskh:
Thanks! Can you please remove the closing `?>` as per the coding guidelines?

lonoak:
Added missing sir2urn file, removed enclosing ?> and duplicated sUS attribute

jaimeperez:
After discussing this a bit offline, I decided not to merge this PR, since the attributes added here are legacy attributes that nobody will use in the not-so-distant future. Therefore, I don't think we should incorporate them into SSP now.

At the same time, it makes sense to provide a solution for use cases like this, so basically [I added the possibility to put attribute mapping files in the `attributemap` directory of a module](https://github.com/simplesamlphp/simplesamlphp/commit/f0d848d7883f1bcaf26bf0124a493563dfbe7b83) (which we can install with composer), and [refer to it with the `module:file` syntax](https://simplesamlphp.org/docs/development/core:authproc_attributemap).

This is now available in master.

PR #358, opened 2016-04-06
Add attribute mappings for eduPersonOrcid

NicolasLiampotis:
This adds the attribute mappings for [eduPersonOrcid](http://software.internet2.edu/eduperson/internet2-mace-dir-eduperson-201602.html#eduPersonOrcid) (defined in eduPerson 201602).


dictionaries/attributes.definition.json
```diff
@@ -53,6 +53,9 @@
 	"attribute_edupersonuniqueid": {
 		"en": "Person's non-reassignable, persistent pseudonymous ID at home organization"
 	},
+	"attribute_edupersonorcid": {
+		"en": "ORCID researcher identifiers belonging to the principal"
```
thijskh: This may be a nitpick, but maybe this should just be "ORCID researcher identifiers", to be more in line with (most) other descriptions in this file (e.g. "Telephone number", not "Telephone number belonging to the principal").

thijskh:
Otherwise, very much in favour of merging. Thanks!

NicolasLiampotis:
> This may be a nitpick, but maybe this should just be "ORCID researcher identifiers", to be more in line with (most) other descriptions in this file (e.g. "Telephone number", not "Telephone number belonging to the principal").

@thijskh thanks for the suggestion. I've updated the English definition and Greek translation so they should now be more user-friendly and in line with the other attribute descriptions.

jaimeperez:
Thanks a lot Nicolas!

PR #362, opened 2016-04-08
IMAPauth module

bilbolodz:
Module to authenticate user against IMAP server. Created for use mainly with filesender software but maybe it can be used in other projects. It's my first simplesamlphp so please don't be to cruel ;-)


jaimeperez:
Hi @bilbolodz, thanks a lot for your contribution!

I think this is a quite interesting module, actually! However, our policy lately has been to try to slim down SimpleSAMLphp, especially removing modules that are not going to be used by the vast majority of our users.

So I would suggest creating a repository for this module on its own, [make it installable with composer](https://github.com/simplesamlphp/composer-module-installer/blob/master/README.md), and telling us so that we can include it in the [list of known modules](https://simplesamlphp.org/modules). Then everybody will be able to find it and install it with composer in their existing SimpleSAMLphp installation. Another advantage would be that you will then be able to update your module with new features or bugfixes as often as you wish, without depending on us at all.

If this sounds fine, just let me know if we can help you somehow with the process :smile: 

Meanwhile, I'll give you some feedback in the code itself :wink:

modules/imapauth/default-disable
```diff
@@ -0,0 +1,3 @@
+This file indicates that the default state of this module
```
jaimeperez: If you make this a module installable with composer, I would recommend to remove this and add a `default-enable` file, as then modules will be enabled as soon as they are installed, minimizing the chances for misunderstandings or trouble.

modules/imapauth/docs/imapauth.txt
```diff
@@ -0,0 +1,48 @@
+ImapAuth module
```
jaimeperez: This could also be the `README.md` in the top of your repository, so that users can see how to install and use your module at first sight. Pretty much like [others](https://github.com/NIIF/simplesamlphp-module-attributeaggregator) are doing.

bilbolodz:
OK, so I'll read about creating composer installable modules

modules/imapauth/lib/Auth/Source/MyAuth.php
```diff
@@ -0,0 +1,173 @@
+<?php
+class sspmod_imapauth_Auth_Source_MyAuth extends sspmod_core_Auth_UserPassBase {
```
jaimeperez: If you want, you can also use namespaces in your module, so that you don't need to give classes names that long. I also recommend to give every class a meaningful name. This could be, for example:

```php
<?php

namespace SimpleSAML\Module\imapauth\Auth\Source;

class IMAP extends sspmod_core_Auth_UserPassBase 
{
    ...
```

(we are slowly migrating all the classes to namespaces, so that means many of them are not there yet, like `sspmod_core_Auth_UserPassBase`)

modules/imapauth/lib/Auth/Source/MyAuth.php
```diff
@@ -0,0 +1,173 @@
...
+
+		if (!is_string($config['imap_port'])) {
+			throw new Exception('Missing or invalid imap_port option in config.');
+		}
+		$this->imap_port = $config['imap_port'];
+
+		if (!is_string($config['imap_security'])) {
+			throw new Exception('Missing or invalid imap_security option in config.');
+		}
+		$this->imap_security = $config['imap_security'];
```
jaimeperez: Very thorough configuration, nice!

modules/imapauth/lib/Auth/Source/MyAuth.php
```diff
@@ -0,0 +1,173 @@
...
+
+		if (is_string($config['imap_additional_options'])) {
+		$this->imap_additional_options = $config['imap_additional_options'];
```
jaimeperez: Wrong indentation?

modules/imapauth/lib/Auth/Source/MyAuth.php
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	protected function login($username, $password) {
+
+	//Defaults if there is no database entry
```
jaimeperez: Also here. We try to enforce PSR-2 these days, using spaces instead of tabs, having the opening braces after the function definition, spaces around `=` sign and things like that. It usually helps if you have an IDE that understands coding standards, so that you can configure it with one coding standard and make it check (and even fix) these kind of things for you.

modules/imapauth/lib/Auth/Source/MyAuth.php
```diff
@@ -0,0 +1,173 @@
...
+	
+	$email=$username . "@" . $this->mail_host;
+	$email_imap = "";
+	$name=$username;
+	$uid=$username;
+
+	$imap = $this->imap_hostname . ":" . $this->imap_port . "/" . $this->imap_security;
+	if (isset($this->imap_additional_options)) $imap= $imap . $this->imap_additional_options;
```
jaimeperez: This is another example, `if` statements should always use braces, even if it's only one statement to be executed.

modules/imapauth/lib/Auth/Source/MyAuth.php
```diff
@@ -0,0 +1,173 @@
...
+imap_close($sess); 
+	}
+
+	$uid=$username . "_" . $this->mail_host;
+	$user_id=$username . "_" . $this->mail_host;
+	$name=$username;
+
+	if ($email_imap != "") $email=$email_imap;
+
+if ($this->use_rc_database == true)
```
jaimeperez: I'm assuming this is to retrieve attributes for the user that you can't get via IMAP, right?

In that case, would it make sense to have this as a separate authentication processing filter, instead of the auth source? That way, those only interested in authentication could just use this, auth source, and if they want additional attributes they can use the authproc filter. And instead of using a configuration option for this (`use_rc_database`), they would just add the appropriate authproc filter if they want to use it.

Does it sound reasonable?

bilbolodz: Where should I read about authproc filter?

jaimeperez: [Here](https://simplesamlphp.org/docs/stable/simplesamlphp-authproc) (by the end of the page, mostly).

modules/imapauth/lib/Auth/Source/MyAuth.php
bilbolodz: I've started to work on new module called unixauth which implements a authentication with a few protocols (at the beginning SSH and IMAP) and tried to use "a new namespace schema" but got an error:

<?php
namespace SimpleSAML\Module\unixauth\Auth\Source;
class UNIX extends sspmod_core_Auth_UserPassBase {

error:
PHP Fatal error:  Class 'SimpleSAML\\Module\\unixauth\\Auth\\Source\\sspmod_core_Auth_UserPassBase' not found in /usr/share/simplesamlphp/modules/unixauth/lib/Auth/Source/MyAuth.php on line 7, referer: https://192.168.110.3/simplesamlphp/module.php/core/authenticate.php

I don't quite understand what's wrong

jaimeperez: Don't worry, it's a very typical error among those who use namespaces for the first time 😉 

When you declare a namespace with the `namespace` keyword, that namespace applies to the entire file, not only to classes or interfaces defined there. Therefore, all the names used or declared in the file are relative to that namespace. The concept is quite similar to a file system. You may have a file in the root of your filesystem, named *file*. However, if you change your current directory to somewhere else with `cd /some/path`, then *file* is not in there of course, and to access it you need to give an absolute path: */file*.

The `namespace` keyword is equivalent to  `cd` in this context. So once you use it, every class, interface or other name used in that file needs to use either a relative *path* to it, or an *absolute* one. Since you are specifying a relative *path*, that's why the autoloader is trying to find `SimpleSAML\Module\unixauth\Auth\Source\sspmod_core_Auth_UserPassBase`. Just prepend those names with a `\`:

```php
<?php
namespace SimpleSAML\Module\unixauth\Auth\Source;

class UNIX extends \sspmod_core_Auth_UserPassBase {
```

jaimeperez:
Hi @bilbolodz!

I'm closing this issue here, given that we are not going to merge the code in our repository anyway. You can, of course, continue to work on it committing new things, or commenting on the code if you need any guidance. I'll be more than happy to help 😉

bilbolodz:
I've made "standalone" version of my unixauth module:
https://github.com/bilbolodz/unixauth

If you be so kind and check what need to be fixed before including it into  list of known modules

jaimeperez:
Hi @bilbolodz!

The first thing I can see is the name of the module. We have a naming convention to differentiate SimpleSAMLphp modules from anything else, and you need to adhere to it in order to install the module with composer. Instead of *bilbolodz/unixauth*, the name should be *bilbolodz/simplesamlphp-module-unixauth*.

bilbolodz:
Done https://github.com/bilbolodz/simplesamlphp-module-unixauth
Any other comments?

jaimeperez:
Great!

I'll try to take a look as soon as I can. Right now I'm in a SimpleSAMLphp hackathon in the US and it'll take me some time to get back to normal business 😉

bilbolodz:
Have you managed to find some time to check my work and put it in list of known modules?

PR #367, opened 2016-04-19
Add 'release bilateral attributes' feature to AttributeLimit authproc

sitya:
AttributeLimit allows to release only that attributes to the SP, which are listed in SP's metadata. In the most federation metadata is generated by a central tool of the national federation, so there could be attributes which are not specified in the federation (so it could not be listed in SP's requiredAttributes), but the SP would need it.

With this PullRequest we implemented this feature. It adds the functionality of specifying bilateral attribute relations in addition to the attribute rules defined in the peer's metadata. You can both specify SP entityIDs where you send some additional attributes (bilateralSPs) and special attributes to send to some SPs (bilateralAttributes).

### DIFF
--- a/modules/core/docs/authproc_attributelimit.txt
+++ b/modules/core/docs/authproc_attributelimit.txt
@@ -8,6 +8,10 @@ metadata. The configuration is a list of attributes that should be allowed. In c
 release some specific values, make the name of the attribute the key of the array, and its value an array with all the
 different values allowed for it.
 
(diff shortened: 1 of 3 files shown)


jaimeperez:
Hi @sitya! Thanks a lot for the PR!

After a very, very quick glance at it, it looks great. I'll take a deeper look at it as soon as I can and hopefully merge it 😉

Issue #366, opened 2016-04-18
CAS Authentication Source LDAP not Binding.


atticus88:
When setting up a CAS authentication source in 1.14.2 I was getting an error like the following.

```
SimpleSAML_Error_AuthSource: AUTHSOURCEERROR('%AUTHSOURCE%' => '\'ldap\'', '%REASON%' => '\'Library - LDAP getAttributes(): Failed to get attributes from DN \\\'cn=user,dc=example,dc=com'\\\'; cause: \\\'Operations error\\\' (0x1); additional: \\\'000004DC: LdapErr: DSID-0C090748, comment: In order to perform this operation a successful bind must be completed on the connection., data 0, v2580\\\'\'')
```

```
Backtrace:
5 /var/simplesamlphp/lib/SimpleSAML/Auth/LDAP.php:176 (SimpleSAML_Auth_LDAP::makeException)
4 /var/simplesamlphp/lib/SimpleSAML/Auth/LDAP.php:536 (SimpleSAML_Auth_LDAP::getAttributes)
3 /var/simplesamlphp/lib/SimpleSAML/Auth/LDAP.php:625 (SimpleSAML_Auth_LDAP::validate)
2 /var/simplesamlphp/modules/cas/lib/Auth/Source/CAS.php:182 (sspmod_cas_Auth_Source_CAS::finalStep)
1 /var/simplesamlphp/modules/cas/www/linkback.php:26 (require)
0 /var/simplesamlphp/www/module.php:127 (N/A)
```

I was able to resolve the issue by adding the bind line before getting the attributes in `/var/simplesamlphp/lib/SimpleSAML/Auth/LDAP.php`.
```php
$this->bind($config['priv_user_dn'], $config['priv_user_pw']);
$attributes = $this->getAttributes($dn, $config['attributes']);
return $attributes;
```

I did not see this issue when just setting up LDAP authentication only LDAP with CAS.

atticus88:
After some more digging it looks like the issue is caused by the following:  When the CAS auth source calls the LDAP validate. Since the no password is passed in the $password variable is set to null.  Then the password is sanitized/slashes added. The $password then becomes a empty string.  When the $password is checked if it is not null, it then tries to bind again with a blank password. 

`/modules/cas/lib/Auth/Source/CAS.php`
```php
$ldapattributes = $ldap->validate($this->_ldapConfig, $username);
```

[...]

PR #370, opened 2016-04-20
fix issue CAS authentication source LDAP binding failure. #366

atticus88:
Added check in ldap validate to check if password is null before it runs addcslashes so it does not get converted to a empty string.


jaimeperez:
Thanks a lot @atticus88!

I've committed your fix with a slightly different approach to reduce the complexity of the method.

PR #376, opened 2016-04-29
Look for HTTP_X_FORWARDED headers set by Reverse Proxy

nathanlenz:
If baseurlpath is not set to a full url in config and the server values are used to construct the url, then it will fail when behind a reverse proxy where the proxy terminates the https request and forwards it on via http.

The workaround is pretty simple, just put a full url into the config. Submitting a pull request just in case you want to add it.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -75,6 +75,11 @@ class HTTP
      */
     private static function getServerHTTPS()
     {
+        // detect https requests terminated by reverse proxy
+        if (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] === 'https') {
+            return true;
+        }
+
         if (!array_key_exists('HTTPS', $_SERVER)) {
             // not an https-request
             return false;
@@ -100,7 +105,14 @@ class HTTP
      */
     private static function getServerPort()
     {
-        $port = (isset($_SERVER['SERVER_PORT'])) ? $_SERVER['SERVER_PORT'] : '80';
+        if (isset($_SERVER['HTTP_X_FORWARDED_PORT'])) {
+            // port from reverse proxy
+            $port = $_SERVER['HTTP_X_FORWARDED_PORT'];
+        } elseif (isset($_SERVER['SERVER_PORT'])) {
+            $port = $_SERVER['SERVER_PORT'];
+        } else {
+            $port = '80';
+        }
         if (self::getServerHTTPS()) {
             if ($port !== '443') {
                 return ':'.$port;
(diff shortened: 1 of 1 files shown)


travishegner:
I'd like to add a +1 for this PR. I ran into the same issue, implemented nearly identical code, and almost created a PR until I found this one.

Might I recommend some logic that if `$_SERVER['HTTP_X_FORWARDED_PROTO'] === 'https'` and `!isset($_SERVER['HTTP_X_FORWARDED_PORT'])` then return an empty string in that case as well? This would allow the generated URLS to be assumed 443, where as it stands the generated urls will be like `https://saml-server.example.com:80/`, which is typically not correct.

This would allow greater compatability without the need to set the X_Forwarded_Port header if you are using standard https ports.

nathanlenz:
Ah, I wasn't aware of that case. I'm only testing behind an Amazon ELB which sets both headers. I think this would make it work:

```
        if (isset($_SERVER['HTTP_X_FORWARDED_PORT'])) {
            // port from reverse proxy
            $port = $_SERVER['HTTP_X_FORWARDED_PORT'];
        } elseif (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] === 'https') {
            // x_forwarded_port is not set, but x_forwarded_proto is set.
            // in this case, assume 443 otherwise you end up with an https://example.com:80/ url.
            $port = '443';
        } elseif (isset($_SERVER['SERVER_PORT'])) {
            $port = $_SERVER['SERVER_PORT'];
        } else {
            $port = '80';
        }
```

travishegner:
@nathanlenz Looks to me like that will work.

jaimeperez:
Hi guys!

I don't think that's a bug that we need to fix. The idea of `baseurlpath` is, precisely, to allow SimpleSAMLphp to be aware of its own full URL, without the need for guessing, which is prone to error. Ok, maybe the name of the configuration option is not the best we could have chosen, but that's not reason not to use it.

The __X-Forwarded-*__ family of headers have one problem: they cannot be trusted. Anyone can set those headers and in most cases (actually I'm not aware of any proxy that gets rid of them, and even if they do, that would be against the definition of those headers, since all values must be kept in order while traversing a proxy) they will just get through.

So, given that the configuration must always be set and customized to your own needs, I really don't get why should we make the code more complicated to support untrusted HTTP headers to fix something that's not really a problem...

Maybe I'm missing something, and in that case, please correct me, but in general I don't think this is needed nor a good idea.

iluuu1994:
This is a valid issue. Here's how Symfony does it:
https://symfony.com/doc/current/deployment/proxies.html

Basically, you have to tell Symfony which proxies you trust to provide the correct `X-Forwarded-*` headers. If the request comes from anywhere else the headers are ignored.

Right now we have this rather horrible workaround for SimpleSaml to work behind a reverse proxy:

```php
if (($_SERVER['HTTP_X_FORWARDED_PROTO'] ?? null) === 'https') {
    $_SERVER['HTTPS'] = 'on';
}

if (($_SERVER['HTTP_X_FORWARDED_PORT'] ?? null) !== null) {
    $_SERVER['SERVER_PORT'] = $_SERVER['HTTP_X_FORWARDED_PORT'];
}
```

PR #378, opened 2016-04-29
fix issue with cas ldap search and active directory referals

atticus88:
When using CAS auth module the LDAP setting for referrals was not being passed, causing the search to fail when LDAP server is Active Directory.  This fix makes sure all ldap configuration fields are passed when using CAS.

### DIFF
--- a/modules/cas/lib/Auth/Source/CAS.php
+++ b/modules/cas/lib/Auth/Source/CAS.php
@@ -78,6 +78,29 @@ class sspmod_cas_Auth_Source_CAS  extends SimpleSAML_Auth_Source  {
 		}else{
 			throw new Exception("cas login URL not specified");
 		}
+
+
+		// Check for the required ldap configuration 
+		if(!isset($this->_ldapConfig['servers'])) {
+			throw new Exception("ldap server not specified");
+		}
+
+		// Check for ldap options in cas configuration if not configured set to default
+		if(!isset($this->_ldapConfig['enable_tls'])) {
+			$this->_ldapConfig['enable_tls'] = TRUE;
+		}
+		if(!isset($this->_ldapConfig['debug'])) {
+			$this->_ldapConfig['debug'] = FALSE;
+		}
+		if(!isset($this->_ldapConfig['timeout'])) {
+			$this->_ldapConfig['timeout'] = 0;
+		}
+		if(!isset($this->_ldapConfig['port'])) {
+			$this->_ldapConfig['port'] = 389;
+		}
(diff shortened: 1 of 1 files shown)


atticus88:
Need to fix issue with undefined index if values is not set in config

modules/cas/lib/Auth/Source/CAS.php
```diff
@@ -78,6 +78,29 @@ public function __construct($info, $config) {
 		}else{
 			throw new Exception("cas login URL not specified");
 		}
+
+
+		// Check for the required ldap configuration 
+		if(!isset($this->_ldapConfig['servers'])) {
```
thijskh: This makes sense because I see that above a "ldap" section in the config is already enforced. However, I noticed that in finalStep(), `_ldapconfig['servers']` is checked whether it is set, making it essentially optional. I think we should either make `_ldapconfig['ldap']['servers'] required, or not (and specify which it is in the docs) but the current code is ambivalent.

modules/cas/lib/Auth/Source/CAS.php
```diff
@@ -78,6 +78,29 @@ public function __construct($info, $config) {
...
+			throw new Exception("ldap server not specified");
+		}
+
+		// Check for ldap options in cas configuration if not configured set to default
```
thijskh: We are setting defaults here that are also specified in the SimpleSAML_Auth_LDAP constructor as well. This seems suboptimal to repeat in different places of the code.

But I don't have a clear-cut solution to that in the context of just the CAS module, so it might be acceptable here. Maybe SimpleSAML_Auth_LDAP's constructor should be changed so that if you pass in `null` for any of the optional params, you get the default?

modules/cas/lib/Auth/Source/CAS.php
atticus88: After a bit of testing it looks like the _ldapconfig['servers'] is not a required field for CAS authentication to work, but the ldap array is required in the config. That is why I was thinking the servers attribute was required.

```php
'example-cas' => array(
    'cas:CAS',
    'cas' => array(
        'login' => 'https://cas.example.edu/login',
        'validate' => 'https://cas.example.edu/validate',
        'logout' => 'https://cas.example.edu/logout'
    ),
    'ldap' => array(),
),
```

modules/cas/lib/Auth/Source/CAS.php
atticus88: I re-work this and try to find a better way of setting the defaults for the the LDAP constructor from the CAS module, or change the LDAP constructor.

PR #380, opened 2016-04-30
Allow regular expression matching of trusted.url.domains.

### LINKED ISSUE #379: Allow wildcards or regular expression matching in trusted.url.domains
Use case:

* Service provider
* A large number of applications that grow and shrink dynamically
* Each application exists on its own subdomain

Currently, `trusted.url.domains` grows very large and looks like:

```
'trusted.url.domains' => array('app-1.example.com', 'app-2.example.com', ..., 'app-n.example.com'),
```

In this scenario, there is constant maintenance of adding and removing values from `trusted.urls.domains`. One solution to help with this would be to allow wildcards or regular expression matching when evaluation `trusted.urls.domain`. For example the values could be:

```
'trusted.url.domains' => array('*.example.com'),
```

Or:

```
'trusted.url.domains' => array('@\.example\.com$@'),
```

This would eliminate the need to either manually maintain this list or write a new tool to rewrite this list regularly.

jdufresne:
Fixes #379

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -105,6 +105,17 @@ $config = array(
     'trusted.url.domains' => array(),
 
     /*
+     * Enable regular expression matching of trusted.url.domains.
+     *
+     * Set to true to treat the values in trusted.url.domains as regular
+     * expressions. Set to false to do exact string matching.
+     *
+     * If enabled, the start and end delimiters ('^' and '$') will be added to
+     * all regular expressions in trusted.url.domains.
+     */
+    'trusted.url.regex' => false,
+
+    /*
      * Enable secure POST from HTTPS to HTTP.
      *
      * If you have some SP's on HTTP and IdP is normally on HTTPS, this option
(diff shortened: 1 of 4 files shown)


jaimeperez:
Hi Jon! Thanks a lot for this! It's not that often that we get PRs so complete, with documentation and tests 😄 

There's one thing I've spotted in your code, though. If an attacker notices that your SimpleSAMLphp installation is fine to redirect to *app.com* (that is, *app.com* is allowed in `trusted.url.domains`), he or she can leverage that knowledge to circumvent the protection mechanism and exploit an open redirect. Let's say this is your configuration:

```php
    'trusted.url.regex' => true,
    'trusted.url.domains' => array('app.com'),
```

Then the attacker could use a domain under his/her control, say *evil.com*, and register a subdomain *app.com.evil.com*, configuring a web server there to redirect to the final destination of the open redirect attack. Then, passing SimpleSAMLphp a URL such as `http://app.com.evil.com/some_URL_ID` would allow the attacker to skip the protection and get the end user redirected to somewhere else without noticing, because *app.com.evil.com* will match the regular expression that you are building, `app.com`, which is lacking regex delimiters as well as begin and end of line delimiters.

The proper way to handle this case would be something like this:

```php
               foreach ($trustedSites as $regex) {
                    if (preg_match('@^'.$regex.'$@', $hostname)) {
                        $trusted = true;
                        break;
                    }
                }
```

I think we would need a test for this corner case too. Would you be ok to add that couple of changes to the PR?

Thanks again!

jdufresne:
I have made the recommended changes. All additional feedback is welcome. Thanks!

jaimeperez:
Great work! Thanks a lot again for the help Jon!

PR #382, opened 2016-05-03
Added OpenLDAP case for UserGroups

grueneedv:
This lines make it possible that the group membership information from an OpenLDAP server can be added as authproc filter. The example is more generally and should work for a basic OpenLDAP setup (at least it is compatbile with our 4 different OpenLDAP servers). It works for us (we use it) but we can't support it.


thijskh:
I think this is fine to merge.

jaimeperez:
Unfortunately, it looks like there are conflicts that need to be fixed before that. I also see that some classes like `SimpleSAML_Logger` are used with their old name, instead of the new one (`SimpleSAML\Logger`).

Issue #320, opened 2016-01-25
[SimpleSamlPHP] Facebook Authentication


priyaekbote:
Hi there,

We have integrated a simplesamlphp with authfacebook module.
authfacebook is enabled and we can use it for login.
It returns the data except the facebook email id.

We need the email id used to facebook to make an entry in our custom database.

Could you please help us out to get the email id from facebook in attributes list ?

Thanks.

jaimeperez:
Hi @priyaekbote!

This is a issue tracker, not an user support forum. Can you please send your question to the [mailing list](https://groups.google.com/forum/#!forum/simplesamlphp)?

Thanks!

priyaekbote:
Hi @jaimeperez 

I though it is a issue in the authfacebook module.

Thanks for the help !!

PR #385, opened 2016-05-07
Fix issue with Facebook authentication retrieving only user id and name

NicolasLiampotis:
As of Facebook Graph API v2.4, the returned user object only contains the `id` and `name`. As a result you need to specify in the request url which additional fields you want the API to return. Effectively, the `/{user-id}` call should become `/{user-id}?fields=email,first_name,etc..`.

This PR adds the `user_fields` option to the authfacebook module configuration:
```
/**
 * A comma-separated list of user profile fields to request.
 *
 * Note that some user fields require appropriate permissions. For
 * example, to retrieve the user's primary email address, "email" must
 * be specified in both the req_perms and the user_fields parameter.
 *
 * When empty, only the app-specific user id and name will be returned.
 *
 * See the Graph API specification for all available user fields:
 * https://developers.facebook.com/docs/graph-api/reference/v2.6/user
 */
private $user_fields;
```

Also note that the `username` field is no longer part of the Facebook user object. The `third_party_id` property is the closest alternative, i.e. 

> A string containing an anonymous, but unique identifier for the person. You can use this identifier with third parties 

As such, the `facebook_user` attribute is now based on the  `third_party_id` (if available) and in conjunction with facebook2name filter this will become the user's ePPN. Ideally, this should be mapped to an eduPersonUniqueId which is also opaque. However, a `third_party_id` string can contain non alphanumeric characters which are not permitted in the uid portion of an ePUID. The plain `id` property on the other hand is app-specific so the ePTID is the right place for it. That being said, my question is whether authfacebook should continue constructing ePPNs based on the plain `id` in the absence of the `third_party_id`. What do you think?

Reported issue #320 might be related to this PR, although not getting the user's email address could be due to the `email` permission missing from the `req_perms` config.

### DIFF
--- a/config-templates/authsources.php
+++ b/config-templates/authsources.php
@@ -201,6 +201,10 @@ $config = array(
         // which additional data permissions to request from user
         // see http://developers.facebook.com/docs/authentication/permissions/ for the full list
         // 'req_perms' => 'email,user_birthday',
+        // Which additional user profile fields to request.
+        // When empty, only the app-specific user id and name will be returned
+        // See https://developers.facebook.com/docs/graph-api/reference/v2.6/user for the full list
+        // 'user_fields' => 'email,birthday,third_party_id,name,first_name,last_name',
     ),
     */
 
--- a/modules/authfacebook/lib/Auth/Source/Facebook.php
+++ b/modules/authfacebook/lib/Auth/Source/Facebook.php
@@ -40,6 +40,21 @@ class sspmod_authfacebook_Auth_Source_Facebook extends SimpleSAML_Auth_Source {
 
 
 	/**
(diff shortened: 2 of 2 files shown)


jaimeperez:
Hi Nicolas! Thanks a lot for this!

I was wondering whether using the [Facebook SDK](https://github.com/facebook/facebook-php-sdk-v4) as a dependency would have made this transparent for us. It would be great to stop shipping with SimpleSAMLphp an external library like the old, deprecated Facebook base SDK, and add a dependency instead. If that would have the side benefit that changes in the API are easier to cope with, I'm all in.

How much work do you think it would be to upgrade to the new SDK? Would you be able to help us with that?

PR #386, opened 2016-05-09
Use REQUEST_URI instead of SCRIPT_PATH to build the URL without params

nicolus:
In some cases (ie : rewriting URLs with Apache2), the script path and URL can be different, which means that HTTP::getSelfURLNoQuery() returns the wrong URL.

Using REQUEST_URI instead of SCRIPT_PATH and PATH_INFO seems more reliable and is consistent with the HTTP::getSelfURL() method.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -739,10 +739,8 @@ class HTTP
     public static function getSelfURLNoQuery()
     {
         $url = self::getSelfURLHost();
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @nicolus! Thanks for your PR!

This was initially resolved in #391. Later on, the code was modified again to rely on whatever is configured in `baseurlpath`. Take a look at commit 636aa66eb2384ce98c17aeb27d258dfd6958f7a9.

PR #387, opened 2016-05-10
General improvements of MemcacheMonitor module

tvdijen:
Fix missing type attribute to comply with W3c XHTML 1.0 Transitional
Also added some missing definitions to the dictionary.


jaimeperez:
Thanks a lot @tvdijen!

PR #391, opened 2016-05-13
add some checks to prevent appending to SCRIPT_NAME when it isn't nec…

travishegner:
…essary

This PR address issue #5 in what is hopefully a compatible way. As mentioned in my comment on that issue, certain environments populate `$_SERVER['SCRIPT_NAME']` differently, so this patch will check that variable before appending to it.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -742,11 +742,13 @@ class HTTP
         $url .= $_SERVER['SCRIPT_NAME'];
 
         /* In some environments, $_SERVER['SCRIPT_NAME'] already ends with $_SERVER['PATH_INFO'].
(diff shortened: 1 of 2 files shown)


travishegner:
If you are interested in another PR, I have a topic branch that puts this commit onto the v1.14.3 tag.

jaimeperez:
Thanks a lot for your contribution @travishegner!

I've just committed your fix (with some slight modification), and hopefully we can forget about this issue for good 😄

jaimeperez:
Hi again!

We've changed a bit the logic of the function, and how we are no longer dependent on how the environment (particularly, `$_SERVER['PATH_INFO']`) are populated. I've been testing the new code and it appears to work perfectly fine with *nginx*, but please test it and make sure to tell us if you notice something wrong.

PR #392, opened 2016-05-13
Fix PHP notices "Undefined index" in as_login.php.

jdufresne:
Sometimes appear in error logs as:

PHP Notice:  Undefined index: ReturnTo in .../modules/core/www/as_login.php on line 9
PHP Notice:  Undefined index: AuthId in .../modules/core/www/as_login.php on line 13

### DIFF
--- a/modules/core/www/as_login.php
+++ b/modules/core/www/as_login.php
@@ -6,11 +6,11 @@
  * @package SimpleSAMLphp
  */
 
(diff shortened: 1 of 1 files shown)


thijskh:
I'm not completely sure about the use of `empty()`. Are we certain that either variable will never contain something that PHP evaluates to false? Maybe better to use `isset()`.

jdufresne:
Good call. Fixed to use `!isset()`.

PR #394, opened 2016-05-23
Make log level consistent to that of LDAP authentication

tvdijen:
Negotiate uses 'info' for succesful/unsuccesful logons, where ldap authentication uses 'stats' level for this. I would like to see this a little more consistent.


jaimeperez:
Hi again Tim!

This is a bit complicated. The `stats` method is not really intended to log messages, but to gather statistics. The fact that there's no such **stats** log level corroborates this, and in fact the method **logs** whatever we pass it in the `NOTICE` level. Since we have the possibility of saving the statistics to a completely separate place, this could lead to these log messages not being logged at all.

On the other hand, consistency is definitely always good and we should aim for that. I'll need to think a bit about this 😃

tvdijen:
I can't move this PR to the negotiate-module, so I'm closing this one..
It's actually pretty simple:
- We either start logging "User 'foobar' successfully authenticated" to the stats-level for negiotate, or we stop logging these messages to the stats-level for LDAP

PR #395, opened 2016-05-23
Memcached extension support

bradjones1:
I'm working on re-deploying an application in PHP 7 and the memcached extension seems to be way further ahead in terms of development and adoption... the implementation in SSP seems simple enough, so here's a stab at supporting using either of the extensions, with memcache being preferred if we find it, first.

This would probably require some additional tests and documentation updates, but starting this PR first to see if the maintainers have any interest.


tvdijen:
Should probably make the move to the newer and well maintained Memcached extension.
Since both extensions store data in a different way, they are not compatible.. It would probably be nice to add support for the Memcached extension in 1.15 and make it the preferred one in 2.0

bradjones1:
@tvdijen As I understand it (and I'm not an expert by any means) I think the two extensions serialize data differently, which would mean you couldn't read/write to the same Memcached array. It's probably not unreasonable for there to be an engineering limitation on using both extensions concurrently, but I don't think that would stop us from being able to support both during some cut-over period.

tvdijen:
Correct and agreed!
By the way, it's probably a good idea to use strict comparisons in your commit:
i.e. self::$extension === 'memcached'

jaimeperez:
Hi @bradjones1! Thanks a lot for this!

I agree with @tvdijen. I definitely think we should start migrating to **memcached** if that's going to take over in PHP 7, and 1.15 is the right candidate to prepare for the switch in 2.0 (the list of things that will break in 2.0 is starting to be so huge, I'm starting to be scared myself... 😊)

Once thing that I'm wondering is if both extensions share the same API, as your code seems to address only the class instantiation, but not the actual use of the API. I think that as long as our code is consistent when making a decision on which extension to use (and it's backwards-compatible for now), it's fine to support both at the same time. An alternative would be to add a temporary configuration option, defaulting to the old **memcache** extension, that allowed people to specify which one to use, if both are available. In that case, people could start migrating in 1.15 by just changing the configuration, without the need to mess with their PHP installation to remove the old extension.

What do both of you think?

I'll make some brief comments in the code too.

lib/SimpleSAML/Memcache.php
```diff
@@ -154,7 +162,12 @@ public static function set($key, $value, $expire = null)
 
         // store this object to all groups of memcache servers
         foreach (self::getMemcacheServers() as $server) {
-            $server->set($key, $savedInfoSerialized, 0, $expire);
+            if (self::$extension == 'memcached') {
```
jaimeperez: Tim already mentioned this, but just for the record: strict comparisons are always preferred, even if we are completely sure that the types are not going to be an issue, it is in general a good practice to be on the safe side.

lib/SimpleSAML/Memcache.php
```diff
@@ -277,7 +290,12 @@ private static function addMemcacheServer($memcache, $server)
         }
 
         // add this server to the Memcache object
-        $memcache->addServer($hostname, $port, true, $weight, $timeout, $timeout, true);
+        if (self::$extension == 'memcached') {
+            $memcache->addServer($hostname, $port);
+        }
+        else {
+            $memcache->addServer($hostname, $port, TRUE, $weight, $timeout, $timeout, TRUE);
```
jaimeperez: We are using the [PSR-2](http://www.php-fig.org/psr/psr-2/) coding standard, which mandates lowercase in all keywords, including `true`, `false` and `null`. The else statement should also be in the same line as the closing brace of the if.

badrange:
Hi everyone; I stumbled across this PR today because I would like to use PHP7+memcached in a project I'm working on. 

Turns out that a 3.0.0 release of php-memcached is *right around the corner* (A commit with the message "Add ChangeLog notes and tentative release date for 3.0.0" was done 4 days ago). 

So if we manage to get this PR merged, many people will start benefitting from it right away!

bradjones1:
@jaimeperez Incorporated your code style changes, thanks @badrange for the update on memcached's status.

thijskh:
Thanks bradjones1, great work! I think this looks good and is ready for the Memcache class. However, I think there are some other places that need to be updated:

- [ ] 1. `bin/memcacheSync.php`
- [ ] 1. `modules/core/www/frontpage_config.php`
- [ ] 1. `docs/simplesamlphp-install.md` and `docs/simplesamlphp-artifact-idp.md`
- [ ] 1. everything related to AuthMemcookie

I think it will be terribly confusing if this is fixed in only one part of SSP.

typhonius:
I've addressed the comments raised by @thijskh above against the branch @bradjones1 has PR'd. That PR is over here: https://github.com/bradjones1/simplesamlphp/pull/1 - if Brad pulls that in, I believe it'll update this PR too.

thijskh:
Great, thanks! Brad, are you available to pull this in?

bradjones1:
Thanks everybody, yes, I will this morning.

typhonius:
@bradjones1 looks like there was a merge conflict here. I rebased against master and fixed locally but I think you'll need to do the same in your branch. The conflict was in `modules/core/www/frontpage_config.php` which I think could be resolved with the following:

```
$funcmatrix[] = array(
    'required' => 'optional',
    'descr' => 'predis/predis (required if the redis data store is used)',
    'enabled' => class_exists('\Predis\Client'),
);

$funcmatrix[] = array(
    'descr' => 'Memcache or Memcached Extension (required if a Memcached backend is used)',
    'enabled' => class_exists('Memcache') || class_exists('Memcached'),
);
```

bradjones1:
@typhonius Thanks... That's why you don't just close the tab after hitting "merge." :-)

thijskh:
Good and useful work, thanks! I was wondering if in the documentation we should provide some more explicit guidance rather than treat the two options equally, e.g. only explain/recommend to install php5-memcached, and mention that php5-memcache is also supported for the time being as a legacy implementation. But that can be addressed separately.

bin/memcacheSync.php
```diff
@@ -3,18 +3,19 @@
...
-    }
-
-    exit(1);
+if(!class_exists('Memcache') && !class_exists('Memcached')) {
+	echo("Error: the memcache and memcached libraries appear to be unavailable.\n");
+	echo("\n");
+	echo("This is most likely because PHP doesn't load it for the command line\n");
+	echo("version. You probably need to enable one of them somehow.\n");
+	echo("\n");
+	if(is_dir('/etc/php5/cli/conf.d')) {
```
tvdijen: I think this is wrong in many ways...
We have nothing to do with CLI-configuration is this case, so we need a different conf-dir here.
Also, this conf-dir can be located in several places, depending on your OS / setup

thijskh: It _is_ CLI-configuration as this is an executable script I think?

It looks like a merge error, since this is the original advice that was updated in de972b674829d24ee6767d2b8fe0be99e686ef0d. I can reapply that commit later.

Nonetheless it remains an heuristic that tries to dole out some advice where it can, it does not have the ambition to support every OS and configuration. Whether that is useful is up for debate.

tvdijen: Ah, my bad!  It should use CLI config indeed.
I'm not saying we have to support every single configuration, but the hardcoded paths are too specific... I like de972b6 a lot more!

thijskh: Fixed now.

tvdijen: 👍

PR #397, opened 2016-05-31
Fix unexpected handling of the 'attributes' setting

tvdijen:
In authsources.php, taken from the ldap-example:
        // Which attributes should be retrieved from the LDAP server.
        // This can be an array of attribute names, or NULL, in which case
        // all attributes are fetched.
        'attributes' => NULL,

This suggests that an empty array would return no attributes, but in fact it returns them all, just like a NULL value.
This change fixes that to make this setting act as you would expect and also takes care of a couple of attributes that were being reused within the same function.

### DIFF
--- a/lib/SimpleSAML/Auth/LDAP.php
+++ b/lib/SimpleSAML/Auth/LDAP.php
@@ -520,22 +520,25 @@ class SimpleSAML_Auth_LDAP {
      * The array of attributes and their values.
      * @see http://no.php.net/manual/en/function.ldap-read.php
      */
-    public function getAttributes($dn, $attributes = NULL, $maxsize = NULL) {
-
+    public function getAttributes($dn, $wanted_attributes = NULL, $maxsize = NULL) {
+		$empty_attributes = is_array($wanted_attributes) && empty($wanted_attributes);
+	
         // Preparations, including a pretty debug message...
-        $description = 'all attributes';
-        if (is_array($attributes)) {
-            $description = '\'' . join(',', $attributes) . '\'';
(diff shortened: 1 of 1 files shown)


tvdijen:
Or, on second thought, would it be better to take care of this case at the end of the validate-function?

jaimeperez:
Hi @tvdijen! Thanks so much for this!

You are right, the documentation is misleading and we should fix it (the issue, not the documentation, as I totally believe we should have a way to tell that we don't want any attributes). On the other hand, I'm a bit concerned about backwards-compatibility, as this change could break existing configurations (people having an empty array there because **it works** to get all the attributes, and all of a sudden getting nothing back from the LDAP after upgrading.

I'm a bit tempted to put this on hold until we have released 1.15 and we can start breaking things properly in 2.0. However, the LDAP class is so incredibly crappy that it is high up in the list for a refactoring (and even substituting it by a new class), and that would break your PR.

What do you think? Is it urgent for you to have this solved?

tvdijen:
No, it's not urgent at all.. It's just something I ran into and it only affects my admin logins.

Another thing I noticed, but don't have a solution for, is that you can brute-force someone's password without the account getting locked out.. This kind of contradicts to our (most?) company's security policy.

jaimeperez:
Thanks for the feedback!

Regarding the brute force attack, yes, I'm aware of that. I myself wrote a script some time ago to brute-force the admin login form 😅

tvdijen:
Moved to https://github.com/simplesamlphp/simplesamlphp-module-ldap/pull/5

PR #398, opened 2016-06-01
Update the authwindowslive:LiveID module to use OAuth 2.0 authorization

ghalse:
Microsoft's support for OAuth WRAP has been discontinued, and the old consent.live.com endpoints no longer exist. Whilst there are now probably other ways to connect to a Microsoft Account/Windows Live ID, it may be useful for some to use the new OAuth 2.0 model.

Attributes are now retrieved from the user's profile using the newer Microsoft Graph API rather than the Messenger API. This required additions to the windowslive2name attribute map, but I've preserved the existing ones for anyone who's using this map for something else.

This was done mainly as an exercise in understanding social bridges. Whilst it does work properly, I've only tried limited test cases and the code could probably stand to be improved. However, the previous version doesn't work at all any more - consent.live.com isn't even published in DNS any more.

### DIFF
--- a/modules/authwindowslive/lib/Auth/Source/LiveID.php
+++ b/modules/authwindowslive/lib/Auth/Source/LiveID.php
@@ -8,130 +8,135 @@
  */
 class sspmod_authwindowslive_Auth_Source_LiveID extends SimpleSAML_Auth_Source {
 
-	/**
-	 * The string used to identify our states.
-	 */
-	const STAGE_INIT = 'authwindowslive:init';
+    /**
+     * The string used to identify our states.
+     */
+    const STAGE_INIT = 'authwindowslive:init';
 
-	/**
-	 * The key of the AuthId field in the state.
-	 */
-	const AUTHID = 'authwindowslive:AuthId';
+    /**
+     * The key of the AuthId field in the state.
+     */
+    const AUTHID = 'authwindowslive:AuthId';
 
-	private $key;
-	private $secret;
-
-
-	/**
-	 * Constructor for this authentication source.
-	 *
-	 * @param array $info  Information about this authentication source.
-	 * @param array $config  Configuration.
-	 */
-	public function __construct($info, $config) {
-		assert('is_array($info)');
-		assert('is_array($config)');
-
-		// Call the parent constructor first, as required by the interface
-		parent::__construct($info, $config);
-
-		if (!array_key_exists('key', $config))
-			throw new Exception('LiveID authentication source is not properly configured: missing [key]');
-
-		$this->key = $config['key'];
-
-		if (!array_key_exists('secret', $config))
-			throw new Exception('LiveID authentication source is not properly configured: missing [secret]');
-
-		$this->secret = $config['secret'];
-	}
+    private $key;
+    private $secret;
 
 
-	/**
-	 * Log-in using LiveID platform
-	 *
-	 * @param array &$state  Information about the current authentication.
-	 */
-	public function authenticate(&$state) {
-		assert('is_array($state)');
+    /**
+     * Constructor for this authentication source.
+     *
+     * @param array $info  Information about this authentication source.
+     * @param array $config  Configuration.
+     */
+    public function __construct($info, $config) {
+        assert('is_array($info)');
+        assert('is_array($config)');
 
-		// We are going to need the authId in order to retrieve this authentication source later
-		$state[self::AUTHID] = $this->authId;
+        // Call the parent constructor first, as required by the interface
+        parent::__construct($info, $config);
 
-		$stateID = SimpleSAML_Auth_State::saveState($state, self::STAGE_INIT);
+        if (!array_key_exists('key', $config)) {
+            throw new Exception('LiveID authentication source is not properly configured: missing [key]');
+        }
 
-		SimpleSAML\Logger::debug('authwindowslive auth state id = ' . $stateID);
+        $this->key = $config['key'];
 
-		// Authenticate the user
-		// Documentation at: https://azure.microsoft.com/en-us/documentation/articles/active-directory-v2-protocols-oauth-code/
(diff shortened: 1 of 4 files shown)


jaimeperez:
Hi @ghalse!

Thanks a lot for this! We are not using this ourselves so didn't know about the change in the API, so it's good someone gave us a heads up.

I'm a bit confused, though. Is this a reimplementation of the old module to work with OAuth2? In that case, was it that simple? I'm curious because looking at the changes it looks like it was pretty simple, even though OAuth2 is quite complex in my opinion...

I'll give you some feedback on the code itself too.

modules/authwindowslive/lib/Auth/Source/LiveID.php
```diff
@@ -96,38 +100,32 @@ public function finalStep(&$state) {
...
-		parse_str($result, $response);
+		$response = json_decode($result,true);
 
 		// error checking of $response to make sure we can proceed
-		if (!array_key_exists('wrap_access_token',$response))
-			throw new Exception('[' . $response['error_code'] . '] ' . $response['wrap_error_reason'] .
-				"\r\nNo wrap_access_token returned - cannot proceed\r\n" . $response['internal_info']);
+		if (!array_key_exists('access_token',$response))
+			throw new Exception('[' . $response['error'] . '] ' . $response['error_description'] .
+				"\r\nNo access_token returned - cannot proceed\r\n" . implode(', ', $response['error_codes']));
```
jaimeperez: Please, avoid control structures without braces, as they open up for bugs that are difficult to find later. Even if you are only having one line inside, surround it with braces.

Of course, I'm conscious that the existing code is doing exactly the same and you just modified a few things inside that, but we are trying to get rid of all the code smells that we already have, which are quite a few...

modules/authwindowslive/lib/Auth/Source/LiveID.php
```diff
@@ -96,38 +100,32 @@ public function finalStep(&$state) {
...
-						$response['wrap_access_token'] . "\r\n"));
-		$data = \SimpleSAML\Utils\HTTP::fetch('https://apis.live.net/V4.1/cid-'. $response['uid'] . '/Profiles',$opts);
+		// Documentation at: http://graph.microsoft.io/en-us/docs/overview/call_api
+		$opts = array('http' => array('header' => "Accept: application/json\r\nAuthorization: Bearer " .
+						$response['access_token'] . "\r\n"));
+		$data = \SimpleSAML\Utils\HTTP::fetch('https://graph.microsoft.com/v1.0/me',$opts);
 		$userdata = json_decode($data, TRUE);
 
+		// This is the simplest case
+		if(!array_key_exists('@odata.context',$userdata) OR array_key_exists('error',$userdata))
```
jaimeperez: Apart from the braces, please, use the `||` and `&&` operators instead of `OR` and `AND`. While they have the same meaning in PHP, they don't have the same precedence, so it's always safer to use the former.

modules/authwindowslive/lib/Auth/Source/LiveID.php
```diff
@@ -96,38 +100,32 @@ public function finalStep(&$state) {
...
-			foreach($userdata['Entries'][0] AS $key => $value) {
-				if (is_string($value))
-					$attributes['windowslive.' . $key] = array((string)$value);
-			}
-
-			if (array_key_exists('Emails', $userdata['Entries'][0]))
-				$attributes['windowslive_mail'] = array($userdata['Entries'][0]['Emails'][0]['Address']);
-
+		$attributes['windowslive_targetedID'] = array('https://graph.microsoft.com!' . (!empty($userdata['id']) ? $userdata['id'] : 'unknown'));
+		foreach($userdata AS $key => $value) {
```
jaimeperez: Keywords should always be in lowercase, according to the [PSR-2 coding standard that we are using](https://github.com/simplesamlphp/simplesamlphp/blob/master/CONTRIBUTE.md).

modules/authwindowslive/www/linkback.php
```diff
@@ -4,30 +4,30 @@
...
 	if (array_key_exists('exp', $_REQUEST))
-		$state['authwindowslive:wrap_exp'] = $_REQUEST['exp'];
+		$state['authwindowslive:exp'] = $_REQUEST['exp'];
 
 } else {
-	// wrap_error_reason = 'user_denied' means user chose not to login with LiveID
+	// error_reason = 'user_denied' means user chose not to login with LiveID
 	// redirect them to their original page so they can choose another auth mechanism
-	if ($_REQUEST['wrap_error_reason'] === 'user_denied') {
+	if ($_REQUEST['error'] === 'user_denied') {
```
jaimeperez: Is this a bug? In the comment, you are mentioning `error_reason`, but here (and in the exception below) you are looking for `error` instead.

modules/authwindowslive/lib/Auth/Source/LiveID.php
ghalse: Thanks for the feedback - I'm offline for the next week or so, but I'll
look at all of this when I get back to somewhere with reliable connectivity.

modules/authwindowslive/www/linkback.php
ghalse: It isn't a bug - I didn't update the comment. I've improved the comment now to explain it better.

ghalse:
The original module was for the OAuth WRAP service. OAuth 2.0 was based on OAuth WRAP. So it seems that only relatively minor tweaks were needed to update the protocol. I was also surprised by how little was requred to get it to work with the new API. The module might not be a complete implementation because its not doing session handling (that's true of the original version), but it has enough to complete authentication and get attributes.

jaimeperez:
Hi Guy!

For some reason, some unrelated commits ended up in the history of this PR. Can you please rebase your commits and make sure we have a clean history with only your changes applied on top of `master` so that I can merge it?

Thanks a lot!

jaimeperez:
Great! Thanks a lot Guy, I really appreciate!

PR #400, opened 2016-06-01
Allow the optional Subject NameID to be set in a SAML AuthnRequest

pmeulen:
Add "saml:NameID" option for use with SimpleSAML_Auth_Simple that allows the Subject NameID to be set in a SAML AuthnRequest.

### DIFF
--- a/modules/saml/docs/sp.md
+++ b/modules/saml/docs/sp.md
@@ -52,6 +52,13 @@ All these parameters override the equivalent option from the configuration.
 
 :   *Note*: SAML 2 specific.
 
+`saml:NameID`
+:   Add a Subject element with a NameID to the SAML AuthnRequest for the IdP.
+    This is an associative array with the fields for the NameID.
+    Example: `array('Value' => 'user@example.org', 'Format' => SAML2_Const::NAMEID_UNSPECIFIED)`
+
+:   *Note*: SAML 2 specific.
+
 
 Authentication data
 -------------------
--- a/modules/saml/lib/Auth/Source/SP.php
+++ b/modules/saml/lib/Auth/Source/SP.php
@@ -205,6 +205,13 @@ class sspmod_saml_Auth_Source_SP extends SimpleSAML_Auth_Source {
 			$ar->setIsPassive((bool)$state['isPassive']);
 		}
 
+		if (isset($state['saml:NameID'])) {
+			if (!is_array($state['saml:NameID'])) {
+				throw new SimpleSAML_Error_Exception('Invalid value of $state[\'saml:NameID\'].');
+			}
+			$ar->setNameId($state['saml:NameID']);
+		}
+
 		if (isset($state['saml:NameIDPolicy'])) {
 			if (is_string($state['saml:NameIDPolicy'])) {
 				$policy = array(


jaimeperez:
Hi Pieter, thanks for the PR!

It's quite surprising how easy this was 😄 

If I understand correctly, the idea is to allow the SP to specify which individual we are requesting authentication for, right? In that case, I think we should also implement the other side, meaning SimpleSAMLphp being able to use the Subject NameID to determine if the authenticating subject is the one required or not, and if not, send an error back to the SP as the standard specifies. However, I don't really know how we could do that, since that would imply persistent NameIDs, right?

trscavo:
> that would imply persistent NameIDs, right?

Not necessarily. Any persistent attribute will do. The MACE-Dir SAML Attribute Profiles show how to formulate an attribute as a NameID.

jaimeperez:
Thanks Tom, I'll take a look and see how we could implement this. Meanwhile, Pieter, would it be too difficult to add a test for this?

pmeulen:
Hi Jaime,
The Subject NameID in the AuthnRequest is for the SP to request authentication of a specific Subject, yes. How the IdP reacts to this is up to the IdP and depends on the usecase I guess. Thinking about implementing this for an IdP in SSP you'd want the Subject NameID to be accessible in SimpleSAML_Auth_ProcessingFilter and SimpleSAML_Auth_Source so you can implement Auth sources and filers that use this. Implementing this should be a different issue from this PR IMO.

Looking though the existing tests, I could not see any tests for sspmod_saml_Auth_Source_SP or SimpleSAML_Auth_Simple did I miss anything? Adding tests for this is a good idea, but I'd have to find some time to do this.

trscavo:
> How the IdP reacts to this is up to the IdP and depends on the usecase I guess.

The spec is pretty clear about this. Either the IdP returns an assertion containing a NameID with the requested Subject or it returns an error. In other words, be careful what you ask for!

pmeulen:
>  > How the IdP reacts to this is up to the IdP and depends on the usecase I guess.
> 
> The spec is pretty clear about this. Either the IdP returns an assertion containing a NameID with the requested Subject or it returns an error. In other words, be careful what you ask for!

Agreed.

What I was thinking about was how you can have multiple ways of dealing with the requested NameID on the IdP side in SSP:

- You can have a generic check in the SSP (e.g. some authproc filter) that asserts that the requested NameID matches the returned NameID.

- The authentication module used by the IdP can use the fact that it knows the NameID of the desired user in advance. For example: we use this feature for an IdP that authenticates using Tiqr (tiqr.org). When this IdP knows the NameID of the user it can sent a push notification to the user's smartphone directly, without having to figure out the ID of the user first. What is possible here depends on the way the IdP handles authentication. SSP should support IdPs that what to do something like this.

May I suggest we move this (useful) discussion to a new issue? IMO the IdP side of this has little to do with the implementation of the SP side that is in this PR.

jaimeperez:
Thanks for the clarification Tom!

That's what I took from the spec too, and it's part of my concerns. I get that this is still pretty uncommon and not so many people use this functionality, but if we allow sending a NameID to request the authentication of a particular user, I think we should handle the case too on the IdP side. Otherwise, we would be breaking the spec.

Using an authproc filter or implementing this in the authentication source itself is a nice approach because it makes this easier (especially because the closer you are to your authentication backend, the easier to *read* and interpret the NameID and use it in a coherent way to identify a subject). However, I think this should really be built-in, so that SSP doesn't break the spec when acting as an IdP.

It is tricky to solve, though, because if you have a generic mechanism enforced always, regardless of the authentication source, you can't easily recognize and match identifiers in general for every possible authentication source. It would need to be configurable, then, and have some flexible defaults that would fit the most common cases.

But I agree this would probably need to be in a separate PR.

Regarding the tests, we are unfortunately very far from a good code coverage yet, but the only way to solve that is to actually add more tests, especially for new functionalities that we incorporate. So as a rule of thumb, we always try to ask for tests covering new code for every PR. If it's too complicated or you don't have what you need to actually implement a test for this, just let me know, you know I'm flexible 😉

thijskh:
Good work Pieter to add the tests, great to see the coverage increase by 2.4%pt.

Also, the original change looks good and safe to me. I've seen @jaimeperez' comments about implementation on the IdP side. However, I think that should be an independent change and independent discussion. This PR as-is will enable SP's to leverage this functionality, something they cannot do with SSP now. So it's definitely something to merge.

PR #401, opened 2016-06-01
Show a user friendly error page

hordijk:
The authenticate.php file checks for the presence of the an exception within the state. If this is present the exception is formatted for logging and shown as plain text to the end user.

This has several disadvantages:
- The end user doesn't receive a user friendly error page.
- The exception is not logged within the authenticate.php
- Internal directory structure is exposed

For an example of such an situation/error see:
http://stackoverflow.com/questions/30511003/simplesamlphp-error

With the usage of SimpleSAML_exception_handler we'll solve the disadvantages.


jaimeperez:
Hi @hordijk!

Thanks a lot for the heads up and the PR. It's quite a shame that we still had such a crappy code printing out the exception as plain text.

I've committed a fix for this, instead of just merging your PR, since the proper way to fix this would be to actually throw the exception and letting the error handler capture it, instead of calling directly the error handler (which is something that we don't need to do).

Again, thanks for letting us know about this!

PR #410, opened 2016-06-24
docs: make it work with Apache 2.4

pasja:
* Add an extra apache access control directive
  to make it work with newer Apaches
  Source: http://stackoverflow.com/questions/23337446/getting-a-403-forbidden-error-for-simplesaml-after-apache-upgrade

### DIFF
--- a/docs/simplesamlphp-install.md
+++ b/docs/simplesamlphp-install.md
@@ -121,6 +121,12 @@ Find the Apache configuration file for the virtual hosts where you want to run S
             Alias /simplesaml /var/simplesamlphp/www
     </VirtualHost>
 
+If you are using Apache 2.4.6 or higher version, you have to add an extra configuration directive to play nicely with the Apache 2.4 security settings
+
+    <Directory /var/simplesamlphp/www/>
+            Require all granted
+    </Directory>
+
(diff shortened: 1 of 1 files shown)


thijskh:
Thanks. The need for a `Require all granted` is a useful addition. I do have some comments on the specifics though. Firstly I think it's unlikely that this is since 2.4.6 and it should rather read "since 2.4".

Additionally, I think I would not make it a separate section but just integrate it into the VirtualHost block example. Something like this would support Apache < 2.4 and >= 2.4 in one go:
```
    <VirtualHost *>
            ServerName service.example.com
            DocumentRoot /var/www/service.example.com

            SetEnv SIMPLESAMLPHP_CONFIG_DIR /var/simplesamlphp/config

            Alias /simplesaml /var/simplesamlphp/www

            <Directory /var/simplesamlphp/www>
                <IfModule !mod_authz_core.c> 
                # For Apache 2.2:
                Order allow,deny
                Allow from all
                </IfModule>
                <IfModule mod_authz_core.c>
                # For Apache 2.4:
                Require all granted
                </IfModule>
            </Directory>
    </VirtualHost>
```

pasja:
I like that you combined both Apache versions' config into one copy-paste-able block.
I have pushed these changes into this branch.

jaimeperez:
Thanks a lot @pasja!

PR #412, opened 2016-07-01
Add an onlyIfEmpty option to core:ScopeAttribute.

ghalse:
This makes the filter a little more useful for people who want to generate default values from another attribute only if none already exist (for instance, you want to create eduPersonScopedAffiliation from eduPersonAffiliation _only_ if eduPersonScopedAffiliation was not returned by the authenticaton source.)

The option defaults to false, thus preserving the existing behavior.


jaimeperez:
Thanks a lot Guy! 😄

PR #414, opened 2016-07-05
Windows dir separator

matt-beamish:
Solved an issue whereby the absolute path was being added at the end of URLs returned from getSelfUrl under IIS on Windows.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -718,13 +718,16 @@ class HTTP
      * @author Andreas Solberg, UNINETT AS <andreas.solberg@uninett.no>
      * @author Olav Morken, UNINETT AS <olav.morken@uninett.no>
      * @author Jaime Perez, UNINETT AS <jaime.perez@uninett.no>
+	 * @author Matt Beamish, GSG AS <matt.beamish@gsg.wa.edu.au>
      */
     public static function getSelfURL()
     {
         $url = self::getBaseURL();
         $cfg = \SimpleSAML_Configuration::getInstance();
-        $baseDir = $cfg->getBaseDir();
-        $rel_path = str_replace($baseDir.'www/', '', realpath($_SERVER['SCRIPT_FILENAME']));
+		/* Maintain 'ends with slash' structure from config but handle IIS/Windows cases where
+		   a straight replace won't work. */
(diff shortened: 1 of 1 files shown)


thijskh:
Hi Matt, I think this could be useful. However, I have the following suggestions:

- indentation uses tabs, while standard is spaces;
- adding a unit test for behaviour with backslashes in basepath is really desirable I think;
- can you squash the two commits?

jaimeperez:
Hi @matt-beamish!

Thanks for bringing this to our attention. There was indeed a bug in this recent code, since it was ignoring Windows platforms that use backslashes instead of regular slashes.

Your PR however did not resolve the problem in the right way, because a fix was also needed in `SimpleSAML_Configuration::getBaseDir()`, since that method was also buggy. Additionally, it doesn't make much sense to convert backslashes to regular slashes both in `$baseDir` and `$_SERVER['SCRIPT_FILENAME']`, since both will use the same in all situations. We can therefore use `$baseDir` to remove the path up to the `www` directory in SimpleSAMLphp in the file path, and convert then backslashes to regular slashes in this resulting relative path.

Regarding tests, @thijskh, I'm afraid we cannot test this. Due to the way we are guessing our URL now, generated URLs were also broken when the web server was configured with an alias pointing to a directory containing symbolic links. `SimpleSAML_Configuration::getBaseDir()` was working fine because  successive calls to `dirname()` return the canonical path, but `SimpleSAML\Utils\HTTP::getSelfURL()` was not, because `$_SERVER['SCRIPT_FILENAME']` contains the path as configured in the web server, potentially with symlinks. This led to a mismatch and therefore incorrect URLs.

This issue was solved in c72b62038e3687b78ac2b4ef76aa4a3f262beeda, by using the `realpath()` function, to make sure we use the canonical path to the file pointed by `$_SERVER['SCRIPT_FILENAME']`. However, this function introduces a limitation, since calling `realpath()` for a file or directory that does not exist returns `false`. This means we cannot fake paths to represent Windows paths, since `realpath()` would not work then. See [build 140568728](https://travis-ci.org/simplesamlphp/simplesamlphp/builds/140568728) for a practical demonstration of this.

matt-beamish:
Jaime,
No problem; I just whacked together a fix for my use case and thought it'd be worth pushing back up the chain.  I'm new at the whole git thing.
Thanks for looking at it!
Matt

PR #429, opened 2016-08-03
Update module docs

dnmvisser:
Enabling/disabling modules through config.php rather than touching 'enable'/'disable' files inside module directories.

### DIFF
--- a/docs/simplesamlphp-modules.md
+++ b/docs/simplesamlphp-modules.md
@@ -34,13 +34,13 @@ directories and files:
 
 default-disable
 :   The presence of this file indicates that the module is disabled
-    by default. This module can be enabled by creating a file named
-    `enable` in the same directory.
+    by default. It can be enabled using the `module.enable`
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi Dick!

Just one comment. Do we want to deprecate the old way to enable/disable modules entirely? Because if not, I think we should mention both ways in the documentation...

thijskh:
I think I discussed it a few years ago on the mailing lists: I personally believe that the touching of the enable/disable files is hard(er) to manage than a config file, and distributes configuration all over the place. Much easier to do that from config.php if you ask me.

So if it's up to me I'd deprecate the touching and promote the configuration in config.php; perhaps removing the touchy stuff in some future release.

jaimeperez:
Thanks for the feedback Thijs!

Let's merge this now, and we can deprecate this way of handling modules in 1.15.0, so that we can remove it in 2.0.

PR #430, opened 2016-08-05
Add the existing Afrikaans translation into the list of supported languages

ghalse:
The Afrikaans translation is complete enough that end-user facing parts of SimpleSAMLphp can use it. This will allow South African users to make use of the translation (indeed, find out it exists) without having to edit code.


thijskh:
:+1:

jaimeperez:
Thanks a lot Guy! 😄

Issue #432, opened 2016-08-05
Deprecate certificate validation by fingerprint


thijskh:
Because we want to remove certificate validation by fingerprint (per #431), we should deprecate the functionality now so people can transition away from it.

trscavo:
In the case of MDQ, there is no alternative to fingerprints AFAIK.

PR #433, opened 2016-08-05
Deprecate the certFingerprint option.

thijskh:
Issue a notice when the option is used nonetheless.

Closes: #432


thijskh:
First proposal to deprecate the `certFingerprint` option in this way. Feedback on the approach welcome.

thijskh:
@jaimeperez Do you think this is mergable like this? I'd like to get the deprecation notice in before we start work on actually removing any code.

jaimeperez:
Yep, I've gone through it now and there's nothing that should create trouble, so let's merge it!

Sorry so much for the delay, by the way.

PR #434, opened 2016-08-05
Add case where https is terminated at proxy but passed through in header

jeremyfrench:
Had an issue with simplesamlphp where a site was https, but saml was not working.

Tracked it down to getHttps not working.

we were behind a proxy and it was terminating SSL, so saml thought that it was receiving an http request. 

X-Forwarded-Proto is a de facto standard for indicating that the request was https (or not) so I put in some code to check for it.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -75,6 +75,11 @@ class HTTP
      */
     private static function getServerHTTPS()
     {
+        if (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @jeremyfrench!

The `X-Forwarded-*` set of headers is broken as they have [serious security issues and cannot be trusted](https://tools.ietf.org/html/rfc7239#section-8) except for very controlled environments.

`SimpleSAML\Utils\HTTP::getServerHTTPS()` works perfectly fine as intended, which is:

```php
/**
 * Retrieve HTTPS status from $_SERVER environment variables.
 *
 * @return boolean True if the request was performed through HTTPS, false otherwise.
```

The proper way to handle this case is to just use the `baseurlpath` configuration option to set the full, canonical base URL of the service as seen from outside the proxy.

PR #436, opened 2016-08-09
add negotiate-ext module implementation

ailin-nemui:
this is like negotiate module, but instead of requiring the unstable php-krb5, you can use the existing proven apache modules for negotiate auth

### DIFF
--- a/modules/.gitignore
+++ b/modules/.gitignore
@@ -29,6 +29,7 @@
 !/metarefresh/
 !/multiauth/
 !/negotiate/
+!/negotiateext/
 !/oauth/
 !/portal/
 !/preprodwarning/
--- a/modules/negotiate/dictionaries/negotiate.definition.json
+++ b/modules/negotiate/dictionaries/negotiate.definition.json
@@ -13,5 +13,14 @@
 	},
 	"info_post": {
 		"en": "<h3>What is automatic login<\/h3><p>Automatic login allows you to log in without being asked for username and password:<ul><li>if the computer is in the Kerberos domain<\/li><li>using a supported browser (IE and some others depending on installation options)<\/li><li>on a supported operating system (Windows, Mac)<\/li><\/ul><\/p>"
+	},
+	"disabled_info": {
+		"en": "You have turned off automatic login."
+	},
+	"failed_info": {
+		"en": "Negotiate login was not accepted."
+	},
+	"retry_link": {
+		"en": "Retry login with ticket."
 	}
 }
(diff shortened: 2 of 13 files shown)


ailin-nemui:
diff against modules/negotiate for easier review: https://github.com/simplesamlphp/simplesamlphp/commit/05c70c633f61fc6eb0f5141ec7d6a829b4d5e98b?w=1

jaimeperez:
Hi @ailin-nemui!

Thanks a lot for this! It looks like an outstanding job!

I think it might probably be a good idea to have this module independent from the core distribution. It seems you put a lot of work on it, and that way people would be able to start using it already by installing it with composer (instead of waiting for 1.15.0, which will take quite some time yet), and if you have improvements you can have new, independent releases that people can use, without us delaying the process.

In order to make your module installable with composer, you just need to follow [these instructions](https://github.com/simplesamlphp/composer-module-installer).

Does that sound reasonable? I can also add an entry with information about your module in the [modules page](https://simplesamlphp.org/modules).

ailin-nemui:
hi @jaimeperez 
thanks for getting back to me.

We need a Negotiate module for the IPA-based Kerberos domain. Unstable modules such as php-krb5 are not allowed for us due to policy and honestly I would expect that this php module is not available on most sites, I had to come up with another solution.

Apache modules for negotiate are readily available, packaged in Debian and Ubuntu, and battle-proven. Since the original negotiate module is part of simpleSAMLphp, and I expect this variation thereof to be much easier to deploy if the users are not familiar with custom php module compilation, and I had to write it anyway, I am trying to give it back to you.

Take it or leave it that call is yours, I'm fine with either way. If required, I will continue maintaining it as far as my org is concerned and simpleSAMLphp updates are required, but I wouldn't otherwise desire to or be able to give customer support

thijskh:
This might be obvious (I'm not using this functionality), but wouldn't it be nice if it can just be a configurable option of the existing negotiate module which method to use (krb5 or ext)? Or are the differences too fundamental for that?

ailin-nemui:
I found it to make the code harder to follow but I don't see a technical reason why it wouldn't work.

tvdijen:
I've created a separate repository for this module: https://github.com/simplesamlphp/simplesamlphp-module-negotiate-ext

As soon as the module is registered in Packagist by @jaimeperez , this PR can be closed.

I had to make a few adjustments because so much has changed in SimpleSAMLphp since this PR was created.. Would you be able to give it a proper test @ailin-nemui ?

tvdijen:
The module was registered and is ready for use. Closing this PR

PR #437, opened 2016-08-10
Remove inline style to make css work

dnmvisser:
Removed inline style so that style defined in `www/resources/default.css` actually works.


thijskh:
This change makes the heading turn from white to grey with me. I think it's because default.css:49 makes a:link color #777, and this is now no longer overridden by the specific style attribute in HTML.

PR #440, opened 2016-08-12
Allow the config page to be protected separately.

ghalse:
The admin.protectindexpage setting allows for the entire web interface to be protected. However, some people might want to allow their users to make use of the auth sources test page and/or display an alternative welcome page without revealing internal configuration information. Thus this change gives more fine-grained control over the config page.

It preserves the existing behaviour by defaulting admin.protectconfigpage to false, and because the config page still respects the admin.protectindexpage setting.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -78,10 +78,11 @@ $config = array(
     'auth.adminpassword' => '123',
 
     /*
-     * Set this options to true if you want to require administrator password to access the web interface
-     * or the metadata pages, respectively.
+     * Set this options to true if you want to require administrator password to access the web interface,
+     * just the config page, or the metadata pages, respectively.
      */
     'admin.protectindexpage' => false,
+    'admin.protectconfigpage' => false,
     'admin.protectmetadata' => false,
 
     /*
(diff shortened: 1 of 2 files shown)


tvdijen:
Wouldn't it be a better idea to make this an array of protected (or public) pages, because the admin interface is extensible through modules?

jaimeperez:
Hi Guy(s)! 😄 

I'm a bit uncertain about this, for two reasons: one, as Tim was saying, it's impossible to get a complete list of pages that are protected, and we shouldn't try either (I really don't want to start creating lots of new configuration options, we have more than enough already). The other reason is that we have in mind a complete revamp of the web interface, leaving the most basic authentication part open for users (plus some more pages with options for users such as consent, once you are authenticated), and moving everything else to an _admin_ interface that would protect all pages by default.

So basically I'm not sure if the effort here is worth it or not, considering we would introduce a new configuration option that would be deprecated from the start...

ghalse:
@tvdijen - now that I think about it, that would be a better way of doing it :-). Thanks for the suggestion.
However, in light of what @jaimeperez says, I think it'd make more sense to close the pull request and for us to keep this as a private patch for now.

jaimeperez:
Thanks for being so understanding!

I really believe we need to change completely the way the web interface works today, and hopefully what we have in mind will solve all these issues in the future.

Of course, suggestions are welcome 😉

PR #441, opened 2016-08-15
Make SHA-256 the default signature algorithm.

thijskh:
SHA-1 is still supported but needs to be configured explicitly if you need it.

SHA-1 has been marked as deprecated in SSP 1.12 (2014-03-24). SSP 2.0 seems like a good moment to change the default.


tvdijen:
Since SHA-1 has been marked deprecated for over four years now, can we perhaps set the milestone to 1.16?

jaimeperez:
As long as it is clearly marked in the upgrade notes, I'm all in! After all, almost everybody is switching to SHA-256 already, so it shouldn't be a big deal...

config-templates/authsources.php
```diff
@@ -28,26 +28,6 @@
...
-         *
-         * If you are uncertain about identity providers supporting SHA-256 or other
-         * algorithms of the SHA-2 family, you can configure it individually in the
-         * IdP-remote metadata set for those that support it. Once you are certain that
-         * all your configured IdPs support SHA-2, you can safely remove the configuration
-         * options in the IdP-remote metadata set and uncomment the following option.
-         *
-         * Please refer to the hosted SP configuration reference for more information.
-          */
-        //'signature.algorithm' => 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256',
```
tvdijen: Perhaps it's a good idea to leave this line in place

metadata-templates/saml20-idp-hosted.php
```diff
@@ -23,26 +23,6 @@
...
-	 *
-	 * If you are uncertain about service providers supporting SHA-256 or other
-	 * algorithms of the SHA-2 family, you can configure it individually in the
-	 * SP-remote metadata set for those that support it. Once you are certain that
-	 * all your configured SPs support SHA-2, you can safely remove the configuration
-	 * options in the SP-remote metadata set and uncomment the following option.
-	 *
-	 * Please refer to the IdP hosted reference for more information.
-	 */
-	//'signature.algorithm' => 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256',
```
tvdijen: Same here

modules/adfs/lib/IdP/ADFS.php
```diff
@@ -96,6 +96,7 @@ public static function ADFS_SignResponse($response, $key, $cert) {
 		$objXMLSecDSig->setCanonicalMethod(XMLSecurityDSig::EXC_C14N);	
 		$responsedom = \SAML2\DOMDocumentFactory::fromString(str_replace ("\r", "", $response));
 		$firstassertionroot = $responsedom->getElementsByTagName('Assertion')->item(0);
+		/* TODO: this hardcodes SHA-1. Can we upgrade to something better? */
```
tvdijen: This can be safely set to SHA-256 for ADFS.. The code has significantly changed on master though

config-templates/authsources.php
thijskh: What do you mean precisely?

thijskh: But this is the global config. Would really not want to encourage to switch back to sha1 globally. But instead add it only to specific entities that need it?

thijskh: Removed it here but added it to `saml20-sp-remote` as the most specific place to enable it.

tvdijen: Okay, my idea was to make upgrading as painless as possible for deployers that need sha1 by giving them an example... On second thought however, they're likely to copy their existing config instead of using the template, so it may not make a whole lot of sense... Proper upgrade notes should suffice.

thijskh: Yes but I'd prefer to give them an example that enables it not globally. Because that might be the quickest fix but it's not "painless" in the sense that the pain is then that SHA-1 proliferates.

Can you add a review?

metadata-templates/saml20-sp-remote.php
```diff
@@ -25,3 +25,17 @@
 	'simplesaml.nameidattribute' => 'uid',
 	'simplesaml.attributes' => FALSE,
 );
+
+$metadata['https://legacy.example.edu'] = array(
+	'AssertionConsumerService' => 'https://legacy.example.edu/saml/acs',
+        /*
+         * Currently, SimpleSAMLphp defaults to the SHA-256 hashing algorithm.
+	 * Uncomment the following option to use SHA-1 for signatures directed
+	 * at this specific service provider if it does not support SHA-256 yet.
```
tvdijen: Just a minor things;  indentation is off here

jaimeperez:
> I was wondering if we need to log a deprecation notice when SHA-1 is being used, like we do with deprecated methods... Other than that, thumbs up!

Completely agree here 👍

PR #442, opened 2016-08-16
Remove inline style sheet

dnmvisser:
Remove inline style sheet so external CSS works.
Updated default.css so that things keep the same look.


thijskh:
:+1:

thijskh:
This appears not to be OK afterall. Now other links are also made white. This makes e.g. links on the Authentication tab invisible.

jaimeperez:
I've just fixed it in master with proper CSS changing only anchor elements in the _header_ div.

PR #444, opened 2016-08-18
Add support for regex in consent.disable

vrioux:
Add support for regular expressions in consent.disable to make it easy to disable consent requirement for an entire domain or for trusted domains.  We have over 100 SP defines internally and would like to have consent disabled for all of them easily and without having to update the IDP metadata each time we add a new SP.

Example consent.disable in IDP metadata :
	// Disable consent for our SPs
	'consent.disable' => array(
		'https://mysp.mypartner.com',
		array('type'=>'regex', 'pattern'=>'/.*\.mycompany\.com.*/i'),
	),

### DIFF
--- a/modules/consent/lib/Auth/Process/Consent.php
+++ b/modules/consent/lib/Auth/Process/Consent.php
@@ -154,30 +154,45 @@ class sspmod_consent_Auth_Process_Consent extends SimpleSAML_Auth_ProcessingFilt
             if (count($option) === count($option, COUNT_RECURSIVE)) {
                 // Array is not multidimensional.  Simple in_array search suffices
                 return in_array($entityId, $option, true);
-            } else {
-                // Array contains at least one element that is an array, verify both possibilities
-                if (in_array($entityId, $option, true)) {
-                    return true;
-                } else {
-                    // Search in multidimensional arrays
-                    foreach ($option as $optionToTest) {
-                        if (is_array($optionToTest)) {
-                            if (array_key_exists('type', $optionToTest)) { 
-                                if ($optionToTest['type'] === 'regex') {
-                                    if (array_key_exists('pattern', $optionToTest)) {
-                                        // Evaluate regular expression and return true if entityId matches
-                                        if (preg_match($optionToTest['pattern'], $entityId) === 1) {
-                                            return true;
-                                        }
-                                    }
-                                }
-                            }
-                        }
-                    }
-                    // Base case : no match
-                    return false;
-                }
             }
+
+            // Array contains at least one element that is an array, verify both possibilities
+            if (in_array($entityId, $option, true)) {
+                return true;
+            }
+            
+            // Search in multidimensional arrays
+            foreach ($option as $optionToTest) {
+                if (!is_array($optionToTest)) {
+                    continue; // bad option
+                }
+
+                if (!array_key_exists('type', $optionToTest)) {
+                    continue; // option has no type
+                }
+
+                // Option has a type - switch processing depending on type value :
+                if ($optionToTest['type'] === 'regex') {
+                    // regex-based consent disabling
+
+                    if (!array_key_exists('pattern'), $optionToTest) {
+                        continue; // no pattern defined
+                    }
+
+                    if (preg_match($optionToTest['pattern'], $entityId) === 1) {
+                        return true;
+                    }
+                    
+                } else {
+                    // option type is not supported
+                    continue;
+                }
+
+            } // end foreach
+            
+            // Base case : no match
+            return false;
+
         } else {
             return (boolean)$option;
         }
--- a/modules/consent/lib/Auth/Process/Consent.php
+++ b/modules/consent/lib/Auth/Process/Consent.php
@@ -144,13 +144,40 @@ class sspmod_consent_Auth_Process_Consent extends SimpleSAML_Auth_ProcessingFilt
     /**
      * Helper function to check whether consent is disabled.
      *
-     * @param mixed $option  The consent.disable option. Either an array or a boolean.
+     * @param mixed $option  The consent.disable option. Either an array of array, an array or a boolean.
      * @param string $entityIdD  The entityID of the SP/IdP.
      * @return boolean  TRUE if disabled, FALSE if not.
      */
     private static function checkDisable($option, $entityId) {
         if (is_array($option)) {
-            return in_array($entityId, $option, TRUE);
+            // Check if consent.disable array has one element that is an array
+            if (count($option) === count($option, COUNT_RECURSIVE)) {
+                // Array is not multidimensional.  Simple in_array search suffices
+                return in_array($entityId, $option, true);
+            } else {
+                // Array contains at least one element that is an array, verify both possibilities
+                if (in_array($entityId, $option, true)) {
+                    return true;
+                } else {
+                    // Search in multidimensional arrays
+                    foreach ($option as $optionToTest) {
+                        if (is_array($optionToTest)) {
+                            if (array_key_exists('type', $optionToTest)) { 
+                                if ($optionToTest['type'] === 'regex') {
+                                    if (array_key_exists('pattern', $optionToTest)) {
+                                        // Evaluate regular expression and return true if entityId matches
+                                        if (preg_match($optionToTest['pattern'], $entityId) === 1) {
+                                            return true;
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    // Base case : no match
+                    return false;
+                }
+            }
         } else {
             return (boolean)$option;
         }
--- a/modules/consent/lib/Auth/Process/Consent.php
+++ b/modules/consent/lib/Auth/Process/Consent.php
@@ -10,7 +10,7 @@
 class sspmod_consent_Auth_Process_Consent extends SimpleSAML_Auth_ProcessingFilter
 {
     /**
-     * Button to recive focus
+     * Button to receive focus
      *
      * @var string|null
      */
--- a/modules/consent/lib/Auth/Process/Consent.php
+++ b/modules/consent/lib/Auth/Process/Consent.php
@@ -175,7 +175,7 @@ class sspmod_consent_Auth_Process_Consent extends SimpleSAML_Auth_ProcessingFilt
                 if ($optionToTest['type'] === 'regex') {
                     // regex-based consent disabling
 
-                    if (!array_key_exists('pattern'), $optionToTest) {
+                    if (!array_key_exists('pattern', $optionToTest)) {
                         continue; // no pattern defined
                     }
 
@@ -189,7 +189,7 @@ class sspmod_consent_Auth_Process_Consent extends SimpleSAML_Auth_ProcessingFilt
                 }
 
             } // end foreach
-            
+
             // Base case : no match
             return false;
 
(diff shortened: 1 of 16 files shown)


thijskh:
```
PHP Parse error:  syntax error, unexpected 'in' (T_STRING) in Consent.php on line 163
```

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,34 @@ public function __construct($config, $reserved)
...
+            if (count($option) == count($option, COUNT_RECURSIVE)) {
+                // Array is not multidimensional.  Simple in_array search suffices
+                return in_array($entityId, $option, TRUE);
+            } else {
+                // Array contains at least one element that is an array, verify both possibilities
+                if (in_array($entityId, $option, TRUE)) {
+                    return true;
+                } else {
+                    // Search in multidimensional arrays
+                    foreach($optionToTest in $option) {
```
jaimeperez: s/in/as/

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,34 @@ public function __construct($config, $reserved)
...
      * @param string $entityIdD  The entityID of the SP/IdP.
      * @return boolean  TRUE if disabled, FALSE if not.
      */
     private static function checkDisable($option, $entityId) {
         if (is_array($option)) {
-            return in_array($entityId, $option, TRUE);
+            // Check if consent.disable array has one element that is an array
+            if (count($option) == count($option, COUNT_RECURSIVE)) {
+                // Array is not multidimensional.  Simple in_array search suffices
+                return in_array($entityId, $option, TRUE);
```
jaimeperez: Our [contribution guidelines](https://github.com/simplesamlphp/simplesamlphp/blob/master/CONTRIBUTE.md) specify the use of PSR-2 coding standard. This means boolean constants, `null` and others must be lowercase. So s/TRUE/true/g.

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,34 @@ public function __construct($config, $reserved)
...
-     * @param mixed $option  The consent.disable option. Either an array or a boolean.
+     * @param mixed $option  The consent.disable option. Either an array of array, an array or a boolean.
      * @param string $entityIdD  The entityID of the SP/IdP.
      * @return boolean  TRUE if disabled, FALSE if not.
      */
     private static function checkDisable($option, $entityId) {
         if (is_array($option)) {
-            return in_array($entityId, $option, TRUE);
+            // Check if consent.disable array has one element that is an array
+            if (count($option) == count($option, COUNT_RECURSIVE)) {
```
jaimeperez: Strict comparisons (`===`) should always be used.

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,34 @@ public function __construct($config, $reserved)
...
+                return in_array($entityId, $option, TRUE);
+            } else {
+                // Array contains at least one element that is an array, verify both possibilities
+                if (in_array($entityId, $option, TRUE)) {
+                    return true;
+                } else {
+                    // Search in multidimensional arrays
+                    foreach($optionToTest in $option) {
+                        if (is_array($optionToTest)) {
+                            if ($optionToTest['type'] == 'regex') {
```
jaimeperez: Is the array guaranteed to have the key "_type_"? If not, we should have a check to make sure it's there before trying to access it.

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,34 @@ public function __construct($config, $reserved)
...
+                                // Evaluate regular expression and return true if entityId matches
+                                if (preg_match($optionToTest['pattern'], $entityId) === 1) return true;
```
jaimeperez: Blocks should always be delimited by braces, even those with only one statement inside. In this case, you could even simplify the code, get rid of the test, and just return the value of `preg_match()`, taking special attention to the value:
```php
    return preg_match($optionToTest['pattern'], $entityId) === 1;
```

Also, same thing as before with `$optionToTest['pattern']`. Is it guaranteed to exist?

vrioux: Well if we do that, we'll return false on the first regex that doesn't match.  That is not what we want - we need to iterate on each one until we have tried them all.  Ok for the braces.

jaimeperez: That's absolutely true, sorry!

thijskh:
This might be a matter of taste. But I do wonder if we should replace all those nested if's:

```php
foreach ($option as $optionToTest) {
    if (is_array($optionToTest)) {
        if (array_key_exists('type', $optionToTest)) { 
            if ($optionToTest['type'] === 'regex') {
                if (array_key_exists('pattern', $optionToTest)) {
                    // Evaluate regular expression and return true if entityId matches
                    if (preg_match($optionToTest['pattern'], $entityId) === 1) {
                        return true;
                    }
                }
            }
        }
    }
} 
```

with just some `&&`:

```php
foreach ($option as $optionToTest) {
    // Evaluate regular expression and return true if entityId matches
    if (is_array($optionToTest) &&
        array_key_exists('type', $optionToTest) &&
        $optionToTest['type'] === 'regex' &&
        array_key_exists('pattern', $optionToTest) &&
        preg_match($optionToTest['pattern'], $entityId) === 1) {
            return true;
        }
} 
```

jaimeperez:
My two cents: sometimes I find it easier to test for the negated conditions sequentially, breaking or continuing the loop to match the original logic. Something like:

```php
// evaluate regular expression and return true if entityId matches
foreach ($option as $optionToTest) {
    if (!is_array($optionToTest)) {
        continue; // bad option
    }

    if (!array_key_exists('type', $optionToTest)) {
        continue; // option has no type
    }

    if ($optionToTest['type'] !== 'regex') {
        continue; // not a regex
    }
    
    if (!array_key_exists('pattern', $optionToTest)) {
        continue; // no pattern
    }
    
    if (preg_match($optionToTest['pattern'], $entityId) === 1) {
        return true;
    }
}
```

This approach has more or less the same advantages as the single `if` statement (mainly to stop testing conditions if one is not met), while helping a bit with readability. But again, this is just personal preferences 😉

vrioux:
Should I do something to have this pull accepted?

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,40 @@ public function __construct($config, $reserved)
...
      * @return boolean  TRUE if disabled, FALSE if not.
      */
     private static function checkDisable($option, $entityId) {
         if (is_array($option)) {
-            return in_array($entityId, $option, TRUE);
+            // Check if consent.disable array has one element that is an array
+            if (count($option) === count($option, COUNT_RECURSIVE)) {
+                // Array is not multidimensional.  Simple in_array search suffices
+                return in_array($entityId, $option, true);
+            } else {
```
jaimeperez: Given that the main `if` clause will always end the execution of this method, the `else` is useless here, it only adds an indentation level and increases complexity. You can just remove the `else` and have the code inside execute next to the `if`:

```php
if (count($option) === count($option, COUNT_RECURSIVE)) {
    // Array is not multidimensional.  Simple in_array search suffices
    return in_array($entityId, $option, true);
 }

// Array contains at least one element that is an array, verify both possibilities
if (in_array($entityId, $option, true)) {
    return true;
}
```

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,40 @@ public function __construct($config, $reserved)
...
+                // Array contains at least one element that is an array, verify both possibilities
+                if (in_array($entityId, $option, true)) {
+                    return true;
+                } else {
```
jaimeperez: Same as before here.

thijskh:
I guess not. Since we don't seem to have a consistent opinion on the if branching question, I think it may be ready to merge.

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,40 @@ public function __construct($config, $reserved)
...
+                    // Search in multidimensional arrays
+                    foreach ($option as $optionToTest) {
+                        if (is_array($optionToTest)) {
+                            if (array_key_exists('type', $optionToTest)) { 
+                                if ($optionToTest['type'] === 'regex') {
+                                    if (array_key_exists('pattern', $optionToTest)) {
+                                        // Evaluate regular expression and return true if entityId matches
+                                        if (preg_match($optionToTest['pattern'], $entityId) === 1) {
+                                            return true;
```
jaimeperez: As Thijs mentioned, this is quite complex and a bit difficult to follow (even though the sequence of `if` statements make full sense and are all necessary), so I would ask to refactor it a bit, either as Thijs suggested, or as I did. Note also that some `if` statements can be merged together (e.g. the two testing for `type` and `pattern` keys in the `$optionToTest` array, as if any of them is missing, you shouldn't continue).

modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -144,13 +144,40 @@ public function __construct($config, $reserved)
...
+                    return true;
+                } else {
+                    // Search in multidimensional arrays
+                    foreach ($option as $optionToTest) {
+                        if (is_array($optionToTest)) {
+                            if (array_key_exists('type', $optionToTest)) { 
+                                if ($optionToTest['type'] === 'regex') {
+                                    if (array_key_exists('pattern', $optionToTest)) {
+                                        // Evaluate regular expression and return true if entityId matches
+                                        if (preg_match($optionToTest['pattern'], $entityId) === 1) {
```
jaimeperez: It might be a good idea to enforce the regex delimiters here, instead of leaving the user the responsibility to do that. In case the user should do that him/herself, the documentation should clearly indicate so. @thijskh, what do you think?

vrioux:
I'm looking for an example of unit tests in a module directory and can't find one.  The tests are all in the tests/ directory at the root.  Should I do it there and create the full tests/modules/consent/lib/Auth/Process/ConsentTest.php path?

jaimeperez:
Thanks so much @vrioux!

Now that I realize, since this module is distributed together with SimpleSAMLphp, it would be hard to have the tests on it. At some point it would be helpful to have the tests self-contained in its own directory, if we ever decide to extract the module. But if that decision is ever made, we'll have time to fix it. So yes, `tests/modules/consent/lib/Auth/Process/ConsentTest.php` would be the right place for it, sorry for the noise, and thanks again for your valuable help!

tests/modules/consent/lib/Auth/Process/ConsentTest.php
```diff
@@ -0,0 +1,109 @@
...
+ * Test for the consent:Process filter.
+ *
+ * @author Vincent Rioux <vrioux@ctech.ca>
+ * @package SimpleSAMLphp
+ */
+
+namespace SimpleSAML\Test\Module\consent\Auth\Process;
+
+
+class ProcessTest extends \PHPUnit_Framework_TestCase
```
jaimeperez: The name of the class here should be `ConsentTest` 😉

vrioux:
I did a very basic test method that seems to indicate that the code actually works...  

I wish I had more time to help out and maybe I will at a later time, but at least it's a start, I guess...

jaimeperez:
Thanks a lot @vrioux! I really appreciate your effort 😄 

I'm fine with the current code. We can still increase a bit the coverage of the method by adding tests for all corner cases, but that shouldn't prevent this from getting merged. @thijskh, do you agree?

thijskh:
Agreed

jaimeperez:
Great! Thanks a lot again for the contribution!

PR #447, opened 2016-08-21
Enable tracking of number of failed authentication attempts

synnove:
Create an element in the state array, 'AuthFailCount', which stores the number
of failed authentication attempts by a user in a session. This can be used to
trigger various follow-up actions when limits are reached, such as showing the
user a captcha or rate-limiting/ blocking login attempts.

### DIFF
--- a/modules/core/lib/Auth/UserPassBase.php
+++ b/modules/core/lib/Auth/UserPassBase.php
@@ -244,6 +244,16 @@ abstract class sspmod_core_Auth_UserPassBase extends SimpleSAML_Auth_Source {
 		}
 
 		/*
+		 * Prior to attempting login, set AuthFailCount to 0 if it doesn't already
+		 * exist in the state array, then save the state.
+		 */
+
+		if (!isset($state['AuthFailCount'])) {
+			$state['AuthFailCount'] = 0;
+			SimpleSAML_Auth_State::saveState($state, self::STAGEID);
+		}
+
+		/*
 		 * $source now contains the authentication source on which authenticate()
 		 * was called. We should call login() on the same authentication source.
 		 */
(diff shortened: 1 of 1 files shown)


hordijk:
To use this change with ease by other developers/users you might one add support for this variable in the data array of the template within simplesamlphp/modules/core/www/loginuserpass.php file. Based on this data variable the other users which only added their own template can use this immediately.
We should not retrieve the state in the template itself. In the template we could show for example a captcha or a warning message based on the data[AuthFailCount] value.

jaimeperez:
Hi!

@hordijk, what you are suggesting is to add logic to templates, which is something you should never do. Templates should only render HTML and have very, very basic logic (if clauses and loops), so if you need to display a captcha, you should have your logic in the page using the variable and then pass just a boolean variable to the template to tell it whether display it or not.

In any case, I'm afraid this implementation falls short, since it tracks the authentication failures in general. It should be auth-source specific, so that if you tried to login way too many times, that doesn't block other auth sources. Imagine you block yourself out trying to login, but you are also able to identify as administrator. Both should be independent.

In general, I would say a captcha or other similar mechanisms are more security theater than actual security measures. Captchas are only useful if they are used **always**, because then there's no way to circumvent them. But that's very inconvenient for end users, so you definitely don't want your legitimate users to go through a captcha every time they login. If you don't have it always, an attacker can always throw the session and start over, and then the count of authentication failures goes down the drain...

hordijk:
I wanted to mention without changing simplesamlphp/modules/core/www/loginuserpass.php the templates won't have information of AuthFailCount and therefore cannot use basic logic to show any message based on this property. Since the change is performed in modules/core/lib/Auth/UserPassBase.php which is probably extended by custom username/password authentication implementations other users probably want to use this new functionality somehow?

The UserPassBase.php is using $url = SimpleSAML_Module::getModuleURL('core/loginuserpass.php');. This file is actually setting the data properties available to the template. It seems we have to override some functions of the UserPassBase.php if we want to implement it properly.

I agree by starting a new sesssion you start all over and you indeed want to limit the counter per auth- source. So this pull request needs some additional work ;-)

jeremyfrench:
I like the idea of this, wouldn't the state be destroyed if an attacker clears session cookies (or has a botnet)

I think the goal is achievable but only if the attempts are stored in a persistent server side store.

jaimeperez:
Hi @jeremyfrench!

As you are saying, this is extremely easy to circumvent. Clearing cookies is enough. Using a botnet with thousands of clients is another possibility, and of course nothing prevents an attacker from discarding cookies on every botnet node after trying a password.

So, as I was saying previously, this would be more security theater than an actual counter meassure.

tvdijen:
I think we have a proper alternative now:  https://github.com/cirrusidentity/simplesamlphp-module-ratelimit
Therefore closing this PR.

PR #453, opened 2016-09-01
Clarify NameIDPolicy configuration

gollmann:
Describe how a NameIDPolicy element can be removed from an
authentication request.

### DIFF
--- a/modules/saml/docs/sp.md
+++ b/modules/saml/docs/sp.md
@@ -224,6 +224,7 @@ Options
 `NameIDPolicy`
 :   The format of the NameID we request from the IdP.
     Defaults to the `transient` format if unspecified.
+    A value of `NULL` suppresses the NameIDPolicy element in the request.
 
 :   If this option is set, its value will be added to the metadata generated for this SP, in the NameIDFormat element.
 


gollmann:
In some situations it is preferable to not specify a NameIDPolicy in the authentication request, e.g when the SP either accepts an eduPersonTargetedID or a persistent NameID. While this can be done in SSP the documentation does not mention it yet.

thijskh:
Are you sure? If I add in authsources.php in the `default-sp` array:
```
'NameIdPolicy' => null,
```

I get:
```
Backtrace:
1 www/_include.php:45 (SimpleSAML_exception_handler)
0 [builtin] (N/A)
Caused by: Exception: [ARRAY]: The option 'Format' is not a valid string value.
Backtrace:
5 lib/SimpleSAML/Configuration.php:690 (SimpleSAML_Configuration::getString)
4 modules/saml/lib/Message.php:421 (sspmod_saml_Message::buildAuthnRequest)
3 modules/saml/lib/Auth/Source/SP.php:190 (sspmod_saml_Auth_Source_SP::startSSO2)
2 modules/saml/lib/Auth/Source/SP.php:335 (sspmod_saml_Auth_Source_SP::startSSO)
1 modules/saml/www/sp/discoresp.php:28 (require)
0 www/module.php:135 (N/A)
```

gollmann:
Works in 1.14.7. It seems Message.php has since lost the ability to send no NameIDPolicy at all in the authentication request. This is unfortunate, since my federation contact maintains, that it is better to request no NameIDPolicy than to request urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified.

thijskh:
It might never have been intended to work that way as an explicit feature, and this also explains why it was not documented, not part of a unit test and lost when refactoring the code.

The code was changed in 412fde9517d3bc961193412b13c11ce86138fdcc, which did in fact aim to make NameIDPolicy more flexible. So I conclude that your request seems to boil down to adding (back) support for no NameIDPolicy at all.

gollmann:
> So I conclude that your request seems to boil down to adding (back) support for no NameIDPolicy at all.

That is correct. TIA

thijskh:
Alright, so the current PR should not be merged. Do you want to create a new PR with a proposed change or an issue with an enhancement request? In any case I'll close this one.

PR #469, opened 2016-09-13
Adfs

pgh70:
Two improvements to the ADFS module:
1. Honour namespace when available... Fall back to use 'http://schemas.xmlsoap.org/claims' if none given... Namespaces can be added to the attributes using the core:AttributeMap name2claim. This enables the use of locally defined claims.
2. Honour adfs:wreply parameter when available... Using with sharepoint apps, the URL of an app is dynamically created... A static config in SSP can't be used. Using the adfs:wreply to let sharepoint dynamically fill in the return url is a solotion for this problem. If no adfs:wreply parameter is present, the default behaviour (use the configured prp url) is used.

### DIFF
--- a/modules/adfs/lib/IdP/ADFS.php
+++ b/modules/adfs/lib/IdP/ADFS.php
@@ -47,7 +47,7 @@ class sspmod_adfs_IdP_ADFS {
 	public static function ADFS_GenerateResponse($issuer, $target, $nameid, $attributes) {
 		$issueInstant = SimpleSAML\Utils\Time::generateTimestamp();
 		$notBefore = SimpleSAML\Utils\Time::generateTimestamp(time() - 30);
-		$assertionExpire = SimpleSAML\Utils\Time::generateTimestamp(time() + 60 * 5);
+		$assertionExpire = SimpleSAML\Utils\Time::generateTimestamp(time() + 60 * 60);
 		$assertionID = SimpleSAML\Utils\Random::generateID();
 		$nameidFormat = 'http://schemas.xmlsoap.org/claims/UPN';
 		$result =
--- a/modules/adfs/lib/IdP/ADFS.php
+++ b/modules/adfs/lib/IdP/ADFS.php
@@ -72,8 +72,8 @@ class sspmod_adfs_IdP_ADFS {
 			if ((!is_array($values)) || (count($values) == 0)) continue;
 			$hasValue = FALSE;
 			if (strpos($name, '/')) {
-				$r = '<saml:Attribute AttributeNamespace="'.substr($name, 0, strrpos($name, '/', -1)).
-					'" AttributeName="'.substr($name, strrpos($name, '/', -1) + 1).'">';
+				$r = '<saml:Attribute AttributeNamespace="'.htmlspecialchars(substr($name, 0, strrpos($name, '/', -1))).
+					'" AttributeName="'.htmlspecialchars(substr($name, strrpos($name, '/', -1) + 1)).'">';
 			} else {
 				$r = '<saml:Attribute AttributeNamespace="http://schemas.xmlsoap.org/claims" AttributeName="' .
 					htmlspecialchars($name) .'">';
--- a/modules/adfs/lib/IdP/ADFS.php
+++ b/modules/adfs/lib/IdP/ADFS.php
@@ -70,7 +70,13 @@ class sspmod_adfs_IdP_ADFS {
 		foreach ($attributes as $name => $values) {
 			if ((!is_array($values)) || (count($values) == 0)) continue;
 			$hasValue = FALSE;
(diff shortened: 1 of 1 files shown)


jaimeperez:
Thanks for the PR @pgh70!

I'll make a couple of comments directly on the code.

modules/adfs/lib/IdP/ADFS.php
```diff
@@ -70,7 +71,13 @@ public static function ADFS_GenerateResponse($issuer, $target, $nameid, $attribu
 		foreach ($attributes as $name => $values) {
 			if ((!is_array($values)) || (count($values) == 0)) continue;
 			$hasValue = FALSE;
-			$r = '<saml:Attribute AttributeNamespace="http://schemas.xmlsoap.org/claims" AttributeName="' . htmlspecialchars($name) .'">';
+			if (strpos($name, '/')) {
```
jaimeperez: Would a strict comparison necessary here?

modules/adfs/lib/IdP/ADFS.php
```diff
@@ -70,7 +71,13 @@ public static function ADFS_GenerateResponse($issuer, $target, $nameid, $attribu
...
+				$r = '<saml:Attribute AttributeNamespace="'.substr($name, 0, strrpos($name, '/', -1)).
+					'" AttributeName="'.substr($name, strrpos($name, '/', -1) + 1).'">';
```
jaimeperez: Here I think we should also use `htmlspecialchars()`, as in the other branch of the `if` clause.

modules/adfs/lib/IdP/ADFS.php
```diff
@@ -157,7 +164,7 @@ public static function sendResponse(array $state) {
 		$wresult = sspmod_adfs_IdP_ADFS::ADFS_SignResponse($response, $privateKeyFile, $certificateFile);
 
 		$wctx = $state['adfs:wctx'];
-		sspmod_adfs_IdP_ADFS::ADFS_PostResponse($spMetadata->getValue('prp'), $wresult, $wctx);
+		sspmod_adfs_IdP_ADFS::ADFS_PostResponse(($state['adfs:wreply']) ? $state['adfs:wreply'] : $spMetadata->getValue('prp'), $wresult, $wctx);
```
jaimeperez: Here we should be explicit, I think. Either by calling `isset($state['adfs:wreply'])` if that key is only set conditionally, or by setting it to false if no `wreply` parameter is present.

modules/adfs/lib/IdP/ADFS.php
pgh70: If it contains a '/' it will assume the attribute name is with namespace... Checking if it is in fact a namespace contained attribute-name seems a challenge... Worst case: a invalid namespace is issued, which simply doesn't work...

modules/adfs/lib/IdP/ADFS.php
```diff
@@ -38,6 +38,7 @@ public static function receiveAuthnRequest(SimpleSAML_IdP $idp) {
 			'ForceAuthn' => $forceAuthn,
 			'isPassive' => $isPassive,
 			'adfs:wctx' => $wctx,
+			'adfs:wreply' => (array_key_exists('wreply',$_GET)) ? $_GET['wreply'] : ''
```
jaimeperez: This is the main problem I see, as it introduces an [_open redirection_ vulnerability](https://www.owasp.org/index.php/Top_10_2013-A10-Unvalidated_Redirects_and_Forwards).

Any URL there could be used by an attacker to force the response from the IdP to be sent to another place, under his/her control. We should therefore restrict the URLs allowed as return URLs, the sooner the better.

This could be fixed by adding the `adfs:wreply` key conditionally, only if the `wreply` query parameter is present, and the URL contained there is a valid URL and matches some whitelisted host, by calling `SimpleSAML\Utils\HTTP::isURLAllowed()` on it.

modules/adfs/lib/IdP/ADFS.php
pgh70: Good point... Should not be nesessary, but better safe than sorry

modules/adfs/lib/IdP/ADFS.php
pgh70: The key is set to an empty string when unavailable, which renders to a false value.. Explicit make is 'false' is perhaps better..

modules/adfs/lib/IdP/ADFS.php
pgh70: The redirection only occurs when successful login, so not a really big issue, but still it must be addressed..

I will look into the whitelisting of hosts... Additionally i think i add a parameter in metadata which must be explicit set to allow the use ud adfs:wreply

tvdijen:
Everything here has been addressed in https://github.com/simplesamlphp/simplesamlphp/pull/760

Issue #455, opened 2016-09-02
Pre-migrate: simplify translations


hmpf:
Ensure that there are no usage of `t()` that uses concatenation in the tag, or deprecated arguments. Use only two arguments: the `$tag`, which must be a string or a variable, and optional `$replacements`, which stays a key-value array.

hmpf:
Also ensure that absolutely all translations uses `t()` and nothing else.

hmpf:
The following modules have templates with translations that needs simplifying:

- authorize
- authX509

(List not exhaustive)

hmpf:
The following modules need no simplification:

- expirycheck
- negotiate

(List not exhaustive)

hmpf:
The very worst modules to fix are:

- consentAdmin
- feide
- core
- saml
- discopower
- consent

**In that order**

hmpf:
See the link for an attempt at an analysis of all existing usage of `t()`: [Translation analysis](https://github.com/simplesamlphp/simplesamlphp/files/458939/translation-analysis.txt)

### Format

Each line starts with the filename of a file the `t()`-usage was found in. If the same content is translated several times in a single file, there will be only one entry in the file.

After the filename comes what was found between the parentheses of the `t()`:

An opening brace (`{`) on a line means the string is namespaced. Two colons after the brace means the domain has a subdomain. **We don't want any subdomains.**

`, array` near the end of a line means the translation has placeholders. If the `array` is followed by anything, the `t()` uses deprecated arguments. **Get rid of the deprecated arguments.**

` . concat` near the end of a line means the string is concatenated and not findable by an extractor. **We don't want any of these!**

` $` right after the leftmost colon (which marks the end of the filename) means there is no string to translate, but a variable. **We need a `noop()` version of the actual contents of the variable somewhere so that extractors can find it!**

`NL` means the contents of the `t()` is spread over multiple lines. Do we want that?

The final possibility is only found in 
`modules/core/templates/frontpage_federation.tpl.php`: translating the returned value from a function, in this case `mtype`. **Let's move that mtype out of the template.**

hmpf:
Keep subdomains for 1.15, switch to strings instead of tags for 2.0 which also solves the subdomain problem.

tvdijen:
@hmpf Would it be possible to run another analysis on this? I asume you've scripted that before...
Right now it's kinda hard to see what's left to be done..

hmpf:
[...]

PR #474, opened 2016-09-16
Errorcode class

hmpf:
This will get rid of half of the concatenated translation strings. For #455.


hmpf:
Other version is in, but the entire concept should be changed so that we no longer use string-based errors.

PR #476, opened 2016-09-20
Change method and variable visibility

erny:
I would like to inherit from this module, to check the certificate against another system, instead of LDAP. It would be of help if some auxiliary methods were public / protected.

### DIFF
--- a/modules/authX509/lib/Auth/Source/X509userCert.php
+++ b/modules/authX509/lib/Auth/Source/X509userCert.php
@@ -13,13 +13,13 @@ class sspmod_authX509_Auth_Source_X509userCert extends SimpleSAML_Auth_Source {
 	 * x509 attributes to use from the certificate
 	 * for searching the user in the LDAP directory.
 	 */
-	private $x509attributes = array('UID' => 'uid');
+	protected $x509attributes = array('UID' => 'uid');
 
 
 	/**
 	 * LDAP attribute containing the user certificate
 	 */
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @erny!

Actually, the two methods you want to use are probably interesting enough to be public in an utilities class, instead of there. We could probably move them to be static methods in, say, `SimpleSAML\Utils\Crypto`, and have the code using the private methods you are changing here use the new static ones. Would that make sense to you?

thijskh:
We should definitely do that, because already we're performing such operations in various places in the code (look for `chunk_split` for example).

erny:
Hi.

Yes, it makes sense to me.

On the other hand, I'm trying to validate the certificate against an
external system, and the authenticate method is too monolithic. I don't
need all the LDAP stuff here. Perhaps I'll propose some minor refactorings.
(I'm not developing against trunk, but an older version.)

Thanks
Best regards.
Erny



Ernesto Revilla
Área Técnica
TangramBPM.es
Tlf: 630 244 136



2016-09-21 9:49 GMT+02:00 Thijs Kinkhorst <notifications@github.com>:

jaimeperez:
¡Hola @erny!

I've just pushed some changes to add a couple of methods, `SimpleSAML\Utils\Crypto::pem2der()` and `SimpleSAML\Utils\Crypto::der2pem()`, to fullfill your needs. They are valid for any kind of DER/PEM-encoded data, not only certificates.

Hope that helps.

Issue #365, opened 2016-04-18
Session problems!


urban-1:
Hi all,

I am very new to SAML so I might be doing things wrong. I have described my problem at SO: http://stackoverflow.com/questions/36655953/simplesamlphp-unable-to-find-the-current-binding. (I have a local copy in case it gets closed...)

All the above errors seem to appear randomly due to faulty state of some sort...  I am using (and modifying) `$_SESSION` in my application and I have noticed that it is getting changed/replaced by SSP. If I remove all the code that changes session (including `_start()`/`_destroy()`), login and logout works without crashing. Is there a specific way I should be handling the session?

I have tried to set the `store.type` to `sql`, in which case, an sqlite file is created but SSP always comes back with "NO STATE" error. Also note that I can see apache segfaulting in the logs

My site login logic:

```
        // If loaded do not load again (depending on force value)
        if (isset($_SESSION['user']['realname']) && !$force) {
           return true;
        }
        
        // 1st check and use SAML
        if (!isset($_SESSION['user']['username']) ) {
            // Try SAML here
            error_log("initUser::checking SAML");
            static::doSAMLAuth();
        }
        
        // 2nd check and FAIL
        if (!isset($_SESSION['user']['username']) ) {
            error_log("Session not loaded???");
            return false;
        }
        
        // Update the SESSION
        
        // LDAP will update stuff!
        $ldap = new LDAP("AD");
        error_log("initUser::Quering LDAP");
        $ldapUser = $ldap->getUser($_SESSION['user']['username']);`
        $_SESSION['user'] = $ldap->ldapUser2Web($ldapUser);
```

and the SSP authentication:

```
       $auth = new SimpleSAML_Auth_Simple(SSAMLSP);
        
        // This should do the correct redirect
        error_log("doSAML::requireAuth");
        $auth->requireAuth();
        
        error_log("doSAML::isAuthenticated");
         
        // Setup session for this SP
        $ad = $auth->getAuthData("saml:sp:NameID");
        
        $_SESSION['user']['username'] = $ad["Value"];
```

Any help would be appreciated

Andreas

Working on:
 RHEL 7
 PHP 5.4.16
 Apache 2.4
 simplesamlphp 1.14.2

jaimeperez:
Hi @urban-1!

[...]

PR #478, opened 2016-09-23
Bug/previous sessions

shuadoc:
Related issues:

https://github.com/simplesamlphp/simplesamlphp/pull/349
https://github.com/simplesamlphp/simplesamlphp/issues/365
https://github.com/simplesamlphp/simplesamlphp/pull/224

This extends the patch for those issues, saving the previous_session across multiple requests.

### DIFF
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -138,6 +138,9 @@ class SimpleSAML_Session implements Serializable
      * Private constructor that restricts instantiation to either getSessionFromRequest() for the current session or
      * getSession() for a specific one.
      *
+     * Note that this constructor is not called with every request. The object
+     * is also loaded from storage via SessionHandler->loadSession()
+     *
      * @param boolean $transient Whether to create a transient session or not.
      */
     private function __construct($transient = false)
@@ -235,7 +238,9 @@ class SimpleSAML_Session implements Serializable
 
 
     /**
-     * Retrieves the current session. Creates a new session if there's not one.
+     * Returns the SimpleSAML_Session for this session
+     * - Retrieves from storage the first time this function is called during a request
+     * - Creates new object if this is the first request of a new session
      *
      * @return SimpleSAML_Session The current session.
      * @throws Exception When session couldn't be initialized and the session fallback is disabled by configuration.
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -135,6 +135,15 @@ class SimpleSAML_Session implements Serializable
 
 
     /**
+     * The session ID used by the application including simplesamlphp
+     * See SessionHandlerPHP->previous_session for more details
+     *
+     * @var array
+     */
+    private $previousSession = array();
+
+
+    /**
      * Private constructor that restricts instantiation to either getSessionFromRequest() for the current session or
      * getSession() for a specific one.
      *
@@ -396,6 +405,7 @@ class SimpleSAML_Session implements Serializable
     {
         SimpleSAML\Logger::setTrackId($session->getTrackID());
         self::$instance = $session;
+        $session->maintainPreviousSession();
         return self::$instance;
     }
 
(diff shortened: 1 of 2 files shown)


jaimeperez:
Hi @shuadoc!

Could you please elaborate a bit on why is this necessary?

shuadoc:
Of course,

This is needed to preserve the previous_session variable across multiple requests.  Currently previous_session is stored in an object which is destroyed at the end of a single request.  During the second request, previous_session is actually set to the current session.

This commit stores previous_session in the active session to be available for subsequent requests.

An example of when this is necessary:

 - User visits service provider
 - Service provider sets session variables
 - User redirected to identity provider
 - User logs in and is redirected back to service provider

Current behavior in detail:

 - User visits service provider
 - Service provider sets session variables
 - Simplesamlphp is loaded, changes session
 - Simplesamlphp redirects user to identity provider
 - (Request ends, SessionHandler and previous_session are destroyed, cannot call cleanup())
 - User returns from identity provider
 - Simplesamlphp is loaded, sesson stays the same
 - (At this point, the SessionHandler sets previous_session to be the same as the active session)
 - Simplesamlphp finishes execution, returns control to service provider 
 - Service provider calls cleanup(), session remains the same
 - Service provider continues without session variables from before

The only options I can see are:
 - Always have simplesamlphp handle our applications sessions, even for clients who aren't using saml
 - Store previous_session in the current session
 - Prevent simplesamlphp from altering active sessions

Thank you!
Josh

jaimeperez:
Thanks for the detailed response Josh!

I think the problem here is due to the application relying on the `cleanup()` method to retrieve the session, while it should retrieve it normally, just as before the application was _domesticated_ to use SimpleSAMLphp. Let me make some comments on the alternatives you are suggesting:

> Always have simplesamlphp handle our applications sessions, even for clients who aren't using saml

You shouldn't mix both. All the changes in the last releases were intended precisely to keep both sessions (SimpleSAMLphp's and the application's) separated. That way you avoid any possible bugs in the application to mess SSP's session, and the other way around. Besides, there are no guarantees that SimpleSAMLphp will respect (today or in the future) whatever information you store on its session.

> Store previous_session in the current session

Maybe the name of the property is a bit misleading. It's not intended to keep track of a session existing before, as there's no relationship in time between both sessions (they are supposed to coexist). On the other hand, this new mechanism is intended to allow an application to keep its own session when using SimpleSAMLphp.

> Prevent simplesamlphp from altering active sessions

SimpleSAMLphp does not alter active sessions. That precisely why it closes the session before opening its own one, to avoid any alterations. Unfortunately, PHP's session handler does not allow to have two sessions open at the same time, and that's why we need to do this.

In any case, to summarize: if you are using PHP sessions both in your application and in SimpleSAMLphp, you need to make sure they use different cookies, and in your application you should do things exactly the same way as you were doing them before, with two exceptions:

1. If you weren't using named sessions, you will need to start doing it (`session_name()`).
2. You will also need to call the `cleanup()` method **after** calling SimpleSAMLphp.

So your application should be doing something like this:

```php
<?php

include('....../lib/_autoload.php');

$as = new SimpleSAML_Auth_Simple('default-sp');

session_start();

if ($_SESSION['is_valid']) {
    echo "We are already authenticated";
} else {
    $as->requireAuth();
    $s = SimpleSAML_Session::getSessionFromRequest();
    $s->cleanup();

    $_SESSION['is_valid'] = true;
}

// do whatever we need to do
```

I'm closing the PR for now. Feel free to comment if you think I'm missing something.

shuadoc:
Thanks Jaime,

After setting 'session.phpsession.cookiename' to a value other than the default, this is working correctly without the above commit.

It would appear that PHP sessions work fine with the default session config, until including simplesamlphp with default session config.   All appears to be going well, until you need a specific edge case to work properly.  This seems undesirable IMHO.

Please rate the following options:

A) A logged warning when 'session_name' and 'session.phpsession.cookiename' are configured such that there will be a collision.

B) Instructions to set either session_name and/or session.phpsession.cookiename here: https://simplesamlphp.org/docs/1.14/simplesamlphp-maintenance#section_5

C) config-templates/config.php to contain instructions on why they might want to set session.phpsession.cookiename (and that if they set this configuration, they must also use the cleanup() function.)

D) Simplesamlphp to cleanup() after itself before exiting, or before returning control to calling application.  (If and only if the two session_names are different.)  (The cost of this one is higher than the benefit.  Still, it's often nice to clearly state what might happen in an ideal world.)

Regards,
Josh

jaimeperez:
Thanks for the feedback @shuadoc, glad you got it working!

I wouldn't say this is a corner case. If your application is already using PHP sessions and you configure SimpleSAMLphp to use them too, you need at least one of them to **not** use the defaults.

In any case, regarding your suggestions:

A) is a very good idea, thanks!

B) should probably be mentioned [here](https://simplesamlphp.org/docs/1.14/simplesamlphp-maintenance#section_2) (the section on session configuration).

C) is the consequence of an unfortunate decision. There shouldn't be two different configuration options to set some properties of the session, depending on whether you are using PHP sessions or something else. What we should do is actually to get rid of the `session.phpsession.*` configuration options, IMHO. In any case, there is already [documentation talking about using the `cleanup()` method](https://simplesamlphp.org/docs/stable/simplesamlphp-sp#section_6).

D) is impossible to do, unfortunately. The reason is that you can't know when the user is done using your API. Should we clean the session up when returning from `requireAuth()`? What if the app calls `getAttributes()` afterwards? If we do that, we would need to do it for each exposed API method, and that means loading and unloading sessions multiple times, with the overhead that implies. So I think the lesser evil here is to provide a method for the user to mark the end of the use of the API, so that the previous status is restored.

Thanks again for your suggestions, they are duly noted and will be addressed as soon as possible 😃

PR #479, opened 2016-09-25
Fix minor XHTML compliancy issues

tvdijen:


modules/core/templates/frontpage_federation.tpl.php
```diff
@@ -10,9 +10,9 @@
         $this->t('{core:frontpage:login_as_admin}').'</a></p>';
 }
 
-$now = time();
-echo '<dl>';
 if (is_array($this->data['metaentries']['hosted']) && count($this->data['metaentries']['hosted']) > 0) {
+    $now = time();
```
thijskh: I think this breaks the current code because `$now` is not used inside this `if()` block but in the next one, so if this `if` is not true, then `$now` will be unset.

tvdijen:
You're right, I don't know what I was thinking :') Should be good now.

tvdijen:
By the way, the self-closing tags are still allowed in HTML5 for some elements (img, hr, br), so this won't PR won't frustrate a future HTML5-migration.

Issue #458, opened 2016-09-05
PSR-2 compliant whitespace


hmpf:
Fix whitespace as per PSR-2.

When fixing whitespace, ONLY do that for that commit, as this makes `git blame -w` work better.
Reference this issue in the commit message, for instance in the body as "For: #458". Don't do too many files in one go, only one module per commit for instance.

If fixing whitespace in a template leads to it rendering differently, the template is bugged. Take screenshots of before and after, add them to a new issue referencing this one, and also mention what browser you are using.

tvdijen:
Whitespace is done..
I've just ran phpcs and there are a handful of other PSR-2 issues that I can address later today.

PR #481, opened 2016-09-28
PSR-2 fix for ADFS module

tvdijen:
I intentionally skipped whitespace-fixes, since there's a separate issue 458 for that.

### DIFF
--- a/modules/adfs/lib/SAML2/XML/fed/Endpoint.php
+++ b/modules/adfs/lib/SAML2/XML/fed/Endpoint.php
@@ -4,28 +4,29 @@
  *
  * @package SimpleSAMLphp
  */
-class sspmod_adfs_SAML2_XML_fed_Endpoint {
-	/**
-	 * Add this endpoint to an XML element.
-	 *
-	 * @param DOMElement $parent  The element we should append this endpoint to.
-	 * @param string $name  The name of the element we should create.
-	 */
-	public static function appendXML(DOMElement $parent, $name, $address) {
-		assert('is_string($name)');
-		assert('is_string($address)');
+class sspmod_adfs_SAML2_XML_fed_Endpoint
+{
+    /**
+     * Add this endpoint to an XML element.
+     *
+     * @param DOMElement $parent  The element we should append this endpoint to.
+     * @param string $name  The name of the element we should create.
+     */
+    public static function appendXML(DOMElement $parent, $name, $address)
+    {
+        assert('is_string($name)');
+        assert('is_string($address)');
 
-		$e = $parent->ownerDocument->createElement($name);
-                $parent->appendChild($e);
+        $e = $parent->ownerDocument->createElement($name);
+        $parent->appendChild($e);
 
-		$endpoint = $parent->ownerDocument->createElement('EndpointReference');
-                $endpoint->setAttribute('xmlns', 'http://www.w3.org/2005/08/addressing');
-		$e->appendChild($endpoint);
+        $endpoint = $parent->ownerDocument->createElement('EndpointReference');
+        $endpoint->setAttribute('xmlns', 'http://www.w3.org/2005/08/addressing');
+        $e->appendChild($endpoint);
 
-		$address = $parent->ownerDocument->createElement('Address', $address);
-		$endpoint->appendChild($address);
+        $address = $parent->ownerDocument->createElement('Address', $address);
+        $endpoint->appendChild($address);
 
-		return $e;
-	}
-
+        return $e;
+    }
 }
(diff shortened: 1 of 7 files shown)


jaimeperez:
Thanks a lot @tvdijen!

#458 is basically a reminder to keep track of what we need to do, but that doesn't mean we need to make all white spaces in the project compliant with PSR-2 at the same time, in the same commit. Actually, I think it's better if we do that slowly, file by file. So I would say that's something you can address here too, and then we won't have files that are PSR-2 compliant except for the whitespace.

You could probably squash the existing commits, and then add another one changing tabs for spaces. Does that sound reasonable?

tvdijen:
Sure, I can fix the whitespace too.. I just don't have any clue on how to squash these commits. Is that even possible with this online GitHub-interface?

jaimeperez:
Great! I'm not sure you can squash them through the web interface yourself, as I think that's only for those with rights to merge the PR, but you can always do that [in the command line](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/).

tvdijen:
That's, I'll check that out at a later time and then add a commit to fix whitespace too!

tvdijen:
I think I screwed up..

modules/adfs/lib/SAML2/XML/fed/Endpoint.php
```diff
@@ -4,28 +4,29 @@
...
+{
 	/**
 	 * Add this endpoint to an XML element.
 	 *
 	 * @param DOMElement $parent  The element we should append this endpoint to.
 	 * @param string $name  The name of the element we should create.
 	 */
-	public static function appendXML(DOMElement $parent, $name, $address) {
+	public static function appendXML(DOMElement $parent, $name, $address)
+    {
```
jaimeperez: Looks like there's an extra space before the `{`?

modules/adfs/lib/SAML2/XML/fed/Endpoint.php
```diff
@@ -4,28 +4,29 @@
...
 		assert('is_string($name)');
 		assert('is_string($address)');
 
 		$e = $parent->ownerDocument->createElement($name);
-                $parent->appendChild($e);
+        $parent->appendChild($e);
```
jaimeperez: Same here and in the next change, looks like something is odd with indentation. Maybe the rest of the code is indented wrong?

modules/adfs/www/idp/prp.php
```diff
@@ -23,6 +23,6 @@
 	// logout response from ADFS SP
 	$assocId = $_GET['assocId']; // Association ID of the SP that sent the logout response
 	$relayState = $_GET['relayState']; // Data that was sent in the logout request to the SP. Can be null
-	$logoutError = NULL; /* NULL on success, or an instance of a SimpleSAML_Error_Exception on failure. */
+	$logoutError = null; /* NULL on success, or an instance of a SimpleSAML_Error_Exception on failure. */
```
jaimeperez: I would update the comment too (s/NULL/null/). In general, I also prefer single-line comment markers (`//`) for single-line comments, and multi-line comment markers (`/* */`) for multi-line comments, but that's more personal preference and you don't have to do that if you don't want to 😄

tvdijen:
Anything else @jaimeperez ?

jaimeperez:
Thanks a lot Tim! Looks good to me 👍 

There's only one thing I've just realized. Since some of the methods are changing their names and they are declared public, refactoring them could break other code. Have you checked that those functions are not used elsewhere? The simplest solution would be to add again the old methods as wrappers of the renamed ones, and mark them as deprecated in the phpdoc block, but it might not be worth the trouble if they are so specific that nobody should use them.

tvdijen:
I did a grep and the functions are only being used within the ADFS.php file.
Could probably make them private too..  I happen to have an ADFS machine available, so I can run a quick test if nothing breaks.. I'll get back to you on this.

jaimeperez:
Sounds great, thanks so much Tim!

thijskh:
Any update @tvdijen?

tvdijen:
No, unfortunately I haven't found the time to run the test..
I'll see if I can fit it in somewhere this week, since I will be abroad for the next two weeks..

thijskh:
Would be great, since I don't have a test environment for this stuff.

tvdijen:
I wonder if this module ever worked at all.. I can't even get a working setup without this PR :(

pgh70:
I am using the ADFS module from 1.14.3, first without modifications, now with modifications from PR #469, so it did work!

tvdijen:
That's great news @pgh70! Since you have a working environment, would you be willing to perform another test after I convert some functions from public to private?

thijskh:
@pgh70 Can you test Tim's latest changes? Thanks!

tvdijen:
This works like a charm!

tvdijen:
This works like a charm!

PR #490, opened 2016-10-12
Fix a cosmetic exception when using loginuserpassorg.php

ghalse:
When using ldap:LDAPMulti, SimpleSAMLphp generates an exception:

 SimpleSAML_Error_Exception: Error 8 - Undefined index: rememberMeEnabled
 Backtrace:
 4 /srv/simplesamlphp-1.14.8/www/_include.php:83 (SimpleSAML_error_handler)
 3 /srv/simplesamlphp-1.14.8/modules/core/templates/loginuserpass.php:107 (require)
 2 /srv/simplesamlphp-1.14.8/lib/SimpleSAML/XHTML/Template.php:577 (SimpleSAML_XHTML_Template::show)
 1 /srv/simplesamlphp-1.14.8/modules/core/www/loginuserpassorg.php:95 (require)
 0 /srv/simplesamlphp-1.14.8/www/module.php:137 (N/A)

This is because the loginuserpass.php template (which is used by both
loginuserpass.php and loginuserpassorg.php) refers to
$this->data['rememberMeEnabled'], but that variable is only defined when
you make use of loginuserpass.php.

Since these don't appear to be relevant to loginuserpassorg.php, this
patch simply sets these to false (the default value), so that they're
no longer undefined.


jaimeperez:
Thanks a lot Guy!

Issue #419, opened 2016-07-15
Rationalize URL-guessing methods in SimpleSAML\Utils\HTTP


jaimeperez:
We have a bunch of methods in `SimpleSAML\Utils\HTTP` handling URLs, mostly the current URL. Most of them were buggy because they were trying to use whatever was configured in the `baseurlpath` configuration option, but completely ignoring the path in there, leading to broken URLs. See #396

Additionally, the URL guessing logic was also broken in different setups, mostly when run with _nginx_ #391, #386 and #5.

However, the fixes for these issues led to some other issues with Windows systems or with setups where web server aliases were used together with symbolic links in the file system. All of that was fixed, and led to yet another issue, reported in #418.

The problem is that we have a set of methods that are not fully coherent. They try to use whatever is specified in `baseurlpath`, but not too hard. In general, we should have:
- [ ] Some method to guess the current URL (and the associated methods to remove the query and other parts). This method should **NOT** take the `baseurlpath` configuration option into account, so that it can be used by scripts not belonging to SimpleSAMLphp.
- [ ] Some method to obtain the URL of a script belonging to SimpleSAMLphp, this time, of course honoring `baseurlpath`.
- [ ] Pages should always try to build URLs based on the path to the script, relative to SimpleSAMLphp's installation. This way, we can just get the `baseurlpath` and append the relative path. A good example of this is `SimpleSAML\Module::getModuleURL()`.

This is still a bit challenging, since sometimes we will **really** want to get our current URL to generate a link, but we don't know what's the script in use. An example of this is the language bar, generating links to change the language of the current page. The links in this bar are generated by a common template file, included from every page, and therefore, you never know what's the main script. In cases like this, we can make use of the logic that we have currently, which tries to see if we are running a SimpleSAMLphp script, and then builds the URL mixing the `baseurlpath` together with the path to the script, relative to SimpleSAMLphp's base directory (`/whatever/path/to/ssp/www/`).

We need to rationalize all of this, probably create new functions, rename some existing ones to make their names more intuitive and clear, and review **all the code generating links**.

calebdelnayatgfs:
I'll share some thoughts on this related to reverse proxies:

[...]

Issue #343, opened 2016-03-07
Fail to detect HTTPS when application is behind an Amazon AWS Elastic Load Balancer 


zmilan:
Hi,

Thanks for this great tool. It works very good, but I found issue with detecting HTTPS when application is behind an Amazon AWS Elastic Load Balancer. Issue is because in that case we don't have $_SERVER['HTTPS'] set, but good thing is that we have other that can be used:
$_SERVER['HTTP_X_FORWARDED_PORT'] => 443
$_SERVER['[HTTP_X_FORWARDED_PROTO'] => https

For our need I add fix, but it will be good if this can be solved on project level. Here is my fix:

[SimpleSAML\Utils\HTTP::getServerHTTPS() function](https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Utils/HTTP.php#L76) I modify it to looks like this:
`
// add support for servers on aws behind load balancer
**if(isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
            $_SERVER['HTTPS']='on';
            return true;
}**

if (!array_key_exists('HTTPS', $_SERVER)) {
            // not an https-request
            return false;
}

if ($_SERVER['HTTPS'] === 'off') {
            // IIS with HTTPS off
            return false;
}
        
// otherwise, HTTPS will be a non-empty string
return $_SERVER['HTTPS'] !== '';
`

[SimpleSAML\Utils\HTTP::getServerPort() function](https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Utils/HTTP.php#L101) I modify it to looks like this:
`
$port = (isset($_SERVER['SERVER_PORT'])) ? $_SERVER['SERVER_PORT'] : '80';

if (self::getServerHTTPS()) {
     // modified first part of this case to support aws servers under loadbalancer
     if (**$port !== '80' &&**$port !== '443') {
          return ':'.$port;
     }
} else {
      if ($port !== '80') {
          return ':'.$port;
      }
}

return '';
`

Thanks again

jaimeperez:
Hi @zmilan!

[...]

PR #491, opened 2016-10-12
Added support for load balancer-terminated SSL.

danepowell:
Currently, if SSL is terminated at a load balancer in front of your server running SimpleSAMLphp, the library thinks the base url is http even if you visit it via https, which is non-intuitive to say the least.

The current workaround is to manually specify the base path url (including https) in these environments, but it would be much more user-friendly, flexible, and harmless to handle this automatically.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -75,6 +75,10 @@ class HTTP
      */
     private static function getServerHTTPS()
     {
+        if (array_key_exists('HTTP_X_FORWARDED_PROTO', $_SERVER) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
+            // Server is behind load balancer.
+            return true;
+        }
         if (!array_key_exists('HTTPS', $_SERVER)) {
             // not an https-request
             return false;
@@ -101,6 +105,7 @@ class HTTP
     private static function getServerPort()
     {
         $port = (isset($_SERVER['SERVER_PORT'])) ? $_SERVER['SERVER_PORT'] : '80';
+        $port = (isset($_SERVER['HTTP_X_FORWARDED_PORT'])) ? $_SERVER['HTTP_X_FORWARDED_PORT'] : $port;
         if (self::getServerHTTPS()) {
             if ($port !== '443') {
                 return ':'.$port;


tvdijen:
Most loadbalancers don't set this HTTP-header, so I doubt this is very useful to be honest.

jaimeperez:
Hi @danepowell!

Using `baseurlpath` is not a workaround, but the proper way to tell SimpleSAMLphp what's its canonical base URL. The `X-Forwarded-*` headers are broken and cannot be trusted, so they shouldn't be used, as discussed in #77, #376, #419, #434, #324 and #343.

Thanks so much in any case for your contribution!

danepowell:
I get that the X-Forwarded-For headers can be spoofed, but is there actually an attack vector here (if we're just using the protocol and port)? Worst case scenario, wouldn't that simply break authentication for the malicious user?

I get that this might only benefit a fraction of users, but judging by the number of times this has been requested, it's a non-trivial number. SimpleSAMLphp already requires such a huge degree of configuration, I think anything we can do to make it "just work" out of the box would improve the developer experience and increase adoption.

jaimeperez:
There are attacks that could be performed exploiting the nature of these headers. Imagine for instance an attacker setting up a MITM attack, attaching the `X-Forwarded-Proto` header to every request, setting it to `http`. Even if SimpleSAMLphp is configured to use `https` unconditionally, if it is available via `http` for convenience, then an attacker could leverage this to force victims to run their entire sessions via `http`, exposing credentials and session identifiers.

I'm pretty sure we could find other attack vectors if we try.

If we add support for `X-Forwarded-*`, that means we'll have to trust the environment (including user-provided data) and give unvalidated data priority over configuration. That's because the `baseurlpath` option still needs to be there to cover all those cases of different domains/ports between the frontend and the backend, URL rewriting, etc.

Having two different ways to tell SimpleSAMLphp how to obtain its canonical base URL is a source of bugs in the code and confusion for the users, so I'd rather avoid that. Furthermore, there's even no agreement on the way to go, with some people suggesting the use of `X-Forwarded-*` and some others suggesting a new configuration option to enforce _https_.

The only way that support for `X-Forwarded-*` could be implemented securely implies having more configuration options to define which headers and values can be trusted. That would make no sense at all if the problem is already resolved with one existing option. Furthermore, it wouldn't _just work_, and if we trust the headers by default so that it _just works_, then you are allowing insecure deployments by default.

So again, I would really like to thank you for your contribution as we always appreciate a lot people taking their time to help, but I'm afraid I don't see a way to include this in SimpleSAMLphp. We could probably enhance the documentation on the other hand, although I think the comments in the configuration template are quite explicit:

```php
    /*
     * Setup the following parameters to match the directory of your installation.
     * See the user manual for more details.
     *
     * Valid format for 'baseurlpath' is:
     * [(http|https)://(hostname|fqdn)[:port]]/[path/to/simplesaml/]
     *
     * The full url format is useful if your SimpleSAMLphp setup is hosted behind
     * a reverse proxy. In that case you can specify the external url here.
     *
     * Please note that SimpleSAMLphp will then redirect all queries to the
     * external url, no matter where you come from (direct access or via the
     * reverse proxy).
     */
    'baseurlpath' => 'simplesaml/',
```

PR #495, opened 2016-10-17
Allow for toggling Kerberos replay cache

tvdijen:
Replay cache is known to cause performance issues ([reference](http://web.mit.edu/KERBEROS/krb5-devel/doc/basic/rcache_def.html#performance-issues)), so it may be desirable to disable it.

Personally I ran into a (known) issue where replay cache-files were not freed/removed, causing Apache to stop working after a while (too many files open). This issue has been fixed in krb5-libs 1.13, but version 1.10 is still the default on many operating systems (RHEL 6).

### DIFF
--- a/modules/negotiate/lib/Auth/Source/Negotiate.php
+++ b/modules/negotiate/lib/Auth/Source/Negotiate.php
@@ -28,6 +28,7 @@ class sspmod_negotiate_Auth_Source_Negotiate extends SimpleSAML_Auth_Source
     protected $admin_user = null;
     protected $admin_pw = null;
     protected $attributes = null;
+    protected $rcache = null;
 
 
     /**
@@ -66,6 +67,7 @@ class sspmod_negotiate_Auth_Source_Negotiate extends SimpleSAML_Auth_Source
         $this->admin_user = $config->getString('adminUser', null);
         $this->admin_pw = $config->getString('adminPassword', null);
         $this->attributes = $config->getArray('attributes', null);
+        $this->rcache = $config->getBoolean('rcache', null);
     }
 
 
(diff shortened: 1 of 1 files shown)


modules/negotiate/lib/Auth/Source/Negotiate.php
```diff
@@ -66,6 +67,7 @@ public function __construct($info, $config)
         $this->admin_user = $config->getString('adminUser', null);
         $this->admin_pw = $config->getString('adminPassword', null);
         $this->attributes = $config->getArray('attributes', null);
+        $this->rcache = $config->getBoolean('rcache', null);
```
jaimeperez: I find the name of the config option a bit obscure, but that could also be due to my lack of knowledge on this 😓 

In any case, we should probably mention this new option (and its rationale) in the documentation.

modules/negotiate/lib/Auth/Source/Negotiate.php
```diff
@@ -136,6 +138,12 @@ public function authenticate(&$state)
                     SimpleSAML\Logger::debug('Negotiate - authenticate(): No "Negotiate" found. Skipping.');
                 }
             }
+            
+            if ($this->rcache === true) {
+                putenv("KRB5RCACHETYPE=dfl");
+            } elseif ($this->rcache === false) {
```
jaimeperez: Would it make more sense to make this a simple `else` without any condition? What's the default in case the config option is not is not set?

jaimeperez:
Thanks Tim!

This issue led me to think about extracting this module from the core distribution and have it in its own repository. We could definitely do that to try to leverage other's knowledge on Kerberos so that it is properly maintained.

modules/negotiate/lib/Auth/Source/Negotiate.php
tvdijen: The default would in this case be the system default...
If you don't set the option, the system default applies. If you do set it, your setting should override the system default, so that's why I did it this way.
I must admit I have not enough knowledge about Kerberos.. I have no clue about other possible values.

modules/negotiate/lib/Auth/Source/Negotiate.php
tvdijen: This is how Kerberos calls it... It stands for replayCache

modules/negotiate/lib/Auth/Source/Negotiate.php
jaimeperez: In that case, I think it makes sense to keep it like this then.

I would only add some documentation explaining the option, why is it necessary, what are its values, and what's the default. Other than that, 👍

tvdijen:
This issue has been properly fixed in the php-krb5 module now in version 1.1.1;  https://pecl.php.net/package/krb5

PR #497, opened 2016-10-17
Fixed incorrect memcache_store prefix docs.

danepowell:
The search and replace in https://github.com/simplesamlphp/simplesamlphp/commit/f25c7815af52a751cb815709ff33abee14e130a4 was a bit overzealous and should not have changed the documentation for this variable, since the default value is still (lowercase) simpleSAMLphp.

See the actual default value here:
https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Store/Memcache.php#L26

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -592,7 +592,7 @@ $config = array(
 
     /*
      * This value allows you to set a prefix for memcache-keys. The default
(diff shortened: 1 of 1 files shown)


jaimeperez:
Thanks a lot @danepowell!

PR #500, opened 2016-10-27
Fix required attributes duplication

andrius-kulbis:
While parsing the XML metadata file of SP, attributes in AttributeConsumerService marked as required where added multiple times causing duplication of attribute keys in `attributes.required` section.

### DIFF
--- a/lib/SimpleSAML/Metadata/SAMLParser.php
+++ b/lib/SimpleSAML/Metadata/SAMLParser.php
@@ -1217,10 +1217,6 @@ class SimpleSAML_Metadata_SAMLParser
             $attrname = $child->Name;
             $sp['attributes'][] = $attrname;
 
-            if ($child->isRequired) {
-                $sp['attributes.required'][] = $attrname;
-            }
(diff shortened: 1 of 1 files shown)


thijskh:
Good catch, thanks. But it would be a great opportunity to add a unit test for this specific functionality. Could you do that?

jaimeperez:
That's absolute true. Thanks a lot @andrius-kulbis!

jaimeperez:
Oops, sorry Thijs, your comment didn't pop up until I hit the "merge" button 😞

thijskh:
No worries, I solved it by creating the test ;-)
I couldn't push directly to master, has this been disabled? Had to create this PR: #501.

jaimeperez:
That's weird. I see something has changed regarding groups in github, and you were not included in the "owners" group (which is no longer a special group, but anyway).

Could you try again and see if you can push to master?

thijskh:
I've now already created the PR.. I'll check it soon.

PR #509, opened 2016-11-09
Add ability to define additional attributes on ContactPerson element

tdiscuit:
This will allow you to configure an attributes array on contact definitions for trust frameworks such as SIRTFI. This is a continuation from the simplesamlphp/saml2 pull [#79](https://github.com/simplesamlphp/saml2/pull/79).

I thought about having a SIRTFI class that could be included onto the contact person element, but I couldn't think of a clean way of adding it on, but I am open to discussion about that. There are also some todo's on the addContact function that I decided was out of scope for this PR that would have to be addressed if I went down the route of having a separate class for SIRTFI.

Also, saml2 doesn't have a new tag with the updated PR that I submitted there, so I've simply updated the commit hash in the composer config. That can be fixed whenever a new tag is created with the changes that were made to support this in saml2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
 		"ext-date": "*",
 		"ext-hash": "*",
 		"ext-json": "*",
-        "simplesamlphp/saml2": "dev-master#00e38f85b417be1e10a2d738dd2f5ea82edb472c as 2.2",
+        "simplesamlphp/saml2": "dev-master#a94403bfe5627c90fe3764e0ada5a44841a11e80 as 2.3.3",
         "robrichards/xmlseclibs": "~2.0",
         "whitehat101/apr1-md5": "~1.0",
         "twig/twig": "~1.0",
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
 		"ext-date": "*",
 		"ext-hash": "*",
 		"ext-json": "*",
-        "simplesamlphp/saml2": "dev-master#a94403bfe5627c90fe3764e0ada5a44841a11e80 as 2.3.3",
+        "simplesamlphp/saml2": "dev-master#00e38f85b417be1e10a2d738dd2f5ea82edb472c as 2.2",
         "robrichards/xmlseclibs": "~2.0",
         "whitehat101/apr1-md5": "~1.0",
         "twig/twig": "~1.0",
--- a/docs/simplesamlphp-reference-idp-hosted.md
+++ b/docs/simplesamlphp-reference-idp-hosted.md
@@ -143,14 +143,14 @@ Common options
 		'contacts' => array(
 		    array(
 		        'contactType'       => 'other',
-		        'emailAddress'      => 'abuse@example.org',
+		        'emailAddress'      => 'mailto:abuse@example.org',
 		        'givenName'         => 'John',
 		        'surName'           => 'Doe',
 		        'telephoneNumber'   => '+31(0)12345678',
 		        'company'           => 'Example Inc.',
 		        'attributes'        => array(
 		            'xmlns:remd'        => 'http://refeds.org/metadata',
-		            'remd:contactType   => 'http://refeds.org/metadata/contactType/security',
+		            'remd:contactType'  => 'http://refeds.org/metadata/contactType/security',
 		        ),
 		    ),
 		), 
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -35,7 +35,7 @@
 		"ext-date": "*",
 		"ext-hash": "*",
 		"ext-json": "*",
-        "simplesamlphp/saml2": "dev-master#00e38f85b417be1e10a2d738dd2f5ea82edb472c as 2.2",
+        "simplesamlphp/saml2": "dev-master#a94403bfe5627c90fe3764e0ada5a44841a11e80 as 2.3.3",
```
thijskh: @jaimeperez Can we 'safely' do this in master w.r.t. the changes to epti-handling?

thijskh:
We can merge this when it's safe to bump the saml2 dependency to after the latest EPTI-changes.

jaimeperez:
Yes, if we want to update the dependency on the SAML2 library, we need to make sure SimpleSAMLphp uses the `SAML2\XML\saml\NameID` objects for _eduPersonTargetedID_. That way we can make sure it will generate correct values and will work properly with updated versions of the SAML2 lib, where we will remove support for `string` and `DOMNodeList` values.

docs/simplesamlphp-reference-idp-hosted.md
```diff
@@ -123,6 +123,37 @@ Common options
...
+		    array(
+		        'contactType'       => 'other',
+		        'emailAddress'      => 'abuse@example.org',
+		        'givenName'         => 'John',
+		        'surName'           => 'Doe',
+		        'telephoneNumber'   => '+31(0)12345678',
+		        'company'           => 'Example Inc.',
+		        'attributes'        => array(
+		            'xmlns:remd'        => 'http://refeds.org/metadata',
+		            'remd:contactType   => 'http://refeds.org/metadata/contactType/security',
```
NicolasLiampotis: remd:contactType is missing closing quote

docs/simplesamlphp-reference-idp-hosted.md
```diff
@@ -123,6 +123,37 @@ Common options
...
+		        'company'           => 'Example Inc.',
+		    ),
+		),
+
+:	If you have support for a trust framework that requires extra attributes on the contact person element in your IdP metadata (for example, SIRTFI), you can specify an array of attributes on a contact.
+
+		'contacts' => array(
+		    array(
+		        'contactType'       => 'other',
+		        'emailAddress'      => 'abuse@example.org',
```
NicolasLiampotis: Perhaps it would be better to add the `mailto:` scheme in the example email address to be consistent with the SIRTFI guide: https://wiki.refeds.org/display/SIRTFI/Guide+for+Federation+Operators

lib/SimpleSAML/Utils/Config/Metadata.php
```diff
@@ -108,6 +115,13 @@ function ($t) {
             throw new \InvalidArgumentException('"contactType" is mandatory and must be one of '.$types.".");
         }
 
+        // check attributes is an associative array
+        if (isset($contact['attributes'])) {
+            if (empty($contact['attributes']) || !is_array($contact['attributes']) || count(array_filter(array_keys($contact['attributes']), 'is_string')) == 0) {
```
jaimeperez: As a rule of thumb, strict comparisons (`===`) are always preferred to keep us in the safe side in case of an unexpected return value. In this case it should be ok, but the strict comparison could protect us from future changes in the signature of `count()`.

Also, I suspect the line is a bit too long. Can you split it to make it a bit more readable? I.e.:

```php
if (empty($contact['attributes']) || !is_array($contact['attributes']) || 
    count(array_filter(array_keys($contact['attributes']), 'is_string')) == 0
) {
```

jaimeperez:
Hi Tyler!

I've just committed an update of the SAML2 lib dependency so that simplesamlphp/saml2#79 is now available in SSP. Could you please update this PR to remove the update to `composer.json` and address the comments made by @NicolasLiampotis?

Thanks and sorry for the delay taking care of this! 😅

tdiscuit:
Hey @jaimeperez 

I think I addressed all the concerns brought up by @NicolasLiampotis. Let me know if I missed anything.

jaimeperez:
Merged, thanks a lot @tdiscuit!

PR #530, opened 2016-12-07
Deal with cases where there is no SubjectConfirmationData.

jeremyfrench:
I had a case where a saml packet contained this 

` 
<SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer" />
`

This caused SSP to fail. As subject confirmation data is not defined.

I have looked at the spec (well the xml schema) and it seems to be valid. 

`
<element ref="saml:SubjectConfirmationData" minOccurs="0"/>
`

I'll agree that it doesn't make any sense to have such a element but if it is not invalid we should not be having a fatal error.

### DIFF
--- a/modules/saml/lib/Message.php
+++ b/modules/saml/lib/Message.php
@@ -667,6 +667,11 @@ class sspmod_saml_Message {
 				}
 			}
 
+			// If no SubjectConfirmationData then don't do anything.
+			if (!$scd) {
+				continue;
+			}
+
 			if ($scd->NotBefore && $scd->NotBefore > time() + 60) {
 				$lastError = 'NotBefore in SubjectConfirmationData is in the future: ' . $scd->NotBefore;
 				continue;
--- a/modules/saml/lib/Message.php
+++ b/modules/saml/lib/Message.php
@@ -669,6 +669,7 @@ class sspmod_saml_Message {
 
 			// If no SubjectConfirmationData then don't do anything.
 			if (!$scd) {
+				$lastError = 'No SubjectConfirmationData provided';
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


longwave:
The SAML errata documentation at http://docs.oasis-open.org/security/saml/v2.0/errata05/os/saml-v2.0-errata05-os.pdf seems to clarify this:

> At least one bearer &lt;SubjectConfirmation> element MUST contain a &lt;SubjectConfirmationData> element that itself MUST contain a Recipient attribute containing the service provider's assertion consumer service URL and a NotOnOrAfter attribute that limits the window during which the assertion can be confirmed by the relying party.

> Note that while multiple bearer &lt;SubjectConfirmation> elements may be present, the successful evaluation of a single such element in accordance with this profile is sufficient to confirm an assertion. However, each assertion, if more than one is present, MUST be evaluated independently.

To me, it looks like the proposed fix covers this as there still must be a valid &ltSubjectConfirmation> and they are still evaluated, this PR simply skips the case where an empty one is encountered.

jaimeperez:
Hi guys!

@longwave is right. The fix is needed not only to avoid failures when an IdP sends an empty `SubjectConfirmation` alongside a valid one (which is a really stupid behaviour, but spec-compliant anyway), but also to ensure SSP behaves correctly when run under PHP 7. Otherwise, in PHP 7 it will log the errors and continue, effectively validating the assertion (even when it shouldn't).

I've committed fixes for this both in master and in the 1.14 branch (with a slight difference between them and your PR, that being explicit checking of the `SubjectConfirmationData` being null).

Thanks for your help!

PR #531, opened 2016-12-12
Add attributes.noconsent option to correspond with the Consent module

ghalse:
The Consent module has a (currently undocumented) noconsentattributes option that allows specified attributes to be removed from the consent hash calculation. The noconsentattribute option was introduced at simplesamlphp/simplesamlphp@1efcfa8 but no corresponding option was added to the consentAdmin module. Thus if people are using the noconsentattribute option to remove attributes, the consentAdmin module will *always* show this as attribute values having been changed.

### DIFF
--- a/modules/consent/docs/consent.md
+++ b/modules/consent/docs/consent.md
@@ -160,6 +160,12 @@ The following options can be used when configuring the Consent module
     the attributes that should have it value hidden. Default behaviour is that 
     all attribute values are shown
 
+`noconsentattributes`
+:   Allows certain attributes to be excluded from the attribute hash when
+    `includeValues` is TRUE (and as a side effect, to be hidden from display
+	as `hiddenAttributes` does). Set to an array of the attributes that should
+	be excluded. Default behaviour is to include all values in the hash.
+
 `showNoConsentAboutService`
 :   Whether we will show a link to more information about the service from the
     no consent page. Defaults to `TRUE`.
--- a/modules/consentAdmin/config-templates/module_consentAdmin.php
+++ b/modules/consentAdmin/config-templates/module_consentAdmin.php
@@ -6,23 +6,23 @@
  * @package SimpleSAMLphp
  */
 $config = array(
-	/*
-	 * Configuration for the database connection.
-	 */
-	'consentadmin'  => array(
-		'consent:Database',
-		'dsn'		=>	'mysql:host=DBHOST;dbname=DBNAME',
-		'username'	=>	'USERNAME', 
-		'password'	=>	'PASSWORD',
-	),
-	
-	// Hash attributes including values or not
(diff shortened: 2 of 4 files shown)


modules/consent/docs/consent.md
```diff
@@ -160,6 +160,12 @@ The following options can be used when configuring the Consent module
     the attributes that should have it value hidden. Default behaviour is that 
     all attribute values are shown
 
+`noconsentattributes`
+:   Allows certain attributes to be excluded from the attribute hash when
+    `includeValues` is TRUE (and as a side effect, to be hidden from display
```
jaimeperez: s/TRUE/\`true\`/

modules/consentAdmin/config-templates/module_consentAdmin.php
```diff
@@ -6,23 +6,26 @@
...
+        'consent:Database',
+        'dsn'       =>  'mysql:host=DBHOST;dbname=DBNAME',
+        'username'  =>  'USERNAME',
+        'password'  =>  'PASSWORD',
+    ),
 
-	// Where to direct the user after logout
-    // REMEMBER to prefix with http:// otherwise the relaystate is only appended 
+    // Hash attributes including values or not
+    'attributes.hash' => TRUE,
```
jaimeperez: The PSR-2 standard we decided to use enforces the user of lowercase for PHP constants, like `true`, `false` or `null`.

modules/consentAdmin/www/consentAdmin.php
```diff
@@ -49,6 +50,13 @@ function driveProcessingChain(
 
     $attributes = $authProcState['Attributes'];
 
+    // Remove attributes that do not require consent
+    foreach ($attributes AS $attrkey => $attrval) {
```
jaimeperez: Same here as the previous comment: s/AS/as/

modules/consent/docs/consent.md
```diff
@@ -160,6 +160,12 @@ The following options can be used when configuring the Consent module
...
+	as `hiddenAttributes` does). Set to an array of the attributes that should
+	be excluded. Default behaviour is to include all values in the hash.
+
 `showNoConsentAboutService`
 :   Whether we will show a link to more information about the service from the
     no consent page. Defaults to `TRUE`.
```
jaimeperez: s/\`TRUE\`/\`true\`/

ghalse:
I was rebasing it against master and messed my rebase up somewhere along the way. I haven't had a chance to figure out why yet. I'll get back to it :)

thijskh:
I’ve made an attempt but it’s rather complicated by all of the psr-2 changes that touch every line. Perhaps easiest to branch anew from master and reapply the desired changes?

ghalse:
That's the conclusion I came to too :)

PR #533, opened 2016-12-15
Allow the language cookie to be secured

ghalse:
There are config options that allow the session cookie to be secured, but there are no equivalent options for the language cookie. As a result it is not possible to secure the language cookie.

Whilst it might be desirable to have a cookie containing language preference more widely accessible, some people may wish to lock down the cookie -- particularly where they are implementing policies like HSTS. Having an insecure cookie triggers a flag in audits of HSTS.

This change adds two additional config options allowing the language cookie to be secured. The default of serving an insecure cookie remains unchanged, meaning the change should not affect existing installs.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -656,6 +656,8 @@ $config = array(
     'language.cookie.name' => 'language',
     'language.cookie.domain' => null,
     'language.cookie.path' => '/',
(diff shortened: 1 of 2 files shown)


jaimeperez:
Merged! Thanks a lot Guy!

PR #537, opened 2016-12-21
bugfix: Discopower ignored filtering of hide.from.discovery metadata property

ondrejvelisek:
Not sure on which branch I should create pullrequest...


jaimeperez:
Thanks @ondrejvelisek!

PR #539, opened 2016-12-23
bugfix: wrapped the building of authsource config with issets.

doedje:
The change in `AttributeAddFromLDAP.php` was needed to discover the bug in the first place. Without it there was never an error in the log while clearly an exception was thrown somewhere during `$this->getLdap()`.

The change in `BaseFilter.php` makes sure that undefined configuration parameters in `$authsource` are not set to `null` in `$authconfig`, but rather stay undefined.

Not doing so did result in the following exeception because `ldap.port` was indeed undefined in my `authsources.php` but was set to `null` because of the code in `BaseFilter.php`. The error never ended up in the log and I was presuming the value to be set to it's default, as it does when using the authsource as a direct authsource. Confusion all along.... =]

```
Dec 23 08:28:10 simplesamlphp WARNING [94b0f44d76] AttributeAddFromLDAP: exception = exception 'Exception' with message 'ldap:AuthProcess: The option 'ldap.port' is not a valid integer value.' in /Users/remy/git/saml-IdP/lib/SimpleSAML/Configuration.php:737
Stack trace:
#0 /Users/remy/git/saml-IdP/modules/ldap/lib/Auth/Process/BaseFilter.php(267): SimpleSAML_Configuration->getInteger('ldap.port', 389)
#1 /Users/remy/git/saml-IdP/modules/ldap/lib/Auth/Process/AttributeAddFromLDAP.php(172): sspmod_ldap_Auth_Process_BaseFilter->getLdap()
....
```


thijskh:
Thanks Remy. If you are using the isset tests, can we then leave off those ugly `@` modifiers that tried to silence any notices?

doedje:
@thijskh I guess so, my php skills are a bit rusty from old days. Didn't bother to google what those `@` do anyway. I removed some, my test still works. Will remove all of them and push them in this PR.

doedje:
No more ugly `@` in the code... =]

thijskh:
Great. I'll review it in more detail next week... (unless someone beats me to it)

doedje:
Fijne kerst! (dutch for 'merry christmas')

doedje:
any change you looked into this issue?

doedje:
accidentally pushed another feature against my master..... will set this straight right away..

modules/ldap/lib/Auth/Process/BaseFilter.php
```diff
@@ -137,19 +141,38 @@ public function __construct(&$config, $reserved)
...
+            if (isset($authsource['port'])) {
+              $authconfig['ldap.port']       = $authsource['port'];
+            }
+            if (isset($authsource['timeout'])) {
+              $authconfig['ldap.timeout']    = $authsource['timeout'];
+            }
+            if (isset($authsource['debug'])) {
+              $authconfig['ldap.debug']      = $authsource['debug'];
+            }
+            if (isset($authsource['search.base']) && $authsource['search.enable']) {
```
thijskh: This will now generate a notice when search.enable is not set. I think that is not desired...

modules/ldap/lib/Auth/Process/BaseFilter.php
```diff
@@ -137,19 +141,38 @@ public function __construct(&$config, $reserved)
...
-            $authconfig['ldap.password']   = (@$authsource['priv.read'] ? @$authsource['priv.password'] : $authconfig['ldap.password']);
-
+            if (isset($authsource['hostname'])) {
+              $authconfig['ldap.hostname']   = $authsource['hostname'];
+            }
+            if (isset($authsource['enable_tls'])) {
+              $authconfig['ldap.enable_tls'] = $authsource['enable_tls'];
+            }
+            if (isset($authsource['port'])) {
+              $authconfig['ldap.port']       = $authsource['port'];
```
thijskh: Spacing does not seem to follow the coding guidelines (4 spaces for indent)

modules/ldap/lib/Auth/Process/BaseFilter.php
thijskh: Same for checks below.

modules/ldap/lib/Auth/Process/BaseFilter.php
```diff
@@ -137,19 +141,38 @@ public function __construct(&$config, $reserved)
...
+            }
+            if (isset($authsource['priv.username']) && $authsource['priv.read']) {
+              $authconfig['ldap.username']   = $authsource['priv.username'];
+            }
+            if (isset($authsource['priv.password']) && $authsource['priv.read']) {
+              $authconfig['ldap.password']   = $authsource['priv.password'];
+            }
             // Only set the username attribute if the authsource specifies one attribute
-            if (@$authsource['search.enable'] && is_array(@$authsource['search.attributes'])
+            if ($authsource['search.enable'] && is_array($authsource['search.attributes'])
```
thijskh: I think we're not sure these two will be set at this point?

modules/ldap/lib/Auth/Process/BaseFilter.php
doedje: sorry 'bout that.... fixed in upcoming push.

modules/ldap/lib/Auth/Process/BaseFilter.php
doedje: I'll change the code on these...

modules/ldap/lib/Auth/Process/BaseFilter.php
doedje: true... both need a check to avoid a notice...

doedje:
fixed indent and restructured code to avoid notices.

jaimeperez:
@thijskh, you've been following this. Do you think the latest changes by @doedje are fine so that we can merge this?

modules/ldap/lib/Auth/Process/AttributeAddFromLDAP.php
```diff
@@ -19,14 +19,17 @@
...
+ *          - Don't crash but fail silently on LDAP errors; the plugin is to complement attributes
+ * Updated: 20161223 Remy Blom <remy.blom@hku.nl>
+ *          - Adjusted the silent fail so it does show a warning in log when $this->getLdap() fails
  *
  * @author Yørn de Jong
  * @author Jaime Perez
  * @author Steve Moitozo
  * @author JAARS, Inc.
  * @author Ryan Panning
+ * @author Remy Blom <remy.blom@hku.nl>
```
thijskh: Would think we'd better get rid of these changelog and @author comments; that info is, in more accurate form, in vcs metadata...

thijskh:
Thanks @doedje!

thijskh:
I have taken the liberty to squash the commits, because the accidental changes+revert cause a lot of noise in the history otherwise. Might want to do that yourself next time.

doedje:
I totally agree with you, I will have to look into how I to do that... thanx!

jaimeperez:
Thanks a lot @doedje & @thijskh! 👏

PR #545, opened 2017-01-18
Bugfix in HTTP Class + Enhance of LinkedIn Auth class

aWuttig:


lib/SimpleSAML/Utils/HTTP.php
```diff
@@ -85,6 +85,10 @@ private static function getServerHTTPS()
             return false;
         }
 
+        if ($_SERVER['HTTPS'] === null) {
```
jaimeperez: What's exactly the bug here?

If `$_SERVER['HTTPS']` is null, there are two possible reasons for that:

* It is undefined, in which case, the previous code will get rid of it (`if (!array_key_exists('HTTPS', $_SERVER)) { return false; }`).
* It is explicitly defined to `null`, which is not something PHP is going to do itself, and therefore it must be someone, somewhere, doing it explicitly.

modules/authlinkedin/lib/Auth/Source/LinkedIn.php
```diff
@@ -125,4 +120,23 @@ public function finalStep(&$state) {
 
 		$state['Attributes'] = $attributes;
 	}
+
+    /**
+     * @param $array
```
jaimeperez: An actual description of what the method does, as well as the parameters and return values, would be helpful here.

modules/authlinkedin/lib/Auth/Source/LinkedIn.php
```diff
@@ -125,4 +120,23 @@ public function finalStep(&$state) {
...
+     * @param string $prefix
+     *
+     * @return array
+     */
+    protected function flatten($array, $prefix = '') {
+        $result = [];
```
jaimeperez: This is not valid syntax in PHP 5.3. It should be `$result = array();` instead.

modules/authlinkedin/lib/Auth/Source/LinkedIn.php
```diff
@@ -125,4 +120,23 @@ public function finalStep(&$state) {
...
+        foreach($array as $key => $value) {
+            if(is_array($value)) {
+                $result = $result + $this->flatten($value, $prefix . $key . '.');
+            }
+            else {
+                $result[$prefix . $key] = [$value];
```
jaimeperez: Same here:

```php
$result[$prefix.$key] = array($value);
```

modules/authlinkedin/lib/Auth/Source/LinkedIn.php
```diff
@@ -125,4 +120,23 @@ public function finalStep(&$state) {
...
+
+    /**
+     * @param $array
+     * @param string $prefix
+     *
+     * @return array
+     */
+    protected function flatten($array, $prefix = '') {
+        $result = [];
+        foreach($array as $key => $value) {
```
jaimeperez: According to the PSR-2 standard that we are following, there must be an space between the `foreach` keyword and the subsequent opening parenthesis.

modules/authlinkedin/lib/Auth/Source/LinkedIn.php
```diff
@@ -125,4 +120,23 @@ public function finalStep(&$state) {
...
+            if(is_array($value)) {
```
jaimeperez: Same here for the `if`:

```php
if (is_array($value)) {
```

modules/authlinkedin/lib/Auth/Source/LinkedIn.php
```diff
@@ -125,4 +120,23 @@ public function finalStep(&$state) {
...
+                $result = $result + $this->flatten($value, $prefix . $key . '.');
+            }
+            else {
```
jaimeperez: And in this case, the else should be surrounded by the braces, like this:

```php
} else {
```

lib/SimpleSAML/Utils/HTTP.php
aWuttig: > It is explicitly defined to null, which is not something PHP is going to do itself, and therefore it must be someone, somewhere, doing it explicitly.
We use it together with shopware which is based on Zend Framework and maybe the framework set the HTTPS to null. I don`t know why. For us the simplest solution is to add this condition, because the key "HTTPS" exists but the value is null which means --> no HTTPS is activated.

modules/authlinkedin/lib/Auth/Source/LinkedIn.php
aWuttig: ah sorry :/ we are working with PHP7 ;-). You can change it if you want.

modules/authlinkedin/lib/Auth/Source/LinkedIn.php
aWuttig: ok.

aWuttig:
In general we need this to authenticate users via linkedIn and get there location after this. The location is stored in an associative array by convention when getting this info from linkedIn response (see: https://developer.linkedin.com/docs/fields/location). 

The old method of putting the $userData into the $attributes array only iterates through the first level and this is not enough for this scenario.

Best Regards
André

lib/SimpleSAML/Utils/HTTP.php
jaimeperez: Is it possible for you to trace this somehow to see who and why is setting this variable to `null`? It doesn't make much sense to me...

In any case, this should be a separate pull request so that we can keep things clean, and the actual fix is to call `empty()` in the return statement, so that any value evaluated by PHP as empty makes the method return `false`, conforming to the PHP manual:

```php
return !empty($_SERVER['HTTPS']);
```

PR #551, opened 2017-01-24
feature: AttributeCopy can take array for an attribute

doedje:
Just like with AttributeMap you can specify a string or array as the destination:
```
50 => array(
   'class' => 'core:AttributeCopy',
   'sn' => 'urn:mace:dir:attribute-def:sn',
   'cn' => array('urn:mace:dir:attribute-def:cn','urn:mace:dir:attribute-def:diplayName')
),
```

### DIFF
--- a/modules/core/docs/authproc_attributecopy.md
+++ b/modules/core/docs/authproc_attributecopy.md
@@ -1,13 +1,13 @@
 `core:AttributeCopy`
 ===================
 
-Filter that renames attributes.
+Filter that copies attributes.
 
 
 Examples
 --------
 
-Copy a single attribute (user's uid will be copied to the user's username):
+Copy a single attribute (user's `uid` will be copied to the user's `username`):
 
     'authproc' => array(
         50 => array(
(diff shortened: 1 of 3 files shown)


doedje:
Checks have failed:
```
There was 1 failure:
1) Test_Core_Auth_Process_AttributeCopy::testWrongAttributeValue
Failed asserting that exception of type "Exception" is thrown.
```
Which is understandable:
```
    /**
     * Test wrong attribute value
     *
     * @expectedException Exception
     */
    public function testWrongAttributeValue()
    {
        $config = array(
            'test' => array('test2'),
        );
        $request = array(
            'Attributes' => array(
                'test' => array('value1'),
            ),
        );
        $result = self::processFilter($config, $request);
    }
```
This test expects to receive an exception when you put in an array, which is exactly what I wanted, being able to give an array with attributeNames.... I will look into how I can fix the test...

thijskh:
Hi Remy. Thanks for the PR. Tests are still failing, are you looking into this? You know you can run them locally (`vendor/phpunit/phpunit/phpunit --configuration tools/phpunit`)?

doedje:
Didn't know that! Thanx! That makes it a bit easier to run them! ;)

doedje:
checks pass! thanx for the tip of running the test locally!

pradtke:
@doedje I didn't see a test to confirm the new behavior works. Could you add one?
If you have the time could you also update the example in `simplesamlphp/modules/core/docs/authproc_attributecopy.md` with your new option.
Thanks!

doedje:
Added a test and an example as @pradtke requested. (sorry for the late response, I was quite busy with more pressing matters at work....)

jaimeperez:
Nice! Thanks a lot @doedje!

PR #552, opened 2017-01-24
Add "isDefault=true" to the AttributeConsumingService

gollmann:
Add "isDefault=true" to the AttributeConsumingService element of the generated SP metadata since some IdPs would not release attributes otherwise.

### DIFF
--- a/lib/SimpleSAML/Metadata/SAMLBuilder.php
+++ b/lib/SimpleSAML/Metadata/SAMLBuilder.php
@@ -404,6 +404,7 @@ class SimpleSAML_Metadata_SAMLBuilder
         $attributeconsumer = new \SAML2\XML\md\AttributeConsumingService();
 
         $attributeconsumer->index = 0;
+        $attributeconsumer->isDefault = true;
 
         $attributeconsumer->ServiceName = $name;
         $attributeconsumer->ServiceDescription = $metadata->getLocalizedString('description', array());


thijskh:
Hi Gollman. Thanks for the PR. I have two questions though. First, if I check the PR out in my install my generated SP metadata does not change. Are you sure it works? (I did not try to debug why it doesn't work for me yet.)

Secondly. I wonder if it's right. We would then be adding a isDefault to all our ACS-locations. While strictly allowed according to the spec, it sends a strange signal if you ask me. I would find it more logical to name one of the locations as "the default", not all of them.

BTW, the spec defines a clear algorithm for determining which ACS location to use, so I'm wondering why the IdP does not want to release attributes in this case. Is this not a bug that needs to be fixed on the IdP side?

gollmann:
> Am 30.01.2017 um 11:03 schrieb Thijs Kinkhorst <notifications@github.com>:
> 
> Hi Gollman. Thanks for the PR. I have two questions though. First, if I check the PR out in my install my generated SP metadata does not change. Are you sure it works? (I did not try to debug why it doesn't work for me yet.)

Works for me in 1.14.7 and 1.14.11

> Secondly. I wonder if it's right. We would then be adding a isDefault to all our ACS-locations. While strictly allowed according to the spec, it sends a strange signal if you ask me. I would find it more logical to name one of the locations as "the default", not all of them.

I am afraid I don´t get the argument: I see no provision in config/authsources.php to set more than one ACS and as I understand the code in simplesamlphp/lib/SimpleSAML/Metadata/SAMLBuilder.php it only creates a single ACS with the fixed index of zero.

Can you clue me in ?
Thanks !

Kind regards 
Georg Gollmann

PR #554, opened 2017-01-29
Raise exception when memcache-server is down

tvdijen:
The exception was never raised because $stats will be array('server' => false) on failure instead of just false.
This eventually led to $ret being NULL when one of the servers is down..

Because of this NULL-value, it also led to an empty memcacheMonitor screen.

### DIFF
--- a/lib/SimpleSAML/Memcache.php
+++ b/lib/SimpleSAML/Memcache.php
@@ -439,8 +439,10 @@ class SimpleSAML_Memcache
 
         foreach (self::getMemcacheServers() as $sg) {
             $stats = $sg->getExtendedStats();
-            if ($stats === false) {
-                throw new Exception('Failed to get memcache server status.');
+            foreach ($stats as $server => $data) {
+                if ($data === false) {
+                    throw new Exception('Failed to get memcache server status.');
+                }
             }
 
             $stats = SimpleSAML\Utils\Arrays::transpose($stats);


thijskh:
Maybe I'm missing something here, but it should be OK if one memcache server is down? As long as there's another one that is available? If we raise an exception when a single one is down, we cannot make use of redundancy.

tvdijen:
This change is limited to the getStats() function, that afaik is only being called from the memcacheMonitor-module..  Anyway, the way it is coded now, the exception is never raised and memcacheMonitor shows a white page..

thijskh:
I was looking at this https://github.com/tvdijen/simplesamlphp/blob/bb6ce9f7c537c406b66cc2230d6b29644979d4a3/lib/SimpleSAML/Memcache.php#L54 but on second thought it seems a different `getstats()` that is called here if I'm correct. Confusing...

thijskh:
Perhaps it would be helpful to rename our own stats methods to not have the same names as the Memcache extension methods.

tvdijen:
Agreed! And perhaps we shouldn't do array_merge, so memcacheMonitor can show the failing memcache-server as being down.. Right now, memcacheMonitor has no clue about the failing servers, since we already remove then from the stats

thijskh:
Alright, merged. Other improvements as discussed are welcome!

PR #556, opened 2017-01-31
Submit the logout response with the submit method. This fixes #555

### LINKED ISSUE #555: The LogoutResponse is not relayed to the IdP
The HTML form which should post the LogoutResponse to the IdP is not submitted when using the click event. This prevents the SLOService from completing the Logout process and the IdP complains that the Service Provider could not be logged out since it did not receive the LogoutResponse. This could be related to Issue #65.

rawroland:
Please do let me know if there are any conventions I missed.

### DIFF
--- a/templates/post.php
+++ b/templates/post.php
@@ -5,12 +5,12 @@
 	<meta http-equiv="content-type" content="text/html; charset=utf-8" />
 	<title>POST data</title>
 </head>
-<body onload="document.getElementsByTagName('input')[0].click();">
+<body onload="document.forms[0].submit();">
 
 	<noscript>
-		<p><strong>Note:</strong> Since your browser does not support JavaScript, you must press the button below once to proceed.</p> 
-	</noscript> 
-	
+		<p><strong>Note:</strong> Since your browser does not support JavaScript, you must press the button below once to proceed.</p>
+	</noscript>
+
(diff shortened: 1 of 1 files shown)


thijskh:
There is precedent for using `submit()` in other places in the code so I'm assuming this is indeed fine. But then you should remove the comment also in your commit. Otherwise it will get very confusing.

thijskh:
On second thought, the change has been quite explicitly documented. https://github.com/simplesamlphp/simplesamlphp/commit/fee5bf7590ce02e2952aa812c18ddb772d98350d

@Christian-G, can you comment on what situation triggers this behaviour? And is it still relevant?

jaimeperez:
Hi @rawroland!

Are you using master? Could it be that the problems you are experiencing with logout were due to the bug fixed in b23bb374bfcaaceb57ca979297604d6678acd9fc?

rawroland:
@jaimeperez I am using master. No, I don't think that was my problem, because the sessions were correctly removed. The problem arose when sending the response to the IdP.

tvdijen:
Hello Roland! Is this issue still present on version 1.15?

thijskh:
Closing due to lack of information. Feel free to reopen when information is provided

PR #561, opened 2017-02-15
Slight improvement of documentation in authsources config template

shamus13:
Generating <md:RequestedAttribute .../> entries in the xml export format for SPs requires 'name' to be set for the authsource in question, however this is not apparent from the documentation or example use of 'attributes' and 'attributes.required' in the authsources config template.


jaimeperez:
Thanks @shamus13!

This is covered by the documentation, but there was no mention to it in the configuration template indeed, so it's probably a good idea to mention it there too.

PR #568, opened 2017-02-27
Apache CGI support

mariendev:
Option to support Apache CGI in combination with custom config directory.

When setting the environment variable via an htaccess while using PHP via Apache CGI the environment variable gets prefixed with "REDIRECT_". This pull request adds support for reading out the other pull request

### DIFF
--- a/lib/SimpleSAML/Utils/Config.php
+++ b/lib/SimpleSAML/Utils/Config.php
@@ -68,6 +68,11 @@ class Config
     {
         $configDir    = dirname(dirname(dirname(__DIR__))) . '/config';
         $configDirEnv = getenv('SIMPLESAMLPHP_CONFIG_DIR');
+        
+        if($configDirEnv === false) {
+            $configDirEnv = getenv('REDIRECT_SIMPLESAMLPHP_CONFIG_DIR');
+        }
+        
         if ($configDirEnv !== false) {
             if (!is_dir($configDirEnv)) {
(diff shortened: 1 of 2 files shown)


jaimeperez:
Hi @mariendev!

Thanks a lot for your PR!

Can you point me to some documentation where this behaviour is described? I find it a bit odd that they add such a prefix...

mariendev:
Hi @jaimeperez ,

Hard to find official documentation on this behavior. It seems that output is redirected in Apache if you run PHP in CGI mode and then this behavior is introduced.

Fact is that we experience it on our servers and that we cannot change the configuration on short term.

Found some reference from a user on this documentation page:
http://httpd.apache.org/docs/current/mod/mod_rewrite.html

A reference to a stackoverflow item with the same issue is available here:
http://serverfault.com/questions/418299/how-to-prevent-apache-from-prepending-redirect-to-environment-variables-when-ru

Thanks!

thijskh:
There's a rather thorough treatment of the issue here:
https://stackoverflow.com/questions/3050444/when-setting-environment-variables-in-apache-rewriterule-directives-what-causes

I'm doubtful that the right way forward is to fix every application to support two variants of all their environment variables. And prefer to look for solutions outside of individual applications. The page I quoted suggests that it's only with specific RewriteRules. Can't it be worked around with different RewriteRules or with SetEnvIf?

Issue #566, opened 2017-02-22
Support for non-default session.sid_length and session.sid_bits_per_character


aberenyi:
Hi,

Let's assume one is using non-default values in `php.ini` for [`session.sid-length`](http://php.net/manual/en/session.configuration.php#ini.session.sid-length)  and [`session.sid_bits_per_character`](http://php.net/manual/en/session.configuration.php#ini.session.sid-bits-per-character) as per http://php.net/manual/en/session.security.ini.php.

Would be nice if the session ID generated [here](https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/SessionHandlerPHP.php#L166) could use the aforementioned config params otherwise users logging in via SSP will have a completely different session ID compared to users logging in via any other (e.g. simple form POST) way.

What do you think?

Cheers

jaimeperez:
Hi @aberenyi!

This is an interesting discussion, and I see arguments both in favour and against it. The most important argument in favor would be to be coherent with users' expectations, because if I'm using custom configuration settings in `php.ini` and I tell SimpleSAMLphp to use the PHP session handler, I would definitely expect it to honour that.

On the other hand, I see several downsides:

1. These two config options are only available from PHP 7.1, so we can't rely on them being available, and very few people would be using them anyway.
2. Blindly honouring these options could allow a user to downgrade the security of their session IDs.
3. SimpleSAMLphp's session ID must not be related to any other session ID, and in particular, sessions must never be shared. So in the scenario where users can login via different methods (a login form or a SAML login), both methods should have their own session cookies and as such, it's irrelevant if their contents look completely different or not.
4. On the other hand, it does make sense to have session IDs generated the same way in SimpleSAMLphp regardless of what session backend we are using. And of course, it doesn't make much sense to use PHP's settings when generating session IDs for the _memcache_ session handler.

So I'm not entirely sure I'm fine with the idea, though of course I'm open to discussion 😄 

All in all, this brings up another (more) important issue in my opinion, which is the current method we use today to generate session IDs:

```
bin2hex(openssl_random_pseudo_bytes(16))
```

[...]

PR #569, opened 2017-02-27
use session_create_id() to comply with session settings

aberenyi:
A possible solution for https://github.com/simplesamlphp/simplesamlphp/issues/566.

### DIFF
--- a/lib/SimpleSAML/SessionHandlerPHP.php
+++ b/lib/SimpleSAML/SessionHandlerPHP.php
@@ -166,7 +166,7 @@ class SimpleSAML_SessionHandlerPHP extends SimpleSAML_SessionHandler
     public function newSessionId()
     {
         // generate new (secure) session id
-        $sessionId = function_exists(session_create_id) ? session_create_id() : bin2hex(openssl_random_pseudo_bytes(16));
+        $sessionId = function_exists('session_create_id') ? session_create_id() : bin2hex(openssl_random_pseudo_bytes(16));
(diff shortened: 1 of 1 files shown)


mariendev:
Just a coding remark: you must write the function name as a string and not as an (undefined) constant:
Your code: function_exists(session_create_id)
Right version: function_exists("session_create_id")

aberenyi:
@mariendev good catch, fixed

jaimeperez:
Hi @aberenyi!

Thanks for your PR and sorry for the delay from my side.

I think this would be nice to have, but at the same time I'm a bit worried about people using bad defaults. Would it be possible to check somehow the parameters used by PHP (those configurable via `php.ini`) and at least log a warning if they are worse than what we have today?

tvdijen:
How about this:

`    public function newSessionId()
    {
        // generate new (secure) session id
        if (function_exists('session_create_id')) {
            $sid_length = (int) ini_get('session.sid_length');
            $sid_bits_per_char = (int) ini_get('session.sid_bits_per_character');

            if (($sid_length * $sid_bits_per_char) < 128) {
                \SimpleSAML\Logger::warning("Unsafe defaults used for sessionId generation!");
            }
            $sessionId = session_create_id();
        } else {
            $sessionId = bin2hex(openssl_random_pseudo_bytes(16));
        }
        SimpleSAML_Session::createSession($sessionId);
        return $sessionId;
    }`

Issue #39, opened 2014-02-27
Attribute values are not sorted when giving consent (was #505)


sspghost:
_From [j...@wayf.dk](https://code.google.com/u/112409101112243306120/) on July 04, 2012 14:36:49_

The attribute values are not sorted before the attribute hash is calculated in the consent module.

If you are using the attribute values in consent, that the attribute hash will change if the order of the attribute values change but the actual values are the same.

The values should be sorted, so the attribute hash persists even if the order of the attribute values change.

The following will fis this:

```
Index: modules/consent/lib/Auth/Process/Consent.php
===================================================================
--- modules/consent/lib/Auth/Process/Consent.php    ( revision 3128 )
+++ modules/consent/lib/Auth/Process/Consent.php    (working copy)
@@ -309,6 +309,9 @@
     {
         $hashBase = null;  
         if ($includeValues) {
+            foreach ($attributes AS &$values) {
+                sort($values);
+            }
             ksort($attributes);
             $hashBase = serialize($attributes);
         } else {
```

_Original issue: http://code.google.com/p/simplesamlphp/issues/detail?id=505_

thijskh:
@moffe42: Will you create a PR for this?

moffe42:
PR is waiting to be merged

PR #570, opened 2017-03-01
Sort attribute values for consent

moffe42:
Fixes simplesamlphp/simplesamlphp#39

This ensures that the attribute hash is the same if the order of the attribute values changes. A user
givers consent to the values, not the order.

### DIFF
--- a/modules/consent/lib/Auth/Process/Consent.php
+++ b/modules/consent/lib/Auth/Process/Consent.php
@@ -374,6 +374,9 @@ class sspmod_consent_Auth_Process_Consent extends SimpleSAML_Auth_ProcessingFilt
     {
         $hashBase = null;
         if ($includeValues) {
+            foreach ($attributes as &$values) {
+                sort($values);
+            }
             ksort($attributes);
             $hashBase = serialize($attributes);
         } else {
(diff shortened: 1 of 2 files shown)


thijskh:
Makes sense, thanks! Also great that you added a test straight away.

jaimeperez:
Hi @moffe42!

This is an issue that we've had around for quite some time. It's not only the order of the values that's irrelevant, but also the order of the attributes themselves, which could also change.

Unfortunately, a change like this would break completely all given consent, so the change needs to be applied a bit more consciously. Probably the way to go is to look for consent w/o sorting, and if found, delete the old consent, sort both values and attributes, and save the new consent.

thijskh:
@jaimeperez The attributes themselves are already sorted, right?

jaimeperez:
It could be that that was fixed some time ago, I honestly haven't checked. If so, my apologies. I just remember that the attribute order was also an issue, and I don't have any memories of fixing that with all the hassle that would entail...

jaimeperez:
Uhm, I think I might be mixing up things here, as from a quick look to the PR it's pretty obvious the attributes are sorted. I'll try to find out what's the issue I was thinking of then...

moffe42:
As far as I remember from way back at my days at WAYT, it was only the order of the values that was an issue.

moffe42:
Safe to merge??

mschwager:
It may be out of scope of this change, but I'd say it's worthwhile to add another test that ensures the keys are sorted as well. If we're depending on that functionality, then we should test it :)

moffe42:
Added test for attribute ordering.

moffe42:
According to Travis CI the build is okay. It seems that the build status was not updated for some reason.

moffe42:
Is this issue going to be fixed (pull request merge into master)? If not, then I will close this pull request.

jaimeperez:
Merged! Thanks @moffe42!

PR #572, opened 2017-03-10
OTP should be a password field on the form

gregharvey:
Colleagues commented the OTP is still a password, and really shouldn't be visible in the entry field. So this is a simple change to make the OTP a password field. It's still visible in logs for debug, etc. :-)

Thanks,

Greg


jaimeperez:
Hi!

As discussed in simplesamlphp/simplesamlphp-module-yubikey#2, this kind of OTP shouldn't be protected as a password.

Thanks a lot for the PR in any case! Much appreciated!

PR #575, opened 2017-03-22
docs: Fix the build status buttons

jornane:
The build status buttons are broken. When clicked, they simply point to their own image. The link where the results can be checked are behind the button.

This is simply fixed by removing newlines in the button code, as GitHub markdown doesn't seem to parse it correctly.


jaimeperez:
Uhm, weird, that used to work. There must have been some change in how github parses markdown. In any case, I've just merged it. Thanks Jørn!

PR #580, opened 2017-03-27
SimpleSAML_XML_* classes refactorized to PSR-4 and PSR-2

sgomez:
Two commits: one refactorized all classes in lib/SimpleSAML/XML directory to PSR-4 and the other to PSR-2.

Some new tests added.

### DIFF
--- a/lib/SimpleSAML/XML/Errors.php
+++ b/lib/SimpleSAML/XML/Errors.php
@@ -12,124 +12,128 @@
 
 namespace SimpleSAML\XML;
 
-class Errors {
+class Errors
+{
 
-	/**
-	 * This is an stack of error logs. The topmost element is the one we are
-	 * currently working on.
-	 */
-	private static $errorStack = array();
+    /**
+     * This is an stack of error logs. The topmost element is the one we are
+     * currently working on.
+     */
+    private static $errorStack = array();
 
-	/**
-	 * This is the xml error state we had before we began logging.
-	 */
-	private static $xmlErrorState;
-
-
-	/**
-	 * Append current XML errors to the the current stack level.
-	 */
-	private static function addErrors() {
-
-		$currentErrors = libxml_get_errors();
-		libxml_clear_errors();
-
-		$level = count(self::$errorStack) - 1;
-		self::$errorStack[$level] = array_merge(self::$errorStack[$level], $currentErrors);
-	}
+    /**
+     * This is the xml error state we had before we began logging.
+     */
+    private static $xmlErrorState;
 
 
-	/**
-	 * Start error logging.
-	 *
-	 * A call to this function will begin a new error logging context. Every call must have
-	 * a corresponding call to end().
-	 */
-	public static function begin() {
+    /**
+     * Append current XML errors to the the current stack level.
+     */
+    private static function addErrors()
+    {
+        $currentErrors = libxml_get_errors();
+        libxml_clear_errors();
 
-		// Check whether the error access functions are present
-		if(!function_exists('libxml_use_internal_errors')) {
-			return;
-		}
-
-		if(count(self::$errorStack) === 0) {
-			// No error logging is currently in progress. Initialize it.
-			self::$xmlErrorState = libxml_use_internal_errors(TRUE);
-			libxml_clear_errors();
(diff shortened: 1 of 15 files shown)


jaimeperez:
¡Gracias Sergio!

This is incredibly useful 😉 

I had a quick look right now at it, but it's a bit difficult to view the actual changes in the diffs here in github. From what I see, I gather it's only changes to the name classes and formatting, no functional changes, right?

sgomez:
Hi Jaime!

Yes, I split up the changes in two differents commits, because PSR-2 changes are too many. PSR-4 is just adding namespaces and import external classes with `use`.

PSR-2 was done automatically with php-cs-fixer and PSR-2 rules. It's a very tested tools so there should be no problems.

lib/SimpleSAML/XML/Errors.php
```diff
@@ -98,8 +101,8 @@ public static function end() {
 	 *
 	 * This function formats the given LibXMLError object as a string.
 	 *
-	 * @param $error  The LibXMLError which should be formatted.
-	 * @return  A string representing the given LibXMLError.
+	 * @param $error \LibXMLError  The LibXMLError which should be formatted.
```
jaimeperez: Shouldn't this be the other way around?

```
 * @param \LibXMLError $error ...
```

I see this happened all over the code, so we should fix it.

lib/SimpleSAML/XML/Shib13/AuthnResponse.php
```diff
@@ -6,12 +6,23 @@
...
+use SimpleSAML\Utils\Config;
+use SimpleSAML\Utils\Random;
+use SimpleSAML\Utils\Time;
+use SimpleSAML\XML\Validator;
+
+class AuthnResponse {
 
 	/**
-	 * This variable contains an XML validator for this message.
+	 * @var Validator This variable contains an XML validator for this message.
```
jaimeperez: I believe here we should use the full name of the class, including the namespace. I've observed problems with documentation that doesn't do that, because the `use` clauses in the beginning of the file are ignored. Therefore, I think we should use:

```
 * @var SimpleSAML\XML\Validator ...
```

lib/SimpleSAML/XML/Shib13/AuthnResponse.php
```diff
@@ -260,9 +273,9 @@ public function getIssuer() {
 		$nodelist = $this->doXPathQuery($query);
 
 		if ($attr = $nodelist->item(0)) {
-			return $attr->value;
+			return $attr->nodeValue;
```
jaimeperez: Was this a bug?

sgomez: Yes, but it's my fault. I change it because I though it is a \DOMElement but it is a \DOMAttr. I tested it and it works fine because internally must be equivalent (both extends from DOMNode). I'm going to revert the change.

lib/SimpleSAML/XML/Shib13/AuthnResponse.php
sgomez: Ok, I'm going to check all this phpdoc issues in a new commit.

lib/SimpleSAML/XML/Shib13/AuthnResponse.php
```diff
@@ -131,10 +125,11 @@ public function validate()
     }
 
 
-    /* Checks if the given node is validated by the signature on this response.
+    /**
+     * Checks if the given node is validated by the signature on this response.
      *
-     * Returns:
-     *  TRUE if the node is validated or FALSE if not.
+     * @param \DOMElement Node to be validated.
```
jaimeperez: The name of the parameter here (`$node`) is missing. I see phpdoc supports that, but they discourage the practice. It's not really a big issue, specially given this is a private method, but it would be nice to correct it 😊

sgomez: Sorry, I trust too much on the IDE syntax highlight. We missed this.

jaimeperez:
No problem Sergio! I do that too... 😉 

Merged! Thanks a lot again!

PR #581, opened 2017-03-27
facebook oauth response format changed

pradtke:
As of March 27 Facebook has deprecated their V2.2 API, and have change the response format from their access token endpoint from being URL encoded to JSON encoded.

> [Oauth Access Token] Format - The response format of https://www.facebook.com/v2.3/oauth/access_token returned when you exchange a code for an access_token now return valid JSON instead of being URL encoded. The new format of this response is {"access_token": {TOKEN}, "token_type":{TYPE}, "expires_in":{TIME}}. We made this update to be compliant with section 5.1 of RFC 6749.

### DIFF
--- a/modules/authfacebook/extlibinc/base_facebook.php
+++ b/modules/authfacebook/extlibinc/base_facebook.php
@@ -790,7 +790,8 @@ abstract class BaseFacebook
     }
 
     $response_params = array();
-    parse_str($access_token_response, $response_params);
+    //parse_str($access_token_response, $response_params);
+    $response_params = json_decode($access_token_response, true);
     if (!isset($response_params['access_token'])) {
       return false;
     }


jaimeperez:
Thanks a lot Patrick!

Incidentally, a colleague of mine just stepped into this same issue, so we were very lucky to have your PR in place 😄 

Still, we should probably stop including this file in our code base and rather pull it with composer. Do you know by any chance if facebook is publishing a package with the library?

pradtke:
It looks like they are:
https://github.com/facebook/php-graph-sdk

but it only supports php 5.4 and greater, and it may have compatibility issues with Guzzle 6 - that could be an issue if someother module pulls in a newer version of guzzle.

The auth module uses a very small subset of the Facebook API. Another option could be to use something more generic like https://github.com/thephpleague/oauth2-client, and which has supporting implementations for doing auth with 15+ other providers (amazon, google, yahoo, facebook, etc). That would provide a more common/consistent interface for building new auth modules with other oauth2 providers. One major downside is probably the php 5.6 requirement.

Issue #425, opened 2016-07-27
LDAP server hostname in authsources.php


maarbu:
When specifying a single LDAP server in the hostname directive, you don't need to put in ldap:// or ldaps://. `'hostname' => 'ldapserver.example.com',` will work fine.  However, if you add a failover server by separating with a space like this: `'hostname' => 'ldapserver1.example.com ldapserver2.example.com',` this will cause a 500 error.  This is remedied by putting in ldap:// or ldaps:// in front of each server name.  I ran across this issue when upgrading.  It used to work the first way.

I haven't sussed out exactly what upgrade caused the issue.  I upgraded PHP and OpenLDAP at the same time, my guess is OpenLDAP.

Should there be coded added to check for the ldaps:// or ldap:// prefix or is it assumed that people will put this in by default?

tvdijen:
According to the PHP documentation, http://php.net/manual/en/function.ldap-connect.php:

>     This field supports using a hostname or, with OpenLDAP 2.x.x and later, a full LDAP URI of the form ldap://hostname:port or ldaps://hostname:port for SSL encryption.
> 
>     You can also provide multiple LDAP-URIs separated by a space as one string
>     Note that hostname:port is not a supported LDAP URI as the schema is missing.

This describes exactly what you are experiencing, so I guess this works as designed.
So, although this is not a bug in any way, It would probably be nice to have SSP only allow LDAP URI's in the 'hostname' setting. That way you can test the format of the setting pretty easily and it would also provide one unmistakeable way for end-users to set it.
Also, this would take away the obscurity of setting a URI as a hostname and a non-standard port-number in the 'port'-setting (which does absolutely nothing when using URI's)

jaimeperez:
Hi @maarbu & @tvdijen!

My impression is that it doesn't make much sense to have multiple ways to end up with the same configuration. I'm not sure whether I prefer forcing URIs in the hostname(s) or not allowing them at all. I guess the fact that you can have multiple servers in the `hostname` directive while you can only specify one port, makes it a bit easier to decide, as that makes it impossible to have different servers listening on different ports.

Probably, I would suggest getting rid of the port option in all places, and having a `servers` configuration option allowing to specify an array of LDAP URIs. It is as flexible as it can get, offers just one canonical way to configure the servers, and it's pretty much in line with the idea to simplify the configuration.

```php
    'servers' => array(
        'ldaps://example.com:636',
        'ldap://example.com:389',
    ),
```

[...]

Issue #167, opened 2015-01-15
Move the SimpleSAML_Auth_LDAP class


jaimeperez:
This is actually more a collection of LDAP-related utilities, so move it to `lib/SimpleSAML/Utils/LDAP.php`. This of course needs refactoring everything using this class.

tvdijen:
Long time since the last comment, but this is still a relevant change for 2.0

jaimeperez:
If it's not a huge amount of work, I'd favour doing this for 2.0.

tvdijen:
A little while back we've discussed this on Whatsapp and decided the class could move to the ldap-module, which I did. The class now only is a wrapper, redirecting to the class in the module for BC-compatibility.
For 2.0 all we have left to do is to remove `lib/SimpleSAML/Auth/LDAP.php` from this repository.

PR #586, opened 2017-03-30
Fix non-standard LDAP ports (assume LDAP URIs in authsources.php)

JohnMaguire:
We fixed an issue we were experiencing where non-standard LDAP ports did not work correctly. This was due to the fact that we were specifying LDAP URIs in `authsources.php` (to support multiple servers, as noted in #425) which causes the `$port` param of [`ldap_connect()`](https://secure.php.net/ldap_connect) to be ignored.

However, there are tradeoffs to this solution: On the plus side, the interface for configuring authsources doesn't change; on the negative side, this only works if you define your `authsources.php` with LDAP URIs, and breaks otherwise.

This is also a little odd because if you were to define your `authsources.php` with LDAP URIs, you could just as simply add the port number into the LDAP URI itself.

I don't really expect this to be merged, but I figured I would put it up for discussion.

### DIFF
--- a/lib/SimpleSAML/Auth/LDAP.php
+++ b/lib/SimpleSAML/Auth/LDAP.php
@@ -75,11 +75,18 @@ class SimpleSAML_Auth_LDAP
                 SimpleSAML\Logger::warning('Library - LDAP __construct(): Unable to set debug level (LDAP_OPT_DEBUG_LEVEL) to 7');
         }
 
+        // Port must be specified in LDAP URI
+        $hostname = explode(' ', $hostname);
+        $hostname = array_map(function ($host) use ($port) {
+            return "{$host}:{$port}";
+        }, $hostname);
+        $hostname = implode(' ', $hostname);
+
         /*
          * Prepare a connection for to this LDAP server. Note that this function
          * doesn't actually connect to the server.
          */
-        $this->ldap = @ldap_connect($hostname, $port);
+        $this->ldap = @ldap_connect($hostname);
         if ($this->ldap === false) {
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @JohnMaguire!

After thinking a bit about this, maybe it makes sense to deprecate the old way of configuring this, and replacing it in 2.0 with just a single configuration option to set the URI(s)?

JohnMaguire:
That's probably a good idea, as long as you aren't concerned with BC (or are you suggesting a new named option?) The tricky part here was trying to work with existing configurations...

One advantage to just using LDAP URIs instead of list of hosts & a port is that each host could be on a different port, for example.

tvdijen:
There's a lot of reasons to properly break stuff regarding LDAP.. There's #425, but I recall some bc-breaking stuff to the ldap-module, as well as the fact that the current ldap-module is prone to brute-forcing.. And then there's #167 as well..

@jaimeperez : Regarding #167, i could put some effort in that, but I'm thinking [symfony/ldap](https://github.com/symfony/ldap) is perfect to replace at least a part of the code. What are your thoughts about this?

JohnMaguire:
Right, I saw #425 when opening this PR -- I wasn't sure that it justified a BC break though. While #425 definitely means some people have issues with LDAP configuration, I am sure many people have a working config based on the (incorrect) assumptions made by the existing code.

I have no horse in this race however. Happy to take whichever route.

thijskh:
Thanks. I also concur that the current PR is not the way forward, so I'll be closing it. The mentioned open issues will track progress on fixing this in a more structural way.

Issue #571, opened 2017-03-06
LDAP Referrals flag, cannot set via config.


tsmgeek:
In file "./modules/ldap/lib/Auth/Process/BaseFilter.php" on 251, I have to manually force referrals for AD lookups else my logins hang till timeout.
It would be good if this attribute can be set via 'ldap.referrals'.
I am having this problem when using 'AttributeAddFromLDAP'

mschwager:
I can look into this for you! Should be a quick enough fix :+1:

tvdijen:
I think this should cover it:
https://github.com/tvdijen/simplesamlphp/commit/9b57e5569f7981bfc3ce06b8f70d792f16520621

mschwager:
@tvdijen that looks good, although I think you want `referrals` not `referals`.

tvdijen:
Ohhhh crap... Today turns out to be not my day :(

New attempt:
https://github.com/tvdijen/simplesamlphp/commit/ac2fb6658df0ac284bcd9c60b9b79052d8bfea8c

mschwager:
That looks good to me! Unfortunately I'm not able to test it right now. I'm sure it'd be fine to open the PR and discuss things there!

tsmgeek:
Just updated to latest release and then realised that this had yet been merged in.

tvdijen:
Stil happy to make a PR if someone, preferably @mschwager, can confirm to me that it works.

jaimeperez:
@tvdijen wasn't #587 addressing precisely this issue? Did @mschwager or @tsmgeek test that?

tvdijen:
Ohhh right, I already did the PR.. No, nothing has been tested yet...

jaimeperez:
It's good to see I'm not the only one who does something and then forgets... 😅 

In any case, this has a 👍  from me as long as someone can test it properly.

PR #587, opened 2017-03-30
Fix issue #571: adhere to authsource referrals-setting

tvdijen:
Fixes issue #571


tvdijen:
It would be great if @tsmgeek could test this

modules/ldap/lib/Auth/Process/BaseFilter.php
```diff
@@ -142,6 +142,7 @@ public function __construct(&$config, $reserved)
             $authconfig['ldap.port']       = @$authsource['port'];
             $authconfig['ldap.timeout']    = @$authsource['timeout'];
             $authconfig['ldap.debug']      = @$authsource['debug'];
+            $authconfig['ldap.referrals']   = (@$authsource['referrals'] ? @$authsource['referrals'] : false);
```
mschwager: I think this should default to `true` since `SimpleSAML_Auth_LDAP::__construct` does.

modules/ldap/lib/Auth/Process/BaseFilter.php
```diff
@@ -233,6 +234,7 @@ protected function getLdap()
         $port       = $this->config->getInteger('ldap.port', 389);
         $enable_tls = $this->config->getBoolean('ldap.enable_tls', false);
         $debug      = $this->config->getBoolean('ldap.debug', false);
+        $referrals   = $this->config->getBoolean('ldap.referrals', false);
```
mschwager: Same as above.

mschwager:
Unfortunately I'm not able to easily test this since I don't have access to an LDAP server where turning off referrals would change anything.

However, I'd be okay with pulling in the changes anyway. We're simply providing a method to access the already existing argument in `SimpleSAML_Auth_LDAP::__construct` via configuration. What do you think @jaimeperez?

thijskh:
I agree with you @mschwager that this is good to merge. However, it seems the branch has lost some content now. And @mschwager has IMO a valid comment. @tvdijen Can you look into that?

tvdijen:
@thijskh What are you missing?
What I did, recently, is resolving a conflict, because the SSP master branch has been changed since I sent in this pull request..
I'll fix the default value later today

thijskh:
I was mistaken, indeed it looks fine otherwise.

tvdijen:
I just set the default value to true.. Tests are running

thijskh:
@tvdijen Do you think this option needs to be added to the documentation?

tvdijen:
Nope, this was an actual bug, not a feature..  You specify an authsource for the authproc-filter to use and expect it to take over the settings defined in that authsource. 

>     * LDAP connection settings can be retrieved from an ldap:LDAP
>     * authsource. Specify the authsource name here to pull that
>     * data from the authsources.php file in the config folder.

Paragraph 3:
https://simplesamlphp.org/docs/stable/ldap:ldap

Same text is also in the doc-file of the ldap-module.

PR #596, opened 2017-04-10
FQDN can be up to 255 chars according to RFC1035 3.1

### LINKED ISSUE #579: Issue with long site names and SQL session Logout store
After switching from memcache to MySQL as session backend, I have been getting an exception when inserting a new row in <prefix>_saml_LogoutStore table.

SimpleSAML_Error_Error: UNHANDLEDEXCEPTION
Backtrace:
0 /simplesamlphp/www/module.php:180 (N/A)
Caused by: PDOException: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column '_authSource' at row 1

After inspecting the code, I saw that _authSource contains the sitename, and in /simplesamlphp/modules/saml/lib/SP/LogoutStore.php the table that is being created limits the size of this value to 30.
In our company we allow for longer names, so the problem is fully understood.

The obvious workaround for us would be increase that value to comply with our company's naming policies, but I would like to ask if you foresee any side effect in doing so.

Cheers.

sergiofcasado:
Fixes #579.

According to [RFC1035 3.1](https://tools.ietf.org/html/rfc1035#section-3.1) and [RFC 2181 11](https://tools.ietf.org/html/rfc2181#section-11), domain names can be up to 255 in size.

### DIFF
--- a/modules/saml/lib/SP/LogoutStore.php
+++ b/modules/saml/lib/SP/LogoutStore.php
@@ -14,12 +14,24 @@ class sspmod_saml_SP_LogoutStore {
 	 */
 	private static function createLogoutTable(\SimpleSAML\Store\SQL $store) {
 
-		if ($store->getTableVersion('saml_LogoutStore') === 1) {
+		$tableVer = $store->getTableVersion('saml_LogoutStore');
+		if ($tableVer === 2) {
+			return;
+		} elseif ($tableVer === 1) {
+			/* TableVersion 2 increased the column size to 255 which is the maximum length of a FQDN. */
+			$query = 'ALTER TABLE ' . $store->prefix . '_saml_LogoutStore MODIFY _authSource VARCHAR(255) NOT NULL';
+			try {
+				$ret = $store->pdo->exec($query);
+			} catch (Exception $e) {
+				SimpleSAML\Logger::warning($store->pdo->errorInfo());
+				return;
+			}
+			$store->setTableVersion('saml_LogoutStore', 2);
 			return;
 		}
 
 		$query = 'CREATE TABLE ' . $store->prefix . '_saml_LogoutStore (
-			_authSource VARCHAR(30) NOT NULL,
+			_authSource VARCHAR(255) NOT NULL,
 			_nameId VARCHAR(40) NOT NULL,
 			_sessionIndex VARCHAR(50) NOT NULL,
 			_expire TIMESTAMP NOT NULL,
@@ -34,7 +46,7 @@ class sspmod_saml_SP_LogoutStore {
 		$query = 'CREATE INDEX ' . $store->prefix . '_saml_LogoutStore_nameId ON '  . $store->prefix . '_saml_LogoutStore (_authSource, _nameId)';
 		$store->pdo->exec($query);
 
-		$store->setTableVersion('saml_LogoutStore', 1);
+		$store->setTableVersion('saml_LogoutStore', 2);
 	}
 
 


JohnMaguire:
It looks like it might be worth adding an alter table here: https://github.com/sergiofcasado/simplesamlphp/blob/e404e609ec73e2883fc7267d4f8e542b938c934a/modules/saml/lib/SP/LogoutStore.php#L17-L19

And setting the table version to 2: https://github.com/sergiofcasado/simplesamlphp/blob/e404e609ec73e2883fc7267d4f8e542b938c934a/modules/saml/lib/SP/LogoutStore.php#L37

rwmanos:
The suggestions of @JohnMaguire seem reasonable. I updated this MR accordingly.

rwmanos:
I can also remove lines 23-24 https://github.com/sergiofcasado/simplesamlphp/blob/10677340987dc120cdba6d5f29dd243696565dc9/modules/saml/lib/SP/LogoutStore.php#L23-L24 and put the next lines up to before `$store->setTableVersion('saml_LogoutStore', 2);` https://github.com/sergiofcasado/simplesamlphp/blob/10677340987dc120cdba6d5f29dd243696565dc9/modules/saml/lib/SP/LogoutStore.php#L27-L41 in an `else` statement. This will be slightly simpler.

rwmanos:
I added a comment and updated the commit message. Can someone have a look?

rwmanos:
@jaimeperez @thijskh Can you please have a look?

modules/saml/lib/SP/LogoutStore.php
```diff
@@ -14,12 +14,19 @@ class sspmod_saml_SP_LogoutStore {
 	 */
 	private static function createLogoutTable(\SimpleSAML\Store\SQL $store) {
 
-		if ($store->getTableVersion('saml_LogoutStore') === 1) {
+		if ($store->getTableVersion('saml_LogoutStore') === 2) {
```
jaimeperez: Can we store the result of `getTableVersion()` in a variable and use it in the conditions instead of evaluating the function twice? Something like:

```php
$tableVer = $store->getTableVersion('saml_LogoutStore');
if ($tableVer === 2) {
    return;
} elseif ($tableVer === 1) {
```

modules/saml/lib/SP/LogoutStore.php
```diff
@@ -14,12 +14,19 @@ class sspmod_saml_SP_LogoutStore {
...
+			return;
+		}
+		elseif ($store->getTableVersion('saml_LogoutStore') === 1) {
```
jaimeperez: I'm being picky here, but if we are modifying the code slightly here, I'd also like to take the chance to follow our coding standard completely. Do you mind moving the `elseif` up one line?

```php
} elseif ($tableVer === 1) {
```

Thanks!

modules/saml/lib/SP/LogoutStore.php
```diff
@@ -14,12 +14,19 @@ class sspmod_saml_SP_LogoutStore {
...
+			/* TableVersion 2 increased the column size to 255 which is the maximum length of a FQDN. */
+			$query = 'ALTER TABLE' . $store->prefix . '_saml_LogoutStore MODIFY _authSource VARCHAR(255) NOT NULL';
+			$store->pdo->exec($query);
+			$store->setTableVersion('saml_LogoutStore', 2);
```
jaimeperez: Should we check the result of the `ALTER TABLE` statement before bumping the table version? If something goes wrong (i.e. this particular syntax is unsupported by the database backend in use), then we should keep the version to `1` (and probably log an error).

rwmanos:
Hello @jaimeperez, I modified the commit according to your feedback. I also added the SQL error checking. I applied the new patch and it worked well for me. Let me know if you want to change anything else!

rwmanos:
Actually, the SQL error check is wrong. It's never evaluated since if there is an error there will be an Unhandled exception. I will rewrite it with a try-catch.

rwmanos:
Done. Let me know if you have another comment.

jaimeperez:
Everything looks good to me! 👍 

Thanks a lot again for your help!

PR #601, opened 2017-04-24
Reduce state spill over between tests

pradtke:
A PHPUnit listener unsets SSP environmental variables and attempts
to restore globals.


jaimeperez:
Hi @pradtke!

Thanks a lot for this and sorry so much for the silence these months. I still need to go through the PR to make sure I follow everything, especially given that you have some doubts too...

I think this is something we definitely need to tackle, and the sooner the better.

tests/Utils/ClearStateTestListener.php
```diff
@@ -0,0 +1,78 @@
+<?php
+namespace SimpleSAML\Test\Utils;
+
+use PHPUnit_Framework_Test;
+use PHPUnit_Framework_TestSuite;
+
+/**
+ * A PHPUnit test listener that attempts to clear global state and cached SSP configuration between test run
+ */
+class ClearStateTestListener extends \PHPUnit_Framework_BaseTestListener
```
jaimeperez: As far as I understand listeners in phpunit, they are executed always, right? That has probably an impact in the time taken to run the tests. I was wondering... wouldn't it be easier to create an abstract class that clears the state before/after each test and then have the tests that need a clear state extend that class? Something like:

```php
class ClearStateTest extends PHPUnit_Framework_TestCase
{
    protected function setUp()
    {
        clearState();
    }

    protected function tearDownAfterClass()
    {
        clearState();
    }

    public static function test1()
    {
         ...
    }

    ...
}

class SomeTestClass extends ClearStateTest
{
    ...
}
```

tests/config/no-options/config.php
```diff
@@ -0,0 +1,8 @@
+<?php
```
jaimeperez: This is not needed. The whole point of being able to load a configuration from an array was, precisely, to avoid having a fixed directory structure.

tests/lib/SimpleSAML/ConfigurationTest.php
```diff
@@ -31,6 +31,13 @@ public function testLoadDefaultInstance()
      */
     public function testCriticalConfigurationError()
     {
+        try {
+            SimpleSAML_Configuration::getInstance();
+            $this->fail('Exception expected');
+        } catch (\SimpleSAML\Error\CriticalConfigurationError $var) {
+            // This exception is expected.
```
jaimeperez: This is normal behaviour. `SimpleSAML_Configuration::getInstance()` attempts to fetch the configuration from disk. If there's no such configuration, it fails with a `CriticalConfigurationError` exception.

tests/lib/SimpleSAML/ConfigurationTest.php
```diff
@@ -31,6 +31,13 @@ public function testLoadDefaultInstance()
...
+        }
+        //TODO: not sure this is correct. First call to getInstance() throw exception and the 2nd works?...
```
jaimeperez: This is _collateral damage_. When a `CriticalConfigurationError` occurs, you are supposed to display that to the user and be done. That's the entire purpose of that exception. However, some classes will still be invoked that need a configuration object (i.e. `SimpleSAML\Logger`). Since there is no valid configuration (that's why we threw the exception in the first place), the only option is to load an emergency configuration that can be used by any class that needs it.

Here you are the victim of such behaviour, as you are capturing the exception and continuing execution after an event that's supposed to end it.

tests/lib/SimpleSAML/ConfigurationTest.php
```diff
@@ -97,6 +104,9 @@ public function testHasValueOneOf() {
      * Test SimpleSAML_Configuration::getBaseURL()
      */
     public function testGetBaseURL() {
+
+        // Need to set a configuration file because the SSP Logger attempts to load it.
+        putenv('SIMPLESAMLPHP_CONFIG_DIR=tests/config/no-options');
```
jaimeperez: This is not necessary. Instead, the next line should read:

```php
$c = SimpleSAML_Configuration::loadFromArray(array(), '[ARRAY]', 'simplesaml');
```

I admit this is quite obscure. The reason for those parameters is that `SimpleSAML_Configuration::getInstance()` gets, but default, an instance of the configuration called `simplesaml`. This is just a name, but it allows us to load multiple configurations at the same time. In any case, this is the default configuration that you get when you invoke the `getInstance()` method without parameters (as most of the code does).

Now, in `SimpleSAML_Configuration::loadFromArray()`, the instance name is the third parameter, and in order to specify it, you need to specify also the second, which is the _location_ of the configuration. In this case, the location defaults to `[ARRAY]` to signal that this configuration was loaded from an array instead of from a file. It doesn't really matter what you set here, the only thing is that you _need_ to set that parameter to something.

Therefore, if you want to load a default configuration in `SimpleSAML_Configuration` from an array, you need to specify those two parameters, because that's what the code will be looking for later.

tests/lib/SimpleSAML/Locale/LocalizationTest.php
```diff
@@ -6,6 +6,11 @@
 
 class LocalizationTest extends \PHPUnit_Framework_TestCase
 {
+    protected function setUp()
+    {
+        // Localization/Language code attempts to load a cookie, and looks to config.php for a name of the cookie
+        putenv('SIMPLESAMLPHP_CONFIG_DIR=tests/config/no-options');
```
jaimeperez: This shouldn't be necessary.

tests/modules/consent/lib/Auth/Process/ConsentTest.php
```diff
@@ -31,6 +31,9 @@ private function processFilter(array $config, array $request)
      */
     public function testCheckDisable()
     {
+        // Consent code path attempts to load metadata, which tries to find a config.php
+        putenv('SIMPLESAMLPHP_CONFIG_DIR=tests/config/no-options');
```
jaimeperez: Same here...

tests/modules/saml/lib/Auth/Source/Auth_Source_SP_Test.php
```diff
@@ -121,6 +121,9 @@ protected function setUp()
                 ),
             ),
         );
+
+        // Modules makes use of HTTP for getting the base url which attempts to load a config.php
+        putenv('SIMPLESAMLPHP_CONFIG_DIR=tests/config/no-options');
```
jaimeperez: ... and here.

tests/Utils/ClearStateTestListener.php
pradtke: In our earlier conversation (https://groups.google.com/forum/#!topic/simplesamlphp-dev/OKfZtXrc3P0) I thought you didn't want to go with the base class because it required more knowledge on the part of the test writer on how the SSP project does testing.

I have no real preference between the approaches. The test listener did not add meaningfully to the test execution time. I think the subclass approach would give a test implementation more control over what they wanted done with the state, and is easier for people to understand.

tests/lib/SimpleSAML/ConfigurationTest.php
pradtke: OK. I'll go through and change my use `putenv('SIMPLESAMLPHP_CONFIG_DIR=tests/config/no-options')` to this.

tests/lib/SimpleSAML/ConfigurationTest.php
pradtke: OK, that makes sense. Thanks for the clarification.

tests/lib/SimpleSAML/ConfigurationTest.php
pradtke: Yup. This part of the code change was related to reducing state spill over.  This test previously was relying on testLoadDefaultInstance() running before it so it would in the correct state for this test. Now the test can be run on its own.

tests/lib/SimpleSAML/Locale/LocalizationTest.php
pradtke: I'll change to `SimpleSAML_Configuration::loadFromArray(array(), '[ARRAY]', 'simplesaml');`

tests/modules/consent/lib/Auth/Process/ConsentTest.php
pradtke: I'll change to `SimpleSAML_Configuration::loadFromArray(array(), '[ARRAY]', 'simplesaml');`

tests/modules/saml/lib/Auth/Source/Auth_Source_SP_Test.php
pradtke: I'll change to `SimpleSAML_Configuration::loadFromArray(array(), '[ARRAY]', 'simplesaml');`

tests/config/no-options/config.php
pradtke: Gotcha. I'll make changes as you suggested and test things out.

pradtke:
I've moved over to using `SimpleSAML_Configuration::loadFromArray(array(), '[ARRAY]', 'simplesaml');` as discussed.

Remaining things are to decide if we want a testing base class vs a listener.  I think the former is more flexible.

There will likely be a bunch of additional places that need to have state cleared. I think it makes sense to expand the use of the interface as tests are created that need to clear the state. For example, in one of our internal modules we currently do

```php
        $reflection = new \ReflectionProperty('SimpleSAML_Store', 'instance');
        $reflection->setAccessible(true);
        $reflection->setValue(null, null);
```

since we need to vary the instance used.

tests/Utils/ClearStateTestListener.php
jaimeperez: Sorry! 😞 

As you can see, my memory doesn't serve me well after a few months... 😓 

I agree with you that the base class that tests can extend is much more flexible, because each test can control what parts of the state it needs clean. Going back to that conversation, I can also see why I preferred transparency, but maybe that benefit comes at a high price. In the end, we should document how to do testing in the project anyway, so maybe transparency isn't that important in this case...

In any case, I don't want you to feel like you have wasted your time because I changed my mind here, so I leave it up to you how to proceed from here. If you don't have time to change the PR to use a base class, we can just merge it now and I'll take responsibility to do the changes myself. If you want to change it before merging the PR, I'll be grateful and owe you a few beers next time we meet each other 😉

lib/SimpleSAML/Configuration.php
```diff
@@ -198,7 +198,7 @@ public static function setConfigDir($path, $configSet = 'simplesaml')
 
 
     /**
-     * Load a configuration file from a configuration set.
+     * Load a configuration file from a configuration set.XM
```
jaimeperez: Typo?

jaimeperez:
Thanks for the last changes Patrick! I agree with you about expanding the use of the interface. Another thing that we've been discussing here lately was to start using _dependency injection_ so that we can properly test everything. I've always been skeptical about DI because it changes completely how things work, and feels a bit obscure for new developers. On the other hand, I think it's the only way forward if we aim at 100% code coverage.

In any case, that would be a huge change in the code and how everything works in SSP, so let's take one step at a time.

lib/SimpleSAML/Configuration.php
pradtke: oops. Removed it. Probably my subconscious doing a keyboard shortcut that didn't exist in my editing app :)

pradtke:
I've moved to a base testcase class that can do the clearing.
I had trouble with the autoloading though and it not resolving any base classes in the 'tests/' folder. I ended up using a bootstrap.php file for phpunit as a workaround. 

I didn't move all the existing tests over (since it would add a lot of noise to the PR), but I did test with them moved over. I think moving them should be a separate commit. They can be moved with

```bash
 perl -p -i -e 's/PHPUnit_Framework_TestCase/SimpleSAML\\Test\\Utils\\ClearStateTestCase/g' `grep -rl PHPUnit_Framework_TestCase tests | grep -v ClearStateTestCase.php`
```

There may be some adjustments needed to individual testcases where the testcase has overridden setup and we want it to also call super to clear the state.

jaimeperez:
Thanks again for your work Patrick!

PR #603, opened 2017-04-27
Add REMOTE_ADDR to successful auth and user name to unsuccessful log …

mholisky:
Because the logging can be inconsistent on the SP side, I'd like to add the REMOTE_ADDR to the logging message of successful logins. This extra info will help administrators identify compromised accounts.
Added the username to the unsuccessful log message to help identify specific accounts that are being attacked.

### DIFF
--- a/modules/core/lib/Auth/UserPassBase.php
+++ b/modules/core/lib/Auth/UserPassBase.php
@@ -256,7 +256,7 @@ abstract class sspmod_core_Auth_UserPassBase extends SimpleSAML_Auth_Source {
 			throw $e;
 		}
 
-		SimpleSAML\Logger::stats('User \''.$username.'\' Successfully authenticated from '.$_SERVER['REMOTE_ADDR']);
+		SimpleSAML\Logger::stats('User \''.$username.'\' successfully authenticated from '.$_SERVER['REMOTE_ADDR']);
 
 		/* Save the attributes we received from the login-function in the $state-array. */
 		assert('is_array($attributes)');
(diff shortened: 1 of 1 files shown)


thijskh:
I like the idea of adding the `REMOTE_ADDR` to the authentication log.

However, for adding the username to the authentication error message, this has been avoided in the past with the argument that users may often erroneously type their password in the username field, leading to logged passwords.

mholisky:
Thanks for the feedback, I've reverted the change on the unsuccessful log message.

jaimeperez:
Hi @mholisky!

Have you seen [this](https://github.com/simplesamlphp/simplesamlphp/blob/master/config-templates/config.php#L235-L261)? You can log the remote IP address in all your messages, so I don't think this is really necessary...

mholisky:
Thanks @jaimeperez, I did not see that configuration option.  I can most certainly start using that. 

Though I would make one argument to accept my change.  If by default the IP isn't being logged, and an account gets compromised, it becomes difficult to preform forensics, block the source, identify any other accounts that maybe compromised, and then trace what the attacker was able access.  

Because the IdP is the central place my users access many different SPs, it would be the ideal location to gather this information when some SPs are not logging the username and/or IP (or I don't have easy access to the logs).  But for the most part there is a good chance the SP will have web server logs, and I can track down what that IP was accessing on the SP.  

Thanks everyone for their hard work on this project!  I'm just trying to help the next sysadmin down the line.

modules/core/lib/Auth/UserPassBase.php
```diff
@@ -256,7 +256,7 @@ public static function handleLogin($authStateId, $username, $password) {
 			throw $e;
 		}
 
-		SimpleSAML\Logger::stats('User \''.$username.'\' has been successfully authenticated.');
+		SimpleSAML\Logger::stats('User \''.$username.'\' Successfully authenticated from '.$_SERVER['REMOTE_ADDR']);
```
thijskh: Nitpick: `S` of `Successfully` should probably be lowercase.

thijskh:
I can see the value in merging this, what do you think @jaimeperez?

jaimeperez:
Merged! Thanks @mholisky for your help!

PR #604, opened 2017-04-28
AttributeLimit: allow options to use regex or ignore case.

pradtke:
Enhancements to AttributeLimit to give additional control over allowed values

### DIFF
--- a/modules/core/docs/authproc_attributelimit.md
+++ b/modules/core/docs/authproc_attributelimit.md
@@ -47,7 +47,7 @@ Only allow specific values for an attribute ignoring case.
         50 => array(
             'class' => 'core:AttributeLimit',
             'eduPersonEntitlement' => array(
-                'ignoreCase' => 'true',
+                'ignoreCase' => true,
                 'URN:x-surfnet:surf.nl:SURFDRIVE:quota:100'
              )
         ),
@@ -59,7 +59,7 @@ Only allow specific values for an attribute that match a regex pattern
         50 => array(
             'class' => 'core:AttributeLimit',
             'eduPersonEntitlement' => array(
-                'regex' => 'true',
+                'regex' => true,
                 '/^urn:x-surfnet:surf/',
                 '/^urn:x-IGNORE_Case/i',
             )
--- a/modules/core/docs/authproc_attributelimit.md
+++ b/modules/core/docs/authproc_attributelimit.md
@@ -32,6 +32,41 @@ Allow `eduPersonTargetedID` and `eduPersonAffiliation` by default, but allow the
         ),
     ),
 
+Only allow specific values for an attribute.
+
+    'authproc' => array(
+        50 => array(
+            'class' => 'core:AttributeLimit',
+            'eduPersonEntitlement' => array('urn:x-surfnet:surf.nl:surfdrive:quota:100')
+        ),
+    ),
+
+Only allow specific values for an attribute ignoring case.
+
+    'authproc' => array(
+        50 => array(
+            'class' => 'core:AttributeLimit',
+            'eduPersonEntitlement' => array(
+                'ignoreCase' => 'true',
+                'URN:x-surfnet:surf.nl:SURFDRIVE:quota:100'
+             )
+        ),
+    ),
+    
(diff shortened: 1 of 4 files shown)


pradtke:
I have changed the code to load an empty configuration rather than have Logger handle a missing configuration.

thijskh:
@jaimeperez "not even sure about the change being the right way to go" → do I understand correctly that you have doubts about the PR/change as a whole?

pradtke:
I'm open to doing it a different way. For background we needed the functionality to provide feature parity with shibboleth's attribute filter stuff when we did a migration for a customer.

modules/core/docs/authproc_attributelimit.md
```diff
@@ -32,6 +32,41 @@ Allow `eduPersonTargetedID` and `eduPersonAffiliation` by default, but allow the
...
+        ),
+    ),
+
+Only allow specific values for an attribute ignoring case.
+
+    'authproc' => array(
+        50 => array(
+            'class' => 'core:AttributeLimit',
+            'eduPersonEntitlement' => array(
+                'ignoreCase' => 'true',
```
thijskh: `'true'` and same on line 61 should not be strings, especially because the code does a strict type comparison.

modules/core/lib/Auth/Process/AttributeLimit.php
```diff
@@ -124,4 +124,41 @@ public function process(&$request) {
...
+                    } elseif ($regexResult === 1) {
+                        $matchedValues[] = $attributeValue;
+                        // Remove matched value incase a subsequent regex also matches it.
+                        unset($values[$index]);
+                    }
+                }
+            }
+            return $matchedValues;
+        } elseif (array_key_exists('ignoreCase', $allowedConfigValues) && $allowedConfigValues['ignoreCase'] === true) {
+            unset($allowedConfigValues['ignorecase']);
```
thijskh: This `ignorecase` has a different casing than the `ignoreCase` on the line right above.
Also, you unset this option but do not unset the regex one above, any reason for that?

modules/core/lib/Auth/Process/AttributeLimit.php
```diff
@@ -124,4 +124,41 @@ public function process(&$request) {
...
+                    */
+                    $regexResult = @preg_match($pattern, $attributeValue);
+                    if ($regexResult === false) {
+                        \SimpleSAML\Logger::warning("Error processing regex '$pattern' on value '$attributeValue'");
+                        break;
+                    } elseif ($regexResult === 1) {
+                        $matchedValues[] = $attributeValue;
+                        // Remove matched value incase a subsequent regex also matches it.
+                        unset($values[$index]);
+                    }
```
thijskh: What if > 1 match is found?

tvdijen:
@Thijskh: I think Jaime was referring to changes to SimpleSAML_Logger in the first commit that have already been reverted in a later commit..

jaimeperez:
I guys! It's hard to remember almost after a year, but I'm quite sure @tvdijen is right and I was referring to the changes to `SimpleSAML_Logger` 😉

modules/core/docs/authproc_attributelimit.md
pradtke: Fixed

modules/core/lib/Auth/Process/AttributeLimit.php
pradtke: Good catch. I don't want to feed the value (true) into the intersect function since php converts into a string '1' and it would match an attribute value of "1".  I added a test for this and fixed the casing. I also added code+tests to confirm it is unset in the non-true case as well.

For regex unset wasn't needed since it was already using is_int to exclude any string keys.

modules/core/lib/Auth/Process/AttributeLimit.php
pradtke: preg_match only returns false, 0 or 1. You're probably thinking of preg_match_all which returns the number of matches. I updated one of the tests and added a regex that would match an attribute value multiple times

PR #606, opened 2017-05-04
Added Dockerfile

dalareo:
I have added docker support with a Dockerfile and also a Docker Hub public image at https://hub.docker.com/r/dalareo/simplesamlphp/


jaimeperez:
¡Hola David!

I must confess even I think this is nice to have, I'm not sure it belongs here in the main project... What do you think @thijskh?

tvdijen:
As an alternative, we could mention the project on the SSP-website, like we do with modules

jaimeperez:
Yeah, that's also a possibility that I was entertaining. In the end, I think this belongs into its own repo. That way the Dockerfile can be improved without us (me) being a bottleneck, and we can just refer there from our website and the documentation.

David, how does that sound? Do you think that would make sense?

jaimeperez:
Oh, by the way! I'm not against having this as a separate repo under the `simplesamlphp` namespace, on the contrary! I just would like to have someone taking care of it with more knowledge of Docker than me, so that people can work on it kind of independently...

dalareo:
I think nowadays a Dockerfile is a must for every project but of course agree with your criteria. A specific repo with Dockerfile only will also make sense

jaimeperez:
Well, as you can see I'm not using Docker that much myself... 😅 

I see your point, and if it's a common habit, then I'm happy to eat my own words 😄 

In the end, we're already including an RPM spec file...

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM php:apache
+MAINTAINER David A. Lareo "dalareo@gmail.com"
+
+RUN apt-get update -y \
+	&& apt-get install -y git libmcrypt-dev libldap2-dev \
+	&& docker-php-ext-configure ldap --with-libdir=lib/x86_64-linux-gnu/ \
+
+	&& docker-php-ext-install mcrypt ldap
+
+RUN git clone https://github.com/simplesamlphp/simplesamlphp.git /var/simplesamlphp
```
jaimeperez: At this point, you have already cloned the repo, so... is it really necessary to clone it again? Wouldn't it make more sense to just make the current dir available to the container, so that it can also be used by developers to get a development version up and running quickly, where they can develop locally and test in the container?

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN rm -rf /var/simplesamlphp
+RUN git clone https://github.com/simplesamlphp/simplesamlphp.git /var/simplesamlphp
```
jaimeperez: Here I'm lost. Why cloning, removing, and cloning again?

etc/apache2/sites-available/default-ssl.conf
```diff
@@ -0,0 +1,148 @@
+<IfModule mod_ssl.c>
```
jaimeperez: I think this file belongs in `extra`. Also, we probably don't need to specify the same name in the origin as in the destination, so maybe something like `extra/docker-apache.conf`? And then we can just rename it in the Dockerfile...

Dockerfile
dalareo: Yes, you're write that first cloning was a mistake. It makes sense to clone the repo instead of just copying the code in order to let the file be able to work aside the original repo.

Dockerfile
dalareo: Totally agree. Just removing before cloning to check everything is clean inside image/container

Dockerfile
jaimeperez: I guess the problem here is that you are thinking about using this to deploy something easily into production, while I'm thinking about this to make it easier for developers... 😄 

Maybe having a separate `Dockerfile` for development?

dalareo: Copying current code is also right for dev and production. I've updated Dockerfile

jaimeperez: The problem with this is that it is not really suitable for development, as you don't have direct access to the code that you are running. You would need to stop and start the container again everytime you want to test something, which is why I was suggesting to mount the directory. But again, I don't think that's really suitable for production, hence my question about having two separate Docker files...

dalareo: That "problem" can be solved using volumes to sync host folders with container ones.

mleklund:
The docker file would more standard if it did not assume it was part of a git checkout and instead pulled the simplesamlphp code from git.

```
ENV SSP_VERSION=1.14.13
ENV SSP_HASH=6e97a9701cf916ab0b49b4622c64208acdd813af6378b9d78193b65607311680
RUN   wget https://github.com/simplesamlphp/simplesamlphp/releases/download/v$SSP_VERSION/simplesamlphp-$SSP_VERSION.tar.gz \
    && echo "$SSP_HASH  simplesamlphp-$SSP_VERSION.tar.gz" | sha256sum -c - \
	&& cd /var \
	&& tar xzf /simplesamlphp-$ssp_version.tar.gz \
    && mv simplesamlphp-$ssp_version simplesamlphp \
    && rm /simplesamlphp-$ssp_version.tar.gz
```

hmpf:
As for @mleklund's Dockerfile, in cases like this I prefer to move all the copy/untar-magic into a shellscript so that the Dockerfile becomes easier to read (and edit!)

tvdijen:
I've recently caught up with Docker and I think the Dockerfile should definitely be included in the extra-directory.. I don't think we should include Apache-config though, as it is just too specific.. Let's add a generic example docker-file

tvdijen:
@jaimeperez I recall us talking about this in Valencia and I believe our conclusion was that docker-files are so specific per environment that we should not try to include some generic file that doesn't suit anyone's needs.. What do we want with this?

jaimeperez:
Well, as we can see in this very same discussion, everyone has their own preferences and opinions on how to do stuff in the Dockerfile, so I definitely think this is not something that we can ship with the software as a _one-size-fits-all_ solution.

So I'm closing the ticket for now (sorry!). That doesn't mean we should ignore this work or trash it. I believe this makes sense as a separate repo, configuring SSP as an IdP. The SP side is a bit more complicated because it is then intended to protect a PHP application, so in that case it's definitely the responsibility of the user. Of course, the resulting Dockerfile can always be used as a guide or an example.

LaurentGoderre:
Hey,
I alternative;y created this Dockerfile:

```Dockerfile
FROM php:7.3-fpm-alpine

ENV SIMPLESAMLPHP_VERSION=1.16.3

RUN cd /var \
  && wget https://github.com/simplesamlphp/simplesamlphp/releases/download/v${SIMPLESAMLPHP_VERSION}/simplesamlphp-${SIMPLESAMLPHP_VERSION}.tar.gz \
  && tar xzf simplesamlphp-${SIMPLESAMLPHP_VERSION}.tar.gz \
  && rm simplesamlphp-${SIMPLESAMLPHP_VERSION}.tar.gz \
  && ln -s simplesamlphp-${SIMPLESAMLPHP_VERSION} simplesamlphp
```

This results in a super small image (90mb) and allows to configure nginx or apache in front of it as needed.

tvdijen:
Hey @LaurentGoderre! Feel free to create a repo for this. We will not include a Docker-file in this repo.

LaurentGoderre:
@tvdijen yeah that's what I was thinking. Are you open to maintain it officially? I can create the repo and transfer the ownership to this org.

tvdijen:
Perhaps @jaimeperez can answer this.. I'm not sure if we have Docker-experience in our team

LaurentGoderre:
We could also implement env var to make configuring the docker setup super easy.

tvdijen:
I've seen an example of that [here](https://hub.docker.com/r/kristophjunge/test-saml-idp)
I have actually used that one recently to mock an IDP and I guess it's useful for development purposes, but other than that I see more downsides than upsides and would never use this in a production environment.

Long story short, I think I see the benefits of a Docker-file that can be used for mocking / devops purposes.. An updated version of Kristoph Junge's would make sense, but actually supporting it may be a bridge too far for our relatively small team.. Let me promise you this:  I will bring up the subject again during our hackathon later this week and discuss it with our contributors

LaurentGoderre:
@tvdijen using env var like this make its very powerful when used with something like Kubernetes to provision the service.

PR #607, opened 2017-05-05
Added tests for SimpleSAML\Utils\Crypto

mschwager:
Hey all! I'm back with some more tests.

This also required adding an additional argument to `SimpleSAML\Utils\Crypto::loadPrivateKey` to ease in testing. Without this additional argument, `SimpleSAML_Configuration::getBaseDir` eventually gets called to determine the private key location.  This doesn't work well with `vfsstream`. This argument shouldn't cause too much trouble, and seems cohesive enough with the function's purpose.


mschwager:
Thanks for the PHP 5.3 fix! Looks like this is good to go!

jaimeperez:
Indeed it does. Thanks a lot again @mschwager! This one was truly helpful!

PR #610, opened 2017-05-07
Updated traditional chinese(zh-tw) translations

jal-tw:
I have update of some Traditional Chinese (zh-tw) to more easier understand and clear.


dictionaries/admin.translation.json
```diff
@@ -885,7 +885,6 @@
 		"sr": "Shib 1.3 Davalac Identiteta (udaljeni)",
 		"ro": "Furnizor de identitate Shib 1.3 (distant)",
 		"ru": "\u041f\u0440\u043e\u0432\u0430\u0439\u0434\u0435\u0440 \u0438\u0434\u0435\u043d\u0442\u0438\u0444\u0438\u043a\u0430\u0446\u0438\u0438 Shib 1.3 (\u0443\u0434\u0430\u043b\u0435\u043d\u043d\u043e\u0435 \u0440\u0430\u0437\u043c\u0435\u0449\u0435\u043d\u0438\u0435)",
-		"eu": "Shib 1.3 Identitate hornitzailea (Urrunekoa)",
```
thijskh: This should probably not be removed, but the line below changed to "el"...

jal-tw: In el translation, some of character should not be translated. Such as '@' \u0040, ';' \u003b, and many symbols.
In dictionaries/logout.translation.json:594, there have one translation forgot to encode.
I do no change any content of other language, except zh-tw. Of other language, I only corrected all of data to same format using some decode and encode tools.

jaimeperez:
Thanks a lot for your contribution @jal-tw!

Since I understand the changes to languages other than traditional Chinese are accidental, due to unicode characeters being translated to their encodings, and given that evaluating all those changes would be time consuming, do you mind to revert the changes to any other languages than the one you are updating?

Thanks again!

jal-tw:
Basically, it is totally not accidental. Other than zh-tw languages modify are actually correct some misses, double space, no encoded and some English should not to encode.
I have to double check every I modify.

As your wish, I revert it to original version.

thijskh:
Thanks, I feel safer merging this now.

Issue #618, opened 2017-05-24
Display issues on mobile devices: no submit (login) button visible


poikilotherm:
Hey guys,

while creating a custom theme, I discovered that on mobile devices the login button is missing, see screenshots below.

I'm pretty sure you guys fixed this in #359, but making valid XHTML in #479 (found via blame) introduced the third `<td></td>`. This makes things valid, but breaks the view.

Shall I open a PR? How did you guys ensure XHTML validity, so not to break goals of #479 by the new fix?

Cheers,
Oliver

![ssp-error-portrait](https://cloud.githubusercontent.com/assets/5468333/26404937/e9239d66-4093-11e7-8883-ef635b838c5c.png)
![ssp-error-landscape](https://cloud.githubusercontent.com/assets/5468333/26404938/e949bac8-4093-11e7-8f10-e17a3b9028ca.png)

jaimeperez:
Hi @poikilotherm! Thanks for reporting this!

Of course, a PR fixing the issue would be really nice, if you don't mind 😄

poikilotherm:
Here you are... ;-) Also this doesn't fix the table width issue, but this is a story for another day...

tvdijen:
I've tried on my own IdP on several phones and they do show a login-button, however it is outside of the screen, so one would have to scroll to the right...

The fix to this is actually a lot less complicated as what you're suggesting in your PR..
My suggestion would be this:

https://github.com/tvdijen/simplesamlphp/commit/bbcbc8c83a33e023183da29bff9102552f22c412

poikilotherm:
Hi @tvdijen,

I agree that my description of "invisible" was not very precise - I sticked with the wording in #359.

About the complexity: I tried to revert the hacky workaround of hiding different table row via CSS introduced in daafee5, as it introduced more HTML and CSS code than necessary. One could argue that it would be even cleaner to use a proper `rowspan="3"`, a `vertical-align: top;` for the icon `<td>` and remove one of the empty cells, but this would break the current design and I intended to avoid that all cost (users complain even about moved icons on their desktop...)

I tried your fix on my IdP and noted the design break (the button moves to the outer left). Was that on purpose?

Cheers,
Oliver

tvdijen:
That was on purpose yes.. I must admit I didn't take the time to see what it looked like on mobile phones before the #359/#479 changes... The username/password fields are also way too wide on my phones..

[...]

PR #620, opened 2017-05-24
Fix for #618, correcting the login layout on mobile screens.

poikilotherm:
As mentiond in #618, a PR :-)

Feel free to request changes. Would be glad to see this merged.

### DIFF
--- a/modules/core/templates/loginuserpass.php
+++ b/modules/core/templates/loginuserpass.php
@@ -36,7 +36,7 @@ if ($this->data['errorcode'] !== null) {
     <form action="?" method="post" name="f">
         <table>
             <tr>
-                <td rowspan="2" id="loginicon">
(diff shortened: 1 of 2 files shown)


poikilotherm:
Sorry for the noise... Is there any particular reason this really simple and small fix is not yet merged?

jaimeperez:
Hi @poikilotherm!

First of all, thanks a lot for the PR and sorry for the delay. Basically, it's not merged because I haven't been able to look at it yet. That, and also that we are working in a new template system and user interface, so this kind of fixes get low priority because they will be temporary...

thijskh:
Tested it, could reproduce the original problem and that your change fixes it, on two browsers on desktop and two browsers on mobile. Thanks!

jaimeperez:
Thanks Thijs!

PR #622, opened 2017-05-25
Added NameID Functions For SimpleSAML_Auth_Simple

rkeplin:
I've added two functions in `SimpleSAML_Auth_Simple ` to more easily gather NameID information.

```php
$auth = new SimpleSAML_Auth_Simple('some-auth-source');
$auth->getNameIDData('Value');
$auth->getNameIDDataArray();
```
These helper functions provide a shortcut to the NameID information, rather than going through `SimpleSAML_Auth_Simple::getAuthData()`.

### DIFF
--- a/lib/SimpleSAML/Auth/Simple.php
+++ b/lib/SimpleSAML/Auth/Simple.php
@@ -293,6 +293,46 @@ class SimpleSAML_Auth_Simple
 
 
     /**
+     * Retrieves NameID data.
+     *
+     * @param string $name The name of the parameter, e.g. 'Value', 'Format'
+     *
+     * @return string|null The value of the parameter, or null if it isn't found or we are unauthenticated.
+     */
+    public function getNameIDData($name)
+    {
+        assert('is_string($name)');
+
+        if (!$this->isAuthenticated()) {
+            return null;
+        }
+
+        $nameIdData = $this->getAuthData('saml:sp:NameID');
+
+        if (isset($nameIdData[$name])) {
+            return $nameIdData[$name];
+        }
+
+        return null;
+    }
+
+
+    /**
+     * Retrieve all the NameID data.
+     *
+     * @return array|null All the NameID data, or null if we aren't authenticated
+     */
+    public function getNameIDDataArray()
+    {
+        if (!$this->isAuthenticated()) {
+            return null;
+        }
+
+        return $this->getAuthData('saml:sp:NameID');
+    }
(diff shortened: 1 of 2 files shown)


jaimeperez:
Hi @rkeplin!

Thanks a lot for your contribution, and especially for providing a test! 👏 

I'm not a big fan of having multiple ways to achieve the same goal, as that tends to generate confusion among users. On the other hand, this is quite typical, and it might deserve the special consideration given here... What do you think @thijskh?

thijskh:
If we would add special functions I think it makes sense to make them even more specific, e.g. getNameIdValue() and getNameIdFormat().

rkeplin:
@jaimeperez @thijskh Sounds good.  I've replaced the more generic methods with more specific functions.

jaimeperez:
Hi!

I've been giving this a second thought, and I have an argument against it (sorry!). Basically, `SimpleSAML_Auth_Simple` is a generic class that's intended as a base for every authentication mechanism in SSP. It's at the core of the SP and part of the SP API provided, but it's also part of many other authentication flows (there are modules using it, and there's even a great example of the class being used for something entirely different in the admin authentication mechanism).

What I'm trying to say is that `SimpleSAML_Auth_Simple` is used generically, not as something SAML-specific. NameIDs, on the other hand, are quite SAML-specific, and therefore having a few methods specific to SAML in an otherwise generic class doesn't fit well, IMHO. In the end, this could be as simple as:

```php
use SAML2\XML\saml\NameID;
...
$nameIdData = $as->getAuthData('saml:sp:NameID');
if ($nameIdData !== null) {
    $nameId = NameID::fromArray($nameIdData);
}
```

So, for now, I think I'll close the PR (sorry again!), but of course that doesn't mean I'm not willing to continue the discussion here and maybe change my mind 😉

rkeplin:
No worries!  That makes sense, if `SimpleSAML_Auth_Simple` is intended to be a generic authentication class, and not necessarily geared towards anything SAML specific.

What sprung this idea, is that I can get SAML attributes via `getAttributes()`.  But to get NameID information, I have to go through `getAuthData()`.

As the attributes are also returned from `getAuthData()`, it seems `getAttributes()` is a shortcut to them.  So I thought to make a shortcut to help expose the NameID information as well.

Anyways, it's easy enough to extend `SimpleSAML_Auth_Simple` for this.  Thank you for taking it into consideration.

PR #623, opened 2017-05-25
Added tests for SimpleSAML\Utils\XML

mschwager:
More tests :)

### DIFF
--- a/tests/lib/SimpleSAML/Utils/XMLTest.php
+++ b/tests/lib/SimpleSAML/Utils/XMLTest.php
@@ -1,0 +1,379 @@
+<?php
+
+namespace SimpleSAML\Test\Utils;
+
+use \SimpleSAML_Configuration as Configuration;
+use \SimpleSAML\Utils\XML;
+
+/**
+ * Tests for SimpleSAML\Utils\XML.
+ */
+class XMLTest extends \PHPUnit_Framework_TestCase
+{
+    /**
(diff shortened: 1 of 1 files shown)


tests/lib/SimpleSAML/Utils/XMLTest.php
```diff
@@ -0,0 +1,379 @@
...
+    {
+        $data1 = 'root value 1';
+        $data2 = 'root value 2';
+        $dom = new \DOMDocument();
+        $element = $dom->appendChild(new \DOMElement('root'));
+        $element->appendChild(new \DOMText($data1));
+        $element->appendChild(new \DOMText($data2));
+
+        $res = XML::getDOMText($element);
+        $expected = $data1 . $data2 . $data1 . $data2;
```
mschwager: I was a little confused as to why this is repeated twice. Does this make sense, or is there a bug here?

thijskh: Yes, because you feed two times an `DOMText` to the element you're constructing, you've created this document in `$element` after line 134:
```
<?xml version="1.0"?>
<root>root value 1root value 2</root>
```

So the output of `getDOMText` is correct. Whether that's the document you wanted to construct for your test, I don't know.

mschwager: Ahh! Thanks for clarifying that. It wasn't clear to me that the `DOMText` was still represented as two objects internally.

I think this PR is good to go! :+1:

PR #627, opened 2017-06-01
Added metadata configuration options for 'index' and 'isDefault' attributes in AttributeConsumingService element

vanhanit:
This patch adds two metadata configuration options for 'index' and 'isDefault' attributes. We had to be able to output the isDefault attribute in the AttributeConsumingService element.
I don't know if 'attributes.index' and 'attributes.isDefault' is the best and correct names for the options, so please feel free to comment on this if needed.

### DIFF
--- a/lib/SimpleSAML/Metadata/SAMLBuilder.php
+++ b/lib/SimpleSAML/Metadata/SAMLBuilder.php
@@ -403,7 +403,11 @@ class SimpleSAML_Metadata_SAMLBuilder
          */
         $attributeconsumer = new \SAML2\XML\md\AttributeConsumingService();
 
-        $attributeconsumer->index = 0;
+        $attributeconsumer->index = $metadata->getInteger('attributes.index', 0);
+
+        if ($metadata->hasValue('attributes.isDefault')) {
+            $attributeconsumer->isDefault = $metadata->getBoolean('attributes.isDefault', false);
+        }
 
         $attributeconsumer->ServiceName = $name;
         $attributeconsumer->ServiceDescription = $metadata->getLocalizedString('description', array());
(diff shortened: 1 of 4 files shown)


askibinski:
+1, works as described. Adding custom index values is sometimes required for creating valid SP metadata.

vanhanit:
It adds the isDefault and (configurable) index attributes to the AttributeConsumingService tag in the generated SP metadata.

With attributes.index = 1 and attributes.isDefault = true the following is generated: <md:AttributeConsumingService index="1" isDefault="true">

What I can understand the generator can only generate one of these tags, but if more are possible the configuration should of course be per generated tag (but I can't find any configuration for that kind of ourput).

I can't answer why the isDefault is needed, I just know that an IdP we connected agains wanted it to be this way (with index = 1 and isDefault = true).

vanhanit:
As an example, this is the relevant parts I have in the working production environment.
```
'attributes' => array(
		'nationalIdentificationNumber' => 'urn:oid:1.2.246.21',
		'givenName' => 'urn:oid:2.5.4.42',
		'sn' => 'urn:oid:2.5.4.4',
		'displayName' => 'urn:oid:2.16.840.1.113730.3.1.241',
		'cn' => 'urn:oid:2.5.4.3',
),

'attributes.NameFormat' => 'urn:oasis:names:tc:SAML:2.0:attrname-format:uri',
'attributes.isDefault' => true,
'attributes.index' => 1,
```
It shows up in the metadata as
`<md:AttributeConsumingService index="1" isDefault="true">`

vanhanit:
Maybe you were looking at `AssertionConsumerService` instead of `AttributeConsumingService`? I named it `attributes.isDefault` and `attributes.index` because they pertain to the same data that other `attributes.*` settings adjust.

thijskh:
Right, I see where I went wrong now. Will re-check the PR

PR #634, opened 2017-06-15
Added tests for SimpleSAML\XML\Signer

mschwager:
Added some more tests! Also, fixed a PSR-2 issue in `Signer`, and had to add some additional `$full_path` arguments since `vfsstream` doesn't play well with `getCertPath`.

### DIFF
--- a/lib/SimpleSAML/XML/Signer.php
+++ b/lib/SimpleSAML/XML/Signer.php
@@ -123,14 +123,22 @@ class Signer
      *                      to the cert-directory.
      * @param string|null $pass  The passphrase on the private key. Pass no value or NULL if the private
      *                           key is unencrypted.
+     * @param bool $full_path  Whether the filename found in the configuration contains the
+     *                         full path to the private key or not. Default to false.
      * @throws \Exception
      */
-    public function loadPrivateKey($file, $pass = null)
+    public function loadPrivateKey($file, $pass = null, $full_path = false)
     {
         assert('is_string($file)');
         assert('is_string($pass) || is_null($pass)');
+        assert('is_bool($full_path)');
 
-        $keyFile = Config::getCertPath($file);
+        if (!$full_path) {
+            $keyFile = Config::getCertPath($file);
+        } else {
+            $keyFile = $file;
+        }
+
         if (!file_exists($keyFile)) {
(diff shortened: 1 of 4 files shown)


mschwager:
I saw these test failures locally as well:

```
1) SimpleSAML\Test\Module\consent\Auth\Process\ConsentTest::testCheckDisable
array_key_exists() expects parameter 2 to be array, null given

simplesamlphp/lib/SimpleSAML/Configuration.php:302
simplesamlphp/lib/SimpleSAML/Metadata/MetaDataStorageHandler.php:56
simplesamlphp/lib/SimpleSAML/Metadata/MetaDataStorageHandler.php:43
simplesamlphp/modules/consent/lib/Auth/Process/Consent.php:232
simplesamlphp/tests/modules/consent/lib/Auth/Process/ConsentTest.php:24
simplesamlphp/tests/modules/consent/lib/Auth/Process/ConsentTest.php:67
```

It seems to be some interaction with the configuration being statically set. I've seen these kinds of errors from other tests as well. Setting the configuration via static member variables makes for some weird interactions :(

Thoughts?

mschwager:
Figured out the configuration error, `tests/modules/consent/lib/Auth/Process/ConsentTest.php` and `tests/modules/saml/lib/Auth/Source/Auth_Source_SP_Test.php` tests work correctly if they're initialized with an empty configuration in the setup.

I'm still not entirely sure why this initial change surfaced this issue. Using static member variables for configuration testing means that if a test forgets to clean up after itself or not initialize the configuration properly it will probably stomp on other tests.

thijskh:
Would #601 help to mitigate that?

mschwager:
@thijskh I think it might! From what I can tell it looks like that PR is ensuring an empty configuration for all tests via some PHPUnit functionality? That seems like a sane thing to do since we've essentially been manually doing that in certain files, e.g. `clearInstance` in `RedisTest.php`, `SystemTest.php`, `SignerTest.php`, etc.

For now I think this PR is good to go! In the future I think it'd be great to have a better, more generalized solution to specifying configuration in tests, whether that's https://github.com/simplesamlphp/simplesamlphp/pull/601 or something else. I'll look into it if I get some time :)

Let me know what you think!

thijskh:
I agree, merging this PR improves the current situation. Thanks a lot for your work!

Issue #593, opened 2017-04-05
Inline JavaScripts do not allow secure Content Security Policy


martin-kuba:
I have an installation of SimpleSamlPHP 1.14 and I have tried to achieve "A" rating of my site at https://securityheaders.io/ The tool checks for having a Content Security Policy defined by a HTTP header. 
The problem is that SimpleSamlPHP uses a lot of in-line JavaScripts and CSS styles, so the policy must contain "style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline';" to keep them working, but such setting is considered as insecure because in-line scripts and styles can be injected by an attacker.

The recommended solution is to move all scripts and style into separate files and link them.

See https://en.wikipedia.org/wiki/Content_Security_Policy and https://hacks.mozilla.org/2016/02/implementing-content-security-policy/ for details on CSP.

thijskh:
There's an ongoing effort to move to Twig templates. The move of JS from inline to separate files will be tackled as part of that effort, and/or be much simpler after that change. You're welcome to help move the Twig changes forward!

tvdijen:
Only two cases left:
- consent/www/getconsent.php for showing/hiding attribute-values
- templates/includes/header.php does an onload for autofocus

thijskh:
Largely solved in new UI.
See #950 for remaining case.
Any other cases found should also have new issues created.

PR #635, opened 2017-06-16
Make POST template compatible with CSP

AJdeMooij:
See issue #593 for a problem description.
SimpleSamlPHP makes use of unsafe inline Javascript and CSS elements.
Although most generated HTML uses SimpleSamlPHP's own headers, the
keepPost option in an authentication request uses the headers of
the PHP application it is sent from. This forces web applications
using SimpleSamlPHP to allow 'unsafe-inline' in their Content
Security Policy.

This commit fixes this issue for the keepPost page ''only'', to
allow PHP applications using SimpleSamlPHP to use a more strict
Content Security Policy. This does not take away from possible
XSS vulnerabilities in other parts of SimpleSamlPHP

In accordance with the coding standards in CONTRIBUTE.md, all tabs in templates/post.php have been replaced by four spaces as well.

One unit test fails:
 There was 1 failure:
 
 1) Test_SimpleSAML_Configuration::testLoadDefaultInstance
 Failed asserting that exception of type "\SimpleSAML\Error\CriticalConfigurationError" is thrown.

As this was already the case on the master branch, and as this seems quite a different problem, I have not taken it upon myself to fix this issue.

### DIFF
--- a/templates/post.php
+++ b/templates/post.php
@@ -1,34 +1,47 @@
+<?php
+$wwwlib = sprintf(
+    "//%s/%sresources/", 
+    htmlentities($_SERVER['SERVER_NAME']), 
+    htmlentities($this->data['baseurlpath'])
+);
+?>
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
-	<meta http-equiv="content-type" content="text/html; charset=utf-8" />
-	<title>POST data</title>
+    <meta http-equiv="content-type" content="text/html; charset=utf-8" />
+    <title>POST data</title>
+    <script type="text/javascript" src="<?php echo $wwwlib;?>post.js"></script>
+    <link 
+        type="text/css" rel="stylesheet" href="<?php echo $wwwlib;?>post.css" />
 </head>
-<body onload="document.getElementsByTagName('input')[0].click();">
+<body>
 
-	<noscript>
-		<p><strong>Note:</strong> Since your browser does not support JavaScript, you must press the button below once to proceed.</p> 
-	</noscript> 
-	
-	<form method="post" action="<?php echo htmlspecialchars($this->data['destination']); ?>">
-	<!-- Need to add this element and call click method, because calling submit()
-	on the form causes failed submission if the form has another element with name or id of submit.
-	See: https://developer.mozilla.org/en/DOM/form.submit#Specification -->
-	<input type="submit" style="display:none;" />
+    <noscript>
+        <p><strong>Note:</strong> 
+        Since your browser does not support JavaScript, 
+        you must press the button below once to proceed.</p> 
+    </noscript> 
+
+    <form method="post" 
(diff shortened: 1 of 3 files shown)


templates/post.php
```diff
@@ -1,34 +1,47 @@
+<?php
+$wwwlib = sprintf(
+    "//%s/%sresources/", 
+    htmlentities($_SERVER['SERVER_NAME']), 
+    htmlentities($this->data['baseurlpath'])
+);
+?>
```
thijskh: Better to use the functions from `lib/SimpleSAML/Utils/HTTP.php` here.

thijskh:
Thanks. As mentioned in #593 the structural solution should come from the templating changes. But I can appreciate the idea that this case is special and that warrants merging this now. Added one comment, think it's fine to merge otherwise. :+1:

AJdeMooij:
Thank you for your reply. I modified this pull request to incorporate the suggested change and I will be looking forward to the Twig templating.

templates/post.php
```diff
@@ -1,19 +1,16 @@
...
 ?>
+
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
     <meta http-equiv="content-type" content="text/html; charset=utf-8" />
     <title>POST data</title>
-    <script type="text/javascript" src="<?php echo $wwwlib;?>post.js"></script>
+    <script type="text/javascript" src="<?php echo HTTP::getBaseURL();?>/resources/post.js"></script>
```
jaimeperez: This is wrong. According to `HTTP::getBaseURL()` documentation:

```php
    /**
     * Retrieve the base URL of the SimpleSAMLphp installation. The URL will always end with a '/'. For example:
     *      https://idp.example.org/simplesaml/
```

This means the URL generated here would be "whatever//resources/post.js".

Additionally, there's no need to use the functions from `SimpleSAML\Utils\HTTP` directly, as the base URL is already available in the data passed to the template. Instead of this, the code should read:

```php
    <script type="text/javascript" src="/<?php echo $this->data['baseurlpath']; ?>resources/post.js"></script>
```

AJdeMooij: Ah, the trailing slash was my mistake, I should have read the documentation better and I shall fix that one.

As for the `$this->data['baseurlpath']` comment; this one only seems to give the relative path to the public folder on the server, whereas `HTTP::getBaseUrl()` gives the fully qualified domain name. This last one is important when during SimpleSamlPHP install a symlink is created for the packages `simplesamlphp/www` directory. Or am I missing something here?

jaimeperez: Hi again!

Sounds like you are mixing the server's path and the URL. It doesn't matter if you used symlinks, aliases or any other mechanisms during installation. `$this->data['baseurlpath']` will contain the relative base path (of the URL) to the script calling the template. Check it [here](https://github.com/simplesamlphp/simplesamlphp/blob/master/templates/includes/header.php#L56) for example.

AJdeMooij: Indeed I did! Thank you for your comment. I forgot the leading slash, which meant the path could not be resolved, but using the same method as in the example solved my issue. I just pushed my third attempt ;)

thijskh:
Thanks!

PR #637, opened 2017-06-19
Add PDO $options support

ElijahLynn:
I may be missing something but I haven't found a way to add the --ssl-ca
MySQL option to my database connection. It appears we just need to pass
in an $options param.

### DIFF
--- a/docs/simplesamlphp-customauth.md
+++ b/docs/simplesamlphp-customauth.md
@@ -273,9 +273,11 @@ The class follows:
                 throw new Exception('Missing or invalid password option in config.');
             }
             $this->password = $config['password'];
-            $this->options = $config['options'];
-            if (!is_array($config['options])) {
-                throw new Exception('Missing or invalid options option in config.');
+            if (isset($config['options']) {
+                if (!is_array($config['options])) {
+                    throw new Exception('Missing or invalid options option in config.');
+                }
+                $this->options = $config['options'];
             }
         }
 
--- a/docs/simplesamlphp-customauth.md
+++ b/docs/simplesamlphp-customauth.md
@@ -253,9 +253,10 @@ The class follows:
          */
         private $dsn;
 
-        /* The database username & password. */
+        /* The database username, password & options. */
         private $username;
         private $password;
+        private $options;
 
         public function __construct($info, $config) {
             parent::__construct($info, $config);
@@ -272,6 +273,10 @@ The class follows:
                 throw new Exception('Missing or invalid password option in config.');
             }
             $this->password = $config['password'];
+            $this->options = $config['options'];
+            if (!is_array($config['options])) {
+                throw new Exception('Missing or invalid options option in config.');
+            }
         }
 
         /**
@@ -294,7 +299,7 @@ The class follows:
         protected function login($username, $password) {
 
             /* Connect to the database. */
-            $db = new PDO($this->dsn, $this->username, $this->password);
+            $db = new PDO($this->dsn, $this->username, $this->password, $this->options);
             $db->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
 
             /* Ensure that we are operating with UTF-8 encoding.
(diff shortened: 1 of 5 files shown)


modules/consent/lib/Consent/Store/Database.php
```diff
@@ -99,6 +104,14 @@ public function __construct($config)
             $this->_password = null;
         }
 
+        if (array_key_exists('options', $config)) {
+            if (!is_array($config['options'])) {
+                throw new Exception('consent:Database - \'password\' is supposed to be an array.');
```
thijskh: 'password' should be 'options'

ElijahLynn: Thanks, fixed with a amendment, --force push.

docs/simplesamlphp-customauth.md
```diff
@@ -272,6 +273,10 @@ The class follows:
                 throw new Exception('Missing or invalid password option in config.');
             }
             $this->password = $config['password'];
+            $this->options = $config['options'];
+            if (!is_array($config['options])) {
+                throw new Exception('Missing or invalid options option in config.');
+            }
```
thijskh: In this example `options` is now mandatory in the configuration while in the code below it remains optional, is that intentional?

modules/consent/lib/Consent/Store/Database.php
```diff
@@ -471,8 +484,13 @@ private function _getDB()
         if (isset($this->_timeout)) {
             $driver_options[PDO::ATTR_TIMEOUT] = $this->_timeout;
         }
+        if (isset($this->_options)) {
+            array_merge($driver_options, $this->_options);
```
thijskh: this misses an assignment and is hence a no-op

modules/sqlauth/lib/Auth/Source/SQL.php
```diff
@@ -68,6 +74,7 @@ public function __construct($info, $config) {
 		$this->dsn = $config['dsn'];
 		$this->username = $config['username'];
 		$this->password = $config['password'];
+		$this->options = $config['options'];
```
thijskh: Here options is used unconditionally thus mandatory part of the config, while above it's optional. This should be consistent (and for BC I'd want it to be optional).

modules/consent/lib/Consent/Store/Database.php
ElijahLynn: Gotcha, I suppose it is working for us without merging in $driver_options. Good catch.

ElijahLynn: Oh, I don't actually think we are using the Consent module, but maybe I just included this for completeness.

ElijahLynn:
Thanks for the review. I think I addressed your comments but please let me know if I missed anything.

docs/simplesamlphp-customauth.md
```diff
@@ -272,6 +273,12 @@ The class follows:
                 throw new Exception('Missing or invalid password option in config.');
             }
             $this->password = $config['password'];
+            if (isset($config['options']) {
+                if (!is_array($config['options])) {
+                    throw new Exception('Missing or invalid options option in config.');
+                }
+                $this->options = $config['options'];
+            }
         }
```
thijskh: A bit of a nitpick because it's documentation, but $this->options will now be unset if you don't have the config option, thus generate a notice below when it's used.

modules/sqlauth/lib/Auth/Source/SQL.php
```diff
@@ -69,6 +75,9 @@ public function __construct($info, $config) {
 		$this->username = $config['username'];
 		$this->password = $config['password'];
 		$this->query = $config['query'];
+                if (isset($config['options'])) {
+                    $this->options = $config['options'];
+                }
```
thijskh: Indentation seems to be off here.

thijskh:
Thanks - nearly done I think

PR #639, opened 2017-06-22
Disco suggest algorithm accepts also -_() characters

tauceti2:
When IdP has e.g. acronym in the name covered by () or there is dash in the name, then the current suggest algorithm doesn't split the name correctly, so disco doesn't show the name of the organization when you type the acronym or something which is in brackets.


thijskh:
Shouldn't that be in `[`brackets`]` to work?

tauceti2:
Yep, you're right. I have it with brackets on our production IdP, but I just forgot to copy those characters. Apologise for such stupid typo and thanks for the check. Pull request is fixed.

jaimeperez:
Hi!

I'm confused here. Is the PR no longer needed?

tauceti2:
Yes, it adds possibility to search also institutions which has ()-_ in its name. I simply forgot to put additional characters which split the word from the institution name into the brackets to be correct regex.

jaimeperez:
Great, thank you both! I was confused as the PR was closed...

PR #643, opened 2017-06-23
IDP SAML2: added 'as:Reauth'

tbenr:
Allow AuthenticationSource to decide, in reauthenticate method, if user has to forcedly reauthenticate.

I don't know it "as:" prefix is correct. If the feature is considered interesting, let's suggest the correct one.

### DIFF
--- a/lib/SimpleSAML/IdP.php
+++ b/lib/SimpleSAML/IdP.php
@@ -413,6 +413,10 @@ class SimpleSAML_IdP
                 assert('FALSE');
             } else {
                 $this->reauthenticate($state);
+                if(isset($state['as:Reauth'])) {
+                    $this->authenticate($state);
+                    assert('FALSE');
+                }
             }
             $this->postAuth($state);
         } catch (SimpleSAML_Error_Exception $e) {


jaimeperez:
Hi!

Again, what's the use case you are aiming for?

In any case, you don't need new code to do this. If you want the auth source to force the user to authenticate again, it just needs to set `$state['ForceAuthn']` to `true`. That would also trigger complete authentication, instead of calling `reauthenticate()` and later `authenticate()`, as in your code.

I'm closing this for now, but of course comments are welcome.

tbenr:
Hi, thanks for comments!

The use case here is this:
I developed an authentication module that delegates authentication to an Access Manager acting as a reverse proxy. Authenticated Userid and attributes arrive in HTTP header variables.

So the method redirect the user to a page (ie /simplesaml/module.php/_authmodule_/callback.php) that is "protected" by the access manager. once the user authenticate, the Access Manager let him access callback.php with username  and attributes in http headers that can be consumed by the module.

When user has already a session, I also want to check session against Access Manager, that is, from ssp stand point, it is actually a reauthentication (going through callback.php again).

I did added "as:Reauth" because i was thinking that, when user has a valid session against simplesamlphp, the framework calls only reauthenticate(). So from an authentication module stand point, there where no other places to "comunicate" to framework to do a reauthentication, Moreover "ForceAuthn" is evaluated before calling reauthenticate(), and it is something related to the request, not actually an internal need of the IDP to manage authentication.

So where i need to set $state['ForceAuthn'] = true in the authentication module to get the same behavior?

tbenr:
Yes, that is my objective. I didn't take in account that solution because internally SSP goes through `SimpleSAML_Auth_Simple::login()` flow when handling `ForceAuthn == true`. So, to be sure, I made it by replicating the same flow. But if you say it is equivalent, that's much better.
Thanks!

jaimeperez:
Well, it's not completely equivalent, you need to take a look at the code and consider if you need anything from the code that won't be run in your case. For example, any of the information added by to the state array by `SimpleSAML_Auth_Simple::login()` or `SimpleSAML_Auth_Source::initLogin()`. Note also that `SimpleSAML_Auth_Source::loginCompelted()` won't be executed, and then if you need to update the data in the session object, that won't get done.

So it really depends on what you need, but in general, it should be possible to tackle this in your own code without much trouble.

tbenr:
Ok. But if it is not a real and complete reauthentication, don't you think that giving the ability to an authentication module to restart an authentication is a feature that increases flexibility?

More generally, _SP_ has its opportunity to force reauthentication via `ForcheAuthn` in request. Should Identity provider have the same opportunity based on its security policy? I believe yes (ie dynamic risk evaluation?). Here i'm proposing that _IdP_ can do it by delegating the decision to the _Authentication Module_. Maybe this is not the ideal place where enforce this... or maybe there is already a place (that I'm currently ignoring) where implement this kind of rules... I'm just thinking loudly...

PR #644, opened 2017-06-23
IDP SAML2: added 'as:NoSession'

tbenr:
Allow AuthenticationSource to decide to avoid SessionIndex generation and association to current IDP session (avoid SLO).

I don't know it "as:" prefix is correct. If the feature is considered interesting, let's suggest the correct one.


jaimeperez:
Hi @tbenr!

I'm a bit confused by this. What's the use case you have in mind?

tbenr:
Hi again :-)

i have some requirements where, for some Authn Requests (heaving particular AuthnContextClassRefs) i dont have to generate a session on IdP side (so no SessionIndex released to SP). But for other AuthnContextRefs the session is still required. So I still need to have a session with SP tracking (for SLO).

So i developed an authentication module (the same module mentioned here #643  :-) ) which decide how to authenticate the user based on RequestedAuthnContext and also if simplesamlphp has to release a SessionIndex and track SP for SLO.

jaimeperez:
Ok, now you definitely got my attention 😄 

Why is all this needed? I mean, you are basically letting an SP control whether users will be able to log out of it or not, from what I understand in your description. What's the purpose?

Also, what's the trouble with `SessionIndex`?

In any case, I feel this is very ad-hoc, so I'm a bit hesitant on how to approach this...

tbenr:
Well.. the "pratical" reason is this: I have to be compliant to [SPID](https://www.spid.gov.it/) regulatory requiring _IdP_ to NOT release `SessionIndex `when authenticating in _SpidL2_ or higher (let's say two factor auth). It also requires to not generating a session on _IdP_ side. Actually, on _IdP_ side you have a session only associated to _SpidL1_ (single factor auth), and eventually other authentication _SpidL2_ or higher MUST NOT interact with the previously established _SpidL1_ SSO session. 

"Theoretically", IMHO these requirements are reasonable and have their underlying philosophy, even if awkward to implement.

I still need to verify\test interaction among _SpidL1_ SSO session and _SpidL2>_ "temporary auth session". In my current implementation, using the feature in this PR, i can avoid SP tracking and SessionIndex creation for _SpidL2>_, but it still create a real session _IdP_ side, that could potentially lead to a successful SSO with a subsequent _SpidL1_. But at the and I think this can be manage within the _Authentication Module_.

jaimeperez:
I somehow expected this being related to SPID. Lots of weird requirements coming from them lately 😞 

Do you have a link where I can read about this policy? It's fine if it's only in Italian, I can probably make some sense out of it...

In any case, I definitely think this belongs in the authentication source. We have something similar in Feide, where we use the current SimpleSAMLphp codebase plus a custom auth source. There, we evaluate if a user with an existing session needs to be authenticated with two-factor again (in the `reauthenticate()` method) and add a `LoginCompletedHandler` that updates the session and runs the `ReturnCallback` set by the IdP.

This same mechanism can be used for your use case, making sure that if _SpidL2_ is required, the user is reauthenticated properly and SSO doesn't happen.

Regarding the `SessionIndex`, that's going to be more problematic, as there's nothing you can do about it yourself and it would require changes to the codebase. That's why I was asking for the specs, to try to understand if they do require that for real, and why. In any case, I don't really know how we should implement that properly, and I guess I need to think about it a little bit...

tbenr:
Thanks for the useful info, I'll go in deep with `LoginCompletedHandler` approach and how can help me.

I agree with you about SPID requirements oddity...

Anyway here is some useful links:

- [General technical rules](http://www.agid.gov.it/sites/default/files/circolari/spid-regole_tecniche_v1.pdf): SAML2 specs "SPID" style :-)
- [Technical rules errata corrige](http://www.agid.gov.it/sites/default/files/documentazione/spid-avviso-n5-regole-tecniche-errata-corrige.pdf)
- [Session and SLO requirements](http://www.agid.gov.it/sites/default/files/documentazione/spid-avviso-n3-chiarimenti-sessioni-singlelogout.pdf): see first paragraph on page 11 for `SessionIndex` restrictions.

jaimeperez:
Thanks for the pointers!

I don't know Italian, but could it be that this is specifying a SHOULD instead of a MUST?

> Tale elemento non dovrà essere presente nel caso di
asserzioni emesse a seguito di richieste di autenticazione per i livelli SPID 2 e SPID 3.

If it can be taken as a SHOULD, then you would be fine to ignore that requirement. In the end, I understand they intended this to prevent service providers from starting single logout because no session has been kept, due to the step-up level of authentication used. However, in that case, I don't see a reason why this should be enforced (as in, MUST), as that can be implicit for SPs joining SPID, and particular implementations not following that requirement might ignore logout messages anyway (like in this case).

Maybe we should find some authoritative voice that can resolve our doubts before we try to implement this?

tbenr:
Unfortunately in italian that means MUST. I'm actually quite into SPID stuff (I already implemented an IDP with other solutions) and I agree with you that in this case there is no real benefit, apart of annoying implementers :) There are many little pitfalls that force implementers to drift from a standard de-facto SAML2 using opensource\vendor software to a customized solution. I heard such oddity could be blunted in the future but they are still officially in place.

PR #657, opened 2017-07-20
Invalidate opcache after writing a file

seggen-ibuildings:
When opcache.validate_timestamps is disabled, then the new metadata will not be read after a metarefresh. This can be solved by adding the metadata file to an opcache blacklist, but calling opcache_invalidate() after writing a file is a nice out-of-the-box solution.

Hopefully, this will enable everybody that is using simplesamlphp to disable opcache.validate_timestamps without running into problems.


thijskh:
Yes, seems like a nice addition to me! @jaimeperez agreed?

jaimeperez:
Sounds safe to me! 👍

seggen-ibuildings:
thanks!

PR #658, opened 2017-07-25
Skip logging statistics when we have a passive request

ghalse:
At the moment, the consentAdmin module generates a lot of spurious saml20-idp-SSO logs as it passes passively through the authproc filters to determine its attribute set. This has the potential to badly skew any statistics of login attempts, such as those generated by the statistics module. Because the passive state is not logged, there's no way to differentiate those generated passively from a genuine login.

This patch causes the StatisticsWithAttribute filter to be skipped when we have a passive request.

### DIFF
--- a/modules/core/docs/authproc_statisticswithattribute.md
+++ b/modules/core/docs/authproc_statisticswithattribute.md
@@ -12,6 +12,9 @@ Parameters
 `type`
 :   The type of the statistics entry.
 
+`skipPassive`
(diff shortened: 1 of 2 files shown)


thijskh:
Of course it's a bit subjective whether or not one should count isPassive requests. Wouldn't it be a better solution to actually log passive requests separately so one can choose to count or ignore them?

jaimeperez:
Hi!

I agree with @thijskh here, I think it's better to log them separately, as that could also be valuable information (i.e. finding misbehaving SPs).

ghalse:
Based on that feedback, I've made it more flexible - and the default is to log them separately.

PR #668, opened 2017-08-07
Consent: PSR-2 / cleanup / logic away from templates

tvdijen:
- Minor improvements / cleanup
- Set httponly on consent-cookies by default
- PSR-2
- Move logic away from templates
- Removed an old translations-file that is not being used anymore


tvdijen:
Now if someone could help me to get rid of those merge commits, I'd be grateful!

PR #669, opened 2017-08-08
Consent: PSR-2 / cleanup / logic away from templates

tvdijen:
Minor improvements / cleanup
Set httponly on consent-cookies by default
PSR-2 / white-space
Move logic away from templates
Removed an old translations-file that is not being used anymore


jaimeperez:
Thanks a lot for this @tvdijen!

I assume you have tested and everything works exactly as before, right?

modules/consent/www/getconsent.php
```diff
@@ -139,16 +183,101 @@
...
+    foreach ($attributes as $name => $value) {
+        $nameraw = $name;
+        $name = $translator->getAttributeTranslation($nameraw);
+
+        if (preg_match('/^child_/', $nameraw)) {
+            // insert child table
+            $parentName = preg_replace('/^child_/', '', $nameraw);
+            foreach ($value as $child) {
+                $str .= "\n" . '<tr class="odd"><td class="td_odd">' .
+                    present_attributes($t, $child, $parentName) . '</td></tr>';
```
jaimeperez: This looks like a bug. The signature of `present_attributes()` has changed, but here we are still calling it with the old signature.

Why did the signature change, by the way?

modules/consent/www/getconsent.php
```diff
@@ -139,16 +183,101 @@
...
 $t->show();
+
+/**
+ * Recursive attribute array listing function
+ *
+ * @param SimpleSAML_XHTML_Template $t          Template object
+ *
+ * @return string HTML representation of the attributes
+ */
+function present_attributes($t)
```
jaimeperez: We already have this function defined in `templates/includes/attributes.php`. Wouldn't it be better to just use that one?

modules/consent/www/getconsent.php
tvdijen: I removed the 2nd and 3rd argument because (i thought) they were never used... Obviously I've missed something 💩 and my tests missed the nested attribute-table case...
I'll revert this and see if we can use the function from `templates/includes/attributes.php`

tvdijen:
I have fixed the build again... Everything properly works.

Regarding the `present_attributes()` functions; they are different in both behaviour and in output.
The one in the module handles 'hidden attributes', the base-version has logics for NameIds
If we want to use the base-version, we have to put some logic in it specific to the consent-module...
Let me know what you want..

thijskh:
Will this not create conflicts with the Xnewui branch? Maybe apply the work there instead of on master, or at least the parts relevant to templating?

tvdijen:
I don't think any work was done on the consent-module yesterday...
Merging this to both master / Xnew-ui will be really helpful for the person who's gonna do the twiggification of this module.
The only issue @jaimeperez had with this PR is that we apparently have two different versions of `present_attributes()`, but perhaps we should worry about that later and take advantage of the work already done before it becomes unmergeable due to conflicts.

tvdijen:
Okay, I have commited this to the Xnew-ui branch, so I'm closing this PR.

PR #672, opened 2017-08-09
Fix XML validation failure with leading/trailing whitespace

dnmvisser:
Converting XML metadata to PHP on `/admin/metadata-converter.php` will fail when the XML contains leading whitespace.
Annoying some browsers seem to add an empty line when copy/pasting, so this error occurs regularly.
Trimming the xmldata fixes this.


jaimeperez:
Thanks Dick!

PR #674, opened 2017-08-11
Update PHP API change gor ldapwhoami

manu0401:
I have been maintaining the PHP LDAP EXOP patch for a few years,
which include the ldapwhoami() function. This has finally made its
way into PHP distribution and will be available in PHP 7.3, but
with a modified prototype.

This changes adapts to this API change. While there, also update
exception handling on par with recent SimpleSAMLphp code

### DIFF
--- a/lib/SimpleSAML/Auth/LDAP.php
+++ b/lib/SimpleSAML/Auth/LDAP.php
@@ -733,8 +733,8 @@ class SimpleSAML_Auth_LDAP
         $authz_id = '';
 
         if (function_exists('ldap_exop_whoami')) {
-            if (ldap_exop_whoami($this->ldap, $authz_id) !== true) {
-                throw $this->getLDAPException('LDAP whoami exop failure');
+            if (($authz_id = ldap_exop_whoami($this->ldap)) === false) {
+                throw $this->makeException('LDAP whoami exop failure');
             }
         } else {
             $authz_id = $this->authz_id;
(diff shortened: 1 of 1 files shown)


thijskh:
Do I misunderstand this change, or will this only work with the new prototype in PHP 7.3? In that case we cannot really merge it like this, it should be made backwards compatible.

manu0401:
I agree backward compatibility is highly desirable, but here the function name is the same and the prototype differs. If we want to add a version tests, all we have is the PHP version.

I can resubmit with a test on (explode(".", phpversion())[0] >= 7) if you consider that acceptable.

tvdijen:
`if (version_compare(phpversion(), '7.3', '>='))`

manu0401:
Even better because it spares function calls: 
if (defined("PHP_MAJOR_VERSION") && (PHP_MAJOR_VERSION >= 7))

thijskh:
I prefer @tvdijen's solution, as that's what's used in other parts of the code.

manu0401:
This is a PHP 7 version guard actually, since the unofficial patch I have being maintaining switched to the PHP repository prototype for PHP 7.0 and 7.1 (I have not yet done 7.2).

thijskh:
Great, one final change: please avoid tabs and use spaces for indent as per `CONTRIBUTING.md`. Thanks.

PR #675, opened 2017-08-13
ADFS-module: Set enctype on post-form

tvdijen:
Without proper enctype, every space in $wresult will be converted to a +
I really wonder if this ever worked at all..

no_enctype


jaimeperez:
Thanks Tim!

thijskh:
This has been reverted in  #481

PR #678, opened 2017-08-18
Flexible namespace

tvdijen:
It doesn't make much sense to have a hard coded AttributeNamespace used for every attribute...
It is a mandatory field in SAML1.1, so we keep the hard coded value as a fallback.

This will convert an attribute called `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddres` to:
`<saml:Attribute AttributeNamespace="http://schemas.xmlsoap.org/ws/2005/05/identity/claims" AttributeName="emailaddress">`


tvdijen:
While I'm at it, I'll see if I can test #481 so we can finally close that one

tvdijen:
Something went wrong here while resolving conflicts due to the merge of #481 ..
Couldn't reopen this, so I had to open #680

PR #679, opened 2017-08-18
Remove obsolete code

tvdijen:
Both methods create the exact same signed document.


thijskh:
I think @jaimeperez created that for a reason, and I don't see a relevant change in xmlseclibs upstream since that was committed. So I'm wondering if this is right to merge now.

tvdijen:
Looks like it's only doing some trimming on the <signature> block, but even that's not doing anything at all for the generated metadata.

jaimeperez:
Hi guys!

TBH, I don't have any idea why this is in place 😅 

I've been looking back at the commits where that file was added and modified, and I can only see that it was created back in the time when we were using Google Code. Given that the author of the class was @dtsosie, I'm pretty sure this was a contribution based on some issue that I can't find right now. I seem to vaguely recall some issue with spacing, but I think the best we can do here is to ask Daniel.

tvdijen:
I think it's an attempt to remove formatting from ...
https://github.com/robrichards/xmlseclibs/blob/master/src/XMLSecurityDSig.php#L64-L68
... which I can relate to, because imo it shouldn't be formatted there at all...

There may be applications that choke on a formatted XML, but it doesn't make sense to do this for ADFS-metadata only then... Either way, this belong inside the adfs-module.

dtsosie:
Just saw this so sorry. But ADFS does really care if something is white space or not. It is dumb, and not to spec, but the fact of the matter this change had to be made to allow us to import the meta properly. If you do any reversions, make sure you try to do a test on a real ADFS system on a Microsoft Box to ensure compatibility, otherwise you will no longer be ADFS compatible.

tvdijen:
I can import metadata on an Win2012R2 ADFS box with or without this code.. It doesn't make a difference.

jaimeperez:
Hi guys,

I really don't have a way to test this myself, other than asking others to do it for me. @tvdijen I assume you have done the test to sign some metadata and try to import on an ADFS, right? @dtsosie did you have a chance to test Tim's changes? Since you contributed this fix, it would be really useful if you could test whether Tim's change works or not.

tvdijen:
Could it be, that this was written for ADFS 2008, @dtsosie? Even Microsoft tends to fix things sometimes :see_no_evil:

tvdijen:
I've tried an 2008 R2 box and successfully imported metadata from `/module.php/adfs/idp/metadata.php`...

dtsosie:
My test was using the visual studio metadata import. I'm no longer
enrolled, but if you have it attempt to use a template project that imports
the data and see if it works.

I suspect the visual studio was more restrictive and I worked hard to make
it not that case because I wanted this project to work in our company.

This is just a whitespace issue, but for auto import, visual studio cared.

tvdijen:
I'm afraid that goes beyond me...

jaimeperez:
It goes way beyond me too...

@dtsosie how is this supposed to work? This is the first time I ever hear about Visual Studio consuming SAML metadata. What's the purpose?

If this is such a corner case as it seems, I'm quite tempted to merge this as it'll make the code more clear and manageable. If Visual Studio needs to consume SAML metadata and still has problems doing so if the XML is formatted with spaces, this is probably something that should be fixed in Visual Studio...

Anyway, I would like to know more about this and have a complete picture of what's going on here before making up my mind.

dtsosie:
To assist in automatic creation of a ws federation based on adfs, there was
a "central auth" import template for a new target with the visual studio
tool.

It would configure the whole app from scratch.

All the work done for this use case was indeed unwarranted. Made for an
awesome demo, but the company already had apps that couldn't be auto
configured.

It was useful though since developers liked seeing something working so
easily before diving into my saml pool.

It would be easy to work around, and new vs might properly interpret xml
now.

I'd probably just make a converter and stuff the metadata into a www
directory today. But at that time I was doing a lot of ssp development and
I wanted to play.

Hope this was helpful.

-Dan Tsosie

thijskh:
Thanks! I'd conclude that we can just remove this code then. I'd not do that right before 1.15 since it's just cleanup, let's do that after so the change has some more time in master just in case.

thijskh:
Merged in 1f4b6e8489d8bf1a8c099574ce133ee84f7cc3ed

PR #680, opened 2017-08-19
Flexible namespace

tvdijen:
It doesn't make much sense to have a hard coded AttributeNamespace used for every attribute...
It is a mandatory field in SAML1.1, so we keep the hard coded value as a fallback.

This will convert an attribute called http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress to:
<saml:Attribute AttributeNamespace="http://schemas.xmlsoap.org/ws/2005/05/identity/claims" AttributeName="emailaddress">

### DIFF
--- a/lib/SimpleSAML/Utils/Attributes.php
+++ b/lib/SimpleSAML/Utils/Attributes.php
@@ -106,4 +106,29 @@ class Attributes
 
         return $newAttrs;
     }
+
+
+    /**
+     * Extract an attribute's namespace, or revert to default.
+     *
+     * This function takes in a namespaced attribute name at splits it in a namespace/attribute name tuple.
+     * When no namespace is found in the attribute name, it will be namespaced with the default namespace.
+     * This default namespace can be overriden by supplying a second parameter to this function.
+     *
+     * @param string $name The namespaced attribute name.
+     * @param string $namespace The default namespace that should be used when no namespace is found (optional).
+     *
+     * @return array The attribute name, split to the namespace and the actual attribute name.
+     */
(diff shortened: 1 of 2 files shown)


jaimeperez:
Thanks Tim!

Would it make sense to move this code to its own function somewhere (something like `getAttributeNamespace()` that takes an attribute name as input and returns a tuple with the namespace -with the current default when no namespace is part of the name- and the attribute name itself)? I'm thinking that by doing that, we can add tests for the function and ensure that everything works fine.

tvdijen:
Sure, I can do that! But where would you put the function, because I don't think it should be a function of the ADFS-class then... Somewhere within the Utils namespace perhaps?

jaimeperez:
That's a very good question indeed. Maybe `SimpleSAML\Utils\Attributes`? If the default is passed as a second parameter, it can then be reused elsewhere, and wouldn't be ADFS-specific...

jaimeperez:
Thanks a lot Tim! 😄

tvdijen:
Ohh, you were quick!  I was about to add a test for it.
Now I have no experience with writing tests... Would this cover it?
https://github.com/tvdijen/simplesamlphp/commit/d3a907670cd9d837fc7ba7b98f6c3dc8ac9f00a3

jaimeperez:
Ops, sorry! Didn't know you were also working on a test, and I was too quick indeed.

Anyway, I made a test myself, and also changed the second parameter to be mandatory, so that the function is not ADFS-specific. Take a look at ec379ac30543c59e64e435c815435d6a2d692ead and tell me if you see anything wrong.

PR #681, opened 2017-08-21
Remove long-ago deprecated code

tvdijen:
I can trace deprecation back to 2009 (https://github.com/simplesamlphp/simplesamlphp/commit/1dc4bc4654b655360fee11789450fdcbb73cb5a1#diff-5b7745b258a8f6625d4dc1b014cadd37) and I can't find any code that relies on it...
I think it's safe to remove!

### DIFF
--- a/lib/SimpleSAML/Configuration.php
+++ b/lib/SimpleSAML/Configuration.php
@@ -288,95 +288,17 @@ class SimpleSAML_Configuration
      * If no configuration file with the given instance name is found, an exception will
      * be thrown.
      *
-     * @param string $instancename The instance name of the configuration file. Deprecated.
-     *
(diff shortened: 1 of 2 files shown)


jaimeperez:
Hi @tvdijen!

Many of our tests make indirect use of that functionality, as you have already seen by the failed build. Currently, the way we allow testing without a configuration file is by loading the desired configuration with the default _instance_ name, using `loadFromArray()`. Later on, when the code being tested calls `getInstance()`, the instance with the default name is loaded, and that way we avoid checking for an existing configuration file.

So what we can do is to remove the `$instance` parameter in `getInstance()`, but the property must remain as it is still heavily used. Still, I would do that for 2.0, given that's a change in the API.

PR #682, opened 2017-08-22
Remove PHP 5.3 support from composer configuration

mschwager:
Looks like PHP 5.3 support was finally dropped! However it also looks like a spot was missed in the update :)


thijskh:
Thanks!

PR #683, opened 2017-08-22
SAML2-SP: Making the AuthnInstant available in the state array

jtgasper3:
The authnInstant is not being set in the saml2-acs.php. Therefore it isn't available for review or usage in auth proc filters, etc.

### DIFF
--- a/modules/saml/www/sp/saml2-acs.php
+++ b/modules/saml/www/sp/saml2-acs.php
@@ -252,5 +252,7 @@ if (isset($state['SimpleSAML_Auth_Source.ReturnURL'])) {
 }
 $state['PersistentAuthData'][] = 'saml:sp:prevAuth';
 
+$state['AuthnInstant'] = $assertion->getAuthnInstant();
+
 $source->handleResponse($state, $idp, $attributes);
 assert('FALSE');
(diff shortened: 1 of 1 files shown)


jtgasper3:
Hi all,

Just checking to see if anything else is needed for this to be merged into baseline?

Thanks

modules/saml/www/sp/saml2-acs.php
```diff
@@ -252,5 +252,7 @@
 }
 $state['PersistentAuthData'][] = 'saml:sp:prevAuth';
 
+$state['AuthnInstant'] = $assertion->getAuthnInstant();
```
jaimeperez: Since this is something completely specific to SAML authentication, I think it would make sense to name it `saml:AuthnInstant`, following the same convention we use elsewhere. I also think it should be set before (around line 234), so that it is co-located with other related data.

Finally, you should also consider whether it would be useful for you to have this data later, after authentication happens (and all authproc filters are run). If you think it would, then you would need to consolidate it into the authentication data saved in the session, by adding it to `PersistentAuthData`:

```php
$state['PersistentAuthData'][] = 'saml:AuthnInstant';
```

jtgasper3: Hi @jaimeperez, 
Thanks for the notes. I'll get the requested changes made. 

As for my needs, accessing it via an authproc filter is fine, but in the hopes of flexibility for others, I think adding it to `PersistentAuthData` is probably a good idea.

Thanks.

jtgasper3:
Changes made. Hopefully, they are as you suggested. :D

jaimeperez:
Looks great! Thanks a lot @jtgasper3!

Issue #595, opened 2017-04-07
Error report emails not delivered because of wrong from address


kjkooistra-youwe:
https://github.com/simplesamlphp/simplesamlphp/blob/master/www/errorreport.php
Lines: 104 and 107

The from address is set to either the email address filled in in the report form or a default address. However these emails are blocked because of SPF records. The from address should be explicitly set to an email address matching the domain. Overriding it to a hard coded value with matching domain solved the issue for me on a production environment.

Possible solution: Introduce a configuration option to set the value.

Relevant information:
http://www.html-form-guide.com/email-form/php-script-not-sending-email.html
http://stackoverflow.com/questions/27198715/spf-issue-when-sending-automatic-email-with-php#answer-27198758

thijskh:
Yes, would be a nice improvement. Could you provide a PR?

kjkooistra-youwe:
Hi. I have no idea how to create a new bugfix branch here, so I've added a patch file instead (git diff).
[patch.patch.zip](https://github.com/simplesamlphp/simplesamlphp/files/909493/patch.patch.zip)

tvdijen:
Would it help if I make a PR from the patch-file so we can close this issue?

thijskh:
Yes, definitely.

tvdijen:
There you go

PR #684, opened 2017-08-23
Add sendmail_from option

tvdijen:
PR for #595

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -55,6 +55,11 @@ $config = array(
     'technicalcontact_email' => 'na@example.org',
 
     /*
+     * Sender email should use the same domain as your application domain to prevent SPF record filtering
+     */
+    //'sendmail_from' => 'no-reply@example.org',
+
+    /*
      * The timezone of the server. This option should be set to the timezone you want
      * SimpleSAMLphp to report the time in. The default is to guess the timezone based
      * on your system timezone.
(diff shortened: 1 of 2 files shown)


tvdijen:
Let me know what you think!

www/errorreport.php
```diff
@@ -101,10 +101,21 @@
...
 } else {
     $replyto = null;
-    $from = 'no-reply@simplesamlphp.org';
+}
+
+$from = $config->getString('sendmail_from', null);
+if ($from === null || $from === '') {
+    $from = ini_get('sendmail_from');
+    if ($from === '' || $from === false) {
+        $from = 'no-reply@example.org';
```
thijskh: Lines 110-113 could perhaps be consolidated into:
```php
$from = ini_get('sendmail_from', 'no-reply@example.org');
```
?

config-templates/config.php
```diff
@@ -55,6 +55,11 @@
     'technicalcontact_email' => 'na@example.org',
 
     /*
+     * Sender email should use the same domain as your application domain to prevent SPF record filtering
```
thijskh: Probably this should read something like "The envelope from address for outgoing emails. This should be in a domain that has your application's IP addresses in its SPF record to prevent it from being rejected by mail filters."

www/errorreport.php
tvdijen: Nope, not unless `ini_get()` has an undocumented 2nd parameter...

thijskh: Right, I was confused with our own config get functions...

PR #685, opened 2017-08-25
Allow CLI invocation of Cron

pradtke:
Make it easier to run SSP cron functionality from CLI.
This allows setting memory_limit and max_execution_time values that are appropriate for cron functionality that you don't wanted enabled for the all requests.

For tests, I'm currently blocked on getting module enabled during testing. See discussion https://groups.google.com/forum/#!topic/simplesamlphp-dev/hqh7VlSSzhw

### DIFF
--- a/modules/cron/bin/cron.php
+++ b/modules/cron/bin/cron.php
@@ -33,7 +33,7 @@ if (!array_key_exists('t', $options)) {
 }
 
 $tag = $options['t'];
-$cron = new sspmod_cron_Cron();
+$cron = new SimpleSAML\Module\cron\Cron();
 if (!$cron->isValidTag($tag)) {
     echo "Invalid tag option '$tag'.\n";
     exit(2);
--- a/modules/cron/bin/cron.php
+++ b/modules/cron/bin/cron.php
@@ -1,0 +1,45 @@
+#!/usr/bin/env php
+<?php
+
+/*
+ * This script can be used to invoke cron jobs from cli.
+ * You most likely want to execute as the user running your webserver.
+ * Example:  su -s "/bin/sh" -c "php /var/simplesamlphp/modules/cron/bin/cron.php -t hourly" apache
+ */
+
+// This is the base directory of the SimpleSAMLphp installation
+$baseDir = dirname(dirname(dirname(dirname(__FILE__))));
+
+// Add library autoloader.
+require_once($baseDir . '/lib/_autoload.php');
+
+if (!SimpleSAML\Module::isModuleEnabled('cron')) {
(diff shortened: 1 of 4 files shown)


modules/cron/bin/cron.php
```diff
@@ -0,0 +1,45 @@
...
+
+if (!SimpleSAML\Module::isModuleEnabled('cron')) {
+    echo("You need to enable the cron module before this script can be used.\n");
+    echo("You can enable it by running the following command:\n");
+    echo('  echo >"' . $baseDir . '/modules/cron/enable' . "\"\n");
+    exit(1);
+}
+
+$options = getopt("t:");
+if (posix_getuid() === 0) {
```
thijskh: Probably add an `function_exists` there because `posix_getuid` is not always available (requires posix module, not on Windows).

pradtke: thanks for the tip. I've added the check.

pradtke:
@thijskh I've updated the documentation with example invocation and with some advice/warnings on file permissions and logging.

I've also incorporated the function_exists() check.

modules/cron/docs/cron.md
```diff
@@ -80,3 +88,35 @@ follow the appropriate links to execute the cron jobs you want. The
...
+same user as the web server.  Several cron hooks created files and
+those files may have the wrong permissions if you run the job as root.
+
+**note:** Logging behavior in SSP when running from CLI varies by
+version. The latest version logs to PHP's error log and ignores any
+logging configuration from `config.php`
+
+Below is an example of invoking the script. It will:
+
+* Run a command as the apache user
```
thijskh: Might be a nit, but I'd write this as `apache` to make it abundantly clear that this is actually the user named 'apache'.

pradtke: Make sense. I've added the `code` tags.

modules/cron/lib/Cron.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+/**
+ * Handles interactions with SSP's cron system/hooks.
+ */
+class sspmod_cron_Cron
```
jaimeperez: Since this is a new class (and the only one in this module), I would actually prefer it to use the PSR-4-compliant naming convention. That would mean naming the class `Cron` inside the `SimpleSAML\Module\cron` namespace.

jaimeperez:
I've added a minor comment on how to name the new Cron class. Sorry for the late reaction from my side!

pradtke:
@jaimeperez no worries. I added in the namespace.

jaimeperez:
Merged! Thanks a lot for your work Patrick, and sorry so much for the delay on my side.

PR #686, opened 2017-08-28
Document cron module

pradtke:
@thijskh I opted to do the cron documentation as a separate PR and then I can update PR-685 to document the new features once this is merge.

I was unsure how to build the documentation to verify that the links from `automated metadata` to `cron` were correct.

### DIFF
--- a/docs/simplesamlphp-automated_metadata.md
+++ b/docs/simplesamlphp-automated_metadata.md
@@ -26,7 +26,7 @@ Preparations
 
 You need to enable the following modules:
 
- 1. cron
+ 1. [cron](./cron:cron)
  2. metarefresh
 
 The cron module allows you to do tasks regularly, by setting up a cron job that calls a hook in SimpleSAMLphp.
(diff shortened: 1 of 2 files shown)


thijskh:
Yes, this looks fine. Also not sure about the links, but looking at other places in the docs, they do the same. @jaimeperez probably knows how the documentation is built so those links resolve.

jaimeperez:
The links are relative from the current path, and for documentation in modules, the last part of the URL should be in the form `<name of the module>:<name of the file>`, so `cron:cron` would translate to a document called `cron.md` located in `modules/cron/docs/`.

thijskh:
Alright, seems fine then!

pradtke:
@thijskh I did a quick grep of the existing convention using `grep -r '](.*:.*)' docs/ | grep -v 'http'`
The vast majority (all but one link) of the markdown links from `docs` into modules start with `./`.

Should I still remove the `./`? or should I keep it to match existing docs?

jaimeperez:
Actually, without the `./` in front, the links won't work. Haven't tested with other browsers than Chrome, but at least with that one, if the `./` is missing at the front of the link, the browser won't follow it.

I've just fixed a couple broken links, in fact.

Issue #687, opened 2017-09-03
sspmod_saml_Message incorrect functionality (or documentation?)


dub357:
According to the 1.14 docs for saml2-idp-remote, both the "sign.authnrequest" and "sign.logout" properties all state: 
"Note that this option also exists in the SP configuration. This value in the IdP remote metadata overrides the value in the SP configuration."
Also in the docs for saml.SP, those same properties state: 
"Note that this option also exists in the IdP-remote metadata, and any value in the IdP-remote metadata overrides the one configured in the SP configuration."

Turns out that this is incorrect.  In sspmod_saml_Message (simplesamlphp/modules/saml2/lib/Message.php) the "addRedirectSign" method first checks the SP config ($srcMetadata) and only checks the IdP config ($dstMetadata) if null was received from the SP config.  Therefore, the value in the remote IdP metadata NEVER overrides the value in the SP config.

Please don't simply change the documentation around :)  Its very helpful to have the Idp metadata be the final say if authn/logout requests are signed.

Note that the redirect.sign option is currently handled correctly, only sign.logout and sign.authnrequest are not.

dub357:
FYI - I "fixed" this locally with the following code in sspmod_SAML_Message.addRedirectSign

if ($message instanceof SAML2_LogoutRequest || $message instanceof SAML2_LogoutResponse) {
			//$signingEnabled = $srcMetadata->getBoolean('sign.logout', NULL);
			$signingEnabled = $dstMetadata->getBoolean('sign.logout', NULL);
			if ($signingEnabled === NULL) {
				//$signingEnabled = $dstMetadata->getBoolean('sign.logout', NULL);
				$signingEnabled = $srcMetadata->getBoolean('sign.logout', NULL);
			}
		} elseif ($message instanceof SAML2_AuthnRequest) {
			//$signingEnabled = $srcMetadata->getBoolean('sign.authnrequest', NULL);
			$signingEnabled = $dstMetadata->getBoolean('sign.authnrequest', NULL);
			if ($signingEnabled === NULL) {
				//$signingEnabled = $dstMetadata->getBoolean('sign.authnrequest', NULL);
				$signingEnabled = $srcMetadata->getBoolean('sign.authnrequest', NULL);
			}
		}

dub357:
pull request created:
https://github.com/simplesamlphp/simplesamlphp/pull/691

jaimeperez:
[...]

PR #691, opened 2017-09-11
sspmod_saml_Message: Fix for conditional message signing

dub357:
Check destination metadata before source metadata when determining whether or not to sign message.  This fix is to bring code in line with documentation and allows different IdP's to have different requirements about message signing.

Fix for issue:
https://github.com/simplesamlphp/simplesamlphp/issues/687


dub357:
I'm going to close this PR...I need to re-do it.  After I had created it I had pushed more commits to my fork that were unrelated to this PR and it seems that GitHub automatically adds them to the open PR which I did not want.

PR #693, opened 2017-09-19
Show AuthData on authentication status page

pradtke:
It can be useful to see the AuthData when testing an authentication source.

Since the AuthData can be a hierarchy of data of various depths it is converted
into a pretty printed JSON structure and put inside &lt;pre>

Since the amount of AuthData can vary greatly, the data is displayed in side
a &lt;details> element which requires clicking to view.

### DIFF
--- a/dictionaries/status.definition.json
+++ b/dictionaries/status.definition.json
@@ -38,10 +38,10 @@
 	"logout": {
 		"en": "Logout"
 	},
-        "authData_header": {
-                "en": "AuthData"
-        },
-        "authData_summary": {
-                "en": "Click to view AuthData"
-        }
+	"authData_header": {
+		"en": "AuthData"
+	},
+	"authData_summary": {
+		"en": "Click to view AuthData"
+	}
 }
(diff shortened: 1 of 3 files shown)


dictionaries/status.definition.json
```diff
@@ -37,5 +37,11 @@
...
 	"logout": {
 		"en": "Logout"
-	}
+	},
+        "authData_header": {
+                "en": "AuthData"
+        },
+        "authData_summary": {
+                "en": "Click to view AuthData"
+        }
```
thijskh: Indent seems to be off here

thijskh: Indent seems to be off here

templates/status.php
```diff
@@ -60,6 +60,13 @@
     echo(present_attributes($this, $list, ''));
 }
 
+$authData = $this->data['authData'];
+if (isset($authData)) {
```
thijskh: This will always be set because it's set right above. Maybe intended was `!empty()`?

PR #697, opened 2017-10-04
Fix duplicate assertion error using Redis Store

matt-catalyst:
Fixes the following error when using the Redis Store:

```SimpleSAML_Error_Error: UNHANDLEDEXCEPTION
Backtrace:
1 www/_include.php:45 (SimpleSAML_exception_handler)
0 [builtin] (N/A)
Caused by: SimpleSAML_Error_Exception: Received duplicate assertion.
Backtrace:
1 modules/saml/www/sp/saml2-acs.php:150 (require)
0 www/module.php:135 (N/A)
```

### DIFF
--- a/lib/SimpleSAML/Store/Redis.php
+++ b/lib/SimpleSAML/Store/Redis.php
@@ -70,7 +70,7 @@ class Redis extends Store
 
         $result = $this->redis->get("{$type}.{$key}");
 
-        if ($result === false) {
+        if ($result === false || $result === null) {
             return null;
         }
 
--- a/tests/lib/SimpleSAML/Store/RedisTest.php
+++ b/tests/lib/SimpleSAML/Store/RedisTest.php
@@ -48,7 +48,7 @@ class RedisTest extends \PHPUnit_Framework_TestCase
 
     public function getMocked($key)
     {
-        return array_key_exists($key, $this->config) ? $this->config[$key] : false;
+        return array_key_exists($key, $this->config) ? $this->config[$key] : null;
     }
 
     public function setMocked($key, $value)


mschwager:
Are you sure this is because of the Redis Store? I don't see anything suggesting that in the backtrace. I'm having a hard time digging up what `$this->redis->get` returns on error, `false` or `null`. It seems like it shouldn't return either on error, but rather one or the other consistently. Thoughts?

tvdijen:
It should return false on error and null on 'key not found'..

Regarding the commit:
Bear in mind that `return unserialize(null);` would return `false` and that other code may actually depend on this behaviour...

mschwager:
Ahh, thanks for the context. This PR makes sense to me, then, as I believe that's the way this function was originally intended to behave:

> @return mixed|null The value associated with that key, or null if there's no such key.

As for backwards compatibility, ugh, we've put ourselves in the unfortunate situation of reversing the error case return values. On error this function returns `null` and when the key isn't found it returns `false` because of `unserialize`!

Another possible solution would be to update the documentation to include the case where `false` is returned. However, I think I'd prefer to accept this PR at the cost of backwards incompatibility. That way we maintain the interface contract in the `Store` class in `lib/SimpleSAML/Store.php` of `@return mixed|null The value.` Changing all implementations to include the possibility of returning `false` seems like a much more heavy-handed solution with more backwards incompatible implications.

Thoughts?

tvdijen:
I think you're mixing it up...

Old situation:
On error: null (because of if-statement)
On key n/f: false (because of unserialize)

New situation:
Both cases would return null

Since support for redis is relatively new, I think it's worth checking out if any code relies on `:get()` returning false and if we need to change any tests.

matt-catalyst:
I spent some time looking at the other store types (memcache, sql) and they do not appear to return false, so if the Redis plugin was to return false it would be inconsistent with the other types.

mschwager:
Great, thanks @matt-catalyst! One suggestion I have, though, would be to make `getMocked` in `RedisTest` look like this instead:

```php
public function getMocked($key)
{
    return array_key_exists($key, $this->config) ? $this->config[$key] : null;
}
```

I think this is more true to the actual API, where `get` should return `null` when the key doesn't exist, and `false` on error. Testing both the `null` and `false` error cases seems like more trouble than it's worth for now.

matt-catalyst:
Thanks for the suggestion @mschwager - I have updated the pull request.

mschwager:
This LGTM! Can we pull this in @tvdijen?

tvdijen:
Fine by me!

mschwager:
Ahh, I was under the impression that @tvdijen had committer's access, but maybe that's not the case? If not, can we pull this in @jaimeperez or @thijskh?

jaimeperez:
Thanks a lot for reporting and fixing this, @matt-catalyst!

As for the lack of write access of @tvdijen, I've just fixed that too 😉

PR #701, opened 2017-10-12
add redis sentinel support

bynare:


thijskh:
Well, the code does extend the Redis class so code duplication is largely avoided.

Still indeed, also from a user standpoint, is it necessary to make redissentinal a separate store type? Can't we enhance the redis type with some additional configuration? What do you think @bynare?

bynare:
@thijskh i kept it separate so we don't have to change config behaviour for current users/versions, once they update. If we don't re-use existing config values, we can probably use the same store type for both redis and sentinel, e.g adding store.redis.sentinelmastergroup, store.redis.sentinelhosts (the prefix config can be shared) ? Then if the current host/port config is empty, redis sentinel will be initialised, if it has config set.

Let me know what you think.

mschwager:
I agree with much of the sentiment already expressed here. It makes sense to me to re-use the existing Redis store type, and do some intelligent parsing of the config options to determine which mode Redis should be in, i.e. Sentinel or not.

> ... we can probably use the same store type for both redis and sentinel, e.g adding store.redis.sentinelmastergroup, store.redis.sentinelhosts (the prefix config can be shared) ? Then if the current host/port config is empty, redis sentinel will be initialised, if it has config set.

This sounds good to me!

tvdijen:
I have an implementation ready [here](https://github.com/simplesamlphp/simplesamlphp/commit/e57536d311330a781d6a1ad8d7c1f2fe60fbc35e)

The way it works now is that as long as `store.redis.sentinelhosts` is an empty list, the host/port-settings are being used.. When sentinelhosts is set, host/port settings are ignored.

Now, since I don't have a setup available to test this, could you please run a few tests with this code @bynare?

jaimeperez:
Since we don't have any Redis instance available, it's difficult for us to test this. @lonoak, would you mind using the implementation suggested by Tim (available [here](https://github.com/simplesamlphp/simplesamlphp/commit/e57536d311330a781d6a1ad8d7c1f2fe60fbc35e)) and testing it with your Redis instance?

PR #708, opened 2017-10-19
Deprecate assert with string

### LINKED ISSUE #573: assert() with string as the assertion is DEPRECATED as of PHP 7.2
Just a heads up. From http://php.net/manual/en/function.assert.php 

Using string as the assertion is DEPRECATED as of PHP 7.2.

SSP is littered with assertions taking a string argument.

fentie:
Fixes #573

Removed string-based usage of assert. Where applicable in the assert calls, I also replaced `is_null` with `=== null`, double quotes to single quotes (to match style elsewhere) and lowercased PHP constants (false/null)

### DIFF
--- a/lib/SimpleSAML/Auth/Default.php
+++ b/lib/SimpleSAML/Auth/Default.php
@@ -47,8 +47,8 @@ class SimpleSAML_Auth_Default {
 	 * @deprecated This method will be removed in SSP 2.0.
 	 */
 	public static function initLogoutReturn($returnURL, $authority) {
-		assert('is_string($returnURL)');
-		assert('is_string($authority)');
+		assert(is_string($returnURL));
+		assert(is_string($authority));
 
 		$session = SimpleSAML_Session::getSessionFromRequest();
 
@@ -72,8 +72,8 @@ class SimpleSAML_Auth_Default {
 	 * @deprecated This method will be removed in SSP 2.0.
 	 */
 	public static function initLogout($returnURL, $authority) {
-		assert('is_string($returnURL)');
-		assert('is_string($authority)');
+		assert(is_string($returnURL));
+		assert(is_string($authority));
 
 		self::initLogoutReturn($returnURL, $authority);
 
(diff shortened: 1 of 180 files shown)


jaimeperez:
Thanks **so much** for this @fentie! This is an awesome contribution, you just saved us plenty of hours of work!

fentie:
No problem at all. It was mostly regex search & replace 😀

bradjones1:
This appears only committed to master, not the 1.x branches? I tried to find an explanation of the branching strategy but came up empty, can a maintainer shed some light on whether this will be in 1.x or if it should be applied as a patch until then? Thanks.

thijskh:
This will be part of 1.16 when that’s released.

alberto56:
Hi all, when I try to apply this as a patch to the current latest version, I get:

    curl -O https://patch-diff.githubusercontent.com/raw/simplesamlphp/simplesamlphp/pull/708.patch
    patch -p1 < ../708.patch

    ...
    patching file modules/saml/lib/Auth/Process/FilterScopes.php
    patching file modules/saml/lib/Auth/Process/NameIDAttribute.php
    Hunk #3 FAILED at 105.
    1 out of 3 hunks FAILED -- saving rejects to file modules/saml/lib/Auth/Process/NameIDAttribute.php.rej
    patching file modules/saml/lib/Auth/Process/PersistentNameID.php
    patching file modules/saml/lib/Auth/Process/PersistentNameID2TargetedID.php
    ...
    
modules/saml/lib/Auth/Process/NameIDAttribute.php.rej contains:

    --- modules/saml/lib/Auth/Process/NameIDAttribute.php
    +++ modules/saml/lib/Auth/Process/NameIDAttribute.php
    @@ -105,16 +105,16 @@ private static function parseFormat($format)
          */
         public function process(&$state)
         {
    -        assert('is_array($state)');
    -        assert('isset($state["Source"]["entityid"])');
    -        assert('isset($state["Destination"]["entityid"])');
    +        assert(is_array($state));
    +        assert(isset($state['Source']['entityid']));
    +        assert(isset($state['Destination']['entityid']));
     
             if (!isset($state['saml:sp:NameID'])) {
                 return;
             }
     
             $rep = $state['saml:sp:NameID'];
    -        assert('isset($rep["Value"])');
    +        assert(isset($rep['Value']));
     
             $rep['%'] = '%';
             if (!isset($rep['Format'])) {

If anyone can suggest a way around this I'd be grateful, thanks!

thijskh:
Edit the file by hand?
Just ignore the deprecation notices as they’re only notices? Turn them off in PHP config?

alberto56:
There is one line which has changed and makes the patch not apply. Here is a version of the patch which can work if anyone needs it. Use with caution!

[708-modified-for-2018-05-23-15-59-22.patch.tar.gz](https://github.com/simplesamlphp/simplesamlphp/files/2032642/708-modified-for-2018-05-23-15-59-22.patch.tar.gz)

PR #714, opened 2017-10-25
Added tests for SimpleSAML\XML\Validator

mschwager:
Hey all,

Validator is another security-critical piece of SSP, so let's test it!

These tests exercise most of the behavior in the module, however I couldn't find a good way to test certificate validation against a CA. Unfortunately `validateCertificate` relies on a CA file on the actual filesystem, and neither `validateCAExec` nor `validateCABuiltIn` play nicely with `vfsstream`. Maybe we can find an efficient way to test this functionality in the future.

`testIsNodeValidatedFailure` should also confirm that the fix from https://simplesamlphp.org/security/201710-01 is working :)

### DIFF
--- a/tests/lib/SimpleSAML/XML/ValidatorTest.php
+++ b/tests/lib/SimpleSAML/XML/ValidatorTest.php
@@ -1,0 +1,367 @@
+<?php
+
+namespace SimpleSAML\Test\XML;
+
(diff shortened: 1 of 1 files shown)


jaimeperez:
Thanks a lot Matt! 🥇

PR #716, opened 2017-11-01
MySQL key value store now handles NULL expiries.

simonbrahan:
Declaration of kvstore _expires field was defaulting to current timestamp in MySQL.
Added mini upgrade script to kvstore init method to add default null value.

This could have been fixed by changing the declaration of the _expire field in the kv store to `_expire TIMESTAMP NULL` but that would not have fixed existing installations.

### DIFF
--- a/lib/SimpleSAML/Store/SQL.php
+++ b/lib/SimpleSAML/Store/SQL.php
@@ -99,25 +99,61 @@ class SQL extends Store
      */
     private function initKVTable()
     {
-        if ($this->getTableVersion('kvstore') === 1) {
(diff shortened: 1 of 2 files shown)


jaimeperez:
Thanks a lot @simonbrahan!

simonbrahan:
Thanks for the merge Jamie.

I've noticed there's a similar upgrade pattern used in other places in the codebase - notably the logoutstore and the table version storage itself. Should I look at changing those to use the same pattern?

jaimeperez:
If you don't mind, we would really appreciate that!

PR #704, opened 2017-10-16
Implement "Remember my Organization" in multiple-ldap

aguvillalba:
This new feature allows the users to have a checkbox in the login form
which works in the same way than "Remember my username" but for the
organization selected in the dropdown
Implements feature requested in issue: Allow "Remember organization" for multiple-ldap authsource #703

PR #717, opened 2017-11-09
new user interface

Olimpiamagliulo:
Change the design of SSP starting from index.php, authentication method and login page. 

- Based on appearance-config the user can display the standard theme or the new one
- With the new theme the user is redirected to login page or can choose an authentication method
- Migrating php templates to twig
- Languagebar becomes a dropdown menu. 
- Restructure css and js files from www/resource/ into www/assets/
- New user interface will focus on intuitive design and responsiveness. 
- Future changes will follow mvc-pattern.
- Use yarn for package management-> .gitignore updated accordingly.


tvdijen:
Today, I've had a lot of difficulties getting the statistics-module to work on Twig..
It seems that the new UI needs jQuery 3.2.1, which is included from the base-template at all times.

```
    {% spaceless %}{% if jquery %}
        {% set version = '1.8' %}
        {% if jquery.version %}
            {% set version = jquery.version %}
        {% endif %}
        {% if version == '1.8' %}
            {% if jquery.core %}
    <script type="text/javascript" src="/{{ baseurlpath }}resources/jquery-1.8.js"></script>
            {% endif %}
            {% if jquery.ui %}
    <script type="text/javascript" src="/{{ baseurlpath }}resources/jquery-ui-1.8.js"></script>
            {% endif %}
        {% endif %}
    {% endif %}{% endspaceless -%}
    <script type="text/javascript" src="/{{ baseurlpath }}assets/js/lib/jquery-3.2.1.min.js"></script>
```

My problem with the above is that if a module specifically enables jQuery, both the 1.8 and the 3.2.1 version is loaded.. The latter is loaded last (and overwrites the jQuery-object), so we always end up with version 3.2.1 no matter what.

This should probably say (semantically):
```
if old templating-system:
  load jquery 1.8
else if new-templating-system:
  load jquery 3.2.1
```

I guess I was just unlucky here, because any module that specifically load jquery will end up with both versions loaded (and the 3.2.1 being effective). I will have to come up with a solution to make the statistics-module work with 3.2.1, but my point here is that the way we load jQuery now can cause issues because more than one version can be loaded.

tvdijen:
Today I've done some more work on Twigification..
I think the list of remainders is:
- [ ] consent
- [ ] consentAdmin
- [ ] core
- [ ] discopower
- [ ] expirycheck
- [ ] memcacheMonitor
- [ ] oauth

tvdijen:
@jaimeperez do you want to keep the old templates in place? Since several are now broken due to necessary changes for Twigification..

jaimeperez:
The old templates and all resources used by them must be kept in place. In 1.16 we'll have both the new user interface and the new one, and both must work independently. That means data can be added to the scripts showing templates, but not removed.

I was pretty sure that we already had a configuration option to enable or disable the use of twig, but it looks like we don't yet. I'll try to reuse the configuration option we have now to use the new UI to tell the `SimpleSAML_XHTML_Template` class to use twig or not.

tvdijen:
Okay, I have some more work to keep everything work with both templating systems then!

tvdijen:
Anything left I can do on this branch?

tvdijen:
We may want to put a little effort in #704 before merging... Other than that, I agree on the above.

tvdijen:
As discussed with @thijskh over email, this branch is horribly broken and previous attempts by us to find the problem have let us nowhere.. The sheer size of this PR is making it difficult to analyze.
I will start working on peeling off this banana and try and merge some bits and pieces to master, like all the Twig-templates and language-files so we can reduce the footprint of this PR and hopefully better understand the issues we're experiencing..

thijskh:
Yes, would be great to split it out into separate things. And merge individual improvements to master when they're ready.

jaimeperez:
Hi Tim!

> Any thoughts on this @jaimeperez before I start working on it?

I agree, it's a good idea to try to clean things up to make it easier to merge this. It might also be an idea to try to work on specific things in sub-branches, then merge them to this one. Another option is to use this branch only as a reference to open new, smaller, targeted PRs that we can merge one-by-one.

tvdijen:
Cleaned up a fair amount already, but the next step feels like a giant leap to me...
I've tried to cherry-pick one of the Twig-templates to a copy of master and it horribly breaks on translations... Aren't Twig and new translations supposed to work flawlessly on master?

lonoak:
Hi Tim,

  I'd say translations are broken in `Xnew-ui too`, at least this is what I see checking out this branch:

https://screenshots.firefox.com/6ZNsGEnoHCtSKL6P/sso.tuorganizacion.es

Is this what you are experiencing ?

By the way... how do you activate the new ui in master? I mean, there are missing pieces such as [`packages.json`](https://github.com/simplesamlphp/simplesamlphp/blob/Xnew-ui/package.json), which I think are necessary.

jaimeperez:
Hi,

It's hard to say because it's been some time since the last time I was able to test the `Xnew-ui` branch, being dragged into many other projects, but it surprises me a bit that translations are broken since they work just fine for Feide using the `master` branch. Do you guys have the `language.i18n.backend` configuration option set to `gettext/gettext`, and the `gettext/gettext` and `jaimeperez/twig-configurable-i18n` dependencies installed?

tvdijen:
Yes @lonoak, I'm having similar issues, also with `master` + `gettext/gettext`.

lonoak:
yes @jaimeperez, it is included in `composer.json`:

>     "require": {
>         "php": ">=5.4",
>         "ext-SPL": "*",
>         "ext-zlib": "*",
>         "ext-pcre": "*",
>         "ext-openssl": "*",
>         "ext-dom": "*",
>         "ext-date": "*",
>         "ext-hash": "*",
>         "ext-json": "*",
>         "ext-mbstring": "*",
>         "simplesamlphp/saml2": "~3.2",
>         "robrichards/xmlseclibs": "dev-master#93f8c07976b36b6050ab341ab5850484dc897ce0 as 3.0.2",
>         "whitehat101/apr1-md5": "~1.0",
>         "twig/twig": "~1.0",
>         "gettext/gettext": "^3.5",
>         "jaimeperez/twig-configurable-i18n": "^1.2"
>       }

are all the versions correct?

jaimeperez:
That's weird Tim. What happens if you fix the version of `gettext/gettext` to exactly 3.5?

The only thing I can imagine is that something is broken in later versions, but it's weird that it works just fine in Feide, where we have been using twig templates and gettext translations for a while now. I'm more inclined to think that there might be some kind of error in a template...

tvdijen:
Fixing the version doesn't help... I've also fiddled a bit with a fixed version of Twig, but it all makes no difference..
If this is a broken template, it has to be something in one of the base-templates..

lonoak:
@jaimeperez 
> What happens if you fix the version of gettext/gettext to exactly 3.5?

In brach `Xnew-ui`, exactly the same...

@tvdijen 
>  If this is a broken template, it has to be something in one of the base-templates..

I support this idea... all the screens seem the same, so it must be something common...

tvdijen:
Compare https://idp.moo-archive.nl/admin/sandbox.php and https://idp.feide.no/simplesaml/admin/sandbox.php... The former does not translate anything (same traceback in the logs as the one above)

tvdijen:
Okay, I'm closing in on something.. On May 1 (2218066), everything was working fine.. On Jun 1 (c6c7f58), everything was broken.. I guess after June 1st, stuff has been fixed, but not properly...

tvdijen:
OK, I think I'm on to something..
I can see it working fine on June 1st (30b56637f18bccb896914b98d3287d7dbd66d776) and being broken on July 12th (f917d4431af8707a39e224a688860aba66379678)... Will try to investigate further

tvdijen:
OK something's playing tricks with me now.. When I go from bb5ad9cf298b184d74580329364cac4d609047e0 to 086881f74d97ff93bccf14cf039e55d640c3caf1, things break, but when I manually go over the changes, F5'ing on every changed file, things remain working..

tvdijen:
There! I nailed it! All it took was a few cups of strong coffee :D
See https://github.com/simplesamlphp/simplesamlphp/commit/1d7d0cb07e32c6c49e8092c6e76ff7c6e9267996

jaimeperez:
Great! Thanks for hunting this down Tim!

> @jaimeperez Apparently your custom classes have the same name as original Twig-classes.. That may not be a good idea!

That's precisely the intention. Having the exact same name makes it possible to use Twig with an alternative gettext implementation without the need to rewrite your code. Ideally, this should be fixed upstream. However, the Twig maintainers don't seem to care much about this... 😞

tvdijen:
Okay, I have no idea how this merge happened :/ 
I guess it's not a bad thing now the issue has been resolved...

lonoak:
Thanks for your efforts, Tim!

PR #722, opened 2017-11-13
Docs: fix typo's

tvdijen:


jaimeperez:
Thanks @tvdijen! I added you to the list of collaborators with write access, so it's not necessary any longer for you to submit PR's to fix stuff like this 😉

tvdijen:
Maybe I'm doing it wrong then... Can commit directly to the master branch

jaimeperez:
It's ok to open PRs if you are in doubt and want to discuss before merging, but otherwise, simple fixes like this are totally ok if you just commit them to master 😄 

Let me know if you have trouble with that, I think everything should be configured correctly, but I might have made a mistake...

tvdijen:
afbeelding

tvdijen:
Perhaps we can figure this out at next month's hackathon

jaimeperez:
It might be that you need to be more than a collaborator. Check your email 😉

tvdijen:
Unfortunately, that didn't make a difference..

tvdijen:
I think this is my problem:
>Can't be edited or have files uploaded to it from the web

jaimeperez:
I think I found the issue. Should be ok now...

PR #725, opened 2017-11-14
Make MDQ more robust on errors (fixes #723)

### LINKED ISSUE #723: MDX breaks metadata query flow
It seems that if an entity is not found during MDX query (or the query fails), the MetadataStorageSource throws an exception instead of returning null. Subsequentially, it prevents additional metadata sources from being consulted.

I can submit a pull request if you agree that it's a bug.

I think the following line is relevant here: https://github.com/simplesamlphp/simplesamlphp/blob/57df7b25fa558531c1e76aa686c5daec199b3fdc/lib/SimpleSAML/Metadata/Sources/MDQ.php#L300

bajnokk:
Don't bail out if the MDQ cache is broken or the query fails, because
later other metadata sources might provide the metadata for the entity.

### DIFF
--- a/lib/SimpleSAML/Metadata/Sources/MDQ.php
+++ b/lib/SimpleSAML/Metadata/Sources/MDQ.php
@@ -262,7 +262,8 @@ class MDQ extends \SimpleSAML_Metadata_MetaDataStorageSource
      *
      * @return array An associative array with metadata for the given entity, or NULL if we are unable to
      *         locate the entity.
-     * @throws \Exception If an error occurs while downloading metadata, validating the signature or writing to cache.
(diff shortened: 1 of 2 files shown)


bajnokk:
Log levels are always subject to debate. I've set cache errors to "ERROR", because they can be considered mostly as incomplete/broken configuration, while errors during the query as INFO, because
1) NOTICE is reserved for statistics
2) No metadata source should complain too much if the entity can not be found.

This is why this commit contains a very little error suppression in Utils\HTTP::fetch as well.

On signature validation errors an exception is thrown, still. This means that the entity can be found but it can not be validated. IMO it should break metadata processing.

bajnokk:
Updated the commit so that function documentation reflects the change.

PR #727, opened 2017-11-15
Bugfix: ldap:AttributeAddUsersGroups ignored the attribute.username setting for OPENLDAP.

thijskh:
The OPENLDAP option uses hardcoded uid to search for. It should instead use the value of attribute.username. This means it also changes the default since attribute.username defaults to samAccountName. So update the docs to make the example work again. Has not been in a release yet so should not break existing installations.


thijskh:
Confirmed working by reporter.

PR #734, opened 2017-11-30
Statistics: use secure endpoint

tvdijen:
Prevent graph from being blocked because of mixed content.
https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content?utm_campaign=default&utm_medium=firefox-console-errors&utm_source=mozilla


tvdijen:
By the way, I've tried to commit this directory to master from the command line, but it still got refused.

PR #735, opened 2017-12-01
NameIDAttribute filter: update to use SAML2\XML\saml\NameID

bynare:
This patch fixes the following error when attempting to use NameIDAttribute filter (https://simplesamlphp.org/docs/1.15/saml:nameidattribute):
```SimpleSAML_Error_Exception: Error 1 - Cannot use object of type SAML2\XML\saml\NameID as array```


thijskh:
This should have been caught by a test... do you think you could add a testcase(s) for this authprocfilter? Many other authprocfilters already have tests you could look at for an example.

bynare:
@thijskh sure thing - I'll look at getting some tests added ;)

Gaaf!

bynare:
Hi @thijskh,

I've now added some tests - see the latest patch.

Cheers,
Eugene.

thijskh:
Excellent, thanks!

thijskh:
Confirmed that it fixes the problem for me as well.

PR #736, opened 2017-12-04
Add new config option

tvdijen:
It's mentioned in the docs, but should be in the config template as well..
Not sure whether this requires extra comments too.


jaimeperez:
Thanks Tim!

PR #738, opened 2017-12-06
Added tests for sspmod_core_Auth_Process_AttributeMap

mschwager:
Added some tests for `sspmod_core_Auth_Process_AttributeMap`!

I managed to get just about everything but the `loadMapFile` functionality. That function was difficult to test without any modifications to the library. The module and config path functionality ends up returning full filesystem paths which doesn't play well with `vfsstream`, and unfortunately there's no way to mock out the path value. Regardless, this is better than nothing, and maybe we can find a way to test the map file in the future :)


thijskh:
Excellent!

PR #743, opened 2017-12-08
oauth: Cleanup unused demo + templates

tvdijen:
It's become useless since the removal of www/requestToken.php in 1.14 - see b51a731


tvdijen:
It also seems that both
```
templates/authorized.php
templates/consent.php
```
are never used.. Can someone please verify?

PR #746, opened 2017-12-10
 Replace deprecated Sqlite-calls with PDO-calls 

### LINKED ISSUE #744: Sqlite has been deprecated
`modules/core/lib/Storage/SQLPermanentStorage.php` relies on an old sqlite PHP-library that doesn't come with PHP anymore since 5.4. The library was moved to PECL, but repositories are dead/abandoned and they don't work with PHP7.

At the moment, only the oauth-module relies on this code.
Should probably convert this to use PDO, like the Store-class does.

tvdijen:
Attempt to fix #744
I have tested it with the oauth-module (currently the only module using it) and it's working like a charm!
Should be fully backwards compatible too!

### DIFF
--- a/modules/core/lib/Storage/SQLPermanentStorage.php
+++ b/modules/core/lib/Storage/SQLPermanentStorage.php
@@ -29,12 +29,11 @@ class sspmod_core_Storage_SQLPermanentStorage {
 			mkdir($sqllitedir);
 		}
 		
-		$dbfile = $sqllitedir . $name . '.sqllite';
-		
-		if ($this->db = new SQLiteDatabase($dbfile)) {
+		$dbfile = 'sqlite:' . $sqllitedir . $name . '.sqlite';
+                if ($this->db = new \PDO($dbfile)) {
 			$q = @$this->db->query('SELECT key1 FROM data LIMIT 1');
 			if ($q === false) {
-				$this->db->queryExec('
+				$this->db->exec('
 		CREATE TABLE data (
 			key1 text, 
 			key2 text,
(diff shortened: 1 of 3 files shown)


modules/core/lib/Storage/SQLPermanentStorage.php
```diff
@@ -9,198 +9,210 @@
...
+        }
+
+        $sqllitedir = $datadir . 'sqllite/';
+        if (!is_dir($sqllitedir)) {
+            mkdir($sqllitedir);
+        }
+
+        $dbfile = 'sqlite:' . $sqllitedir . $name . '.sqlite';
+        if ($this->db = new \PDO($dbfile)) {
+            $q = @$this->db->query('SELECT key1 FROM data LIMIT 1');
```
LukeLeber: Error suppression is hardly ever a desirable thing.  I know that the original source code had it as well, but shouldn't failures be loud and obvious?

tvdijen: I think I agree with you here, however I'm a bit anxious to break behaviour here.. That would mean we have to wait for the 2.0 release before we can merge this.

tvdijen: Now you've made me wonder if I should use the SimpleSAML\Database class here instead of directly calling PDO...

tvdijen: At this point I see no easy way to do this, so I'd like to stick with this PR 'as is'.
@thijskh could you please review this?

tvdijen:
I have no clue where to start on that, since I'm totally unfamiliar with writing unit tests.

tvdijen:
How does this look @thijskh?
You think I can add 'writing unit tests' to my resume?

thijskh:
Very nice work! "Coverage increased (+0.7%) to 29.503%" \o/
Hope you now see that especially for rather 'self-contained' classes this is very doable and adds a lot of value.

One comment: I see that you create the sqlite database but do not remove it from disk at teardown. This is of course fine for travis but leaves cruft on the machine for people running the testsuite locally.

tvdijen:
Fixed that!

PR #748, opened 2017-12-14
Add 'no-store' to the cache-control header

LukeLeber:
With the release of Google Chrome version 63, we started noticing that the 302 redirect to the Identity Provider was being served from disk cache randomly.  This caused a stale request condition at the Shibboleth Identity Provider.

By adding 'no-store' to the cache-control header, Google Chrome will no longer be able to store the 302 response in the local disk cache.

We will most likely be patching this on our production system, but would greatly appreciate it if this could be merged in so we can go back to having a clean install.

Regards,
Luke

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -176,7 +176,7 @@ class HTTP
 
             // disable caching of this response
             header('Pragma: no-cache');
-            header('Cache-Control: no-cache, must-revalidate');
+            header('Cache-Control: no-cache, no-store, must-revalidate');
         }
 
         // show a minimal web page with a clickable link to the URL


jaimeperez:
Hi @LukeLeber! Thanks a lot for your PR!

It's kind of weird that nobody brought this to our attention yet. I have one question in any case: wouldn't it be enough with the `no-store` directive? As far as I understand, both `no-cache` and `must-revalidate` would be redundant then...

LukeLeber:
Hi @jainrperez,

Unfortunately I don't know enough about the history of all major browsers to say for sure if it would be redundant or not.  I'd hate to inadvertently break support elsewhere.

I believe that this PR should have some exhaustive browser testing associated with it should any of the existing directives be removed.

On a side note, it appears that the 'Use a prediction service to load pages more quickly' setting in Chrome 63 may also be a culprit.  Disabling this feature also seems to cause the incorrect disk caching to go away -- but does little for our users that are affected by this.

thijskh:
I was reading the RFC on this topic (also prompted by an earlier conversation at SAML-Tracer) about 302. Might it not be better to switch to 303 instead? About 303:

```
The response to the request can be found under a different URI and SHOULD be retrieved using a GET method on that resource. This method exists primarily to allow the output of a POST-activated script to redirect the user agent to a selected resource. The new URI is not a substitute reference for the originally requested resource. The 303 response MUST NOT be cached, but the response to the second (redirected) request might be cacheable.
```

Because a request to anything protected by SSP could in fact be a POST, 303 seems to be a more accurate match and at the same time it explicitly forbids any caching.

I would btw be fine with that we merge this now and discuss changing the code in a separate issue...

PR #756, opened 2018-01-01
Set entityId as array-key too

tvdijen:
To me it doesn't make sense to not also replace the array-key with the proper entityId, especially given there is no easy way to do it afterwards;   generateDynamicHostedEntityID is a private method.

I'm not entirely confident this doesn't break anything, so that's why I made this PR.

### DIFF
--- a/lib/SimpleSAML/Metadata/MetaDataStorageHandlerFlatFile.php
+++ b/lib/SimpleSAML/Metadata/MetaDataStorageHandlerFlatFile.php
@@ -113,6 +113,8 @@ class SimpleSAML_Metadata_MetaDataStorageHandlerFlatFile extends SimpleSAML_Meta
         foreach ($metadataSet as $entityId => &$entry) {
             if (preg_match('/__DYNAMIC(:[0-9]+)?__/', $entityId)) {
                 $entry['entityid'] = $this->generateDynamicHostedEntityID($set);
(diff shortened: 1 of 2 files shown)


thijskh:
It breaks the API and people might depend on that right now. Is there a big gain by doing this?

tvdijen:
For my monitoring-module, I loop through all available metadata-sets and entities to verify them.

Right now this leads to stupid checks like this ...
https://github.com/tvdijen/simplesamlphp-module-monitor/blob/v1.1.0/lib/TestSuite/Metadata.php#L32
... and a copy of the `generateDynamicHostedEntityID`-function because it's currently private:
https://github.com/tvdijen/simplesamlphp-module-monitor/blob/v1.1.0/lib/TestSuite/Metadata.php#L66

thijskh:
So it's to construct a mock set. Not sure that justifies breaking this API...

thijskh:
I'm leaning towards not merging this as per above...

jaimeperez:
I agree with Thijs. Tim, is there a way you can do what you need without breaking the API?

tvdijen:
I can do what I need, I just rather don’t copy code..
Would it be an option to make `MetadataStorageHandlerFlatFile::generateDynamicHostedEntityID()` a public method?

PR #760, opened 2018-01-12
ADFS improvements

tvdijen:
This PR addresses the remainders of https://github.com/simplesamlphp/simplesamlphp/pull/469, so we can finally close that one.

### DIFF
--- a/modules/adfs/lib/IdP/ADFS.php
+++ b/modules/adfs/lib/IdP/ADFS.php
@@ -9,6 +9,7 @@ class sspmod_adfs_IdP_ADFS
 
             $requestid = $query['wctx'];
             $issuer = $query['wtrealm'];
+
             $metadata = SimpleSAML_Metadata_MetaDataStorageHandler::getMetadataHandler();
             $spMetadata = $metadata->getMetaDataConfig($issuer, 'adfs-sp-remote');
 
@@ -23,10 +24,11 @@ class sspmod_adfs_IdP_ADFS
             'ForceAuthn' => false,
             'isPassive' => false,
             'adfs:wctx' => $requestid,
+            'adfs:wreply' => false
         );
 
-        if (isset($_GET['wreply']) && !empty($_GET['wreply'])) {
-            $state['adfs:wreply'] = SimpleSAML\Utils\HTTP::checkURLAllowed($_GET['wreply']);
+        if (isset($query['wreply']) && !empty($query['wreply'])) {
+            $state['adfs:wreply'] = SimpleSAML\Utils\HTTP::checkURLAllowed($query['wreply']);
         }
 
         $idp->handleAuthenticationRequest($state);		
@@ -184,7 +186,8 @@ MSG;
         $wresult = sspmod_adfs_IdP_ADFS::signResponse($response, $privateKeyFile, $certificateFile);
 
         $wctx = $state['adfs:wctx'];
-        sspmod_adfs_IdP_ADFS::postResponse($spMetadata->getValue('prp'), $wresult, $wctx);
+        $wreply = $state['adfs:wreply'] ?: $spMetadata->getValue('prp');
+        sspmod_adfs_IdP_ADFS::postResponse($wreply, $wresult, $wctx);
     }
 
     public static function sendLogoutResponse(SimpleSAML_IdP $idp, array $state)
--- a/modules/adfs/lib/IdP/ADFS.php
+++ b/modules/adfs/lib/IdP/ADFS.php
@@ -25,14 +25,18 @@ class sspmod_adfs_IdP_ADFS
             'adfs:wctx' => $requestid,
         );
 
+        if (isset($_GET['wreply']) && !empty($_GET['wreply'])) {
+            $state['adfs:wreply'] = SimpleSAML\Utils\HTTP::checkURLAllowed($_GET['wreply']);
+        }
+
         $idp->handleAuthenticationRequest($state);		
     }
 
-    private static function generateResponse($issuer, $target, $nameid, $attributes)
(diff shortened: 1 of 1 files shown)


tvdijen:
@pgh70 could you take a look?

- The namespace-issue has already been addressed in https://github.com/simplesamlphp/simplesamlphp/pull/678
- I've made the assertion lifetime configurable, similar to the way we configure them on SAML2-IdP's.
- wsreply is being checked against trusted url's

tvdijen:
I don't think @pgh70 is active on GitHub anymore..

@thijskh could you perhaps take a look at this?

modules/adfs/lib/IdP/ADFS.php
```diff
@@ -163,19 +171,25 @@ public static function sendResponse(array $state)
...
+
+        $response = sspmod_adfs_IdP_ADFS::generateResponse($idpEntityId, $spEntityId, $nameid, $attributes, $assertionLifetime);
 
         $privateKeyFile = \SimpleSAML\Utils\Config::getCertPath($idpMetadata->getString('privatekey'));
         $certificateFile = \SimpleSAML\Utils\Config::getCertPath($idpMetadata->getString('certificate'));
         $wresult = sspmod_adfs_IdP_ADFS::signResponse($response, $privateKeyFile, $certificateFile);
 
         $wctx = $state['adfs:wctx'];
-        sspmod_adfs_IdP_ADFS::postResponse($spMetadata->getValue('prp'), $wresult, $wctx);
+        $wreply = $state['adfs:wreply'] ? : $spMetadata->getValue('prp');
```
thijskh: `? :` that doesn't look right.

modules/adfs/lib/IdP/ADFS.php
```diff
@@ -98,14 +107,14 @@ private static function signResponse($response, $key, $cert)
...
-        $responsedom = \SAML2\DOMDocumentFactory::fromString(str_replace ("\r", "", $response));
+        $responsedom = \SAML2\DOMDocumentFactory::fromString(str_replace("\r", "", $response));
         $firstassertionroot = $responsedom->getElementsByTagName('Assertion')->item(0);
         $objXMLSecDSig->addReferenceList(
             array($firstassertionroot), XMLSecurityDSig::SHA1,
             array('http://www.w3.org/2000/09/xmldsig#enveloped-signature', XMLSecurityDSig::EXC_C14N),
             array('id_name' => 'AssertionID')
         );
-        $objKey = new XMLSecurityKey(XMLSecurityKey::RSA_SHA1, array('type'=>'private'));
+        $objKey = new XMLSecurityKey(XMLSecurityKey::RSA_SHA1, array('type' => 'private'));
```
thijskh: Is SHA1 hardcoded on purpose? Should it be?

modules/adfs/lib/IdP/ADFS.php
```diff
@@ -1,5 +1,8 @@
 <?php
 
+use RobRichards\XMLSecLibs\XMLSecurityDSig;
+use RobRichards\XMLSecLibs\XMLSecurityKey;
```
thijskh: What change are these necessary for?

thijskh:
I've added some comments on the code. I don't have any ADFS knowledge so cannot give any judgement on whether the proposed changes are the right thing to do.

modules/adfs/lib/IdP/ADFS.php
tvdijen: They're to silence Scrutinizer complaining about missing namespaces.
We also do this in `modules/saml/lib/Message.php` and `modules/saml/lib/IdP/SAML2.php`, so it makes sense, I think?

modules/adfs/lib/IdP/ADFS.php
jaimeperez: It actually is. It's commonly known as the [_elvis operator_](http://php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary).

modules/adfs/lib/IdP/ADFS.php
tvdijen: Well, it should probably be configurable instead of hardcoded, in a similar way we can do this for SAML-entities.. I'll see if I can add this, although it is out of scope for this PR.

jaimeperez: IIRC, that's the way you initialize `XMLSecurityKey`, and then you _cast_ the key to use the actual intended algorithms...

modules/adfs/lib/IdP/ADFS.php
tvdijen: That's perfectly valid PHP??
"(expression1) ?: expression2 will return expression1 if it evaluates to true or expression2 otherwise."

tvdijen: Could've used `||` too, I guess it's just a matter of coding style...

thijskh: Obviously I know about the ternary operator, but I was not aware of this "improvement" since PHP 5.3 where you can leave out the middle part.

modules/adfs/lib/IdP/ADFS.php
thijskh: Love it

tvdijen:
I've straightened out the way signature algorithm can be set for ADFS-entities with SAML-entities...

thijskh:
I think this should be OK, but my comment about ADFS knowledge stands

pgh70:
Still here, but sick at the moment :-(

tvdijen:
Get well soon!

tvdijen:
@pgh70 Any chance you can have look at this?

tvdijen:
I went through everything again and I'm confident about the changes made. Merging.

PR #763, opened 2018-01-15
Adds support for creation of Auth Source Factories

unt01d:
An attempt to make Auth Source classes more unit-testable by providing a framework to instantiate them via Factories.

When parsing the auth config, we first look to see whether or not a Factory has been created for the Auth Source and, if so, use it to create the Auth Source instance.  If not, we fall back on the previous method of explicitly handling object creation.

Auth Source Factories must be declared in the `modules/<module>/lib/Auth/Source/Factory` directory, and implement the `SimpleSAML\Auth\SourceFactory` interface.


unt01d:
That sounds more than reasonable @thijskh, given that testing is the driver for the change :grinning:  I'll spend some time looking into this.

unt01d:
Here you go @thijskh - added a [simple test](https://github.com/simplesamlphp/simplesamlphp/pull/763/commits/249c2cbe8cf51c36f7cc272fb2b964b61e583a31) to cover instantiation of auth sources both directly and via factories.

thijskh:
Great, thanks!

PR #767, opened 2018-01-18
authfacebook compatability with Facebook strict URI match

pradtke:
Per https://developers.facebook.com/docs/facebook-login/security/#strict_mode the
`state` parameter should be used for any state needed after the redirect, and the
redirect URI should remain constant.

### DIFF
--- a/modules/authfacebook/lib/Facebook.php
+++ b/modules/authfacebook/lib/Facebook.php
@@ -146,4 +146,11 @@ class sspmod_authfacebook_Facebook extends BaseFacebook
     }
     return implode('_', $parts);
   }
+
+  protected function establishCSRFTokenState() {
+     if ($this->state === null) {
+          $this->state = SimpleSAML_Auth_State::getStateId($this->ssp_state);
+          $this->setPersistentData('state', $this->state);
+     }
+  }
 }
--- a/modules/authfacebook/extlibinc/base_facebook.php
+++ b/modules/authfacebook/extlibinc/base_facebook.php
@@ -698,7 +698,7 @@ abstract class BaseFacebook
         $this->clearPersistentData('state');
         return $_REQUEST['code'];
       } else {
-        self::errorLog('CSRF state token does not match one provided.');
+        self::errorLog('CSRF state token does not match one provided. ' . $this->state . '!=' . $_REQUEST['state']);
         return false;
       }
     }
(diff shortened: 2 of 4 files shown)


modules/authfacebook/lib/Facebook.php
```diff
@@ -146,4 +146,11 @@ protected function constructSessionVariableName($key) {
     }
     return implode('_', $parts);
   }
+
+  protected function establishCSRFTokenState() {
```
tvdijen: You're not using this function in your committed code??

pradtke: I'm overriding the behavior from base_facebook->establishCSRFTokenState
I suppose it could use a comment to describe why I'm overriding.

tvdijen: Gah, sorry... It's too late to review code...
But don't we just need a newer version of the BaseFacebook-class??

pradtke: Get some sleep :) Facebook integrations won't break for 6 more weeks :)

I believe BaseFacebook was copied from https://github.com/facebookarchive/facebook-php-sdk/blob/master/src/base_facebook.php and that project has been deprecated. 
My quick read through the latest from that project was that it still wouldn't let us control the state query param.

As to the long term viability of the authfacebook module - my current thinking is to get rid of it, and build a generic Oauth2 auth module based on http://oauth2-client.thephpleague.com/ ( it would then support their 20+ social integrations ). I have an experimental version that works with Facebook and Amazon. I hope to finish it sometime this spring.

tvdijen: Yes! I was reading the same and apparently this library is old and abandoned..

At this point, I feel all our social ID modules should move to separate repositories...
If we could replace them with one generic auth-module in the near future, that would be great!

jaimeperez:
Thanks a lot Patrick!

I understand you've tested this and it works fine with the new Facebook API, right?

pradtke:
Corrected, it has been tested. We are running the patch in production now, and enabled the 'strict redirect uri' option on one of the Facebook apps we use for monitoring. The other Facebook app we use for monitoring we left with the current setting. Both work. So it's compatible with the current Facebook settings and with the ones coming in March.

tvdijen:
I took the liberty to make a minor adjustment...

PR #768, opened 2018-01-21
rewrite getServerPort

tvdijen:
As per mailinglist-discussion:
https://groups.google.com/forum/#!topic/simplesamlphp-dev/sGTsDkDroLE

Also fixes edge-case situation where $_SERVER['SERVER_PORT'] is not set for an HTTPS connection, this function would return an explicit port 80 i.e. ":80" rather than an empty string.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -108,15 +108,11 @@ class HTTP
      */
     public static function getServerPort()
     {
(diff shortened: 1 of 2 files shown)


thijskh:
Looks good! I’d definitely would like to see this put in a testcase though. To ensure this edge case does not get lost in any future changes.

tvdijen:
Yes, I had the same idea...  But I needed a second pair of eyes on the code first... It's quite an important function.
I'll see if I can write a test for it... Not really my expertise

tvdijen:
If you'd be so kind to look at my test, @thijskh?

thijskh:
Excellent!

PR #772, opened 2018-01-26
Fix missing cookie params in authtoken cookie when using remember-me mechanism

renklaf:
Fixes that the configured cookie settings are not correctly set for the auth token cookie when remember-me mechanism is enabled and checked.

Cookie ends up getting set with the default params in `HTTP::setCookie`  
https://github.com/simplesamlphp/simplesamlphp/blob/a5437da2dbbb8dc6348cac02dc343c03bc7f3e14/lib/SimpleSAML/Utils/HTTP.php#L1114-L1120

### DIFF
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -755,13 +755,19 @@ class SimpleSAML_Session implements Serializable
     {
         $sessionHandler = \SimpleSAML\SessionHandler::getSessionHandler();
 
+        if (is_array($params) && !empty($params)) {
+            $params = array_merge($sessionHandler->getCookieParams(), $params);
+        } else {
+            $params = $sessionHandler->getCookieParams();
+        }
+
         if ($this->sessionId !== null) {
             $sessionHandler->setCookie($sessionHandler->getSessionCookieName(), $this->sessionId, $params);
         }
 
         if ($this->authToken !== null) {
             $globalConfig = SimpleSAML_Configuration::getInstance();
-            $sessionHandler->setCookie(
+            \SimpleSAML\Utils\HTTP::setCookie(
                 $globalConfig->getString('session.authtoken.cookiename', 'SimpleSAMLAuthToken'),
                 $this->authToken,
                 $params
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -761,7 +761,7 @@ class SimpleSAML_Session implements Serializable
 
         if ($this->authToken !== null) {
             $globalConfig = SimpleSAML_Configuration::getInstance();
-            \SimpleSAML\Utils\HTTP::setCookie(
+            $sessionHandler->setCookie(
                 $globalConfig->getString('session.authtoken.cookiename', 'SimpleSAMLAuthToken'),
                 $this->authToken,
                 $params


jaimeperez:
Now that I look closer into this, I think your original patch was correct. This can't be done as `SessionHandler::setCookie()` is intended to set precisely a **session** cookie, but the auth token is a regular cookie, not a session one. While this could be ok depending on the session backend in use, it won't be if you are using PHP sessions since you can only have one open, and _setting the cookie_ twice will ruin the actual session.

So I'm afraid the code was rightfully using `SimpleSAML\Utils\HTTP`. The problem is that the right parameters must be passed to it.

Now, `SessionHandler` implements `getCookieParams()`, which fetches the right parameters from configuration. Custom handlers may override it, and all of them **should use it** in any case to configure cookie settings. Both `SessionHandlerPHP` and `SessionHandlerCookie` do. It should therefore be safe to assume that calling `$sessionHandler->getCookieParams()` will always return the right params for you.

The problem is that `SessionHandlerPHP` is failing to comply with its own contract:

```php
/**
     * Set a session cookie.
     *
...
     * @param array|null $cookieParams Additional parameters to use for the session cookie.
...
     */
    public function setCookie($sessionName, $sessionID, array $cookieParams = null)
    {
        if ($cookieParams === null) {
            $cookieParams = session_get_cookie_params();
        }
...
        session_set_cookie_params(
            $cookieParams['lifetime'],
            $cookieParams['path'],
            $cookieParams['domain'],
            $cookieParams['secure'],
            $cookieParams['httponly']
        );
...
    }
```

Essentially, even though `$cookieParams` should **add** parameters when setting the session cookie, the implementation behaves as _either one or the other_. If the parameter is present, it will be used ignoring configuration. If not, `session_get_cookie_params()` is called and the configuration is then used as it was loaded previously in the constructor. This is wrong and we need to fix it. The effect of this is that when _remember me_ is enabled and checked, the session cookie will be renewed with wrong parameters. It might _accidentally_ work if PHP uses current settings for the session without need to specify them again, but that's just an accident.

Once we are sure that all session handlers comply with the contract of `setCookie()`, we can safely call it knowing the configuration parameters are used. Then, we should get the config parameters, merge them with the parameters passed to `SimpleSAML_Session::updateSessionCookies()` and pass the resulting array to `SimpleSAML\Utils\HTTP::setCookie()`.

renklaf:
Hi @jaimeperez 

Thanks for investigating - I missed that. I've reverted the PR to the original patch discussed in the issue.

jaimeperez:
Hi @renklaf!

I see I'm getting sloppy lately 😞 

The one-liner I was suggesting in the issue was literally saying: "_If there are params specified, use those. Otherwise merge them with the configured cookie params._", where _otherwise_ means `$params` is empty (or null, or false), and in that case, what sense does it make to merge it with an existing array?

In any case, the code should read:

```php
if (is_array($params) && !empty($params)) {
    $params = array_merge($sessionHandler->getCookieParams(), $params);
}
```

I'm so sorry for the back and forth...

renklaf:
Hi @jaimeperez 

No worries! :smiley: I've amended the commit.

jaimeperez:
Thanks a lot! I've modified slightly the code, mostly to move it to where it is actually needed, avoiding a double merge for session cookies.

PR #777, opened 2018-01-30
Document IDPList option on saml20-idp-remote.php

pradtke:
Per https://groups.google.com/d/msg/simplesamlphp/ATKQksysR5s/3VhjdLbTAgAJ
there is a feature to allow IdPs to respond to requests sent to different
IdP entityIDs.

### DIFF
--- a/docs/simplesamlphp-reference-idp-remote.md
+++ b/docs/simplesamlphp-reference-idp-remote.md
@@ -120,6 +120,9 @@ The following SAML 2.0 options are available:
 `hide.from.discovery`
 :   Whether to hide hide this IdP from the local discovery or not. Set to true to hide it. Defaults to false.
 
+`IDPList`
+:   The IdP is allowed to respond to an `AuthNRequest` originally sent to entityIDs in this list.
+
 `nameid.encryption`
 :   Whether NameIDs sent to this IdP should be encrypted. The default
     value is `FALSE`.


tvdijen:
It is already documented here: https://github.com/simplesamlphp/simplesamlphp/blob/master/docs/simplesamlphp-scoping.md

And I don't think it was designed to specify which IdP's may respond... I think it was meant for proxying IdP's to be able to adjust the WAYF

pradtke:
@tvdijen I think scoping is a different feature, but unfortunately with the same parameter name.
The feature I'm referring to is from [these lines](https://github.com/simplesamlphp/simplesamlphp/blob/e85780ef9cc60a47a7d304642d4bb4b0d235d9a0/modules/saml/www/sp/saml2-acs.php#L99-L107), when a response comes back to the SP

```php
    if ($state['ExpectedIssuer'] !== $idp) {
        $idpMetadata = $source->getIdPMetadata($idp);
        $idplist = $idpMetadata->getArrayize('IDPList', array());
        if (!in_array($state['ExpectedIssuer'], $idplist, true)) {
            throw new SimpleSAML_Error_Exception(
                'The issuer of the response does not match to the identity provider we sent the request to.'
            );
        }
    }
```

which is different from scoping.

tvdijen:
Okay... Is it even legit to receive a response from an entity other than the one you sent your request to? Other than in a idp-first scenario...

pradtke:
¯\_(ツ)_/¯  I looked through some of the specs and I didn't see any mention that the issuer had to match the entityID that the original authNRequest was sent to.

I don't use the feature, and I don't know how other SP software (such as shib) behaves - I just thought it deserved documenting since (per my original link the the mailing list) some people use it. 
If we want to document it, I think I should at least adjust the text to mention that it is unrelated to scoping.

trscavo:
Jumping in the middle here...

I don't understand the use case but any IdP can push a response to an SP at any time. (At the end of the day, that's all this is...an IdP-initiated response.) It's up to the SP to verify the signature and accept the response content.

tvdijen:
I do understand it... I've heard it before:  "Why can't an IdP respond back to the requester right away, instead of following back the route through all those SAML-proxies??"
Architects see it as a way to reduce traffic...

But those intermediate proxies may handle MFA, attribute aggregation, attribute release, etc...  That's why my first response was:  is that even legal??

thijskh:
I'm merging this since it correctly documents how SSP works. Discussion on whether it's OK to work this way could be had separately - but as long as it does work this way, it should be documented.

PR #779, opened 2018-01-31
Update saml2 dependency per security/201801-01

forevermatt:
Based on the versions of `simplesamlphp/saml2` mentioned on the [201801-01 security vulnerability announcement](https://simplesamlphp.org/security/201801-01), I decided to attempt updating the composer.json file in simplesamlphp 1.14 to request an updated version of saml2.

All the tests pass for me locally (both before and after I made this change).

**NOTE:** I am not familiar enough with the differences between saml2 versions 1.9 and 1.10 to know if there are possibly-breaking changes that would not be caught by tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     },
     "require": {
         "php": ">=5.3",
-        "simplesamlphp/saml2": "~1.9.1",
+        "simplesamlphp/saml2": "~1.10.4",
         "robrichards/xmlseclibs": "~1.4.1",
         "whitehat101/apr1-md5": "~1.0",
         "ext-SPL": "*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "b454686b4114c6478db8a6f5168ba756",
+    "content-hash": "29a1750c1ae72f602af218c4b656e981",
     "packages": [
         {
             "name": "psr/log",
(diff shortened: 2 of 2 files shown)


tvdijen:
That may actually have some consequences because since 1.10 we've changed the way EPTI's are being handled...
Also, the simplesaml-1.14 branch isn't really being maintained anymore...

@jaimeperez do you see harm in merging this? It would resolve some security-related issues since saml/1.9

thijskh:
If this is to be merged, indeed this needs to be tested specifically with epTID and with AllowCreate in authnrequest (those are the changes in saml2 between those releases).

My objection would be that to keep merging things into simplesamlphp-1.14 suggests that branch is supported, while we do not create new releases of it. At least for me goes that given the amount of time I can spend, I don't think keeping 1.14 alive is something I'll work on.

jaimeperez:
Hi!

I agree with Thijs here. I don't think merging this is a good idea, for the following reasons:

* Traditionally, once we've released a new stable version (X.Y), the previous stable has always been deprecated and unmaintained.
* While the `NameIDPolicy` changes might just work, I have serious doubts about `eduPersonTargetedID`.
* Maintaining two (or more) stable versions at the same time requires a lot of time that I'm afraid none of us have. Additionally, it encourages people to keep on the version they already have installed and never upgrade to the next stable version.
* It doesn't even make sense from a security perspective. That security advisory was released together with two others. [One of them](https://simplesamlphp.org/security/201801-02) is, in my opinion, worse than SSPSA 201801-01. If we merge this, we would be merging a fix for a security issue that's hard to exploit in practice, while ignoring another issue that's almost trivial to exploit.

So I think I'm closing this PR for now, but of course I'm open to keep discussing it and hear arguments against my position 😄

forevermatt:
You all make excellent points. When I submitted this, I hadn't fully realized that the full release of 1.15 meant that the 1.14 branch would no longer be supported (though I wondered). Thank you for taking the time to consider it and give thoughtful responses.

I'll resume work upgrading our various installations to 1.15. My initial (brief) attempt yesterday resulted in an error when I tried to begin the login process, but I haven't confirmed whether the problem was in my own code or from a change in 1.15.

Thanks again.

jaimeperez:
No worries!

Remember also that you always have the [mailing list](https://groups.google.com/forum/#!forum/simplesamlphp) in case you need help with the migration 😉

PR #780, opened 2018-02-01
Fix example in the ScopeFromAttribute doc

rmisev:
Change the source attribute to `eduPersonPrincipalName` and add missing commas.


thijskh:
Thanks, useful!

PR #564, opened 2017-02-21
Improve handling of paths containing Windows-style drive letters

LukeCarrier:
This port of catalyst/moodle-auth_saml2#85 fixes #559. There are two changes:
* `SimpleSAML_Configuration::resolvePath()` should resolve paths via `SimpleSAML\Utils\System::resolvePath()` for consistency and to reduce duplication
* `System::resolvePath()` should treat paths beginning with a drive letter as absolute

PR #782, opened 2018-02-06
Fix for Windows machines.

rvoicu:
*Actual bug*
The library cannot load the composer autoloader

*Preconditions of bug replication*
- Windows machines.
- SAML 2.0 sp
- local embedded php server or LAMP
- php version: 7.2.2

*Replication*
- start the application
- simulate a login
- the application fails when receiving the SAML message on the ACS servlet

### DIFF
--- a/lib/_autoload.php
+++ b/lib/_autoload.php
@@ -9,12 +9,12 @@
  */
 
 // SSP is loaded as a separate project
-if (file_exists(dirname(dirname(__FILE__)).'/vendor/autoload.php')) {
-    require_once dirname(dirname(__FILE__)).'/vendor/autoload.php';
-} else {  // SSP is loaded as a library
-    if (file_exists(dirname(dirname(__FILE__)).'/../../autoload.php')) {
-        require_once dirname(dirname(__FILE__)).'/../../autoload.php';
-    } else {
-        throw new Exception('Unable to load Composer autoloader');
-    }
+if (file_exists(dirname(dirname(__FILE__)) . '/vendor/autoload.php')) {
+    require_once dirname(dirname(__FILE__)) . '/vendor/autoload.php';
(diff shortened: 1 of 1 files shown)


tvdijen:
I think the proper way to fix this would be to apply https://github.com/simplesamlphp/simplesamlphp/pull/564 and then run the paths through `System::resolvePath()` before passing them to `file_exists()`.
Right now, on Windows, the path being checked would be something like `C:\dir\to\simplesamlphp/vendor/autoload.php` and you don't have to be an expert to see why that doesn't work..

tvdijen:
@thijskh / @jaimeperez can I get your opinion on this?

thijskh:
This now fails for me.

```
Fatal error: Uncaught Error: Class 'SimpleSAML\Utils\System' not found in /srv/simplesamlphp/lib/_autoload.php:12 Stack trace: #0 /srv/simplesamlphp/www/_include.php(4): require_once() #1 /srv/simplesamlphp/www/saml2/idp/SSOService.php(11): require_once('/srv/simplesaml...') #2 {main} thrown in /srv/simplesamlphp/lib/_autoload.php on line 12
```

I guess the problem is trying to use classes before the autoloader has loaded them.

tvdijen:
Oh man... 🤦‍♂️
Any suggestions on how to fix this?

thijskh:
I do not know what the exact problem is on Windows (and cannot test that).

So generally speaking you'd need to tackle this problem with only PHP's built-in functions.

jaimeperez:
> I guess the problem is trying to use classes before the autoloader has loaded them.

Trying to use a class does in fact trigger all registered autoloaders. If the class is available, the autoloader will load it and then you'll be able to use it.

tvdijen:
Will do a Windows install later today to confirm the fix

tvdijen:
I have reverted the changes to the autoloader.. Windows install is running fine with just the changes from #564

PR #795, opened 2018-02-27
[MOD] Add X_FORWARDED_PROTO and X_FORWARDED_PORT when behind a reverse proxy

changi67:
Add X-FORWARDED variables check when you are behind a SSL reverse proxy

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -83,18 +83,22 @@ class HTTP
      */
     public static function getServerHTTPS()
     {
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @changi67!

Thanks a lot for your PR. Unfortunately, this is not something that we are going to merge. We've discussed this [many](https://github.com/simplesamlphp/simplesamlphp/pull/376) [times](https://github.com/simplesamlphp/simplesamlphp/pull/491), and the conclusion is that the `X-Forwarded` set of headers are not to be trusted. The proper way to resolve this is to set the `baseurlpath` configuration option to whatever full URL matches your deployment.

PR #797, opened 2018-02-27
PSR-2 compliant whitespace in error classes

mschwager:
For #458.

Accomplished via: `phpcbf -n --standard=PSR2 lib/SimpleSAML/Error/`


mschwager:
Hmm, looks like the code climate failures were unrelated to this PR?

Issue #781, opened 2018-02-01
Theming documentation does not cover Twig


ghalse:
The [upgrade notes for 1.15](https://simplesamlphp.org/docs/stable/simplesamlphp-upgrade-notes-1.15) note that a new templating system has been introduced. However, the [theming documentation](https://github.com/simplesamlphp/simplesamlphp/blob/master/docs/simplesamlphp-theming.md) makes no reference to Twig at all.

A cursory check shows that if I've customised `header.php` and/or `footer.php` in `modules/mymodule/themes/fancytheme/default/includes/` as suggested in the theming documentation, then at least the sanitycheck module ignores the old templates from the theme in favour of the default Twig template. I would imagine that the number of things broken in this way will grow as modules migrate and that, for some time to come, we need to be maintaining both Twig and legacy versions of the templates.

Since theming to an institutional look and feel is a very common requirement, it would make sense for the upgrade notes to give some guidance on how to handle this in the upgrade notes. It would also make sense to update the theming documentation to provide updated guidance to people trying to produce new themes.

lonoak:
Hi Guy,  not sure [these](https://github.com/simplesamlphp/simplesamlphp/wiki/Twig:-Migrating-templates) [two](https://github.com/simplesamlphp/simplesamlphp/wiki/New-UI) articles by Olimpia could be a starting point. 

In any case, we have plans to create a new skin using Twig, so maybe we could help on documenting this.

@thijskh , I think @jaimeperez  mentioned the current documentation is not yet included in github... any conventions to contribute documentation to SimpleSAMLphp?  could we use the github wiki for this purpose? is markdown a good format that could be easily adapted to the official documentation pages?

ghalse:
I should have referenced this issue from #800, where I had a go at producing some documentation.

lonoak:
Oooops... true. I didn't see that other issue. Thanks Guy!

MicWit:
I am new to simplesamlpphp and have the test authentication running, but am now up to creating my own theme. I want to use the twig template (as I have used Symfony a lot in the past so know twig), but can not get it to take over. I have made my own theme which has the following files (in modules/modulename/themes/themename/default/:
includes/footer.php
includes/header.php
base.twig
_footer.twig
_header.twig
index.twig
[...]

PR #800, opened 2018-02-28
Document the new Twig templating system

ghalse:
Version 1.15 introduces a new templating system based on Twig, but the theming documentation has not been updated to reflect how this works. There is at least one module (sanitycheck) that only uses Twig templates, so custom themes need to be updated or that module is not themed. This problem will increase as more modules switch to Twig.

This is an attempt to reverse engineer some of that, and provide some stub documentation. It could probably stand to be improved by people more in the know :)


tvdijen:
Perhaps add a link to the Wiki?
https://github.com/simplesamlphp/simplesamlphp/wiki/Twig-conventions
https://github.com/simplesamlphp/simplesamlphp/wiki/Twig:-Migrating-templates
https://github.com/simplesamlphp/simplesamlphp/wiki/Migrating-translation-in-Twig

ghalse:
This might close #781

MicWit:
Note that I just left a comment on #781 as even with all the details in these 2 issues, I can still not get the theme to use twig over php.

PR #801, opened 2018-02-28
Fix missing translation for metarefresh

ghalse:
Metarefresh had an inadvertent dependency on the translation files for the aggregator module that was removed in v1.15, resulting in a "not translated" error. 

This patch removes that dependency and provides the basis for the rest of the module to be translated. I have deliberately not ported the existing translations from the aggregator module as they set the header of metarefresh to "Aggregator", which in hindsight seems wrong to start with.

### DIFF
--- a/modules/metarefresh/dictionaries/metarefresh.definition.json
+++ b/modules/metarefresh/dictionaries/metarefresh.definition.json
@@ -1,0 +1,12 @@
+{
+    "frontpage_link": {
+        "en": "Metarefresh: fetch metadata"
+    },
+    "metarefresh_header": {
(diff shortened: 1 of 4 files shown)


tvdijen:
Perhaps it is a better idea to rebase this PR for the Xnew-ui branch.. What do you think, @thijskh?

thijskh:
I was more hoping that Xnew-ui is merged into master so we avoid continuing this split brain situation.

tvdijen:
Agreed, but merging this in master now will introduce more merging conflicts... That's why I asked..

PR #802, opened 2018-02-28
superfluous code

restena-sw:
SSP 1.15 and above have PHP 5.4 as a minimum version, so the differentiation for magic quotes on versions < 5.4 is not needed any more.
Will also prepare a PR for master immediately.

### DIFF
--- a/www/_include.php
+++ b/www/_include.php
@@ -1,33 +1,5 @@
 <?php
 
-/**
- * Disable magic quotes if they are enabled.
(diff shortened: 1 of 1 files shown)


tvdijen:
Also, it has already been done on `master`;  https://github.com/simplesamlphp/simplesamlphp/commit/e8a89aafa614227c6acb03b7e421bee96ffcf128

PR #805, opened 2018-03-02
Add attributemap claim2name.php to map MS .NET Framework Claim Types to Names

gfaust-qb:
Hi,

second, clean try ;)

I needed a mapping for claim types to names several times and started to create my own attributemap.

I added an attributemap that maps Microsoft .NET Framework (standard) ClaimTypes to Names.
At the moment Names are used as used by Microsoft in their description. 
@see https://msdn.microsoft.com/en-us/library/hh159803.aspx
I recognized that these Names do not map to the names in name2claim.php or other attributemaps - but that there are inconsistencies in other mappings also.
Do you have a suggestion what Names I should use to make a valuable contribution to this project?

Guido


gfaust-qb:
Hi, 

I don't know why codeclimate fails, log shows nothing depending to this file.
Does it fail because the file is new?

Guido

tvdijen (approved):

tvdijen:
Thanks @gfaust-qb ! Very useful!!
You can safely ignore the codeclimate-failure.

tvdijen (changes requested):
To get to your question.. I think you should use the names of the AD attributes..
So you would use `c` instead of `country` for example

gfaust-qb:
Thank you!

I opened an Issue (#807) to discuss this and to offer my help to generate an mapping overview.

Guido

PR #811, opened 2018-03-13
change redis store expire time to duration seconds instead of timestamp

### LINKED ISSUE #810: redis session store expire time bug
https://github.com/simplesamlphp/simplesamlphp/blob/da119effdd152d6d9bbbec842eb1d517c768b563/lib/SimpleSAML/Store/Redis.php#L101

Redis `setex` expire times are in seconds (not unix timestamp).  When using the redis store the sessions are set to expire with a timestamp; resulting in expire times decades into the future.

This caused our redis server to eventually run out of memory and stop working.

lumine:
fixes #810

### DIFF
--- a/lib/SimpleSAML/Store/Redis.php
+++ b/lib/SimpleSAML/Store/Redis.php
@@ -91,8 +91,7 @@ class Redis extends Store
     {
         assert(is_string($type));
         assert(is_string($key));
-        $now = time();
-        assert($expire === null || (is_int($expire) && $expire > $now));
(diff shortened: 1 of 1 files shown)


lumine:
As you suggested, I have changed the assert to check if the expire timestamp is null or in the future.

thijskh:
`$time()` should be `time()`.

thijskh:
Fixed in 096162d5b4e84a8a3cba5951a707e66968c461fb, thanks!

PR #816, opened 2018-03-20
Presence and value of HTTP_X_FORWARDED_PROTO

slanthier1:
Consider that the server behind a a ssl-offloading reverse proxy is truly secure.
En consequence, that makes valid calls for the curl's cron against http://127.0.0.1 while the cookies are secured.

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -83,6 +83,11 @@ class HTTP
      */
     public static function getServerHTTPS()
     {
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi @slanthier1!

Thanks a lot for your PR, but this has been [discussed many times](https://github.com/simplesamlphp/simplesamlphp/pulls?utf8=%E2%9C%93&q=is%3Aclosed+x-forwarded+) and rejected, as the `X-Forwarded` set of headers cannot be trusted. See for example #491.

PR #817, opened 2018-03-20
Check responseHeaders presence before use them

slanthier1:
And so avoid some messages in log file.


tvdijen:
Thanks @slanthier1, this makes sense!
I've slightly modified your fix to be more efficient.

PR #819, opened 2018-03-21
Support preload virtual config file

dombenson:
The main practical motivation here is to support dynamic (virtual) population of authsources.php - since existing code paths will always attempt to load this from a literal file, via SimpleSAML_Configuration::loadFromFile


tvdijen:
Makes total sense to me @dombenson !

PR #820, opened 2018-03-21
Support metadataxml in config

dombenson:
Make the XML metadata storage source support loading from a supplied XML string, rather than necessarily opening/fetching a file.
Useful when generating configuration dynamically, as it allows the XML to be passed at that point, rather than having another abstraction to load it.

### DIFF
--- a/tests/lib/SimpleSAML/Metadata/MetaDataStorageSourceTest.php
+++ b/tests/lib/SimpleSAML/Metadata/MetaDataStorageSourceTest.php
@@ -39,18 +39,18 @@ ServiceDisplayName=\"SimpleSAMLphp Test\"
 xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"
 xmlns:fed=\"http://docs.oasis-open.org/wsfed/federation/200706\">
 <NameIDFormat>urn:oasis:names:tc:SAML:2.0:nameid-format:persistent</NameIDFormat>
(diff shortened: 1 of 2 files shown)


tests/lib/SimpleSAML/Metadata/MetaDataStorageSourceTest.php
```diff
@@ -0,0 +1,48 @@
...
+ServiceDisplayName=\"SimpleSAMLphp Test\"
+xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"
+xmlns:fed=\"http://docs.oasis-open.org/wsfed/federation/200706\">
+<NameIDFormat>urn:oasis:names:tc:SAML:2.0:nameid-format:persistent</NameIDFormat>
+<SingleSignOnService Binding=\"urn:oasis:names:tc:SAML:.0:bindings:HTTP-Redirect\" Location=\"https://saml.idp/sso/\"/>
+<SingleLogoutService Binding=\"urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect\" Location=\"https://saml.idp/logout/\"/>
+</RoleDescriptor>
+</EntityDescriptor>
+";
+        SimpleSAML_Metadata_MetaDataStorageSource::getSource(["type"=>"xml", "xml"=>$strTestXML]);
```
thijskh: Great to see these tests included! However, what is missing in the test is some asserts to actually verify the result of getSource(). So to check if you indeed find the expected entityid with some of its metadata in the result array. Could you add that?

PR #821, opened 2018-03-28
Add PHP 7.2 into Travis

Trejjam:


Trejjam:
Are there any reason why is committed composer.lock file?

thijskh:
Good to see this is already catching some actual bugs!

The discussion about composer.lock is probably best had in #628

Trejjam:
Actually, it is no longer a bug, because of merged fix in [saml2](https://github.com/simplesamlphp/saml2/pull/111). The only bug is a presence of the lock file in a library and maybe not using master of saml2 in tests.

Trejjam:
Hi, @thijskh 

please take a look.

thijskh:
Thanks!

PR #822, opened 2018-03-28
Https fixes

markwick09:
Looking to be able to configure the application url with more granularity.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -58,7 +58,7 @@ $config = array(
         //'baseURL' => 'https://example.com',
         //'protocol' => 'https',
         //'host' => 'example.com',
-        //'port' => '8443'
+        //'port' => '8443',
     //),
 
     /*
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -35,12 +35,19 @@ $config = array(
      */
     //'application' => array(
         /*
-         * The 'baseURL' configuration option allows you to specify a protocol,
-         * host and optionally a port that serves as the canonical base for all
-         * your application's URLs. This is useful when the environment
-         * observed in the server differs from the one observed by end users,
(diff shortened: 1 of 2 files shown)


tvdijen:
Honestly, all I see here is added complexity without added functionality.. What's the benefit here?

markwick09:
The specific use case is:
- One pool of web servers, behind an LB. HTTPS is terminated at the LB.
- Two different host names resolve to the same pool of servers.

Current configuration restricts that we can only override the whole URL, or nothing. The current means of detecting HTTPS traffic in the above scenario, leaving us no options.
The new config option allows overriding the protocol without necessarily overriding the host name. I'd certainly be open to implementation suggestions as alternatives to handle the above scenario.

tvdijen:
Well, the configuration is a PHP-file, so you could render the baseurl setting dynamically by replacing the domain name with `<?= $_SERVER['SERVER_NAME'] ?>`

tvdijen:
Hi @markwick09! I'm still curious about your use-case... Maybe you can elaborate on it a bit more!
Looking at what you're sketching right now, that's already possible, so I'm not sure what you're trying to accomplish...

I see you're having two nodes behind a load balancer that is terminating TLS and will forward request to the backend-nodes. In this case, your LB would host something like https://idp.example.org and that's exactly what you'd configure your backend nodes to when it comes to the `baseurlpath`... It doesn't matter if your nodes run on HTTP/80 in this case, SSP is smart enough to handle this... So what is it that SSP doesn't handle for your use-case?

jaimeperez:
I agree with Tim. I think the particular use case that justified this PR in the first place can easily be covered with the current configuration option by using variables and PHP code to build the actual value in `baseurlpath`. Therefore, I'm closing the PR for now. Feel free to argue in favor of it being reconsidered, of course.

PR #823, opened 2018-03-29
[FIX] ms-ds-consistencyguid need also to be Base64 encode

changi67:
In case of ADAzure synchronization, the main attribute used is ms-ds-consistencyguid.
It need to be Base64 encode.

### DIFF
--- a/lib/SimpleSAML/Auth/LDAP.php
+++ b/lib/SimpleSAML/Auth/LDAP.php
@@ -569,7 +569,7 @@ class SimpleSAML_Auth_LDAP
                 }
 
                 // Base64 encode binary attributes
-                if (strtolower($name) === 'jpegphoto' || strtolower($name) === 'objectguid') {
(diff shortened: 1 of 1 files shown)


tvdijen:
Just as a side-note:  wouldn't it be better to encode those attributes on their way out, using an authprocfilter, rather than on their way in?

changi67:
Why not, but i don't know how to do it. This is just a kick fix to help people using O365.

thijskh:
I'm not sure this is the wrong place per se, they're "on their way" in to the SimpleSAMLphp processing, from that point on we can assume I think we want them in the format that SAML-related stuff expects. I'd regard it as a conversion step between LDAP and SAML, which this seems to be the place to do that, no?

In any case it might be better to have some kind of globally defined list in the LDAP class with the attribute names.

tvdijen:
Perhaps an  option in the LDAP authsource configuration to define which attributes need encoding?
With all the different LDAP backends and custom schemes, it's impossible for us to keep an up-to-date list..

PR #825, opened 2018-04-03
Missing SubjectConfirmationData must not break validation

m0ark:
SubjectConfirmationData is optional , but must not break validation if absent!

### DIFF
--- a/modules/saml/lib/Message.php
+++ b/modules/saml/lib/Message.php
@@ -739,8 +739,8 @@ class sspmod_saml_Message
 
             // if no SubjectConfirmationData then don't do anything.
             if ($scd === null) {
-                $lastError = 'No SubjectConfirmationData provided';
-                continue;
+                $found = true;
+                break;
             }
 
             if ($scd->NotBefore && $scd->NotBefore > time() + 60) {


trscavo:
Jumping in here...SubjectConfirmationData is NOT optional for SSO, at least according to the SAML2 Profiles specification. I must be missing something.

m0ark:
Well, according to the SAML2 SSO Profiles you are right, it is not optional.
But according to this documentation (https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) it clearly is.
I will fix it asap.

trscavo:
FWIW, I believe the Profiles specification takes precedence here.

thijskh:
This has been changed specifically to be this way. See the discussion in #530 and commit message 980b34c782739d140afadd42bb196821adf4c66d. As Tom says SimpleSAMLphp is about the SSO profile.

Is there something not working for you that you want to change this?

m0ark:
Actually I tried to get [module-attributeaggregator](https://github.com/NIIF/simplesamlphp-module-attributeaggregator) working with encrypted assertions by using this function to decrypt and validate the response received after querying the attribute authority.
The response only contains one assertion with an empty `<SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer" />` which should be fine according to the specs, but breaks due to this restriction.

PR #831, opened 2018-04-16
Fix missing single quote in example SP Config

JosephRedfern:


jaimeperez:
Thanks a lot @JosephRedfern!

PR #834, opened 2018-04-18
Log requesterid instead of entityid for stats

tvdijen:
When it comes to statistics, I'd like to now the number of logins per originating SP, instead of per SP-proxy...

### DIFF
--- a/modules/core/lib/Auth/Process/StatisticsWithAttribute.php
+++ b/modules/core/lib/Auth/Process/StatisticsWithAttribute.php
@@ -88,7 +88,11 @@ class StatisticsWithAttribute extends \SimpleSAML\Auth\ProcessingFilter
         }
 
         $source = $this->setIdentifier('Source', $state);
-        $dest = $this->setIdentifier('Destination', $state);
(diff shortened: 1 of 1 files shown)


tvdijen:
@ghalse could you have a look and see if this doesn't conflict with your recent changes to the file?

tvdijen:
Excellent, thanks very much!

thijskh:
Do you want to know only that? So when there's at least one requesterid, you do not known which proxy was used. Or if there was more than 1 requesterid.

Would it make sense to keep logging our 'direct' peer (Destinaton) in the current place, but add also the list of requesterid's to that logline when available?

tvdijen:
That would probably be the best way, but I'm worried that it implies a shitload of work on the statistics-module. At the very least, such change would invalidate any old stats...

thijskh:
I think one way to be able to merge this is make it a configurartion of the authproc which defaults to the current behaviour?

tvdijen:
I'll just close it.. It's been such a long time since I created it

PR #837, opened 2018-04-22
add require admin auth in authenticate.php

falco76:
now authenticate.php isn't protected like others frontend pages
and anyone can discover all authsources
without this code only way to protect it is add some filter in webserver config

### DIFF
--- a/modules/core/www/authenticate.php
+++ b/modules/core/www/authenticate.php
@@ -1,6 +1,16 @@
 <?php
 
+// Load SimpleSAMLphp, configuration
 $config = SimpleSAML_Configuration::getInstance();
+$session = SimpleSAML_Session::getSessionFromRequest();
+
+// Check if valid local session exists.
+if ($config->getBoolean('admin.protectindexpage', false)) {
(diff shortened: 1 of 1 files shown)


jaimeperez:
Hi!

I'm not really sure about this, for the following reasons:

- There’s nothing really sensitive there, except for the configured authentication sources. Hiding them would be security by obscurity, so to me that’s not a good-enough argument in favour.
- I’m not entirely sure requiring admin access wouldn’t break other stuff.
- Given that in the new user interface, this authenticate.php page would be the one and only entrypoint for users to authenticate, spending time on this is kind of wasting it. Specially given that if there are multiple auth sources configured, the user will need to be able to choose (see the first item).

tvdijen:
Do we agree, that with the work on the new admin-module, this PR should be closed?
@falco76 FYI:  we're working on an admin-module for 2.0 that will hold the current admin-pages.. This will allow you to disable the admin module completely

jaimeperez:
I do agree 😄 Thanks for bringing this up Tim!

PR #840, opened 2018-04-25
Opt in to accept empty subject confirmation data.

m0ark:
By default the behaviour is unchanged to ensure SAML sso profiles are handled properly.

Enables utilization for processing other SAML messages, e.g. attribute queries, where SubjectConfirmationData is optional.


thijskh:
So this is a variant of #825.
@jaimeperez Can you shed your light in this? Since you changed this initially in #530 ?

On the code itself my comment would be that I do want to avoid negatively worded parameters, so not `deny`… but `allow`…

PR #847, opened 2018-05-08
Use DiscoHints instead of custom configuration to enable IdP discovery

m0ark:
Now using DiscoHints->IPHint to support specified SAML extension for IdP discovery


m0ark:
I re-added support for hint.cidr as proposed.

PR #848, opened 2018-05-09
Add authproc filter to test correct cardinality

ghalse:
Attribute schemas often have specific requirements for the number of
values a particular attribute may have, and violating the cardinality
rules associated with these may have unexpected and undesirable results.

The two authproc filters this change introduces allow the cardinality of
attributes to be tested. core:Cardinality is the more general case, and
allows for both a minimum and maximum number of values to be tested.
Processing will abort with an error if too few or too many are received.

core:CardinalitySingle is a special case of this for single-valued
attributes, and allows various corrective actions to be taken rather
than aborting. This allows for cardinality mistakes to be "fixed" in a
predictable way, and can also be used to construct single-valued
attributes from multi-valued ones.

### DIFF
--- a/tests/modules/core/lib/Auth/Process/CardinalitySingleTest.php
+++ b/tests/modules/core/lib/Auth/Process/CardinalitySingleTest.php
@@ -1,0 +1,139 @@
+<?php
+// Alias the PHPUnit 6.0 ancestor if available, else fall back to legacy ancestor
+if (class_exists('\PHPUnit\Framework\TestCase', true) and !class_exists('\PHPUnit_Framework_TestCase', true)) {
+    class_alias('\PHPUnit\Framework\TestCase', '\PHPUnit_Framework_TestCase', true);
+}
+
+/**
+ * Test for the core:CardinalitySingle filter.
+ */
+class Test_Core_Auth_Process_CardinalitySingleTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * Helper function to run the filter with a given configuration.
+     *
+     * @param  array $config The filter configuration.
+     * @param  array $request The request state.
+     * @return array  The state array after processing.
+     */
+    private static function processFilter(array $config, array $request)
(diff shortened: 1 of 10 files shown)


ghalse:
This pull request is a refinement of the [simplesamlphp-module-cardinality](https://github.com/safire-ac-za/simplesamlphp-module-cardinality) module I wrote for SAFIRE's own use case, after a brief discussion with @jaimeperez.

ghalse:
Hmm, the PHPunit tests fail here but not in my dev environment nor via [travis-ci in the original module](https://travis-ci.org/safire-ac-za/simplesamlphp-module-cardinality/builds/377010427). If anyone knows why they fail here/how to fix them/how to do the tests better, I'll happily amend the pull request.

tests/modules/core/lib/Auth/Process/CardinalitySingleTest.php
```diff
@@ -0,0 +1,141 @@
+<?php
+// Alias the PHPUnit 6.0 ancestor if available, else fall back to legacy ancestor
+if (class_exists('\PHPUnit\Framework\TestCase', true) and !class_exists('\PHPUnit_Framework_TestCase', true)) {
```
tvdijen: The only thing I'm not sure about is the necessity of this if-statement...
Why not use the namespaced version after `extends`? Is this some backwards compatibility-thing for 1.14?

tvdijen:
Gah, was that intentional?!  It looked like a typo

ghalse:
There was a typo - a missing i. You just introduced another whilst fixing the first :)

tvdijen:
Hah, I see it now yes!
Anyway, to give you an update.. I have made a couple of minor changes / additions and gave the seal of approval :)  I leave it up to Thijs of Jaime to merge this after giving this a second pair of eyes

thijskh:
Great addition!

Issue #20, opened 2014-02-27
OAuth DataStore: Access-tokens don't expire (was #298)


sspghost:
_From [thomas.z...@freenet.de](https://code.google.com/u/108330137531469242352/) on March 16, 2010 17:53:35_

The simplesamlphp OAuthDataStore implementation 

sspmod_oauth_OAuthStore

assigns a validity to OAuth access tokens upon inserting them into the DB.
By default that is 24 hours, like here

``` php
$this->store->set('access', $token->key, $consumer->key, $token,
$this->config->getValue('accessTokenDuration', 60*60*24) );
```

However, upon looking up the access tokens, that validity is not checked in
the sspmod_core_Storage_SQLPermanentStorage, giving OAuth access tokens an
indefinite validity.

A very fast fix is adding

``` php
$this->store->removeExpired();
```

to the beginning of function lookup_token($consumer, $token_type, $token).

_Original issue: http://code.google.com/p/simplesamlphp/issues/detail?id=298_

PR #852, opened 2018-05-10
SQLPermanentStorage:  Check for expiration

tvdijen:
Adresses https://github.com/simplesamlphp/simplesamlphp/issues/20

The suggested fix in the issue is not a proper one.. There's a cron hook for removing expired entries.
Calling `$this->store->removeExpired();` on every single lookup would be horrible overkill.

We may never expose expired data in the first place and that's what this PR does; filtering it out.


thijskh:
So... now that this class had tests, add testcase(s) for this change aswell?

Issue #482, opened 2016-09-30
Return the group name rather than the group DN in ldap:AttributeAddUsersGroups


vittala:
At the moment, it seems that the group values returned from ldap:AttributeAddUsersGroups are always the DNs of the group. The configuration for this module allows for `attribute.groupname`, but changing this doesn't seem to have any effect on the module's results.

If the purpose of `attribute.groupname` is to select which LDAP attribute to return as the group's name, then this commit will do so:
https://github.com/vittala/simplesamlphp/commit/f731853a100846c7c48d93bdc4f4598fcb1c6817

Regards
Vittal

tvdijen:
Fixed on master through fb0e79265abe0530b4ed48608c3b110a552b967d

PR #853, opened 2018-05-10
ldap: fix groupname-setting in AttributeAddUsersGroups

tvdijen:
Fixes https://github.com/simplesamlphp/simplesamlphp/issues/482

If attribute.groupname is set in the module configuration use the configured attribute as the group name. The setting wasn't used at all...


tvdijen:
Pushed to master through fb0e79265abe0530b4ed48608c3b110a552b967d

PR #857, opened 2018-05-15
 Return XML parser error in metarefresh

ghalse:
If metarefresh refuses to load remote XML, it would be useful to be able to see the error message from the underlying SAML2 or XML parsers in order to determine the root cause of the error.


tvdijen:
Great @ghalse, thanks!

Issue #793, opened 2018-02-22
Session_cache_limiter(): Cannot change cache limiter when session is active (PHP 7.2)


tonis-ls:
We are updating SSP 1.15.2 from PHP 5.5 to 7.2 and in the log files we find errors like:

Apache log:

[Thu Feb 22 15:41:09.835044 2018] [php7:warn] [pid 11754] [client 172.17.15.231:52937] PHP Warning:  session_cache_limiter(): Cannot change cache limiter when session is active in /var/www/simplesamlphp/lib/SimpleSAML/SessionHandlerPHP.php on line 122

And simplesamlphp log:

Feb 22 15:41:09 idp ERROR [9a1f0134d6] SimpleSAML_Error_Exception: Error 2 - session_cache_limiter(): Cannot change cache limiter when session is active

[https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-1.15/lib/SimpleSAML/SessionHandlerPHP.php#L122](url)

tvdijen:
Thanks for reporting this @tonis-ls !
It seems that `session_cache_limiter()` must be called before `session_start()`...
Could you swap line 121 and 122 and see if that fixes the issue for you?

tonis-ls:
Thank you @tvdijen for the quick response. 
The errors are gone after switching lines 121 and 122 in SessionHandlerPHP.php.

tvdijen:
That's great, thanks for testing!
I've fixed this on both the `master` and the `simplesaml-1.15` branch.

iamtankist:
Could be that I'm doing something wrong, but on latest 1.15.4 I'm still getting right after installation at `simplesaml/module.php/core/frontpage_welcome.php`

``` 
Warning: session_set_cookie_params(): Cannot change session cookie parameters when session is active in /srv/simplesaml/lib/SimpleSAML/SessionHandlerPHP.php on line 355

Warning: session_id(): Cannot change session id when headers already sent in /srv/simplesaml/lib/SimpleSAML/SessionHandlerPHP.php on line 363

Warning: session_cache_limiter(): Cannot change cache limiter when headers already sent in /srv/simplesaml/lib/SimpleSAML/SessionHandlerPHP.php on line 119

Warning: session_cache_limiter(): Cannot change cache limiter when headers already sent in /srv/simplesaml/lib/SimpleSAML/SessionHandlerPHP.php on line 121

[...]

PR #863, opened 2018-05-23
Make sure that session is closed and only then update cookie params

smotornyuk:
As mentioned by @iamtankist in #793 , there is a potential problem if you are trying to update cookie parameters having an active session(https://secure.php.net/session_set_cookie_params) 

There are not much to do for a fix - I just put the call to `session_set_cookie_params` right after closing the previous session, not before as it was

### DIFF
--- a/lib/SimpleSAML/SessionHandlerPHP.php
+++ b/lib/SimpleSAML/SessionHandlerPHP.php
@@ -343,6 +343,11 @@ class SessionHandlerPHP extends SessionHandler
             );
         }
 
+        if (session_id() !== '') {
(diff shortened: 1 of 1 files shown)


thijskh:
I'm not sure it's complete - we call `session_set_cookie_params(...)` also in other places, and also `session_id(<some-value>` seems affected by this?

aelliott1485:
how soon could this change be merged into other branches (e.g. the 1.16-rc1)? Can I assist with that?

tvdijen:
Merged it into 1.16...

Issue #860, opened 2018-05-22
Unit test suite not executing fully


unt01d:
It looks like there are several issues with the unit test suite at the moment which are producing unwanted log output and causing the test run to exit early.

Here's the output I'm getting (this is reflected in the builds too):

```
$ vendor/bin/phpunit -c tools/phpunit/phpunit.xml
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

................May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
May 22 13:00:41 SimpleSAMLphp WARNING [CL78e7bc2e] SimpleSAML_Configuration::getBaseURL() is deprecated, please use getBasePath() instead.
[...]

PR #864, opened 2018-05-25
Fixes cardinality auth filter unit tests (#860)

unt01d:
This fixes the unit tests for the Cardinality/CardinalitySingle auth process filters.

The intention is to assert that certain requests result in a redirect.  In order to do this, we need to mock the HTTP helper class, and in order to do that we need to create a non-static HTTP wrapper which we can inject into the cardinality filter instances.

### DIFF
--- a/tools/phpunit/phpunit.xml
+++ b/tools/phpunit/phpunit.xml
@@ -23,6 +23,7 @@
             <exclude>
                 <directory>./../../vendor/</directory>
                 <directory>./../../tests/</directory>
+                <file>./../../lib/SimpleSAML/Utils/HttpAdapter.php</file>
             </exclude>
         </whitelist>
     </filter>
--- a/lib/SimpleSAML/Utils/HttpAdapter.php
+++ b/lib/SimpleSAML/Utils/HttpAdapter.php
@@ -1,0 +1,211 @@
+<?php
+
+namespace SimpleSAML\Utils;
+
+/**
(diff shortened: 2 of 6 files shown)


tools/phpunit/phpunit.xml
```diff
@@ -23,6 +23,7 @@
             <exclude>
                 <directory>./../../vendor/</directory>
                 <directory>./../../tests/</directory>
+                <file>./../../lib/SimpleSAML/Utils/HttpAdapter.php</file>
```
unt01d: For some reason, the code coverage process was trying to include this file multiple times, which was killing it.  So the simplest thing to do was to ignore it (it doesn't need to be included in any case...).

thijskh:
Testsuite now fails (again…) on psalm. I've committed a fix on master, maybe if you rebase on current master the tests will be green...

thijskh:
@ghalse

unt01d:
Done, but the 5.4 build seems to be failing now @thijskh...  :confused:

thijskh:
That is an occasional transient problem in one test, fixable by restarting that job (which I did). Strange that coverage does not seem to report correctly anymore. But otherwise fine :+1:

Would like to see a comment from @ghalse since it mainly concerns his work.

ghalse:
Sorry I've taken so long to respond - from what I've read recently, this certainly looks like the right sort of way to do things. And perhaps provides people with a useful hint on how to do unit tests for redirects later on :)

PR #866, opened 2018-05-28
Improve MetaDataStorageHandlerPdo::getMetaData() performance

manu0401:
By using an appropriate WHERE clause in the SQL statement,
we can avoid loading the whole dataset in getMetaData(),
which brings an welcome performance improvement with large
Metadata sets.

### DIFF
--- a/lib/SimpleSAML/Metadata/MetaDataStorageHandlerPdo.php
+++ b/lib/SimpleSAML/Metadata/MetaDataStorageHandlerPdo.php
@@ -142,6 +142,48 @@ class SimpleSAML_Metadata_MetaDataStorageHandlerPdo extends SimpleSAML_Metadata_
         return $metadataSet;
     }
 
+    /**      
+     * Retrieve a metadata entry.
+     *      
+     * @param string $entityId The entityId we are looking up.
+     * @param string $set The set we are looking for metadata in.
+     *
+     * @return array An associative array with metadata for the given entity, or NULL if we are unable to
+     *         locate the entity.
+     */
+    public function getMetaData($entityId, $set)      
+    {       
+        assert('is_string($entityId)');
+        assert('is_string($set)');
+
+        $tableName = $this->getTableName($set);
+
+        if (!in_array($set, $this->supportedSets, true)) {
+            return null;
+        }
+
+        $stmt = $this->db->read("SELECT entity_id, entity_data FROM $tableName WHERE entity_id=:entityId", array('entityId' => $entityId));
+        if ($stmt->execute()) {
+            $rowCount = 0;
+
+            while ($d = $stmt->fetch()) {
+                if (++$rowCount > 1) {
+		    SimpleSAML\Logger::warning("Dulicate match for $entityId in set $set");
+		    break;
+                }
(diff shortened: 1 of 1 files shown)


lib/SimpleSAML/Metadata/MetaDataStorageHandlerPdo.php
```diff
@@ -142,6 +142,48 @@ public function getMetadataSet($set)
...
+        }
+
+        $stmt = $this->db->read("SELECT entity_id, entity_data FROM $tableName WHERE entity_id=:entityId", array('entityId' => $entityId));
+        if ($stmt->execute()) {
+            $rowCount = 0;
+
+            while ($d = $stmt->fetch()) {
+                if (++$rowCount > 1) {
+		    SimpleSAML\Logger::warning("Dulicate match for $entityId in set $set");
+		    break;
```
tvdijen: Indentation is off here

lib/SimpleSAML/Metadata/MetaDataStorageHandlerPdo.php
```diff
@@ -168,8 +168,8 @@ public function getMetaData($entityId, $set)
 
             while ($d = $stmt->fetch()) {
                 if (++$rowCount > 1) {
-		    SimpleSAML\Logger::warning("Dulicate match for $entityId in set $set");
-		    break;
+                    SimpleSAML\Logger::warning("Dulicate match for $entityId in set $set");
```
jaimeperez: Looks like a typo: _Duplicate_

lib/SimpleSAML/Metadata/MetaDataStorageHandlerPdo.php
```diff
@@ -142,6 +142,48 @@ public function getMetadataSet($set)
...
+     * @param string $entityId The entityId we are looking up.
+     * @param string $set The set we are looking for metadata in.
+     *
+     * @return array An associative array with metadata for the given entity, or NULL if we are unable to
+     *         locate the entity.
+     */
+    public function getMetaData($entityId, $set)      
+    {       
+        assert('is_string($entityId)');
+        assert('is_string($set)');
```
thijskh: assert should not have string arguments (since PHP 7.2). Should be like:
```
        assert(is_string($entityId));
        assert(is_string($set)); 
```

thijskh:
> (all SimpleSAMLphp source code should be updated, not only this contribution)

It already has...

thijskh:
Great improvement! One question: according to the original implementation and the docblock it should return `null` when the entity is not found. Does this implementation actually do that?

manu0401:
Thijs Kinkhorst <notifications@github.com> wrote:

> Great improvement! One question: according to the original implementation
> and the docblock it should return `null` when the entity is not found.
> Does this implementation actually do that?

On no match, we return an unset $data which means the value is null, but
there will be a PHP Notice:  Undefined variable: data. 

That should be improved, but I do not have my repo at hand right now.

-- 
Emmanuel Dreyfus
http://hcpnet.free.fr/pubz
manu@netbsd.org

thijskh:
I've merged it and added the initialization of `$data` with `null`.

Issue #841, opened 2018-04-25
add_authority not working in smartattributes


djdevin:
https://github.com/simplesamlphp/simplesamlphp/blob/57df7b25fa558531c1e76aa686c5daec199b3fdc/modules/smartattributes/lib/Auth/Process/SmartID.php#L85-L89

It looks like `$request['saml:AuthenticatingAuthority'][0]` should be `$request['saml:sp:State']['saml:AuthenticatingAuthority'][0]`

This is the only way I could get the default configuration on https://simplesamlphp.org/docs/1.15/smartattributes:smartattributes to work.

tvdijen:
Proposed change makes sense.. I've created a PR for this

PR #867, opened 2018-05-28
smartattributes: fix bug

tvdijen:
Fixes https://github.com/simplesamlphp/simplesamlphp/issues/841

### DIFF
--- a/modules/smartattributes/lib/Auth/Process/SmartID.php
+++ b/modules/smartattributes/lib/Auth/Process/SmartID.php
@@ -1,117 +1,120 @@
 <?php
 
-class sspmod_smartattributes_Auth_Process_SmartID extends SimpleSAML_Auth_ProcessingFilter {
+class sspmod_smartattributes_Auth_Process_SmartID extends SimpleSAML_Auth_ProcessingFilter
+{
+    /**
+     * Which attributes to use as identifiers?
+     *
+     * IMPORTANT: If you use the (default) attributemaps (twitter2name, facebook2name,
+     * etc., be sure to comment out the entries that map xxx_targetedID to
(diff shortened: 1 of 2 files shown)


modules/smartattributes/lib/Auth/Process/SmartID.php
```diff
@@ -1,116 +1,123 @@
...
+     * eduPersonTargetedID, or there will be no way to see its origin any more.
+     */
+    private $_candidates = array(
+        'eduPersonTargetedID',
+        'eduPersonPrincipalName',
+        'openid',
+        'facebook_targetedID',
+        'twitter_targetedID',
+        'windowslive_targetedID',
+        'myspace_targetedID',
```
jaimeperez: Is this related?

tvdijen: It's just whitespace fixes you're looking at.. This code is not new.

jaimeperez: No, I was skipping whitespace changes explicitly, and then the `myspace_targetedID` appeared as new. That's why I was wondering whether it was related to the bugfix or not... 🤔

thijskh: I guess a merge error since the line has been removed in 852ee7e32a75aac93263e6b465567eedfd0f02d5.

jaimeperez: In that case, I think it's a good idea to remove any leftovers. Other than that: 👍

tvdijen: I'll fix it later today

PR #868, opened 2018-05-28
consentAdmin: Add attributes.exclude option to correspond with the Consent module

ghalse:
The consent:Consent authproc filter has a (currently undocumented) noconsentattributes option that allows specified attributes to be removed from the consent hash calculation. This has no corresponding option in the consentAdmin module, so using the noconsentattribute option to remove attributes means the consentAdmin module will always show this as attribute values having been changed.

This patch changes that behaviour by adding a new 'attributes.exclude' configuration option in consentAdmin, that allows people to specify attributes that should not be used for hash calculation and match their corresponding option in consent:Consent. It also deprecates the noconsentattributes option in consent:Consent, replacing it with a similarly named 'attributes.exclude' option.

The default behaviour of the consentAdmin module should be unchanged for anyone who doesn't explicitly update their configuration.

Closes: #531 - this pull request is effectively a rebase of #531 which has got corrupted over time.

### DIFF
--- a/modules/consent/lib/Auth/Process/Consent.php
+++ b/modules/consent/lib/Auth/Process/Consent.php
@@ -116,7 +116,15 @@ class sspmod_consent_Auth_Process_Consent extends SimpleSAML_Auth_ProcessingFilt
             $this->_hiddenAttributes = $config['hiddenAttributes'];
         }
 
-        if (array_key_exists('noconsentattributes', $config)) {
+        if (array_key_exists('attributes.exclude', $config)) {
+            if (!is_array($config['attributes.exclude'])) {
+                throw new SimpleSAML_Error_Exception(
+                    'Consent: attributes.exclude must be an array. '.
+                    var_export($config['attributes.exclude'], true).' given.'
+                );
+            }
+        } elseif (array_key_exists('noconsentattributes', $config)) {
+            SimpleSAML\Logger::warning("The 'noconsentattributes' option has been deprecated in favour of 'attributes.exclude'.");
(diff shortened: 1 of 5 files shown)


modules/consent/lib/Auth/Process/Consent.php
```diff
@@ -116,7 +116,15 @@ public function __construct($config, $reserved)
...
         }
 
-        if (array_key_exists('noconsentattributes', $config)) {
+        if (array_key_exists('attributes.exclude', $config)) {
+            if (!is_array($config['attributes.exclude'])) {
+                throw new SimpleSAML_Error_Exception(
+                    'Consent: attributes.exclude must be an array. '.
+                    var_export($config['attributes.exclude'], true).' given.'
+                );
+            }
```
thijskh: Nothing is actually set/saved in this case?

ghalse: Well spotted! Seems I lost a line when I copied over the original diff.

thijskh: Given that the module already has some test cases, would it be easy to add two for the two different settings so something like this would have been caught?

ghalse: Looking at exactly that right now :)

ghalse: There doesn't seem to be an easy way to test the functionality within process() - at least not without some substantive changes, because it issues a redirect and we'd need to examine the state after that completes (the problem with exit). But I've just added one that tests the correct setting of the instance private variables using a ReflectionClass.

PR #869, opened 2018-05-29
Namespace remaining non-namespaced classes

tvdijen:


tvdijen:
Will do! Do you want me to remove the class completely since it's been deprecated for so long already? Or just get rid of the references..

tvdijen:
If anyone could run a few tests with this branch that would be wonderful..

PR #872, opened 2018-06-11
Ignore the enable file for modules

clinta:
Ignore the enable file in modules


tvdijen:
Shouldn't the pattern be like `/modules/*/enable`?
And shouldn't we do the same for 'disable'?

thijskh:
This file is inside the `modules/` directory already. But I agree on your idea about `disable` (although it will be more rare).

tvdijen:
Alright, I've added it

PR #875, opened 2018-06-15
Getting Signature algorithm from SP and overriding default SHA256

lampdev:
XMLSecurityKey::RSA_SHA256 algorithm is hardcoded in current master for SP signature verification. Got a case when SP sends RSA1 in response and it fails.

### DIFF
--- a/modules/saml/lib/Message.php
+++ b/modules/saml/lib/Message.php
@@ -199,7 +199,7 @@ class Message
 
         $lastException = null;
         foreach ($pemKeys as $i => $pem) {
-            $key = new XMLSecurityKey(XMLSecurityKey::RSA_SHA256, array('type' => 'public'));
+            $key = new XMLSecurityKey($element->getSignatureMethod() ? $element->getSignatureMethod() : XMLSecurityKey::RSA_SHA256, array('type' => 'public'));
             $key->loadKey($pem);
 
             try {


lampdev:
Hi @jaimeperez ,

thanks for your response. I'm able to recreate this issue now on dev-master, but wondering if I can replace the hostnames/company names with placeholders. I wouldn't like to disclose the organizations working as IdP and SP in this integration. What I can tell now is that in SAML response I'm seeing Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1" value and the code tries to validate the signature with RSA-SHA256

jaimeperez:
Hi @lampdev!

When you say:

> the code tries to validate the signature with RSA-SHA256

What code is it exactly that you think is trying to validate with RSA-SHA256?

lampdev:
I mean RobRichards\XMLSecLibs\XMLSecurityKey::verifyOpenSSL() and specifically return openssl_verify($data, $signature, $this->key, $algo); where $algo comes as RSA-SHA256 when the IdP returns SHA1

jaimeperez:
I have just verified it myself with an SP in `dev-master`, and it works perfectly fine. If the IdP signs the response with RSA-SHA1, SHA1 is used as the digest algorithm. If RSA-SHA256 is specified in the signature, then SHA256 is used, and so on with the rest of the options available.

Bear in mind that *RSA-SHA256* is **not** an algorithm that you will find in `$algo`. That's a digital signature algorithm, while `$algo` identifies a digest algorithm. Possible values there will be `1` (integer, indicating SHA1), `SHA256`, `SHA384` or `SHA512`.

What is it that the IdP is signing? The response or the assertion? Also, can you paste an example signature here?

PR #892, opened 2018-07-24
Update logoutstore table migration syntax for PostgreSQL

### LINKED ISSUE #891: Logoutstore migration is not compatible with PostgreSQL
sspmod_saml_SP_LogoutStore->createLogoutTable will attempt to change the column length of _authSource using the MODIFY syntax. This is not compatible with PostgreSQL.

akinspe:
Fixes #891

### DIFF
--- a/modules/saml/lib/SP/LogoutStore.php
+++ b/modules/saml/lib/SP/LogoutStore.php
@@ -19,7 +19,17 @@ class sspmod_saml_SP_LogoutStore {
 			return;
 		} elseif ($tableVer === 1) {
 			/* TableVersion 2 increased the column size to 255 which is the maximum length of a FQDN. */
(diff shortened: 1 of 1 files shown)


thijskh:
Is it necessary to make a switch statement, can't it be one SQL query that is more generally accepted?

akinspe:
Unfortunately I am not aware of a common DDL syntax that is valid in all supported drivers. This is one area of SQL that is not universal.

thijskh:
Thanks. I've ported it to 1.16 and master aswell (could not merge/cherry-pick it automatically because files have changed completely re formatting in the meantime).

akinspe:
Thanks! I look forward to the release of 1.15.5

thijskh:
I am not sure that that will happen, I rather expect 1.16 to be released soon.

PR #902, opened 2018-08-01
Import assets from Xnew-ui branch

tvdijen:
Reduce footprint of #717 

With these assets in place, Twig-templates can be imported on a per-module basis..


tvdijen:
Just want your approval on merging this into master

Issue #905, opened 2018-08-06
PHP Fatal error:  Call to a member function loadTranslations()


kamoex84:
PHP Fatal error:  Call to a member function loadTranslations() on null in /var/www/html/simplesamlphp/lib/SimpleSAML/Locale/Localization.php on line 221

tvdijen:
Hey @kamoex84 ! I assume you're running the `master` branch?

tvdijen:
Well, the short-term solution would be to use a released version..
`master` currently has some issues that I'm unable to resolve

tvdijen:
@jaimeperez what would be the appropriate way to fix this?
`master` is working fine with backend set to `gettext/gettext`, not with `SimpleSAMLphp`. This is because the current code will always use Twig-templates when available (which rely on gettext)..
Should we add another condition so that Twig-templates will only be the preferred case when backend is `gettext/gettext`?

kamoex84:
Duly noted, @tvdijen . Thanks.

jaimeperez:
I think the proper solution would be to actually fix this so that it works. I'm pretty certain that it was working back in the day, but wouldn't bet money on it 😜 

Should be relatively simple. This is one of the advantages of patching twig to use our own translation functions: we can substitute the call to Oscar Otero's gettext implementation with a call to our own translation methods. Still, I don't have the code right in front of me so it might be that it's a bit more complicated than it should.

At the same time, I'm starting to wonder whether it's a good idea to have automatic detection of twig templates or not, since that would make it impossible to have both the old and the new UI coexist... It might actually be easier to disable that _automagical_ detection and force each UI to use its own translation subsystem.

tvdijen:
A proper fix would always be better, but that probably goes over my head...
I think the best I can do right now is open a new PR as a successor of #717 and document the things that are broken on `master` right now..

tvdijen:
@kamoex84: If you don't mind I'm closing this issue and we're continuing this in the bigger picture at #906...
Thanks for reporting this!

tvdijen:
@jaimeperez Isn't this what's messing us up here?
https://github.com/simplesamlphp/simplesamlphp/blob/89bff5a399d26055f96ed100cb6995043c32a9ba/lib/SimpleSAML/XHTML/Template.php#L230

jaimeperez:
Most likely, yes...

jaimeperez:
[...]

PR #906, opened 2018-08-06
Fix new UI

tvdijen:
Last week, the Xnew-ui branch was accidentally merged into master. There are however still a couple of issues that need to be taken care of. I will keep track of the remaining issues here, making this PR the successor of #717. Help is very much appreciated!

The aim here is to make `master` working properly ASAP.

Tagging previous contributors:  @Olimpiamagliulo, @thijskh, @dnmvisser, @pmeulen, @joostd


tvdijen:
Checklist:

- [ ]  Old style templates do not work any more, since the new Twig-templates take precedence over the old ones. `language.i18n.backend => 'SimpleSAMLphp'` therefore is not doing what it's supposed to do.
See Jaime's comment on this issue here: https://github.com/simplesamlphp/simplesamlphp/issues/905#issuecomment-410634318

- [ ] consent: templates/getconsentform.twig was fixes in the first commit of this PR but still looks like sh*t and needs work
- [ ] consent: pressing the 'no consent' button leads to a broken redirect
- [ ] NOSTATE errors are not translated properly (tags are shown)

tvdijen:
I'd like to perform an intermediate merge to master, since I've done a ton of work on whitespace/indentation for the complete `modules/` directory and I'd hate to see that causing merging issues. Any objections @thijskh @jaimeperez ?

jaimeperez:
> - [ ]  Old style templates do not work any more, since the new Twig-templates take precedence over the old ones and `language.i18n.backend => 'SimpleSAMLphp'` is therefore not doing what it's supposed to do.
See Jaime's comment on this issue here: https://github.com/simplesamlphp/simplesamlphp/issues/905#issuecomment-410634318

See [this comment](https://github.com/simplesamlphp/simplesamlphp/issues/905#issuecomment-425088284) as well.

> - [ ] add documentation on currently undocumented `theme.controller`-setting

This was introduced to allow modules to perform certain tasks when displaying pages. I can probably try to document it, although I think it really belongs together with the documentation on custom theming.

> - [ ] Modules that have hooks on the config frontpage need a translation in core/locales... Right now, the hooks for statistics, sanitycheck and memcachemonitor (list not limited) are broken. I think this undermines the 'hooks'-model.. Any module should be able to hook into the config-page.

Fully agree. We need to find a solution to this. Moving away from pure-PHP templates makes it harder to run hooks. We need to investigate whether twig provides a way to offer some similar functionality.

I think we also need to work on a proper deployment mechanism using _webpack_ or something like that. I'd move all source code (including CSS and JS) outside of `www/`, then have a build process that pulls the dependencies, packs everything together and dists the bundles to `www/`. This is essential to optimise the use of CSS and JavaScript, such as:

* we have the optimal amount of JS/CSS files.
* we can minimise bundles, making them as small as possible.
* we can implement cache-busting mechanisms.
* we can also implement proper content security policies, forbidding any inline JS.

Since I've been working a lot with all that stuff lately, that's also something I can probably work wit

Maybe we should start using [projects](https://github.com/simplesamlphp/simplesamlphp/projects/2) again to organise what's needed and coordinate the work?

PR #911, opened 2018-08-23
Upgrade gettext to latest version

jasverix:
Checked all usages and found no incompatibilities.

I cannot use this package in my project because of the gettext limitation.


tvdijen:
On second thought, the version constraint should probably be changed to `^4.6`

PR #916, opened 2018-08-29
Make authfacebook compatible with new facebook response query params

pradtke:
Do not include query parameters when constructing redirect uri for use with
the `access_token` endpoint.

Background:
Depending on your Facebook app configuration Facebook may return additional query
parameters after a a user authorizes an app. The module was incorrectly adding
these extra params to its redirect uri and generating a redirect uri that is
not in the app's whitelist. The new param that Facebook has added to response is enforce_https=1

### DIFF
--- a/modules/authfacebook/extlibinc/base_facebook.php
+++ b/modules/authfacebook/extlibinc/base_facebook.php
@@ -142,16 +142,6 @@ abstract class BaseFacebook
     );
 
     /**
-     * List of query parameters that get automatically dropped when rebuilding
-     * the current URL.
-     */
-    protected static $DROP_QUERY_PARAMS = array(
-        'code',
-        'state',
-        'signed_request',
-    );
-
-    /**
      * Maps aliases to Facebook domains.
      */
     public static $DOMAIN_MAP = array(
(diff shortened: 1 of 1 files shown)


thijskh:
Makes me wonder why the code was there originally. i.e. which url params it intended to keep which are now disregarded.

In any case, I’m fine with the changes if you’ve tested them - I cannot verify this myself.

tvdijen:
Maybe because we're using a version of the PHP library that isn't even available on Github anymore...
They've moved on two major versions already;  https://github.com/facebook/php-graph-sdk

pradtke:
@thijskh 
> Makes me wonder why the code was there originally. i.e. which url params it intended to keep which are now disregarded

It was used before PR #767.  Before then the module added extra query parameters to the redirect uri that were used for determining state.  Facebook made changes (that were addressed in that PR) where any dynamic stuff needed to be passed to them in a 'state' query parameter and that the redirect uri needed to remain static.

> Facebook tends to change their API pretty often

Yeah, though I feel part of the issue is that the underlying Facebook moudle code had functions for all aspects of the Facebook API which aren't needed for doing authentication. More code == more bugs. I hoping this fall to move our Facebook logins to our generic [OAuth2 client module](https://github.com/cirrusidentity/simplesamlphp-module-authoauth2) which mostly just wraps phpleague Oauth libraries, which in turn that are more widely used and tested..

PR #919, opened 2018-09-03
Fixed interoperable SAML 2 profile link.

RoSk0:


thijskh:
I don't think I agree with this change. The new location is just a draft, still under heavy discussion. And simpleSAMLphp is not necessaily compatible with that draft since many things have changed.

PR #927, opened 2018-09-12
Modification to checkURLAllowed

gavinjackson:
Hello,

I hit a problem connecting my IDP to a remote SP that had an underscore in the host name.
The following stackoverflow article provided an alternate method (using parse_url rather than filter_var) for checking domain validity.
https://stackoverflow.com/questions/39539468/php-filter-validate-url-not-finding-subdomains-with-underscore

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -328,7 +328,7 @@ class HTTP
         $url = self::normalizeURL($url);
 
         // check the validity of the URL
-	if (parse_url($url,PHP_URL_HOST) === false) {
+        if (parse_url($url,PHP_URL_HOST) === false) {
             throw new \SimpleSAML_Error_Exception('Invalid URL: '.$url);
         }
 
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -327,8 +327,9 @@ class HTTP
         }
         $url = self::normalizeURL($url);
 
-        if (filter_var($url, FILTER_VALIDATE_URL) === false) {
-            throw new Error\Exception('Invalid URL: '.$url);
+        // check the validity of the URL
+	if (parse_url($url,PHP_URL_HOST) === false) {
+            throw new \SimpleSAML_Error_Exception('Invalid URL: '.$url);
         }
 
         // get the white list of domains


thijskh:
Well, hostnames with a `_` in it are in fact invalid. Not so sure we should be 'downgrading' the entire check to support hostnames that are invalid to begin with?

gavinjackson:
> Well, hostnames with a `_` in it are in fact invalid. Not so sure we should be 'downgrading' the entire check to support hostnames that are invalid to begin with?

This is the SP I am trying to integrate with (this is an external vendor that I need to integrate with):

https://ffms_virtual_pr291.qa.theretentionpeople.com/api/users/login-saml-metadata

So - simplesamlphp fails to logout with the existing code (login is fine though ...). I guess my point is that I am seeing this convention used in the wild (and would prefer not to manually patch simplesamlphp every time there is an update).

Should I go back to the vendor and tell them to use a valid domain name?

nathanjrobertson:
They're not completely invalid in DNS. In fact, they're required by SRV and TXT records. BIND (the DNS server) has flip-flopped a number of times over the years on whether to be strict to RFC 952 or not on this one that it, to the point where it is now configurable (the "check-names" directive). Indeed, RFC 1035 section 2.3.1 uses very weak language around adherence to earlier RFC naming conventions (the example quoted below speaks to SMTP, but is only an example to be followed in a wider context):

> For example, when naming a mail domain, the user should satisfy both the rules of this memo and those in RFC-822.  When creating a new host name, the old rules for HOSTS.TXT should be followed.  This avoids problems when old software is converted to use domain names.

Note, it is **only** a SHOULD (not a MUST), and really they're only saying that for backwards compatibility with older software (keeping in mind RFC 1035 was written in November 1987, such older software is now way past obsolete). Now, it is true that that RFC does reinforce the "digits, letters and hyphon" of RFC 952, but that was before Active Directory and Kerberos SRV and TXT records made underscores compulsory. We now live in a world where underscores exist, and strictly speaking the RFC is recommending we implement to avoid problems with other software.

Yes, the attached patch breaks an ancient RFC. However, because it is configurable in the largest DNS server, largely configured in the wild to allow the underscore (the fact you can click on Gavin's link and it works says that your web browser, operating system DNS resolver, ISP DNS server, etc. all allow underscores), says that it is it is now largely accepted (despite what RFC 952 originally specified). The pragmatic view should be to follow Postel's law of robustness (be conservative in what you send, lenient in what you accept), and admit that this ship has sailed. Even if you wanted not to take the pragmatic approach, there are better places than an SP for a SAML SSO to implement this strictness. That's the job of BIND, web browsers, OS DNS resolvers, etc. to do. The folks who write those daemons, libraries and applications have considered this many times over the years, and taken the pragmatic approach. I suggest simpleSAMLphp follow this approach and accept the patch Gavin has provided.

tvdijen:
I think the only RFC that applies here is [RFC 3986](http://www.faqs.org/rfcs/rfc3986.html). It's not allowed.

lib/SimpleSAML/Utils/HTTP.php
```diff
@@ -327,8 +327,9 @@ public static function checkURLAllowed($url, array $trustedSites = null)
         }
         $url = self::normalizeURL($url);
 
-        if (filter_var($url, FILTER_VALIDATE_URL) === false) {
-            throw new Error\Exception('Invalid URL: '.$url);
+        // check the validity of the URL
+        if (parse_url($url,PHP_URL_HOST) === false) {
+            throw new \SimpleSAML_Error_Exception('Invalid URL: '.$url);
```
jaimeperez: Please note this is incorrect in `master`. The `SimpleSAML_Error_Exception` class has been migrated to namespaces, and therefore exceptions should be thrown as it is currently, by creating a new instance of `Error\Exception`.

jaimeperez:
I fully agree with @thijskh here. I'm fine with allowing URLs that reflect standard practice, but I'm not fine with weakening the test. I would strongly prefer relying on a trustworthy third-party check, rather than doing it ourselves.

tvdijen:
Since this topic is dead and the remote branch was deleted, I'm inclined to close this..
It's an edge-case scenario that appears to be not important enough for the contributor to get the PR up to our standards.

PR #930, opened 2018-09-14
Add initial support for SAML Subject Id Attributes

peter-:
The OASIS spec [SAML V2.0 Subject Identifier Attributes Profile](https://wiki.oasis-open.org/security/SAMLSubjectIDAttr) defines two new standard attributes intending to replace use of persistent NameIDs and the eduPersonTargetedID and eduPersonUniqueId SAML attributes. The updated Kantara [SAML V2.0 Interoperability Deployment Profile](https://kantarainitiative.github.io/SAMLprofiles/saml2int.html) will also standardize on these new attributes (cf. SDP-SP15 ibid.).

This commit adds the two new attributes:

* to the URN attribute maps
* to attribute definitions and translations of the locale system
* to the smartattributes:SmartID auth proc filter (after any other SAML attributes, but before the non-SAML ones).

Support for saml2int SDP-SP16 (attribute requirements signalling via Entity Attributes) is not included here.


peter-:
Thanks. A simpler PR would have been possible had I used an internal attribute name without a dash in its name, I guess. But I was trying to follow the spec and the [examples](https://git.shibboleth.net/view/?p=java-identity-provider.git;a=blob;f=idp-conf/src/main/resources/conf/attribute-resolver-full.xml;h=b5b61ae009bffe9d9ae2096c55b0608dc1430a9d;hb=HEAD#l238) from [Shibboleth](https://git.shibboleth.net/view/?p=java-identity-provider.git;a=blob;f=idp-conf/src/main/resources/conf/attribute-filter.xml;h=888a8a7f32ac6290dbcbb8cf4cb6fec409efac09;hb=HEAD#l17) for consistency.

As for translations I was inclined to just copy the existing ones for `attribute_edupersontargetedid` to `attribute_pairwise_id` and likewise `attribute_edupersonuniqueid` to `attribute_edupersonuniqueid`, since the latter ones are essentially replacements for the former ones, and this would allow to jump-start translations with something that's probably good enough.  
But the ePTID case isn't a perfect match since `pairwise-id` is scoped and SSP has a specific naming/translation convention for all scoped attributes ("at the home organization"), and ePUID only has translations in 2 languages available (`zh-tw` and `el`), so there's little to gain here.  

The existing attribute definitions in English are not perfect, though, and most of the translations then follow those strings. E.g. `attribute_edupersonuniqueid` is long and very thorough:

> Person's non-reassignable, persistent pseudonymous ID at home organization

but `attribute_edupersontargetedid`, which is as just as non-reassignable and quite as persistent, but _additionally_ also differs per Service Provider, is simply given as:

> Persistent pseudonymous ID

So for the new attributes I've tried to start with a slightly different way to phrase things, based on WAYF.dk's description for targeted/pairwise attributes, IIRC ("service-specific"). Maybe that'll catch on in  any new translations, too.

PR #938, opened 2018-09-25
PasswordProtectedTransport authncontext when HTTPS

peter-:
Following up on the idea mentioned in #937: If the transport is secure fall back to the `PasswordProtectedTransport` authn context class ref, otherwise keep the current default of `Password`.

Requires a version of the SAML2 library with simplesamlphp/saml2#129 merged due to referencing a newly defined Constant.


jaimeperez:
This makes full sense, thanks a lot @peter-!

Issue #655, opened 2017-07-19
Allow core:PHP to manipulate the whole $request/$state


pradtke:
Based on https://groups.google.com/forum/#!topic/simplesamlphp-dev/PDENkRUnfA8

**Patrick**:

There are frequent requests on the mailing list to be able to override certain service provider metadata settings. When that metadata comes from metarefresh it is not easily achievable.

One option may be to make core:PHP more versatile and let it adjust the entire $request, not just just the attributes array.

For example if it has access to the $request/$state you could do something like:

```php
10 => array(
    'class' => 'core:PHP',
    'code' => '
        switch ($request["Destination"]["entityid"]) {
            case "https://entity1":
                // Add to existing required attributes
                $request["Destination"]["attributes"][] = "extra-required-attribute";
                break;
            case "https://something-else":
                // Replace existing required attributes
                $request["Destination"]["attributes"] = ["name","favorite_sandwhich"];
                break;
         }
    ',
    ),
```

I suppose the obvious downside is that it is easy to typo things in the code block, and people may start doing other weird things to the internal state that becomes hard to support.

Anyhow, I'm curious what other people think.

**Jaime**:
Do you mind opening an issue in Github for this? 

In any case, I see the benefit of this and we could even pass more information to the function. This could be a nice addition to motivate people to use a callback instead of the (soon to be deprecated) “code” option of the filter :-)

PR #942, opened 2018-09-27
#655 allow core:php to manipulate all of the state

pradtke:
#655 has the initial discussion on allowing more functionality from the core:php module.

I think some things to consider:
1.  Should I call the variable $state or $request when passing to function?  I generally think of it as $state though the `process` method signature calls in`$request`
2. How much should we worry about backwards compatibility? In theory someone could already have code using core:php that declares a variable $state. With this change their code would now interact with the authproc $state instead of a locally defined $state. 

I'll update the documentation once we decide on the above 2 things.

### DIFF
--- a/modules/core/lib/Auth/Process/PHP.php
+++ b/modules/core/lib/Auth/Process/PHP.php
@@ -49,7 +49,7 @@ class PHP extends \SimpleSAML\Auth\ProcessingFilter
         assert(is_array($request));
         assert(array_key_exists('Attributes', $request));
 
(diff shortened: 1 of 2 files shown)


jaimeperez:
Thanks a lot Patrick!

I'd say `$state`, as in the end, that's what it is. Regarding backwards compatibility, I would just document it in the upgrade notes making it very clear that people using the `core:PHP` filter might find them breaking. In any case, I wouldn't give the variable a misleading name just because someone _might_ be using a variable named `$state` in their filters.

That said, I think if we merge this, it should definitely wait for 1.17 to see the light of day.

jaimeperez:
I don't see why not 😄

PR #958, opened 2018-10-09
Added return statements, updated return types

ArnoVanDerVegt:
- Updated the phpDoc comments to match the code
- Added return statements
- Added return values

### DIFF
--- a/modules/cas/lib/Auth/Source/CAS.php
+++ b/modules/cas/lib/Auth/Source/CAS.php
@@ -178,7 +178,6 @@ class CAS extends \SimpleSAML\Auth\Source
     /**
      * Called by linkback, to finish validate/ finish logging in.
      * @param array $state
-     * @return array username, casattributes/ldap attributes
      */
     public function finalStep(&$state)
     {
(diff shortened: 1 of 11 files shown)


modules/cas/lib/Auth/Source/CAS.php
```diff
@@ -207,6 +207,8 @@ public function finalStep(&$state)
         $state['Attributes'] = $attributes;
 
         \SimpleSAML\Auth\Source::completeAuth($state);
+
+        return $attributes;
```
tvdijen: If this were to match the phpdoc, it should be:
return array($username, $attributes);

modules/cas/lib/Auth/Source/CAS.php
```diff
@@ -208,7 +208,7 @@ public function finalStep(&$state)
 
         \SimpleSAML\Auth\Source::completeAuth($state);
 
-        return $attributes;
+        return [$username, $attributes];
```
tvdijen: OK, you're probably not gonna like me, but I've taken a closer look here and I think the return-statement should go away completely, along with the phpdoc..
This function is called only once [here](https://github.com/simplesamlphp/simplesamlphp/blob/9a988ee0e2f7bc7d997a47188d87a43e5928e61e/modules/cas/www/linkback.php#L26) and a return-value is never used..

ArnoVanDerVegt: No problem, thanks for the info. I've updated the phpDoc/code.

PR #959, opened 2018-10-09
Converted array declarations

ArnoVanDerVegt:
Updated array declarations to the short version.


tvdijen:
This + namespaces might actually justify a 1.17 release with only non-functional changes

PR #960, opened 2018-10-09
Removed unused variables

ArnoVanDerVegt:


lib/SimpleSAML/XHTML/Template.php
```diff
@@ -638,8 +638,6 @@ public function getTranslation($translations)
      */
     private function includeAtTemplateBase($file)
     {
-        $data = $this->data;
```
tvdijen: This variable is actually used in the included file on line 645

tests/modules/core/lib/Auth/Process/AttributeAddTest.php
```diff
@@ -153,7 +153,7 @@ public function testWrongFlag()
                 'test' => array('value1'),
             ),
         );
-        $result = self::processFilter($config, $request);
+        self::processFilter($config, $request);
```
tvdijen: @thijskh while reviewing this PR I noticed that you're relying on PHPdoc for catching an exception instead of calling `$this->setExpectedException('\Exception');`...
While perfectly legal, it makes the code harder to understand imho... I had to look twice (why aren't we asserting anything here??) before I figured it out.. Would you mind if I take the opportunity to replace the phpdoc with code?

thijskh: Sure. I think this phpdoc method is a mostly legacy construct.

tvdijen: I see the use of this construct is spread across the tests, so I'll keep it separate

Issue #753, opened 2017-12-21
Cannot integrate 1.15.0 with projects using Twig 2


RinkAttendant6:
With SimpleSAMLphp 1.15.0, I cannot get the ["Integrating authentication with your own application"](https://simplesamlphp.org/docs/stable/simplesamlphp-sp#section_6) steps working with my application that uses Twig 2.

I am using the Composer autoloader in my application and I've narrowed it down to the `require` line which tries to autoload Twig in my application's vendor dependencies (`vendor`) instead of from where SimpleSAMLphp is installed (`/var/simplesamlphp`).

```php
require '/var/simplesamlphp/lib/_autoload.php';
$as = new \SimpleSAML\Auth\Simple('default-sp');
// also tried the old way: new \SimpleSAML_Auth_Simple('default-sp');
```

This is the error I get:

```
[Thu Dec 21 16:34:09.020522 2017] [php7:error] [pid 16036] [client <snip>] PHP Fatal error:  Uncaught Error: Cannot instantiate abstract class Twig_Test in /var/www/app1/vendor/twig/twig/lib/Twig/Extension/Core.php:203
Stack trace:
#0 /var/www/app1/vendor/twig/twig/lib/Twig/ExtensionSet.php(449): Twig_Extension_Core->getTests()
#1 /var/www/app1/vendor/twig/twig/lib/Twig/ExtensionSet.php(429): Twig_ExtensionSet->initExtension(Object(Twig_Extension_Core))
#2 /var/www/app1/vendor/twig/twig/lib/Twig/ExtensionSet.php(398): Twig_ExtensionSet->initExtensions()
#3 /var/www/app1/vendor/twig/twig/lib/Twig/Environment.php(936): Twig_ExtensionSet->getUnaryOperators()
#4 /var/simplesamlphp/vendor/twig/twig/lib/Twig/Lexer.php(382): Twig_Environment->getUnaryOperators()
#5 /var/simplesamlphp/vendor/twig/twig/lib/Twig/Lexer.php(68): Twig_Lexer->getOperatorRegex()
#6 /var/www/app1/vendor/twig/twig/lib/Twig/Environment.php(505): Twig_Lexer->__construct(Object(Twig_Environment))
#7 /var/www/app1/vendor/twig/twig/lib/Twig/Environment.php(561): Twig_Environment->tok in /var/www/app1/vendor/twig/twig/lib/Twig/Extension/Core.php on line 203
```

[...]

PR #965, opened 2018-10-12
Backport Twig2 support

tvdijen:


tvdijen:
That's why I've been asking for feedback [here](https://github.com/simplesamlphp/simplesamlphp/issues/753), but like you said yourself on the discussion board, usually the only way to get feedback us to just release it

thijskh:
Alright but is this not somethink that should only be part of an upgrade 1.16 to 1.17, not of an upgrade from 1.16.2 to 1.16.3?

thijskh:
I understand that you tested it, but what about people’s own websites and modules? They are prepared for some breakage potential with a new 1.17 but I think not with an upgrade from 1.16.2 to 1.16.3

tvdijen:
I agree, I think we should have a 1.17 release.. Lot's of changes on master since 1.16 was released.
Jaime thought it was too soon for a 1.17 release and suggested backporting to 1.16 due to high demand of this feature, see discussion board...

jaimeperez:
Hi guys!

As Tim said, I suggested backporting it as there's quite some people complaining about conflicts because they are already using Twig 2, so in that sense, I see this as a bugfix. However, it is also true that upgrading to a new major version of the library could cause unforeseen trouble, and I also have to admit I haven't tested really extensively.

So I'd say let's release this as part of 1.17, and use it as a way to encourage people to upgrade.

PR #966, opened 2018-10-17
Add routing & dependency injection support for easier coding and testing

jaimeperez:
This PR adds support for routing (automatically dispatching requests to controllers based on the URL) and dependency injection, so that coding and testing are easier and self-contained. This includes (roughly, and so far), the following changes:

* Bump the minimum PHP version supported to 5.5.
* Add some symfony components as a dependency.
* Add support for Twig 2 (coexisting with Twig 1).
* Make the new user interface and the old one completely independent. Enable the new one by setting the `usenewui` configuration option to `true`.
* Twig templates are **not** used unless `usenewui` is set to `true` **or** the template belongs to a theme. This allows us to have the two user interfaces coexist, while avoiding breaking themes that have already been migrated to twig.
* Modules can now add a `routes.yaml` file to the root of the module, where the different routes are defined and mapped to a method in a class. The class is completely customisable, and the URLs are paths relative to the module's root. E.g.: `/bar` in `modules/foo/routes.yaml` is translated to `module.php/foo/bar`.
* The `SimpleSAML\XHTML\Template` class is now a `Response` object (PSR-7 compatible), meaning you can now use it as a response that you can `send()`.
* Fully independent new user interface. When `usenewui` is set to `true`, visiting the root of the SimpleSAMLphp installation gets you redirected to `module.php/core/login`. If there is more than one authentication source defined (excluding _admin_), the list of them is shown, so that the user can choose. If there is only one (or the user chooses one), authentication is started with that auth source. After authentication succeeds, the user is redirected to `module.php/core/account/<auth_source_id>`, where the user can see his own attributes, session length, etc. We might still want to change this page a bit, and it should allow us to include links to other pages relevant to the user (e.g. consent management). Finally, the `module.php/core/logout/<auth_source_id>` allows us to log out from the given auth source.
* The _controller_ classes are defined as _autowired services_ in Symfony. This means they can have certain (_type-hinted_) arguments in their constructor, so that they get their dependencies _automagically_ injected. So far, `SimpleSAML\Configuration` and `SimpleSAML\Session` are the only two supported types of arguments. This way, controller classes no longer need to manually get the instances, so that we can use the configuration and session objects we want while we are testing.


lib/SimpleSAML/Module.php
```diff
@@ -12,6 +18,44 @@
...
+        'jpe'   => 'image/jpeg',
+        'jpeg'  => 'image/jpeg',
+        'jpg'   => 'image/jpeg',
+        'js'    => 'text/javascript',
+        'pdf'   => 'application/pdf',
+        'png'   => 'image/png',
+        'svg'   => 'image/svg+xml',
+        'svgz'  => 'image/svg+xml',
+        'swf'   => 'application/x-shockwave-flash',
+        'swfl'  => 'application/x-shockwave-flash',
```
tvdijen: Do we need to include the flash MIME-types?

lib/SimpleSAML/Module.php
```diff
@@ -63,6 +107,172 @@ public static function isModuleEnabled($module)
...
+     * configuration and the actual request, it will run a PHP script and exit, or return a Response produced either
+     * by another controller or by a static file.
+     *
+     * @param Request|null $request The request to process. Defaults to the current one.
+     *
+     * @return Response|BinaryFileResponse Returns a Response object that can be sent to the browser.
+     * @throws Error\BadRequest In case the request URI is malformed.
+     * @throws Error\NotFound In case the request URI is invalid or the resource it points to cannot be found.
+     */
+    public static function process(Request $request = null)
```
tvdijen: Type-hinting objects is only supported on PHP7

jaimeperez: See here: http://php.net/manual/en/migration72.new-features.php

tvdijen: Ah, my bad!

thijskh:
I've committed the PHP version bump separately. So you might want to rebase this PR to current master.

thijskh:
Merged with current master & conflicts resolved

composer.json
```diff
@@ -41,10 +41,15 @@
         "whitehat101/apr1-md5": "~1.0",
         "twig/twig": "~1.0 || ~2.0",
         "gettext/gettext": "^4.6",
-        "jaimeperez/twig-configurable-i18n": "^2.0"
+        "jaimeperez/twig-configurable-i18n": "^2.0",
```
thijskh: Should we move this to simplesamlphp namespace or something else with shared responsibility? As to prevent blockers if one person is not available.

lib/SimpleSAML/Module.php
```diff
@@ -63,6 +107,172 @@ public static function isModuleEnabled($module)
...
+        if (preg_match('#\.php$#D', $path)) {
+            // PHP file - attempt to run it
+
+            /* In some environments, $_SERVER['SCRIPT_NAME'] is already set with $_SERVER['PATH_INFO']. Check for that
+             * case, and append script name only if necessary.
+             *
+             * Contributed by Travis Hegner.
+             */
+            $script = "/$module/$url";
+            if (stripos($request->getScriptName(), $script) === false) {
```
thijskh: `ipos`? or better just `pos`?

lib/SimpleSAML/Module.php
```diff
@@ -310,4 +520,21 @@ public static function callHooks($hook, &$data = null)
...
+     *
+     * @param Request $request The request to process by this controller method.
+     *
+     * @return RedirectResponse A redirection to the URI specified in the request, but without the trailing slash.
+     */
+    public static function removeTrailingSlash(Request $request)
+    {
+        $pathInfo = $request->getPathInfo();
+        $url = str_replace($pathInfo, rtrim($pathInfo, ' /'), $request->getRequestUri());
+        return new RedirectResponse($url, 308);
```
thijskh: Nice to use the 'correct' http response code!

lib/SimpleSAML/Module.php
jaimeperez: Nice catch! Fixed as well 😄

PR #967, opened 2018-10-18
PSR2 / Whitespace

### LINKED ISSUE #458: PSR-2 compliant whitespace

tvdijen:
- Fixes a couple of whitespace issues
- Namespaced tests

Closes #458

There are some issues left that need refactoring, so another run should be done as part of a 2.0 release

### DIFF
--- a/bin/memcacheSync.php
+++ b/bin/memcacheSync.php
@@ -3,19 +3,19 @@
 
 
 // Check that the memcache library is enabled
-if(!class_exists('Memcache') && !class_exists('Memcached')) {
-	echo("Error: the memcached (or memcache) PHP extension appears to be unavailable.\n");
-	echo("\n");
-	echo("This is most likely because PHP doesn't load it for the command line\n");
-	echo("version. You probably need to enable it somehow.\n");
-	echo("\n");
-	if(is_executable('/usr/sbin/phpenmod')) {
-		echo("It is possible that running one of the following commands as root will fix it:\n");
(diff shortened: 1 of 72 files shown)


thijskh:
Only userpassorgbase fails with 5.4; probably not worth fixing because already decided 5.5 will be minimum for this release.

tvdijen:
Yes, since the PR Jaime's is working on also relies on 5.5 I've decided to remove the legacy-code while at it

thijskh:
I think a PHP version bump is good to commit explicitly. So I've committed this separately now. So if you rebase this PR I think we should be ok (and you could drop the special casing in userpassorgbase).

tvdijen:
I'm not sure I follow.. What's there to rebase? And what special casing?

thijskh:
Right, this was already fixed. So good to go!

tvdijen:
P.S.: Should we bump the version on the SAML-library as well?

thijskh:
Not necessarily I think since we're not currently changing anything there that makes it so

PR #969, opened 2018-10-21
Type hints

tvdijen:
- Added type hints wherever possible
- Removed obselete calls to assert()
- Added some phpdocs
- Removed some tests that tested for invalid input (they make no sense any more with the type hints)


thijskh:
Adding type hints on the Auth Source and Auth Processing Filters will break every Authsource or Authproc currently in use because the signatures will change, which are defined API's with been publishing for people to implement. So we should not do that without in a new major. I believe many of the "internal" functions already have type hints.

tvdijen:
Closing this, because it's not gonna get merged for 1.18..
For 2.0 I am preparing a new PR that will fully typehint the source, including scalars, return types, etc.

PR #974, opened 2018-10-30
Add a note about developing and testing Auth Proc Filter

cornelinux:


cornelinux:
I don't think this is necessary - at least would not for me.
When I planned to write a auth proc filter, I read the docs page to the end. So having a note there should be sufficient (at least for me)

But it might not hurt to add a static text to the UI.

jaimeperez:
Thanks a lot for the PR @cornelinux!

I don't think we need a message in the web interface. We're completely changing the interface anyway, so it would be kind of a waste of time. At the same time, it might be a good idea to take the chance to rethink the way this works and start running authproc filters when authenticating without an SP asking for it. Of course, we won't be able to run filters specific to a service provider, but those configured in the IdP or general configuration are reasonable, I think. In the end, from the user's point of view, it is useful to have information that reflects what can actually be sent to a service, rather than the _raw_ information contained in the backend, which is more useful for debugging purposes for administrators.

cornelinux:
@jaimeperez Thanks for the feedback.
Of course it would be great, if this could run without a SP. Till that, 3rd party module developers have to deal with that.

Could this redisgn of auth proc filters be a discussion topic for the hackaton end of november?

jaimeperez:
I think that would definitely be reasonable. Given that we are going to focus mostly on the user interface, why not taking a look into it?

thijskh:
Reworded and reformatted the text a bit, and committed to master, thanks!

Issue #975, opened 2018-10-31
Remove unnecessary SP's request state check


lhaemmerle:
**Is your feature request related to a problem? Please describe.**
We have the case that a user wants to access an SSP SP, which first lets the user pick an IdP A in the Discovery Service. The user then is sent to IdP A to authenticate but in our case the IdP A can also act as IdP B (with a different entityID). If the user is then returned with an authentication response from IdP B (instead of A), this response currently is rejected by the SSP SP even though it would have accepted the response if the user had initially selected IdP B in the Discovery Service and even though the SSP SP would have also accepted it if it was an unsolicited authentication response. 

**Describe the solution you'd like**
Given that the check only provided a benefit if unsolicited authentication responses were completely blocked and given that unsolicited authentication responses are already in use by some SSP SPs, the proposal is to remove the (unncessary) state check and accept the assertion from IdP B even though the user was sent to IdP A by the SP initially. The check of the entityID before and after authentication provides no benefit but creates issues for IdPs that act as different (virtual) IdPs with several entityIDs.

**Describe alternatives you've considered**
Alternatively to remove this check, SSP could change its behaviour to block unsolicited authentication responses. However, this would most likely cause forward-compatibility/upgrade issues as currently some SSP SPs expect unsolicited authentication responses to work. Also, this would not solve our initial problem but would make things worse for some existing SSP deployments.

**Additional context**
The Shibboleth SP as well as Microsoft ADFS allow unsolicited authentication responses and they don't block authentication responses that are received from a different IdP than what the user selected on the Discovery Service. So, changing the behaviour of SSP in this regard would make it more consistent with these two other popular SAML SP implementations. 
More details on why this check does not provide an extra benefit is also available in this thread where Scott Cantor (one of the SAML authors) explains when and how such a check would make sense: https://marc.info/?l=shibboleth-dev&m=154091453202576&w=2

tvdijen:
Would you be able to provide a SAML-trace that covers your scenario?
I'm wondering if your IdP A is answering as IdP B as an 'InReponseTo' the message ID from the authentication request.. If this is the case, then I'm not surprised SSP is detecting this as a fishy situation..

lhaemmerle:
The InResponseTo is containing the correct value but the issuer is IdP B instead of A. The authentication does not contain the information at which IdP the user should authenticate.
[...]

PR #976, opened 2018-10-31
Disabled unnecessary issuer check as proposed in issue #975

lhaemmerle:
As described in https://github.com/simplesamlphp/simplesamlphp/issues/975 this removes the issuer check, which is not needed as it does not provide any extra security because SSP also supports Unsolicited/IdP-initated login.

### DIFF
--- a/modules/saml/lib/Auth/Source/SP.php
+++ b/modules/saml/lib/Auth/Source/SP.php
@@ -280,9 +280,6 @@ class SP extends Source
             $ar->setExtensions($state['saml:Extensions']);
         }
 
-        // save IdP entity ID as part of the state
-        $state['ExpectedIssuer'] = $idpMetadata->getString('entityid');
-
         $id = State::saveState($state, 'saml:sp:sso', true);
         $ar->setId($id);
 
--- a/modules/saml/www/sp/saml2-acs.php
+++ b/modules/saml/www/sp/saml2-acs.php
@@ -95,18 +95,6 @@ if ($state) {
             'The authentication source id in the URL does not match the authentication source which sent the request.'
         );
     }
-
-    // check that the issuer is the one we are expecting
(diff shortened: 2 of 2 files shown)


tvdijen:
This would solve your problem, but it would also break SSP's functionality to set a list of trusted IdP's (IDPlist-setting)..
Why don't you add IDP B's entityID to that IDPlist-array?

tvdijen:
It's meant to restrict access to IdP's behind a SAML-proxy and in that scenario you don't have their metadata loaded on the SP-side.

lhaemmerle:
Ok, so can one then assume that IDPList is only set in deployments that act as proxies?
If so, maybe instead of removing the code, changing the line 103 in saml2-acs.php to
```
if (count($idplist) > 0 && !in_array($state['ExpectedIssuer'], $idplist, true)) {
```
would make everybody happy :-)

tvdijen:
I think that would work theoretically, yes, unless someone is stupid enough to set IDPlist to `null`..
Should probably use getArrayizeString to collect the value instead of getArrayize..

lhaemmerle:
Ok, well then this is probably the best way to move forward. I'm not sure if I should change my pull request now (and with which method) so I leave things as they are hoping that the person having a look at these pull requests reads the comments above :-)

tvdijen:
Yes, I will discuss this with Jaime by the end of the month.. I also want to discuss the ability to disable unsolicited responses

lhaemmerle:
I'm not sure who is meant by "would you be willing" (I guess not me :-) ) but I think I would be happy to keep things as they are in he current code but change this one line 104 of modules/saml/www/sp/saml2-acs.php to:
`if (count($idplist) > 0 && !in_array($state['ExpectedIssuer'], $idplist, true)) {`

thijskh:
Fixed in e543fe5f8ac225734cf745db5ab440b328ab99af

PR #979, opened 2018-11-07
Create a setting for the allowed assertion offset

Wittev1:
This was an hardcoded setting that we would like to change. Therefore making it configurable seems the right thing.

We have a customer with an Microsoft ADFS server that's about one minute ahead of time a few times per day. With this setting, we can allow them to be ahead instead of having the customers of the customers complain.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -176,6 +176,15 @@ $config = [
      */
     'enable.http_post' => false,
 
+    /*
+     * Set the allowed time difference between encrypting/decrypting assertions
+     *
+     * If you have an server that is constantly out of sync, this option
+     * allows you to adjust the allowed time-frame.
+     *
+     * Defaults to 60.
+     */
+    'assertion.allowed_offset' => 60,
 
 
     /************************
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -177,14 +177,15 @@ $config = [
     'enable.http_post' => false,
 
     /*
-     * Set the allowed time difference between encrypting/decrypting assertions
+     * Set the allowed clock skew between encrypting/decrypting assertions
      *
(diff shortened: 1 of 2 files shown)


tvdijen:
I suggest we add a filter to this to remain [SAML2INT](https://kantarainitiative.github.io/SAMLprofiles/saml2int.html#_clock_skew)-compliant even when the user enters some stupid value here.. Standard mandates a maximum offset of 5 minutes..

```
$options = [
    'options' => [
        'default' => 60, // value to return if the filter fails
        // other options here
        'min_range' => 0,
        'max_range' => 300,
    ],
];
$var = filter_var($allowed_assertion_offset, FILTER_VALIDATE_INT, $options);
```

Wittev1:
Sounds good. It's added.

tvdijen:
Thanks @Wittev1! A second pair of eyes will review this PR and hopefully merge it

jaimeperez:
Hi guys!

First of all, thanks a lot for your contributtion @Wittev1. I'm a bit skeptical about this though. I'm not sure I like the idea, for two main reasons:

- The issue here is the ADFS server, which is not keeping its own time properly. It should be the IdP fixing its own issue, not the rest of the world accommodating for their broken setup. In a way, fixing this on our side is a way to encourage them not to fix their broken setup.
- I would like to avoid adding configuration options if they are not really necessary. We already have way more than I would like, as that makes it more difficult to configure the software. So if we are going to increase the allowed offset, I think it might be better to just set it to the maximum allowed by SAML2Int.

tvdijen:
Did we forget to change https://github.com/simplesamlphp/simplesamlphp/blob/c94d1758615ba6e593b9271d205b1d203a108053/modules/saml/www/sp/saml2-acs.php#L158-L163 to use this new setting?

PR #981, opened 2018-11-08
Auth:LDAP: Add method getResourceLink

rcalpha:
This would add a method in SimpleSAML_Auth_LDAP that can be used in custom authentication module to do additional operation, returning the LDAP resource link. This can be used using ldap_modify to update some attribute value (like zimbraLastLogonTimestamp) or another operation

### DIFF
--- a/lib/SimpleSAML/Auth/LDAP.php
+++ b/lib/SimpleSAML/Auth/LDAP.php
@@ -838,3 +838,13 @@ class LDAP
         return $dn;
     }
 }
+
+/**
+ * Get LDAP resource link
+ *
+ * @return resource
+ */
+public function getResourcelink()
+{
+    return $this->ldap;
+}


tvdijen:
The resource is a protected property that can be accessed by any descendant of the class, so I must be missing the point here?..

tvdijen:
Why don’t you inherit from this class in your custom module?
Resource link is stored in a protected variable for that reason

tvdijen:
@rcalpha The LDAP-module has been removed from this repository
If you still feel this should be addressed on our end, please open a new PR @ https://github.com/simplesamlphp/simplesamlphp-module-ldap
Personally I think you can use custom modules to fix your specific use case

tvdijen:
Unfortunately I cannot reopen this PR since at was removed upstream

Issue #771, opened 2018-01-25
NameIDPolicy is always sent in authentication requests


jaimeperez:
Before SimpleSAMLphp 1.15.0, it was possible (but undocumented) to set `NameIDPolicy` to `null` in order for the SP to **not** send a `NameIDPolicy` in the authentication request. Starting with the aforementioned version, setting the configuration option to `null` triggers an exception.

A way to skip the `NameIDPolicy` in the authentication request is very much desired as that element is optional in the standard, and always specifying it could cause interoperability issues.

This issue was originally reported and discussed [here](https://groups.google.com/forum/#!topic/simplesamlphp/wCnZGwSBRic).

jaimeperez:
The main issue I see with the _trick_ used to unset the NameIDPolicy is that in PHP, in general, not setting a variable and setting it to `null` is equivalent. Therefore, I think setting `NameIDPolicy` to `null` to _**not** request a NameID format_, while leaving it unset to _request a transient NameID format_ is a bad idea, and prone to future bugs.

Since the _null trick_ was undocumented, we have the freedom to choose how to tell SimpleSAMLphp not to request a specific NameID format. I’m inclined to use the boolean `false` for that. That way, we would have the following possibilities

- Using the old, deprecated behaviour to set `NameIDPolicy` to a **string**. In that case, we use that string as the format we request.
- Using an **array** to define the policy, which translates directly into what’s requested. If no options are specified inside the array (i.e. it’s empty), then we default to request a transient NameID.
- Leaving `NameIDPolicy` unset or explicitly setting it to `null` (which should be considered equivalent). Here a transient NameID will be requested, as per the default behaviour.
- Explicitly setting `NameIDPolicy` to the boolean `false`. In this case, a NameID is not requested at all.

@thijskh, Peter, would you two mind applying the following (updated) patch and test? As far as I’ve tried, it works as intended.

The main problem I have with this is that fixing this issue goes well beyond this simple patch. We need to:

- Update the documentation to reflect the preferred array format for the `NameIDPolicy` configuration option.
- Update the documentation to tell how to avoid sending out a NameIDPolicy (setting the option to `false`).
- Make sure the SP metadata page doesn’t break when using an array instead of a string.
- Replicate the same behaviour in the API, so that the same functionality can be achieved either by configuration or calls to `requireAuth()`.
- Look for any other uses of `NameIDPolicy` in the code and make sure nothing breaks and the behaviour is consistent.
[...]

PR #984, opened 2018-11-13
Bugfix/name id policy

thijskh:
Updated the branch to current master, codestyle and fixed the remaining issues.

It works fine for all the use cases I tested, the ability to not send any NameIDPolicy which I need indeed also is available again.


thijskh:
This is to fix #771

thijskh:
Build only fails on PHP 7 and seems unrelated to this change?

tests/lib/SimpleSAML/Utils/Config/MetadataTest.php
```diff
@@ -259,4 +259,44 @@ public function testIsHiddenFromDiscovery()
...
+
+    
+    /**
+     * Test \SimpleSAML\Utils\Config\Metadata::parseNameIdPolicy().
+     */
+    public function testParseNameIdPolicy()
+    {
+        // Test null or unset
+        $nameIdPolicy = null;
+        $this->assertEquals(Metadata::parseNameIdPolicy($nameIdPolicy), ['Format' => \SAML2\Constants::NAMEID_TRANSIENT]);
```
jaimeperez: I think the `assert*()` methods receive the expected value first, then the actual value. It's really a minor detail, but could make it harder to debug when getting errors during testing, as the values would then be inverted.

tests/lib/SimpleSAML/Utils/Config/MetadataTest.php
```diff
@@ -259,4 +259,44 @@ public function testIsHiddenFromDiscovery()
...
+
+        // Test array
+        $nameIdPolicy = [
+            'Format' => 'urn:oasis:names:tc:SAML:1.1:nameid-format:persistent',
+            'AllowCreate' => false
+        ];
+        $this->assertEquals(Metadata::parseNameIdPolicy($nameIdPolicy), [
+            'Format' => 'urn:oasis:names:tc:SAML:1.1:nameid-format:persistent',
+            'AllowCreate' => false
+        ]);
```
jaimeperez: This could be simplified:

```php
        $nameIdPolicy = [
            'Format' => 'urn:oasis:names:tc:SAML:1.1:nameid-format:persistent',
            'AllowCreate' => false
        ];
        $this->assertEquals($nameIdPolicy, Metadata::parseNameIdPolicy($nameIdPolicy));
```

PR #992, opened 2018-11-20
Replace preprodwarning

dnmvisser:
Use the `production` config value to display a pre-production status banner in the base template, which is shown on all pages. This is to replace the functionality that the `preprodwarning` module offers. Current UX states that yellow is the appropriate "warning" colour.

Note that one locale message from the `preprodwarning` module has been used to add to the main locale file. If this PR gets accepted, I'll add all the other translations as well. 
Once that is done it's time to clean up the preprodwarning module.


thijskh:
Looks good indeed.

You only added it to the new template. Which means we can only remove the module after we've completely removed the old templating system. Alternatively you could add the same warning to the old templates aswell.

tvdijen:
I'd say we mark the module as being deprecated and to be removed for 2.0.
People can use the module now and have similar functionality in the new system, so we already provide an upgrade path...

dnmvisser:
i';m preparing a different PR

PR #993, opened 2018-11-21
Update OpenSSL RSA bit length in docs

ghalse:
Change documentation from recommending 2048 bit keys to using 3072 bit
keys.

Given that we're recommending people generate ten-year keys, 2048 bit
keys are probably a bit short. Almost all commercial certificate
authorities now recommend 4096 bit keys, and eduGAIN requires 3072 bit
keys for new federations.

This change aligns the SimpleSAMLphp documentation with the lower
eduGAIN requirement, since that still meets most standards bodies
recommendations for 2028 (ten years from now). cf
https://www.keylength.com/

### DIFF
--- a/docs/simplesamlphp-artifact-sp.md
+++ b/docs/simplesamlphp-artifact-sp.md
@@ -11,7 +11,7 @@ This is used for SSL client authentication when contacting the IdP.
 
 To generate a private key and certificate, you may use the `openssl` commandline utility:
 
-    openssl req -newkey rsa:2048 -new -x509 -days 3652 -nodes -out sp.example.org.crt -keyout sp.example.org.pem
+    openssl req -newkey rsa:3072 -new -x509 -days 3652 -nodes -out sp.example.org.crt -keyout sp.example.org.pem
 
 You can then add the private key and certificate to the SP configuration.
 When this is done, you can add the metadata of your SP to the IdP, and test the authentication.
--- a/docs/simplesamlphp-googleapps.md
+++ b/docs/simplesamlphp-googleapps.md
@@ -42,7 +42,7 @@ Edit `config.php`, and enable the SAML 2.0 IdP:
 You must generate a certificate for your IdP.
 Here is an example of an openssl command to generate a new key and a self signed certificate to use for signing SAML messages:
 
-    openssl req -newkey rsa:2048 -new -x509 -days 3652 -nodes -out googleappsidp.crt -keyout googleappsidp.pem
+    openssl req -newkey rsa:3072 -new -x509 -days 3652 -nodes -out googleappsidp.crt -keyout googleappsidp.pem
 
 The certificate above will be valid for 10 years.
 
--- a/docs/simplesamlphp-idp.md
+++ b/docs/simplesamlphp-idp.md
@@ -127,7 +127,7 @@ Here is an example of an `openssl`-command which can be used to generate a new p
 
 This key and certificate can be used to sign SAML messages:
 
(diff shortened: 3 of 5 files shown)


tvdijen:
I missed this one earlier, but I'd like to add my two cents to this:

I think the issue is more the ten-year lifetime of certificates then the key strength..
You're referring to commercial CA's; they usually don't supply certificates with a lifetime over 2-3yrs, for good reasons!
I think it's naive to anticipate on private keys being kept private for ten years (Heartbleed anyone?).. Only CA's can somewhat guarantee this due to very strict policy, auditing and very costly security measures.

So, that beoing said, should we perhaps advise people to use shorter term certificates?

trscavo:
Deployers should be advised to use **long-lived, self-signed certificates** for signing and encryption. To quote the [InCommon documentation](https://spaces.at.internet2.edu/display/InCFederation/X.509+Certificates+in+Metadata) on this subject (that I wrote):

> A certificate's expiration date has nothing to do with the security of the corresponding private key, which is an ongoing concern.

ghalse:
As of March, the CA/Browser forum doesn't permit subscriber certs to be signed for longer than 825 days, which equates to about two years. However, at the same time, the major browsers recommend that CA and intermediate certificates be signed for between 8 and 25 years. So I guess the question here is whether we're equating the SAML signing cert with a CA certificate or a subscriber certificate...

The explicit trust model used by the R&E community is closer to a root certificate store than a subscriber certificate. This plays out in the complexity of rolling a SAML cert -- if I update my CA cert, I have to re-embed it in every browser; if I update my SAML cert, I have to re-embed it in every relying party.

Really this boils down to a compromise between risk and complexity: LetsEncrypt can roll certs every three months because the process is easy and well defined; SAML providers cannot because the process is very site dependent.

IMHO creating a ten year (or equally long-lived --- ours is longer) certificate is the right thing to do.

And just because it's signed for ten years doesn't mean I have to *use* it for ten years. If the next heartbleed comes along and I think it's been compromised, I'll roll it early... And we'll all be using EC certs in ten years anyway :-)

tvdijen:
Hehe, that's the exact opposite to my world @trscavo
And I really disagree that a cert's expiration date has nothing to do with security.. We cannot foresee any risks involved over the next couple of years..

@ghalse: the R&E community is a big one when talking federation, but there's a world outside as well! I just went over the process of a SAML cert rollover.. Couldn't be easier! I really don't think that argument holds..

Just to sketch my world:  I have to use CA-signed (2yr) certs.. And I have to use separate ones for TLS, SAML-signing and SAML-decryption. I guess there's also a huge difference between a hub&spoke federation vs. a mesh federation.. Risks in a mesh are a lot smaller compared to hub&spoke

trscavo:
The original PR affects self-signed certs, not CA-signed certs, so bringing up the latter in this context is off-topic at best.

@ghalse: I can't find a recommendation from eduGAIN that deployers should use certificates with 3072-bit keys. Are you perhaps confusing signing assertions with signing metadata? In any case, I don't have a problem with 3072-bit keys but I would push back on 4096-bit keys since AFAIK there is no evidence to suggest that larger key sizes are warranted. There's a trade-off between key size and operational efficiency, as you know.

tvdijen:
@trscavo I know that the PR is about self-signed, but at the same time @ghalse uses the argument that commercial CA's recommend 4K now.. So if considered off-topic, it already started in the initial post..

I'm just raising a few questions here... If the worlds of self-signed / CA-signed is so far apart, shouldn't we at least come with two recommendations?

Issue #995, opened 2018-11-21
"authorize" denies access with a static message


dnmvisser:
When using the [authorize authproc filter](https://simplesamlphp.org/docs/stable/authorize:authorize), unauthorized users are served a 403 that uses a static message, which (in English) reads:

`You don't have the needed privileges to access this application. Please contact the administrator if you find this to be incorrect.`

This is better than just the 403, but it would be really helpful if the rejection message could be customised. Especially when `authorize` is used for specific SPs, which may require custom contact details/procedures/etc. An idea would be to introduce a new parameter `rejection_message`:

```php
'authproc.sp' => [
   60 => [
     'class' => 'authorize:Authorize',
     'regex' => FALSE,
     'reject_message' => [
        'en' => 'Please contact the HR department at hr@orgname.org',
        'nl' => 'Neem contact op met de afdeling PZ op hr@orgname.org'
      ],
     'group' =>  [
         'CN=SimpleSAML Students,CN=Users,DC=example,DC=edu',
         'CN=All Teachers,OU=Staff,DC=example,DC=edu',
     ],
   ],
],
```

PR #997, opened 2018-11-22
Show custom rejection message to unauthorised users

dnmvisser:
Fixes https://github.com/simplesamlphp/simplesamlphp/issues/995.

### DIFF
--- a/modules/authorize/docs/authorize.md
+++ b/modules/authorize/docs/authorize.md
@@ -24,18 +24,18 @@ There are three configuration options that can be defined: `deny`, `regex`, and 
 
 Unauthorized will be shown a 403 Forbidden page.
 
-### Deny ###
+### `deny` ###
 The default action of the filter is to authorize only if an attribute match is found (default allow). When set to TRUE, this option reverses that rule and authorizes the user unless an attribute match is found (default deny), causing an unauthorized action.
 
 Note: This option needs to be boolean (TRUE/FALSE) else it will be considered an attribute matching rule.
 
-### Regex ###
+### `regex` ###
 Turn regex pattern matching on or off for the attribute values defined. For backwards compatibility, this option defaults to TRUE, but can be turned off by setting it to FALSE.
 
 Note: This option needs to be boolean (TRUE/FALSE) else it will be considered an attribute matching rule.
 
-### Rejection message ###
-Optionally provide a localised, custom message to an unauthorised user. This can be used to provide tailored instructions on how to fix the authorisation issue, supply specific contact details, etc.
+### `reject_msg` ###
+This option can be used to provide a localised, custom message to an unauthorised user. For example: tailored instructions on how to fix the authorisation issue, specific contact details, etc.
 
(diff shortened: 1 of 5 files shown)


modules/authorize/lib/Auth/Process/Authorize.php
```diff
@@ -93,6 +107,10 @@ public function process(&$request)
         assert(array_key_exists('Attributes', $request));
 
         $attributes = &$request['Attributes'];
+        // Store the rejection message array in the $request
+        if(!empty($this->reject_msg)) {
+          $request['authprocAuthorize_reject_msg'] = $this->reject_msg;
```
thijskh: indent = 4 spaces (also below)

modules/authorize/docs/authorize.md
```diff
@@ -20,20 +20,26 @@ This module provides a user authorization filter based on attribute matching for
 `authorize:Authorize`
 ---------------------
 
-There are two configuration options that can be defined; deny and regex. All other filter configuration options are considered attribute matching rules.
+There are three configuration options that can be defined: `deny`, `regex`, and `reject_msg`. All other filter configuration options are considered attribute matching rules.
 
-The users not authorized will be shown a 403 Forbidden page.
+Unauthorized will be shown a 403 Forbidden page.
```
thijskh: "Unauthorized *users*"

dnmvisser:
Should be all good now @thijskh

modules/authorize/templates/authorize_403.twig
```diff
@@ -2,7 +2,11 @@
 
 {% block content %}
   <h1>{{ '{authorize:Authorize:403_header}'|trans }}</h1>
+{% if reject_msg[currentLanguage] is defined %}
+  <p>{{ reject_msg[currentLanguage]|raw }}</p>
```
jaimeperez: This isn't necessary. We have a function that obtains a localized string from an array containing the available translations, indexed by the language code. This is accessible from twig via the `translateFromArray` filter, so this code would be:

```twig
    {%- if reject_msg %}

    <p>{{ reject_msg|translateFromArray }}</p>
    {%- else %}

    <p>{% trans %}{authorize:Authorize:403_text}{% endtrans %}</p>
    {%- endif %}
```

A side note, also. Spacing is the hardest part of twig, IMHO. We should have good spacing / indentation in the twig templates, so that we can clearly see how the code flows, but the resulting HTML should also be properly indented. When we use twig tags, there can be space before and after them. If we have a newline right after the closure of a twig tag, that newline is removed from the output. The easiest way to control space in both the template and the resulting HTML is to use `-`. Adding it right after the begining of an opening tag (`{%`) tells twig to remove **all** the white space before, including newlines. Similarly, adding it before the end of a closing tag (`-%}`) removes all white space after it. So the simplest way for me to keep things working is to remember to add two newlines after a tag, and clear all white space before a tag that's standing on it's own line. E.g.:

```twig
    {%- if something %}

    <p>Something</p>
    {%- else %}

    <p>else</p>
    {%- endif %}
```

Issue #948, opened 2018-10-05
newUI: translation issues


tvdijen:
- NOSTATE errors are not translated properly
![43726604-e56c7e98-999f-11e8-9e46-301424214532 1](https://user-images.githubusercontent.com/841045/46558417-4da0fa80-c8ed-11e8-9c66-32d2ee0fcd61.png)

- Metadata not found are errors not translated properly
https://user-images.githubusercontent.com/841045/43832207-ee10efee-9b06-11e8-883a-d261d77aa6c6.png

This issue is probably bigger than the two mentions above...

tvdijen:
I remember us talking about this in Valencia, but was this fixed eventually? @thijskh @jaimeperez

jaimeperez:
I haven't really worked on this, not even tested. If, by any chance, it is already fixed, that was accidental 😄

tvdijen:
I can confirm this is still an issue

tvdijen:
Marked for 1.17 milestone since old UI is working perfectly

tvdijen:
Mixing up old/new UI... All is fine

PR #1000, opened 2018-11-23
Translations for preprodwarning replacement

dnmvisser:
This completes the work from https://github.com/simplesamlphp/simplesamlphp/pull/994.


tvdijen:
Off-topic:
@jaimeperez This kind of illustrates my aversion against the full-sentence msgid's over a tag.. The overhead in the .po-files is huge.

jaimeperez:
[I just fixed the text](https://github.com/simplesamlphp/simplesamlphp/commit/12159498e818c13ba16f5f95c8ae0786afbdd37d) so that we don't have to pass it through a filter. The usual way to translate strings is to use `{% trans %}` tags.

Regarding the overhead in `*.po` files, that's absolutely true. However, using the actual strings instead of IDs has two immediate advantages:

* You can actually see the text in its context, inside the HTML template. This makes it easier to translate and to have an overview of the entire page while you are developing it.
* If there is no translation available for a message, you still get the original message in english, instead of a weird message ID like the ones in #948.

tvdijen:
Downside is when you change a text, you have to change it in _all_ .po-files + the template..

PR #1005, opened 2018-11-29
Optional preselect in multiauth source config

sgomez:
The use case is: We want to use multiauth but without show the select source page. This a extra step than can confuse to our users.

So, if no `source` or `multiauth:preselect` option is present, the optional `preselect` value could be used to redirect directly to default source.

It is responsibility of the custom theme to detect than the request comes from 'multiauth' module and show a way to delegate the authentication to the others sources.

If no `preselect` option is added, the module works as usual.

The proposed config for this is:

```php
<?php
array(
    'example-multi' => array(
        'multiauth:MultiAuth',

        /*
         * The available authentication sources.
         * They must be defined in this authsources.php file.
         */
        'sources' => array(
            'example-saml' => array(
                // ...
            ),
            'example-admin' => array(
                // ...
            ),
        ),
        // New option (optional)
        'preselect' => 'example-saml',
    ),
);
```

### DIFF
--- a/tests/modules/multiauth/lib/Auth/Source/MultiAuthTest.php
+++ b/tests/modules/multiauth/lib/Auth/Source/MultiAuthTest.php
@@ -1,0 +1,203 @@
+<?php
+
+namespace SimpleSAML\Test\Module\multiauth\Auth\Source;
+
+use PHPUnit\Framework\TestCase;
+use SimpleSAML\Configuration;
+use SimpleSAML\Module\multiauth\Auth\Source\MultiAuth;
+
+class MultiAuthTest extends TestCase
+{
+    /** @var Configuration */
+    private $sourceConfig;
+
+    public function setUp()
+    {
+        $this->config = Configuration::loadFromArray(['module.enable' => ['multiauth' => true]], '[ARRAY]', 'simplesaml');
+        Configuration::setPreLoadedConfig($this->config, 'config.php');
+
+        $this->sourceConfig = Configuration::loadFromArray(array(
+            'example-multi' => array(
+                'multiauth:MultiAuth',
+
+                /*
+                 * The available authentication sources.
+                 * They must be defined in this authsources.php file.
+                 */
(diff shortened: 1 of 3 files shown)


jaimeperez:
¡Hola Sergio!

If I understand correctly, the goal here would be to have an IdP using multiauth so that several auth sources can be used with it, but enforcing a specific source by default and leaving the rest unusable unless they are selected programatically. Is that correct?

What I'm struggling to see is how this can be effectively used by a remote SP, since SAML doesn't provide any standard way to do this. If you set the `preselect` configuration option, who would you choose other auth sources at some point?

jaimeperez:
I see! Now that makes a lot of sense.

So essentially, multiauth would always proceed with the "default" or "preselected" source, and then that source would allow you to either login with it or choose one of the alternatives defined in multiauth, going back to it and manually selecting them, right?

sgomez:
Yes. `MultiAuth::delegateAuthentication` checks that the new AS is a valid one and defined in   `authsource.php`. So, it is like we selected the new AS in the first place.

There are not BC, it is just a new feature.

tests/modules/multiauth/lib/Auth/Source/MultiAuthTest.php
```diff
@@ -0,0 +1,203 @@
+<?php
+
+namespace SimpleSAML\Test\Module\multiauth\Auth\Source;
+
+use PHPUnit\Framework\TestCase;
+use SimpleSAML\Configuration;
+use SimpleSAML\Module\multiauth\Auth\Source\MultiAuth;
+
+class MultiAuthTest extends TestCase
```
jaimeperez: Since the test is manipulating the configuration, it should probably extend `\SimpleSAML\Test\Utils\ClearStateTestCase` in order to avoid state spill over tests.

jaimeperez:
Merged! ¡Gracias Sergio!

PR #1012, opened 2018-12-10
Add Password Support to Redis library

benrj:
For Redis instances that require authentication, the current Redis store does not work.  So, this commit adds optional support for utilizing the Predis password option.  It will pick up the password from the config file and pass it along when it instantiates the Predis client.

The test simply checks that adding the password option doesn't break the constructor. Since Predis parameters end up as protected, we can't verify that the instance contains the password.

### DIFF
--- a/lib/SimpleSAML/Store/Redis.php
+++ b/lib/SimpleSAML/Store/Redis.php
@@ -31,13 +31,14 @@ class Redis extends Store
             $host = $config->getString('store.redis.host', 'localhost');
             $port = $config->getInteger('store.redis.port', 6379);
             $prefix = $config->getString('store.redis.prefix', 'SimpleSAMLphp');
+            $password = $config->getString('store.redis.password', '');
 
             $redis = new \Predis\Client(
                 [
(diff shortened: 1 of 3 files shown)


lib/SimpleSAML/Store/Redis.php
```diff
@@ -31,13 +31,14 @@ public function __construct($redis = null)
...
             $prefix = $config->getString('store.redis.prefix', 'SimpleSAMLphp');
+            $password = $config->getString('store.redis.password', '');
 
             $redis = new \Predis\Client(
                 [
                     'scheme' => 'tcp',
                     'host' => $host,
                     'port' => $port,
-                ],
+                ] + (!empty($password) ? ['password' => $password] : []),
```
tvdijen: Wow, I'm really surprised this is valid code..  It really feels like a comma is missing!

tvdijen:
Thanks for you contribution @benrj !

benrj:
You bet.  I actually added one more commit with a documentation note.

PR #1021, opened 2018-12-19
Multiauth: The preselect param is not being saved in the state

dbollaer:
I was testing the preselect method and I noticed that it was not being used.
When I changed the order in which the state was being saved, it worked.

I also corrected on small spelling mistake (Source)


tvdijen:
@sgomez could you please take a look at this?

sgomez:
Yes, @dbollaer is right. I did the test in the wrong environment, because I need this in the stable branch. This is a bug. Please, accept it.

Issue #1030, opened 2019-01-02
Infinite recursion in autoloader


weirdan:
**Describe the bug**
`temporaryLoader` is prone to infinite recursion

**To Reproduce**
1. `git clone git@github.com:weirdan/simplesaml-autoloader-recursion.git`
2. `cd simplesaml-autoloader-recursion`
3. `composer install`
4. `php test.php`

**Expected behavior**
Shouldn't produce any errors

**Screenshots or logs**
```console
$ php test.php 
PHP Fatal error:  Uncaught Error: Maximum function nesting level of '256' reached, aborting! in /home/weirdan/src/simplesaml-autoloader-recursion/vendor/composer/ClassLoader.php:335
Stack trace:
#0 /home/weirdan/src/simplesaml-autoloader-recursion/vendor/composer/ClassLoader.php(321): Composer\Autoload\ClassLoader->findFile('RobRichards\\XML...')
#1 [internal function]: Composer\Autoload\ClassLoader->loadClass('RobRichards\\XML...')
#2 [internal function]: spl_autoload_call('RobRichards\\XML...')
#3 /home/weirdan/src/simplesaml-autoloader-recursion/vendor/simplesamlphp/simplesamlphp/lib/_autoload_modules.php(24): class_exists('\\RobRichards\\XM...', true)
#4 [internal function]: temporaryLoader('RobRichards\\XML...')
#5 [internal function]: spl_autoload_call('RobRichards\\XML...')
#6 /home/weirdan/src/simplesaml-autoloader-recursion/vendor/simplesamlphp/simplesamlphp/lib/_autoload_modules.php(24): class_exists('\\RobRichards\\XM...', true)
#7 [internal function]: temporaryLoader('RobRichards\\XML...')
#8 [internal function]: s in /home/weirdan/src/simplesaml-autoloader-recursion/vendor/composer/ClassLoader.php on line 335
```

**Additional context**
Refs vimeo/psalm#1162

tvdijen:
Nice catch!

weirdan:
What I've actually been doing is a research on the bug reported in Psalm, so I really have no idea  :laughing:

tvdijen:
Yeah, I'm pretty sure that would fix it..  With `true` and your fix, we would eventually end up here:

https://github.com/simplesamlphp/simplesamlphp/blob/2042707a59f0febba831086982d4fbd8b5c17903/lib/_autoload_modules.php#L64-L69

[...]

PR #1031, opened 2019-01-02
Autoloader infinite recursion fix

weirdan:
Fixes simplesamlphp/simplesamlphp#1030
Refs vimeo/psalm#1162


weirdan:
@tvdijen I may need some help here. It appears tests need some prior configuration, otherwise `SimpleSAML\Logger` fails with `CriticalConfigurationError`. Do I need to extend some base test class instead of inheriting from `PHPUnit\Framework\TestCase` directly?

weirdan:
>  I may need some help here.

Ok, I've found the required bit I need to add to test `setUp`.

PR #1036, opened 2019-01-10
Allow StandardErrorLoggingHandler to be used

ghalse:
There is an existing StandardErrorLoggingHandler logging class, but no way to use it. Since logging to stderr is potentially useful to people trying to use SimpleSAMLphp in containers, this change makes the existing logging class usable.


jaimeperez:
Thanks a lot Guy!

PR #1043, opened 2019-01-23
allow paths to use stream wrappers

npoepperl:
Currently the function `System::resolvePath` assumes paths, which are using stream wrappers like `s3://` to be relative to a base path.

To allow the usage of stream wrappers the `System` class has been extended. This gives the ability to store **certificates** and **private keys** in a different location than the server's filesystem but can also be used for other scenarios where configured paths are used (e.g. logdir, datadir, tempdir).


tvdijen:
Very useful! Thanks @npoepperl

PR #1052, opened 2019-02-18
Make metadata validateFingerprint algorithm configurable

francoisfreitag:
Some metadata publishers offer SHA256 fingerprints to validate the fingerprint of the certificate used to sign metadata. Allow users to configure the fingerprinting algorithm to facilitate validation.

For backward compatibility, the default algorithm remains SHA1.

---

Out of convenience, I rebased this work on https://github.com/simplesamlphp/simplesamlphp/pull/1051. While I prefer to reuse the existing certificate setup architecture, I can rework the PR use master as its base.

### DIFF
--- a/docs/simplesamlphp-automated_metadata.md
+++ b/docs/simplesamlphp-automated_metadata.md
@@ -162,6 +162,10 @@ Each metadata source has the following options:
     don't need this option if you don't want to validate the signature
     on the metadata.
 
+`validateFingerprintAlgorithm`
+:   Algorithm used to compute the signing certificate's fingerprint. Defaults to
+    `XMLSecurityDSig::SHA1`.
+
 `template`
 :   This is an array which will be combined with the metadata fetched to
     generate the final metadata array.
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -48,6 +48,11 @@ Released TBD
   * Allow `core:PHP` to manipulate the entire state array.
   * IdP initiated login: add compatibility with Shibboleth parameters.
 
+### metarefresh
+  * The algorithm to compute the fingerprint of the certificate that signed
+    metadata can be specified with the new `validateFingerprintAlgorithm`
+    configuration option.
+
 ### multiauth
   * Added a `preselect` configuration option to skip authsource selection (#1005).
 
--- a/lib/SimpleSAML/Metadata/SAMLParser.php
+++ b/lib/SimpleSAML/Metadata/SAMLParser.php
@@ -2,6 +2,7 @@
 
 namespace SimpleSAML\Metadata;
 
+use RobRichards\XMLSecLibs\XMLSecurityDSig;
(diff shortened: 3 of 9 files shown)


francoisfreitag:
Thanks for the review!

No problem changing the `SHA*` strings to `XMLSecurityKey` constants. Shouldn't the code be using the digital signatures constants though?
https://github.com/robrichards/xmlseclibs/blob/master/src/XMLSecurityDSig.php#L54-L57

> PS: SHA1 is perfectly fine for thumbprinting.. The NIST-source you're referring to applies to SHA-signatures only.

I am not clear on this, would you mind explaining a bit more? My understanding is that the public part of the certificate is hashed to obtain a fingerprint. A malicious user would attempt to produce metadata signed with a forged certificate. If a collision can be found for the public part of the certificate, a fake public certificate can be attached to the metadata and it would match the signature in `validateFingerprint`. I do not see why SHA1 is fine in this case?

I updated the source to link to https://shattered.io/ instead of the NIST, as it includes Digital Certificate signatures in the affected systems list.

tvdijen:
You are absolutely right about the constants.. I wasn't thinking.. Thanks for fixing.

> My understanding is that the public part of the certificate is hashed to obtain a fingerprint.
Correct

> A malicious user would attempt to produce metadata signed with a forged certificate. If a collision can be found for the public part of the certificate, a fake public certificate can be attached [...]
So far correct

> [...] to the metadata and it would match the signature in `validateFingerprint`.
The fingerprint is not a signature.. It's just a short representation of the public key that can be used for (human) comparison of certificates and see if they're the same.

Yes. collisions can happen when comparing SHA1-thumbprints, but that's why we check actual (strong algo) signatures to verify messages instead .

jaimeperez:
Hi,

@tvdijen is right. Calculating the fingerprint (with SHA-1) of a certificate in order to identify it is not part of any signature verification. The attack surface you are describing is not possible because the trust model in SAML does not work as you are describing. Metadata is either exchanged manually (in which case there's no point in mounting such a complex attack when you can simply exert some social engineering), or automatically from well-known locations where metadata is published on an HTTPS URL that was verified by the underlying layer.

francoisfreitag:
I understand now, many thanks to both of you for the explanation!

thijskh:
Good initiative. I'm a bit worried that the implementation instantly breaks all existing configurations (is not BC).

You could know from the length of the configured fingerprint what the intended algorithm is, right? So if it's e.g.  `D0:E8:40:25:F0:B1:2A:CC:74:22:ED:C3:87:04:BC:29:BB:7B:9A:40`, use sha1.

francoisfreitag:
Perhaps the default algorithm should remain `SHA1`, since it's fine for certificate fingerprinting? That way, existing implementations don't need to change anything, but users who would rather use `SHA256` for fingerprinting can opt-in?

thijskh:
Yes!

francoisfreitag:
Thanks for the review. Updated to use SHA1 as the default algorithm.

PR #1053, opened 2019-02-21
Prevent SP Metadata to be saved in Session

m0ark:
Currently SP Metadata is saved in the session for every AuthN request. This can lead to heavily polluted sessions in case of unprocessed AuthN requests (in my case some services frontend trying to access protected data once every 5 seconds).
Combine this with the default configured `session.state.timeout` of 60 minutes and we had to face sessions of ~72mb size for every user per 100kb of SP metadata.

This patch prevents saving the whole SP Metadata and only saves the EntityId.

### DIFF
--- a/lib/SimpleSAML/IdP.php
+++ b/lib/SimpleSAML/IdP.php
@@ -167,16 +167,14 @@ class SimpleSAML_IdP
         return $this->config;
     }
 
-
     /**
-     * Get SP name.
+     * Get SP object.
      *
      * @param string $assocId The association identifier.
      *
-     * @return array|null The name of the SP, as an associative array of language => text, or null if this isn't an SP.
+     * @return SimpleSAML_Configuration The SP, as a configuration object or null if this isn't an SP.
      */
-    public function getSPName($assocId)
-    {
+    public static function getSP($assocId) {
         assert('is_string($assocId)');
 
         $prefix = substr($assocId, 0, 4);
@@ -200,6 +198,25 @@ class SimpleSAML_IdP
                 return null;
             }
         }
+
+        return $spMetadata;
+    }
+
+
+    /**
+     * Get SP name.
+     *
+     * @param string $assocId The association identifier.
+     *
+     * @return array|null The name of the SP, as an associative array of language => text, or null if this isn't an SP.
+     */
+    public function getSPName($assocId)
+    {
+        assert('is_string($assocId)');
+        $prefix = substr($assocId, 0, 4);
+        $spEntityId = substr($assocId, strlen($prefix) + 1);
+
+        $spMetadata = self::getSP($assocId);
 
         if ($spMetadata->hasValue('name')) {
             return $spMetadata->getLocalizedString('name');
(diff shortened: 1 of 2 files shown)


thijskh:
Seems reasonable. However, it does change the API quite a bit. But might be good to merge for 2.0?

jaimeperez:
Hi @m0ark!

The metadata is getting into the state array by design. With your approach, a change in the metadata that happens during an ongoing authentication flow would lead to an inconsistency in the state, and as such, to unexpected behaviour. This is something we would like to avoid, and therefore, the entire metadata array ended up in the state array which is in turn stored in the session.

Now, I find it really weird that you have 100Kb of SP metadata stored in the state array, not to talk about the ~72Mb size sessions. I suspect there must be something wrong going on in your deployment, because what I see here is around 4Kb size per session. Of course, if you start the authentication flow again without having completed it before, that would lead to a new state array with its corresponding metadata saved to the session. But I don't think that's a normal situation. What is it that you have in the metadata that weights so much? Why are you ending up with SP metadata being stored to the session numerous times per user?

That said, I noticed a huge possibility for improvement in the way we do garbage collection in the `SimpleSAML\Session` class. Basically, garbage collection is being done only when some data is actually stored into the session, and at no other time. Would you mind applying the following patch (you might need to adapt it if you are not running the latest stable version of the software) and telling me how that works for you?

```patch
diff --git a/lib/SimpleSAML/Session.php b/lib/SimpleSAML/Session.php
index cd158b3fe..6f19fc0d1 100644
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -445,6 +445,9 @@ class Session implements \Serializable, Utils\ClearableState
      */
     public function save()
     {
+        // clean out old data
+        $this->expireData();
+
         if (!$this->dirty) {
             // session hasn't changed, don't bother saving it
             return;
@@ -871,9 +874,6 @@ class Session implements \Serializable, Utils\ClearableState
         assert(is_string($id));
         assert(is_int($timeout) || $timeout === null || $timeout === self::DATA_TIMEOUT_SESSION_END);
 
-        // clean out old data
-        $this->expireData();
-
         if ($timeout === null) {
             // use the default timeout
             $timeout = self::$config->getInteger('session.datastore.timeout', null);
@@ -928,6 +928,7 @@ class Session implements \Serializable, Utils\ClearableState
 
                 if ($ct > $info['expires']) {
                     unset($typedData[$id]);
+                    $this->markDirty();
                 }
             }
         }
@@ -953,8 +954,6 @@ class Session implements \Serializable, Utils\ClearableState
             return null;
         }
 
-        $this->expireData();
-
         if (!array_key_exists($type, $this->dataStore)) {
             return null;
         }
```

jaimeperez:
@tvdijen I think we haven't applied this patch to master yet, right? Even though we didn't get confirmation from the OP, I think it makes sense to apply it. The way it is today, `expireData()` doesn't do anything unless there's any other modification in the session, so it makes much more sense to do as in the patch, mark the session as dirty if we have expired data, and then run expiration right before.

PR #1055, opened 2019-03-04
Supress warning on absent NameIDFormat

tvdijen:
After updating to 1.17 I started noticing a lot of warnings in the log:
\Mar  4 12:47:16 sv1810934 IDP-BZK[5791]: 3 [c11b0ed290] Unable to generate NameID. Check the userid.attribute option.

SP doesn't ask for a specific NameIDFormat and none is set in the metadata...
current([]) would return false, but we never catch that, so it falls through,..

### DIFF
--- a/modules/saml/lib/IdP/SAML2.php
+++ b/modules/saml/lib/IdP/SAML2.php
@@ -1194,17 +1194,17 @@ class SAML2
             $a->setAttributes($attributes);
         }
 
+        $nameIdFormat = null;
+
         // generate the NameID for the assertion
         if (isset($state['saml:NameIDFormat'])) {
             $nameIdFormat = $state['saml:NameIDFormat'];
-        } else {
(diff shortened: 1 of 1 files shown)


tvdijen:
Also see referenced commit above

foaadnami:
@tvdijen hopefully the last issue before 1.17 release :) thanks a lot for all the great work !

tvdijen:
Yes, I expect 1.17 final to be released later this week..

thijskh:
So this is the history:
- PR by @ghalse to make this functionality possible.
- problem is that it does `array_shift()` on function output which generates notices.
- I replaced that with `current()` which does the same as array_shift but can be used with array output.
- however, I failed to notice that the return value behaviour is different when no items are found (of course, because this is PHP): `array_shift` will return null, `current` will return false.
- so the solution is to let the if statement check for false instead of null.

PR #1058, opened 2019-03-07
Replace e-mail class with one that uses PHPMailer

jornane:
The old mail class uses the built-in `mail` function in PHP, and builds a payload by itself using hard-coded HTML and MIME-snippets. Let's use a library for that; PHPMailer has new dependencies and does all the heavy lifting.

E-mail is used for two things in SimpleSAMLphp; to allow users to send error reports to the admin, and for cron to tell the admin it has run. Both of these have a need to embed key/value data in the e-mail, so I thought it made sense to let the e-mail class handle formatting key/value data as well. This cleans up errorreport.php and cron.php.


lib/SimpleSAML/Utils/EMail.php
```diff
@@ -0,0 +1,133 @@
...
+     * It will refuse to return the SimpleSAMLphp default address,
+     * which is na@example.org.
+     *
+     * @return string Default mail address
+     */
+    public static function getDefaultMailAddress() {
+        $config = Configuration::getInstance();
+        $address = $config->getString('technicalcontact_email', 'na@example.org');
+        if ('na@example.org' === $address) {
+            throw new \Exception('technicalcontact_email not set');
```
thijskh: To prevent confusion: maybe make more explicit inthe message that it is in fact set but not changed from default value

lib/SimpleSAML/Utils/EMail.php
```diff
@@ -0,0 +1,133 @@
...
+     * Get the default e-mail address from the configuration
+     * This is used both as source and destination address
+     * unless something else is provided at the constructor.
+     *
+     * It will refuse to return the SimpleSAMLphp default address,
+     * which is na@example.org.
+     *
+     * @return string Default mail address
+     */
+    public static function getDefaultMailAddress() {
```
thijskh: PSR-2 dictates that { is on the next line for methods

lib/SimpleSAML/Utils/EMail.php
```diff
@@ -0,0 +1,133 @@
...
+    private $text;
+    /** @var PHPMailer The mailer instance */
+    private $mail;
+
+    /**
+     * Constructor
+     *
+     * @throws PHPMailer\PHPMailer\Exception;
+     */
+    public function __construct($subject, $from = null, $to = null) {
```
thijskh: can we add type hints?

lib/SimpleSAML/Utils/EMail.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->mail = new PHPMailer(true);
+        $this->mail->Subject = $subject;
+        $this->mail->setFrom($from ? $from : static::getDefaultMailAddress());
+        $this->mail->addAddress($to ? $to : static::getDefaultMailAddress());
```
thijskh: nit: this can be elvis operator

lib/SimpleSAML/Utils/EMail.php
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * Set the data that should be embedded in the e-mail body
+     *
+     * @param array $data The data that should be embedded in the e-mail body
+     */
+    public function setData($data) {
+        $this->data = array_map(function($v){return is_array($v) ? $v : [$v];}, $data);
```
thijskh: this is not completely transparent code, not a problem but the comment could be a bit more explicit as to what it does

lib/SimpleSAML/Utils/EMail.php
```diff
@@ -0,0 +1,133 @@
...
+    public function generateBody($template) {
+        $config = Configuration::loadFromArray([
+            'usenewui' => true,
+        ]);
+        $t = new Template($config, $template);
+        $twig = $t->getTwig();
+        if (is_bool($twig)) {
+            throw new \Exception('Even though we explicitly configure that we want Twig, the Template class does not give us Twig. This is a bug.');
+        }
+        // Twig does HTML escaping for us, right?
```
thijskh: yes. so remove ths question or make it a statement

www/errorreport.php
```diff
@@ -41,90 +39,18 @@
...
+$data['directory'] = dirname(dirname(__FILE__));
 
-// send the email
-$toAddress = $config->getString('technicalcontact_email', 'na@example.org');
-if ($config->getBoolean('errorreporting', true) && $toAddress !== 'na@example.org') {
-    $email = new \SimpleSAML\XHTML\EMail($toAddress, 'SimpleSAMLphp error report', $from);
-    $email->setBody($message);
-    $email->send();
-    SimpleSAML\Logger::error('Report with id '.$reportId.' sent to <'.$toAddress.'>.');
+if ($config->getBoolean('errorreporting', true)) {
```
thijskh: this does change behaviour if na@example.org is still set. but may be acceptable

jornane: Yes. When trying to send an error report, na@example.org will prevent this from working. Instead of attempting and thereby silently failing, we will now have a hard-fail for trying to send an error report. But maybe we should treat `errorreporting` as false already when we must decide whether to render the form? Ref. https://github.com/simplesamlphp/simplesamlphp/blob/2081fc116dee8f1b6898aa635bb3ee3a2a34879f/templates/error.php#L48

jornane: Apparently, the user is already prevented from writing an error report when na@example.org. I did not find the code that's responsible for this, but black-box testing shows that the form disappears.

tvdijen:
@jornane How am I supposed to use this to send custom (html) mails?  I've been struggling with this for a bit and it appears the only way is to override the template in my theme?

tvdijen:
I've been playing around with this for some time now and I feel like there are a few things missing that are crucial for such class:

- There's no debug-logging at all.. Try setting your smtp hostname to something stupid and you'll have the time of your life debugging it (you can't without editting the EMail-class and enable it from there and then the output of the SMTP-transaction will go to the browser)
- There's not way to pass a Twig-template, leaving us with just one possibility;  overriding the base template, which can only be done one (themes have precedence over modules).
- Can't really use the class without diving into the API of PHPMailer

IMHO our EMail-class should be a wrapper that is smart enough to take the complexity of the PHPMailer-API away and can handle the most common use-cases (Like processing a \SimpleSAML\XHTML\Template). Maybe it should even extend from the PHPMailer class..

@jaimeperez we've discussed this out-of-band, but do you agree with this? I feel this is worth a new Issue and addressing in 1.19

PR #1061, opened 2019-03-08
Externalize authwindowslive

tvdijen:
TODO:
- Release v1.0 of the module
- Register the module in Packagist

Build fails because module doesn't exist in Packagist yet


tvdijen:
All actions are done

PR #1062, opened 2019-03-08
Externalize authtwitter

tvdijen:
TODO:

    Release v1.0 of the module
    Register the module in Packagist

Build fails because module doesn't exist in Packagist yet


tvdijen:
All tasks have been taken care of

PR #1069, opened 2019-03-12
Set max Twig-version

tvdijen:
It appears that Twig 2.7 has a bc-breaking change that breaks our unit tests:

Fatal error: Declaration of Twig_Extensions_TokenParser_Trans::parse(Twig_Token $token) must be compatible with Twig\TokenParser\TokenParserInterface::parse(Twig\Token $token) in /home/travis/build/simplesamlphp/simplesamlphp/vendor/twig/extensions/lib/Twig/Extensions/TokenParser/Trans.php on line 86

This PR locks the max version to 2.6.x.
Needs backporting to 1.17 & 1.16-branch as well

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "symfony/http-kernel": "^3.4 || ^4.0",
         "symfony/dependency-injection": "^3.4 || ^4.0",
         "symfony/yaml": "^3.4 || ^4.0",
-        "twig/twig": "~1.0 || ~2.0",
+        "twig/twig": "<1.38 || >2.0 <2.7",
         "whitehat101/apr1-md5": "~1.0"
     },
     "require-dev": {


tvdijen:
See https://github.com/twigphp/Twig/issues/2886

kiler129:
What you really need is to wait for the PR fixing the NS BC break and add conflict (see https://github.com/nelmio/NelmioSecurityBundle/issues/197#issuecomment-472028322 for details).

stof:
Twig already fixed this

tvdijen:
@kiler129 We're still struggling with an issue here..
It doesn't seem to affect runtime, but our Psalm-tests fail and it started after the release of 1.38/2,7...

It starts here:
https://travis-ci.org/simplesamlphp/simplesamlphp/jobs/509120456#L733

We extend a Twig-class here:
https://github.com/jaimeperez/twig-configurable-i18n/blob/master/src/Twig/Environment.php#L12

It might be a Psalm-issue, but the only clue I have right now is that it broke @ the changes at Twig a few days ago.. We're completely lost so if you have an idea about this, it would help us very much!

stof:
this looks like an issue in Psalm

tvdijen:
Thanks for taking a look! It seems I'm no longer the only one facing this issue:
https://github.com/vimeo/psalm/issues/1486

PR #1070, opened 2019-03-13
Use the list of configured tabs instead of a hardcoded set.

dnmvisser:
This avoids having to change the code when adding custom tabs.
Note that it still requires to add translations.


dnmvisser:
Hi the travis build seems to have failed but AFAICT it's because of something unrelated: https://travis-ci.org/simplesamlphp/simplesamlphp/jobs/505680782

dnmvisser:
Perhaps discopower user @ghalse could have a look at this?

modules/discopower/lib/PowerIdPDisco.php
```diff
@@ -248,21 +248,9 @@ public function handleRequest()
...
-            'iceland' => \SimpleSAML\Locale\Translate::noop('{discopower:tabs:iceland}'),
-            'incommon' => \SimpleSAML\Locale\Translate::noop('{discopower:tabs:incommon}'),
-            'kalmar' => \SimpleSAML\Locale\Translate::noop('{discopower:tabs:kalmar}'),
-            'misc' => \SimpleSAML\Locale\Translate::noop('{discopower:tabs:misc}'),
-            'norway' => \SimpleSAML\Locale\Translate::noop('{discopower:tabs:norway}'),
-            'sweden' => \SimpleSAML\Locale\Translate::noop('{discopower:tabs:sweden}'),
-            'switzerland' => \SimpleSAML\Locale\Translate::noop('{discopower:tabs:switzerland}'),
-            'ukacessfederation' => \SimpleSAML\Locale\Translate::noop('{discopower:tabs:ukacessfederation}'),
-        ];
+        foreach($this->discoconfig->getValue('tabs', null) as $tab) {
```
tvdijen: I'd use getArray here and use `[]` or `['misc']` as the default value.
Now this code breaks if the setting is not there, because we cannot loop over `null`

Issue #1067, opened 2019-03-12
"Federation" admin page / metadata generation


restena-sw:
**Describe the bug**

I see a discrepancy when I display the IdP metadata. In config, and in the "simpleSAMLphp flat file format", I have:

'UIInfo' => 
  array (
    'DisplayName' => 
    array (
      'en' => 'RESTENA Staff (internal)',
      'fr' => 'RESTENA Employees (interne)',
      'de' => 'RESTENA Mitarbeiter (intern)',
      'lb' => 'RESTENA Matärbeschter (intern)',
    ),
[...]

but the accompanying "in SAML 2.0 Metadata format" listing does not include any DisplayName. If one were to copy&paste that into a federation metadata, this would result in the entity not having a <DisplayName> at all (but I'm not doing that on this IdP).

**To Reproduce**

1. Use the following saml20-idp-hosted metadata:

$metadata['https://clueless-private.restena.lu/simplesaml/saml2/idp/metadata.php'] = array (
  'metadata-set' => 'saml20-idp-remote',
  'entityid' => 'https://clueless-private.restena.lu/simplesaml/saml2/idp/metadata.php',
  'SingleSignOnService' => 
  array (
    0 => 
    array (
      'Binding' => 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect',
      'Location' => 'https://clueless-private.restena.lu/simplesaml/saml2/idp/SSOService.php',
    ),
  ),
  'SingleLogoutService' => 
  array (
    0 => 
    array (
      'Binding' => 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect',
      'Location' => 'https://clueless-private.restena.lu/simplesaml/saml2/idp/SingleLogoutService.php',
    ),
  ),
  'keys' => 
  array (
    0 => 
    array (
      'type' => 'X509Certificate',
      'signing' => true,
      'encryption' => true,
[...]

PR #1076, opened 2019-03-18
Fix logic error in UIInfo generation

ghalse:
There is a simple boolean logic error in the handling of mdui:Keywords and mdui:Logo that results in those elements never being output into metadata.

This impacts any user interface that displays a logo for an entity (such as discovery interfaces). It is a regression that affects the 1.17.1 release.

### DIFF
--- a/lib/SimpleSAML/Metadata/SAMLParser.php
+++ b/lib/SimpleSAML/Metadata/SAMLParser.php
@@ -1089,8 +1089,8 @@ class SAMLParser
 
                     foreach ($e->getKeywords() as $uiItem) {
                         if (!($uiItem instanceof \SAML2\XML\mdui\Keywords)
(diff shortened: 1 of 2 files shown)


thijskh:
Would this be easy to cover with a testcase?

tvdijen:
I wonder if this is a different approach to fix issue #1067

ghalse:
> @ghalse does this issue persist on master or a checkout of the 1.17 branch?

My patch was to master as at 55e3ad19292c0057b684c79bf86ed20299a8d407. See https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Metadata/SAMLParser.php#L1101

tvdijen:
Ouch, I just switched logics there :(

thijskh:
Great work @ghalse, thanks!

Issue #950, opened 2018-10-05
newUI: discopower-module has inline javascript


tvdijen:
discopower still has on the fly composed javascript in lib/DiscoPower.php (handleRequest)

ghalse:
This can be fixed by moving the logic to generate the tab list into a standalone PHP file that outputs a JSON structure. The rest of the code goes into a static .js file, and the tab list is built dynamically using $.ajax(). Ideally the latter gets bundled in with other Javascript to avoid too many <script> references.

PR #1078, opened 2019-03-19
discopower module tabs don't work

ghalse:
The template for discopower ends up with broken tabs because the call to
```javascript
$("#tabdiv").tabs
```
is missing.

Looking through the equivalent twig template, the whole function has migrated to [PowerIdPDisco.php](https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/discopower/lib/PowerIdPDisco.php#L288), so rather than fixing it in this template, it makes sense to duplicate what's happening in the twig template. Thus the first commit removes the javascript references to liveUpdate at the bottom in favour of simply including `$this->data['search']` which produces the same result albeit with the tabs call.

At the same time I noticed that both `String.prototype.score` functions were being included (quicksilver and suggest) rather than honouring the one selected in the config file. The second commit removes the duplication and replaces the hardcoded quicksilver.js with the configured version.

Although my patches are against master, this affects the released 1.17.1.

### DIFF
--- a/modules/discopower/templates/disco.tpl.php
+++ b/modules/discopower/templates/disco.tpl.php
@@ -7,9 +7,10 @@ $this->data['head'] = '<link rel="stylesheet" media="screen" type="text/css" hre
     SimpleSAML\Module::getModuleURL('discopower/assets/css/disco.css').'" />';
 
 $this->data['head'] .= '<script type="text/javascript" src="'.
-    SimpleSAML\Module::getModuleURL('discopower/assets/js/jquery.livesearch.js').'"></script>';
+    SimpleSAML\Module::getModuleURL('discopower/assets/js/jquery.livesearch.js').'"></script>'."\n";
 $this->data['head'] .= '<script type="text/javascript" src="'.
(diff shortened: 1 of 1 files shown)


tvdijen:
Ah, that's what we couldn't figure out last week... Loading both js files instead of just one.. thanks a lot @ghalse!

dnmvisser:
Yup this fixes the tabs on the legacy UI!!

tvdijen:
@ghalse Do you happen to see a way to get rid of https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/discopower/lib/PowerIdPDisco.php#L286-L294 and fix #950

ghalse:
> @ghalse Do you happen to see a way to get rid of https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/discopower/lib/PowerIdPDisco.php#L286-L294 and fix #950

I might do. I actually looked at that yesterday and decided against tackling it because it was too big a change to get into the 1.17 branch and fix the current regression. But I'd like to see it gone too :).

PR #1079, opened 2019-03-20
Remove the inline javascript from discopower-module

ghalse:
Fixes #950

PR #1080, opened 2019-03-20
Make the module-discopower twig template usable

ghalse:
simplesamlphp@b3ee105e720a3d243c1a7c8247b4687816542e03 undid simplesamlphp@d13304e8cbc8ecdb83315f8118e4b4dc88d05912

This is a regression that makes the twig template in 1.17.1 unusable. However, the change below is against master and thus incorporates the change from #1079 too.


tvdijen:
I'm starting to think I break more than I fix ...

tvdijen:
@ghalse since you know this module through and through...
We're in the process of externalizing modules, giving them their own repository. Would you by any chance want to adopt the discopower-module and become its curator together with @dnmvisser ?

dnmvisser:
I'm OK with adopting it

PR #1085, opened 2019-03-22
Fix build

tvdijen:
Two weeks ago an upstream change in Twig triggered a bug in Psalm.
We are running a pretty old version of Psalm and the bug was actually resolved in a newer version. The old version will not be patched anymore

This change involves:
- Upgrading Psalm from 0.3.92 to 1.1.9
- Upgrading PHPunit to 5.7 because the newer Psalm-version doesn't support PHPunit 4.8 anymore
- Stop testing PHP 5.5, because PHPunit 5.7 has a minimum required version of PHP 5.6
- All Psalm-errors related to typehints are suppresed, because we can't really fix them without being fully typehinted throughout the codebase.
- The new version of Psalm detected stuff that the old version didn't;  these issues have been resolved

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -16,7 +16,6 @@
     <filter>
         <whitelist processUncoveredFilesFromWhitelist="true">
             <directory suffix=".php">./lib/</directory>
-            <directory suffix=".php">./modules/consent/lib/</directory>
             <directory suffix=".php">./modules/core/lib/</directory>
             <directory suffix=".php">./modules/saml/lib/</directory>
             <exclude>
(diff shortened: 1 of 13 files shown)


phpunit.xml
```diff
@@ -16,7 +16,6 @@
     <filter>
         <whitelist processUncoveredFilesFromWhitelist="true">
             <directory suffix=".php">./lib/</directory>
-            <directory suffix=".php">./modules/consent/lib/</directory>
```
tvdijen: This line actually broke the PHPunit coverage tool, probably because the module isn't enabled by default. I have removed it, since we're in the process of externalizing the consent-module

PR #1086, opened 2019-03-22
Initial tests for authorize:authorize

pradtke:
Let me know if you have a recommended way for testing code that calls out `HTTP::redirectTrustedURL`. I avoided that code path in those tests.  For our internal modules we use the technique outlined in https://github.com/cirrusidentity/simplesamlphp-test-utils#mock-redirects

### DIFF
--- a/modules/authorize/docs/authorize.md
+++ b/modules/authorize/docs/authorize.md
@@ -50,14 +50,17 @@ Note: If regex is enabled, you must use the preg_match format, i.e. you have to 
    (as far as I know), you have to close the browser.
 
 ### Examples ###
-To use this filter configure it in `config/config.php`:
+To use this filter configure it in `config/config.php`.
+For unstructured attributes use `^` and `$` to anchor your regex as necessary:
+
 ```php
 'authproc.sp' => [
     60 => [
         'class' => 'authorize:Authorize',
         'uid'   =>  [
-            '/.*@example.com/',
-            '/(user1|user2|user3)@example.edu/',
+            '/^.*@example.com$/',
+            // Use anchors to prevent matching 'wronguser1@exampled.edu.attacker.com'
+            '/^(user1|user2|user3)@example.edu$/',
         ],
         'schacUserStatus' => '@urn:mace:terena.org:userStatus:' .
         'example.org:service:active.*@',
@@ -72,10 +75,10 @@ An alternate way of using this filter is to deny certain users. Or even use mult
 'authproc.sp' => [
     60 => array[
         'class' => 'authorize:Authorize',
-        'deny'  => TRUE,
+        'deny'  => true,
         'uid'   =>  [
-            '/.*@students.example.edu/',
-            '/(stu1|stu2|stu3)@example.edu/',
+            '/.*@students.example.edu$/',
+            '/^(stu1|stu2|stu3)@example.edu$/',
         ]
     ]
 ]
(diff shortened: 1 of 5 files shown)


tvdijen:
I think we use the builtin webserver and test for a HTTP/302 like we do here:
https://github.com/simplesamlphp/simplesamlphp/blob/ad1184b203b31e61c8b93e4dff3ad7f2653e9bc9/tests/www/IndexTest.php#L102

One remark I have is that PSR-2 mandates one class per file, so if you could split the file in two that'd be great

tests/modules/authorize/lib/Auth/Process/AuthorizeTest.php
```diff
@@ -0,0 +1,162 @@
+<?php
+/**
+ * Test for the authorize:Authorize authproc filter.
+ */
+
+namespace SimpleSAML\Test\Module\consent\Auth\Process;
```
tvdijen: consent?

pradtke: Ooops! I was copying the namespace form another file. I ran into another problem, after moving the Authorize class into it's own file I get this error `Class 'SimpleSAML\Test\Module\authorize\Auth\Process\TestableAuthorize' not found`   I'm guessing the autoloader doesn't know how to load test classes that follow the module namespace convention. I'm not sure how to address that: Make `_autoload_modules.php` aware of the `SimpleSAML\Test` namespace? Add a dev autoloader to composer? Or am I completely off base and the cause it something else?

tvdijen: I'd say the autoload-dev option would be the nicest way to fix this. I'm surprised we don't have that in place yet to be honest.

pradtke: I've added the `autoload-dev` option

dbollaer:
Hello @pradtke ,
I encountered similar problem with my modules.
I think of using a custom loader (_autoload_modules.php), 
You should be able to fix it with:
>  "SimpleSAML\\Test\\Module\\authorize\\": "tests/lib/"

pradtke:
@tvdijen Any chance this can be merged in soon? I have some other work I want to do in the authorize module and I'd like to have some base line unit tests committed first.

modules/authorize/docs/authorize.md
```diff
@@ -50,14 +50,17 @@ Note: If regex is enabled, you must use the preg_match format, i.e. you have to
...
+
 ```php
 'authproc.sp' => [
     60 => [
         'class' => 'authorize:Authorize',
         'uid'   =>  [
-            '/.*@example.com/',
-            '/(user1|user2|user3)@example.edu/',
+            '/^.*@example.com$/',
+            // Use anchors to prevent matching 'wronguser1@exampled.edu.attacker.com'
```
thijskh: s/exampled/example/

tests/_autoload_modules.php
```diff
@@ -0,0 +1,42 @@
...
+        return; // it can't be a module test class
+    }
+    if (array_shift($elements) !== 'SimpleSAML') {
+        return; // the first element is not "SimpleSAML"
+    }
+    if (array_shift($elements) !== 'Test') {
+        return; // the second element is not "test"
+    }
+    if (array_shift($elements) !== 'Module') {
+        return; // the second element is not "module"
```
thijskh: third

tvdijen:
It’s done! Thanks @pradtke

tvdijen:
Oh, I missed Thijs’ comments.. mobile app doesn’t show them.. I’ll take care of them!

PR #1088, opened 2019-04-01
rememberme: Set cookie lifetime iso expire fixing the remember me fea…

fingerspitziluijks:
rememberme: Set cookie lifetime iso expire fixing the remember me feature and respective errors.

Currently the remember me functionality does not work correctly and using it results in severable undefined index errors as the expire cookie parameter is passed along the SessionHandlerPHP which does not accept this one. Using lifetime instead of expire, effectively doing the same thing, this can be fixed in a pretty simple way. Next to that the params given to the session handler are merged with the current ones before given to the session handler instead of after.


tvdijen:
Since this will break the API, it will have to wait until 2.0.. Until then we can fix the description in `config.php`

fingerspitziluijks:
Thanks and excellent, love to see this feature working (again)!

thijskh (approved):
I've reviewed it and use outside core SSP that would break seems not so likely or would be quickly fixed. So I've opted not to add more 'bc' code lines. Did add some unit tests though, and merged. Thanks!

Issue #1093, opened 2019-04-10
Web Server rewrite not taken into account to access module web interface


dweeves:
## Abstract case 
 - cannot access a module with a web interface (not bound to SAML session) through a server rewrite without redirect to real (module.php/modulename/moduleroute) on a url with parameters.

## Concrete case: 

module selfregister, that allows for handling user registration.
 
## Apache Rewrite Configuration:
`RewriteRule ^/register module.php/selfregister/newUser.php [QSA,L]`

## Why it bothers
In some deployments, the "exposed" url needs to be "polished" (having /register) instead of (/module.php/selfregister/newUser.php) is a valid expectation i think.

## To Reproduce
1. Install a module that has its own web interface
2. create an apache rewrite to this module web interface
3. You'll get a 404 not found (because module does not match initial url)

## Expected behavior
The real module name (selfregister) should be matched using $_SERVER['PATH_INFO'] (which is the apache rewritten url) rather than using the left part of original non apache rewritten URL.

So , having /register able to process module

## Additional context
The problem lies in module detection using $request->getPathInfo() , which uses symfony request class, which tries to be a bit too "smart"

File:
/lib/SimpleSAML/Module.php , method process(), line 123

Real cause:
use of /vendor/symfony/http-foundation/Request.php , which computes the pathinfo from the request uri.

## Current workaround (i.e : not fitting my requirement)
 - Force a real redirect in apache config [R,QSA,L] in the end of the redirect rule line 
  - This is really not what i want, since exposing the "SAML module" internal url which i tried to avoid.

tvdijen:
This is a duplicate of https://github.com/simplesamlphp/simplesamlphp/issues/1023

dweeves:
submitted PR #1095 to solve this issue

Issue #1023, opened 2018-12-19
New Routing: breaks Apache Aliases (module


dbollaer:
I use apache aliases in Simplesamlphp to allow easier access to my custom module.

1. Create an alias in your apache config:
Alias /login.php "/var/www/simplesamlphp/www/module.php/core/www/authenticate.php"
2. Restart apache
3. Go to the page

**Expected behavior**
You will receive a page not found exception, with the following message:

SimpleSAML\Error\NotFound: The requested page 'http://localhost/login.php' could not be found. The URL must at least contain a module name followed by a slash.

In the past this used to just show the authenticate page.

When you look in the logs, you see that login.php page is in the front, normally it should not be part of the path and it replaced module.php.

**Screenshots or logs**
```
Dec 19 19:12:21 localhost simplesamlphp[8138]: 2 [a9caccd0cf] /login.php/core/www/authenticate.php - Localization directory missing/broken for langcode 'nl' and domain 'themes'
Dec 19 19:12:21 localhost simplesamlphp[8138]: 3 [a9caccd0cf] /login.php/core/www/authenticate.php - Something went wrong when trying to get path to language file, cannot load domain 'themes'.
Dec 19 19:12:21 localhost simplesamlphp[8138]: 3 [a9caccd0cf] SimpleSAML\Error\NotFound: The requested page 'https://localhost/login.php' could not be found. The URL must at least contain a module name followed by a slash.
Dec 19 19:12:21 localhost simplesamlphp[8138]: 3 [a9caccd0cf] Error report with id 9709a3dd generated.
```

**Additional context**
I rewrote some parameters, (privacy)

I use php7.1-fpm

If it is possible to use the routing directly, the aliases would no longer be necessary.

tvdijen:
As a workaround/fix you could use:
`Redirect "/login.php" "/module.php/core/authenticate.php"`

We should probably mention this in the upgrade notes, but other than that I don't think there's a solution for this if we want to move forward with routing.

dbollaer:
I understand you want to move forward and I will use your redirect in the meantime, but I will probably end up fixing it in the coming months. Could you please provide me some hint where I can start looking.

Also I was wondering why you did not go for full routing or is this something you plan for in the future.
I think having nice custom urls is one of the major benefits of using routing.

tvdijen:
[...]

PR #1095, opened 2019-04-10
Add support for "rewritten urls" in module detection

dweeves:
This is a fix for issue #1023 & duplicated #1093

The code adds a new method to get module information from url.
This method is first used on the actual symfony request pathinfo
if no valid module is found, then the same method is applied on the real 'PATH_INFO' (which will be the rewritten on) .

The last commit is now supporting urls without trailiing slash (/register)

### DIFF
--- a/lib/SimpleSAML/Module.php
+++ b/lib/SimpleSAML/Module.php
@@ -106,21 +106,28 @@ class Module
         return self::isModuleEnabledWithConf($module, $config->getArray('module.enable', []));
     }
 
-
+    /**
+     * Performs module detection from a given url
+     * @param $url url to detect module from
+     * @return array , associative array with the following keys:
+     *          - module : name of module
+     *          - isvalid : the module above is a valid module
+     *          - url : part of url that is found after module
+     * @throws Error\NotFound
+     */
     public static function getModuleInfoFromRequest($url)
     {
         assert(substr($url, 0, 1) === '/');
-
+        $module=substr($url,1);
         /* clear the PATH_INFO option, so that a script can detect whether it is called with anything following the
          *'.php'-ending.
          */
         $modEnd = strpos($url, '/', 1);
-        if ($modEnd === false) {
+        if ($modEnd !== false) {
             // the path must always be on the form /module/
-            throw new Error\NotFound('The URL must at least contain a module name followed by a slash.');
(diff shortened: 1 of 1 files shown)


dweeves:
Any news on final approbation ?

tvdijen:
Hi @dweeves !

I've asked Jaime again to take a look at this PR. I know there are several version of 'routing improvement' lying around and we need to pick a strategy here. A finalized version of the routing-feature will be part of a  v1.18-release, hopefully somewhere in Q4 this year depending on resources.

lib/SimpleSAML/Module.php
```diff
@@ -126,31 +149,20 @@ public static function process(Request $request = null)
...
-         *'.php'-ending.
-         */
-        unset($_SERVER['PATH_INFO']);
-
-        $modEnd = strpos($url, '/', 1);
-        if ($modEnd === false) {
-            // the path must always be on the form /module/
-            throw new Error\NotFound('The URL must at least contain a module name followed by a slash.');
+        if ($basepathinfo === '/') {
+            throw new Error\NotFound('No PATH_INFO to module.php');
```
jaimeperez: This error would be shown to the end user, but won't help much, I'm afraid. Should we have an error message similar to the old one?

dweeves: I agree that the error message is not giving much valuable information, so using the old one is a good option.

About the use of  $_SERVER['path_info'] , i didn't find a way to have it through Symfony request. i had a look at the code inside symfony and the thing is they are trying to be "too smart" about request parsing, providing a "too high" abstraction level above the initial request. so loosing information with no way to get it back from the original request. So the only way i found was to ask the original request details directly through $_SERVER.

jaimeperez: Uhm, I see...

Do you think this is actually wrong behaviour in Symfony? Should we report it upstream for them to consider a change (regardless of whether we work around it or not)?

dweeves: I think this is a behaviour that is not ideal for the purpose we want to achieve (ie, detecting which module is targeted by the  rewritten url) , i cannot say it's a "bad" behaviour from symfony, it's just a choice that was made that is making the request object not well suited for low level url access but very interesting to process request content & structure.

PR #1096, opened 2019-04-12
prefill username supplied in initial request into forcedUsername

melanger:
O365 includes the supplied username (e.g. "username@example.com") in POST data. This feature can extract the username (e.g. "username") and fill it into `forcedUsername`.

The feature is off by default and can be configured in `config.php`:
- boolean `forced_username_enabled` - controls this feature, default `false`
- string `forced_username_field` - the name of GET/POST field, default `'username'`
- string `forced_username_pattern` - regular expression with a named group username, default `'/^(?<username>.*)$/'`

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -829,7 +829,29 @@ $config = [
      */
     'production' => true,
 
+    /*
+     * Set this option to true to force username when present in GET or POST
+     * in the initial request.
+     *
+     * Defaults to false.
+     */
+    //'forced_username_enabled' => false,
 
+    /*
+     * This option allows you to specify the field of GEST or POST
+     * which holds the forced username.
+     *
+     * Defaults to 'username'.
+     */
+    //'forced_username_field' => 'username',
+
+    /*
+     * This option allows you to specify a regular expression
+     * for the forced username. The RE must contain a named group username.
+     *
+     * Defaults to '/^(?<username>.*)$/' (the entire string).
+     */
+    //'forced_username_pattern' => '/^(?<username>.*)$/',
 
     /*********************
      | DISCOVERY SERVICE |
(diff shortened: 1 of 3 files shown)


melanger:
O365 asks for the user's email address (account), which in our case is "username@organisation.tld" and this is sent in POST, so the username can be extracted. I am not against any other way to achieve this without altering SSP code, but I did not find any.

O365 login page

I will take a look at the improvements, thanks.

jaimeperez:
Hi,

From the standard (profiles document, line 529, section 4.1.4.1 – &lt;AuthnRequest&gt; usage):

> Note that the service provider MAY include a `<Subject>` element in the request that names the actual identity about which it wishes to receive an assertion. This element MUST NOT contain any `<SubjectConfirmation>` elements. If the identity provider does not recognize the principal as that identity, then it MUST respond with a `<Response>` message containing an error status and no assertions.

Therefore, an SP already has a standard way to specify that it is expecting an assertion about a given subject, and the IdP should be able to respect that.

This implementation, however, is completely tailored to Microsoft, who is not using the standard for something that is already allowed by it. Therefore I'm quite hesitant to merge this...

melanger:
Hi,

I understand your concerns, but I think that requiring this from MIcrosoft would be also quite specific, as not all IdPs have the same relation between emails at Microsoft and usernames at IdP (e.g. `username@example.com`) so the (presumed) username cannot be globally forced with SAML's `Subject`, since that has to fail with error if the username is not found.

I am searching SSP for the Subject/NameId functionality, but I only found setting it in the request (SP side, PR #400), no handling when SSP is used as IdP. Is it correct that this functionality is missing from SSP right now?

melanger:
OK, I would be able to solve my problem without this PR, using `<saml:Subject>` with `<saml:NameID>` inside, but how to access it from a template? Is it possible to access the Authnrequest data including `<saml:Subject>` somehow?

PR #1100, opened 2019-04-20
fixed stats being logged despite loglevel being <= WARNING

unixtam:
The config templates reads:

```
     * - SimpleSAML\Logger::ERR     No statistics, only errors
     * - SimpleSAML\Logger::WARNING No statistics, only warnings/errors
     * - SimpleSAML\Logger::NOTICE  Statistics and errors
```

However, statistics were being logged regardless of selected loglevel. With this PR statistics are no longer logged when ERR or WARNING are specified.


tvdijen:
Hi @unixtam! Thanks for reporting this..
I think however the fix is not the way to go.. We have to find out why `$statsLog = true` despite the fact that `logLevel < NOTICE`.

Can you give us a few more hints about your situation? What version are you on? Have you experienced this with earlier versions as well? (since we've just released 1.17)

unixtam:
Hi @tvdijen,

> We have to find out why $statsLog = true despite the fact that logLevel < NOTICE.

As far as I can tell, there is simply no check in place. For example, [UserPassBase.php:299](https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/core/lib/Auth/UserPassBase.php#L299) calls `\SimpleSAML\Logger::stats()`, which, in turn, calls `\SimpleSAML\Logger::log()` with `$statsLog` set to `true` in [Logger.php:251](https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Logger.php#L251). You could do the check there, if you prefer.

unixtam:
I've just force pushed an update where the check is moved up one function call.

unixtam:
As for whether this bug has been present in previous versions, I haven't tested it, but this commit from 2008 would suggest the bug went unnoticed for more than 11 years: https://github.com/simplesamlphp/simplesamlphp/commit/972c5df4d221d24265efb859901c1969def011f3.

tvdijen:
> but this commit from 2008 would suggest the bug went unnoticed for more than 11 years
Why am I not surprised 😅

The way you have fixed this now is much better imho and I see no reason to not merge it!

thijskh:
Fix looks good if indeed the intention is for it to work as described in config-templates. Obviously an alternative is to change the documentation to match the behaviour :-). I think there is something to say for both. @jaimeperez ?

unixtam:
@thijskh, I'd recommend to keep these issues separate. Currently, actual behavior differs from documented behavior. Since the documented behavior is stated clearly, it would seem reasonable to assume that this is also the intended behavior, which means the actual behavior is faulty and should be fixed. If you wish to rethink how stats logging is *supposed* to work, it would seem sensible to first merge this PR and then create a new issue for rethinking the intended behavior.

if you are going to rethink the logging process, please bear in mind that not everyone is interested in logging stats mixed with regular log messages. Personally, I would prefer to keep stats separate altogether, so that I can log on NOTICE level without statistics polluting the log. A config option to set whether stats should be logged at all and whether they should be logged with the regular log would be nice. There already *is* separate statistics logging in place, so I wonder why these stats aren't just logged with those statistics instead. But again, that's a separate issue.

PR #1105, opened 2019-04-30
Per-IDP configurable AuthnContextClassRef/AuthnContextComparison

tvdijen:
Allows to set a different authnContextClassRef per IDP.
Can be used in cases where one SP has multiple IDP's and one-or-more IDP's require specific non-default values (i.e. Dutch eHerkenning).

It would be very desirable if we can backport this to 1.17


tvdijen:
Second commit is unrelated, but I was at it anyway...

PR #1106, opened 2019-05-02
Add authData

macmenco:
Make authData available in template

### DIFF
--- a/modules/core/lib/Controller.php
+++ b/modules/core/lib/Controller.php
@@ -82,6 +82,8 @@ class Controller
         }
 
         $attributes = $auth->getAttributes();
+        
+        $session = Session::getSessionFromRequest();
 
         $t = new Template($this->config, 'auth_status.twig', 'attributes');
         $l = $t->getLocalization();
@@ -92,6 +94,7 @@ class Controller
             ? $auth->getAuthData('saml:sp:NameID')
             : false;
         $t->data['authData'] = $auth->getAuthDataArray();
+        $t->data['trackid'] = $session->getTrackID();
         $t->data['logouturl'] = Module::getModuleURL('core/logout/'.urlencode($as));
         $t->data['remaining'] = $this->session->getAuthData($as, 'Expire') - time();
         $t->setStatusCode(200);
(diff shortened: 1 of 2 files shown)


tvdijen:
I've discussed this with Menco and he basically wants the output of /admin/test/{as} (behind admin-pass) that has a little more info, instead of /core/account/{as} (public)

I can see why we don't display the authData on the account-page, but perhaps we can add a /debug route for this on the account-page? Admins can send this link to users to provide debug-info and it would save Menco the trouble of overriding templates

thijskh:
On the auth test it’s a collapsible. Why not do the same on the account oage but with a better description. E.g. “Show technical details” or “show debug information for use by support staff”

tvdijen:
@thijskh Thanks, I totally agree..   @macmenco could you include an updated `auth_status.twig` to this PR? Be aware that this page is for end-users and that it shouldn't display scary debug-data by default (as it does in the admin-module)

macmenco:
Sorry for the delay. I've added the requested change to display the extra info.

tvdijen:
No worries, we have bigger priorities as you can imagine ;)

tvdijen:
I took care if the comments now, but I'm not entirely happy with the result..

The way it is now:

afbeelding

What I'd prefer:

afbeelding

However, I cannot think of an easy way to get rid of the 'general'-block of the Expander-template.. Any suggestions @thijsk or @jaimeperez ?

jaimeperez:
Yeah, I've been there as well. Maybe we should make an alternative expander HTML layout for cases like this, where there's no block at all? Then we can just embed that one instead.

tvdijen:
We can do this easier:  https://stackoverflow.com/a/13806784

jaimeperez:
That's a good option actually, having some little logic in the template itself to decide what kind of widget is needed. However, I think we need a little bit more than just skipping the block with the visible contents, as in that case we would just have another block that we use to expand it, and it would look weird, IMO. Maybe drop the border of that block, and also remove padding from the hidden contents?

In any case, I think it's worth exploring this path and testing how it looks if we just skip the top block.

jaimeperez:
Hi!

I've taken the liberty to do a few modifications myself. This is how it would look like now, taking advantage of what we already have in other places:

image

jaimeperez:
@tvdijen does it look good? Shall we merge?

tvdijen:
Excellent, thanks!

PR #1107, opened 2019-05-03
Update SSOService.php

mrvanes:
The SSOService info log prints late (in Logger:flush()) because the Logger has no valid $trackid before \SimpleSAML\IdP:getById(). This caused me some confusion and head-scratching and can easily be remedied by logging the info after Session initialisation.


thijskh:
Shouldn't we instead flush the deferred log messages as soon as there's a trackid, i.e. add that to Logger:setTrackId()?

PR #1110, opened 2019-05-06
look it up from the metadatadir from configuration variable

### LINKED ISSUE #1109: importPdoMetadata.php and the metadata directory
**Describe the bug**
I am attempting to import my metadata from flat files into a MySQL database using the pdo option in config.php.  However, in my setup the metadatadir is not the default 'metadata/' directory inside the SimpleSAMLphp distribution, I have it set to a custom path.  I managed to create the database tables using initMDSPdo.php, however when I went to import my existing metadata into the database I had to modify importPdoMetadata.php on line 16 to correctly pull my metadata from the custom directory.

**To Reproduce**
1. Store your flatfile metadata in a non-standard directory.

1. Configure your metadata for type 'pdo' and run initMDSPdo.php.

1. Attempt to import your metadata into the database with importPdoMetadata.php.

**Expected behavior**
The metadata is imported into the database.

brycejlowe:
Fixes #1109

Instead of using a hardcoded "metadata" folder, use the metadatadir to import metadata from flatfile into a database.

### DIFF
--- a/bin/importPdoMetadata.php
+++ b/bin/importPdoMetadata.php
@@ -13,7 +13,8 @@ foreach ($config['metadata.sources'] as $s) {
         $mdshp = new \SimpleSAML\Metadata\MetaDataStorageHandlerPdo($s);
         $mdshp->initDatabase();
 
-        foreach (glob("metadata/*.php") as $filename) {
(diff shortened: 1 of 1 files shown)


thijskh:
Thanks. Looking at the code I think there are more things that could be fixed, but this makes it better than it was, so merging.

PR #1111, opened 2019-05-08
Fixes #926: PDO Metadata Source Doesn't Load __DYNAMIC:1__ Metadata Sets

### LINKED ISSUE #926: PDO Metadata Source Doesn't Load __DYNAMIC:1__ Metadata Sets
It looks like version 1.16.x will not load any of the __DYNAMIC:1__ metadata sets with the PDO Metadata Storage Handler (I haven't tested it with the other types, but I believe those are unaffected as they don't override the `getMetaData` function from the `MetadataStorageSource` class).

On `frontpage_federation.php` it shows the metadata set with an Entity ID of `__DYNAMIC:1__` and Index of `__DYNAMIC:1__`. If you click on show metadata, it returns:

```
Backtrace:
0 www/saml2/idp/metadata.php:237 (N/A)
Caused by: SimpleSAML_Error_MetadataNotFound: METADATANOTFOUND('%ENTITYID%' => '\'https://idp.example.com/saml2/idp/metadata.php\'')
Backtrace:
2 lib/SimpleSAML/Metadata/MetaDataStorageHandler.php:299 (SimpleSAML_Metadata_MetaDataStorageHandler::getMetaData)
1 lib/SimpleSAML/Metadata/MetaDataStorageHandler.php:319 (SimpleSAML_Metadata_MetaDataStorageHandler::getMetaDataConfig)
0 www/saml2/idp/metadata.php:28 (N/A)
```

On `frontpage_federation.php` page in 1.15.4, it lists the IdP metadata with an Entity ID of `https://idp.example.com/saml2/idp/metadata.php` and an Index of `__DYNAMIC:1__`. Viewing the metadata set will properly show the metadata set as expected.

brycejlowe:
Fixes #926

I recently ran into the issue outlined in #926 and upon doing some research this bug was introduced in commit 7dc9d91403c00b5b2cada54ac6dea14ff01baa6a when optimizations were made to the PDO Metadata Storage Handler to improve performance by using a WHERE clause to lookup a specific metadata entry by entity id.  This obviously doesn’t work with __DYNAMIC:*__ entries because dynamic entity ids are generated and merged into the metadata array when the metadata is fetched.

Rather than rollback these changes and lose the full benefit of the optimizations (I see why the change was made, as the load method fetches all metadata from a given table and then processes it all at once), when an idp_hosted metadata set is queried, I changed the query to use a wildcard to lookup rows that match __DYNAMIC%.  This way, when idp_hosted metadata entries are fetched from the database they always include the dynamic ones as well.  Then it is the responsiblity of the lookupIndexFromEntityId method to select the appropriate metadata entry to return.  The solution to this issue could be handled in a couple of ways, I chose a way that didn’t require schema changes to the overall SQL tables.  Instead of using a LIKE statement we could add a discrete boolean column to flag whether an entry is dynamic or not and tweak the query.

As a part of this fix, I also tried to clean0up some duplicate code and push it into the base class rather than having it copy and pasted in the FlatFile and Pdo handler classes.

### DIFF
--- a/lib/SimpleSAML/Metadata/MetaDataStorageHandlerFlatFile.php
+++ b/lib/SimpleSAML/Metadata/MetaDataStorageHandlerFlatFile.php
@@ -68,7 +68,7 @@ class MetaDataStorageHandlerFlatFile extends MetaDataStorageSource
      *
      * @return array|null An associative array with the metadata, or null if we are unable to load metadata from the given
      *     file.
-     * @throws Exception If the metadata set cannot be loaded.
+     * @throws \Exception If the metadata set cannot be loaded.
      */
     private function load($set)
     {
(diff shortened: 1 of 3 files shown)


brycejlowe:
I have run the test suite against a Docker image of CentOS 7 with PHP 7.0 installed from the Remi repository and am unable to re-produce this error.

tvdijen:
@brycejlowe Excellent work! Will cherry-pick this onto the 1.17-branch as well

ct27stf:
so when will it be released for the 1.17 ?

PR #1112, opened 2019-05-09
Enabeling the handling of unhandled exceptions

Wittev1:
Sometimes there are unhandled exceptions within simplesaml, the end-user gets a default "simplesaml error screen" which confuses them.
This expansion allows us to display an application-matching error and trigger things like Bugsnag.

### DIFF
--- a/www/_include.php
+++ b/www/_include.php
@@ -15,7 +15,8 @@ function SimpleSAML_exception_handler($exception)
         $exception->show();
     } elseif ($exception instanceof \Exception) {
         try {
-            $unhandled_exception_handler = (\SimpleSAML\Configuration::getInstance())->getArray('errors.unhandled_exception_handler', null);
+            $config = \SimpleSAML\Configuration::getInstance();
+            $unhandled_exception_handler = $config->getArray('errors.unhandled_exception_handler', null);
         } catch (\Exception $e) {
             $unhandled_exception_handler = null;
         }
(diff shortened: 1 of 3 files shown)


www/_include.php
```diff
@@ -14,8 +14,17 @@ function SimpleSAML_exception_handler($exception)
     if ($exception instanceof \SimpleSAML\Error\Error) {
         $exception->show();
     } elseif ($exception instanceof \Exception) {
-        $e = new \SimpleSAML\Error\Error('UNHANDLEDEXCEPTION', $exception);
-        $e->show();
+        try {
+            $unhandled_exception_handler = (\SimpleSAML\Configuration::getInstance())->getArray('errors.unhandled_exception_handler', null);
```
tvdijen: This has to go on two lines to remain PHP 5.6 compatible

jaimeperez:
Hi @Wittev1!

Thanks a lot for your contribution. However, I'm not entirely sure this is needed. It has been possible to capture unhandled exceptions and do whatever you need with them since 1.15.0. You just need a custom module (that you can make installable with composer and add as a dependency) implementing the `exception_handler` hook.

It might be more intuitive to have a configuration option for this, I admit that, but at the same time we already have an explosion of configuration options and I would very much like to avoid adding new ones, especially if they are kind of redundant...

So for now I'm closing the PR, but of course that doesn't mean we're not open for discussion if you disagree! 😄

PR #1115, opened 2019-05-14
Feature/add smtp mail

### LINKED ISSUE #1108: Add support for sending email via SMTP rather than calling mail()
I am going through my SimpleSAMLphp environment to docker-ize it for hosting inside of Amazon ECS.  Currently I host it just on a Linux server which has access to the mail command, but my plan is to move it to a Docker container that might not.  It looks like the core email functionality in /SimpleSAML/XHTML/Email.php is just a wrapper around the mail() function.  Have you thought of replacing it with something like or PhpMailer or SwiftMailer?

The end goal is to provide email connectivity from a Docker container with minimal dependencies and modification of php.ini.

**Additional info**
I would be happy to open a pull request to integrate PhpMailer into the project if you would like someone to do it, I wasn't sure what the roadmap for email was.

brycejlowe:
Fixes #1108 

Adding the ability to use SMTP through PHPMailer in SimpleSAMLphp.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -87,24 +87,18 @@ $config = [
      * Set the transport options for the transport method specified.  The valid settings are relative to the
      * selected transport method.
      */
-    //// smtp mail transport options
+    // // smtp mail transport options
     // 'mail.transport.options' => [
-    //     'host' => 'mail.example.org',
-    //     'port' => 25,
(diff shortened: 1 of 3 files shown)


thijskh:
Good work, thanks!

Only 'worrysome' part to me is that config.php keeps growing and growing with a myriad of options. But there's a separate issue about this and this needs a broader fix than the scope of this PR.

PR #1132, opened 2019-05-29
Webmozart assertions

tvdijen:
Start using Webmozart for assertions, like we do in all externalized modules and the saml2-library already.

### DIFF
--- a/lib/SimpleSAML/Auth/Simple.php
+++ b/lib/SimpleSAML/Auth/Simple.php
@@ -9,6 +9,7 @@ use SimpleSAML\Error;
 use SimpleSAML\Module;
 use SimpleSAML\Session;
 use SimpleSAML\Utils;
+use Webmozart\Assert\Assert;
 
 /**
  * Helper class for simple authentication applications.
@@ -41,7 +42,7 @@ class Simple
      */
     public function __construct($authSource, Configuration $config = null, Session $session = null)
     {
-        assert(is_string($authSource));
+        Assert::string($authSource);
 
         if ($config === null) {
             $config = Configuration::getInstance();
@@ -167,7 +168,7 @@ class Simple
 
         $as = $this->getAuthSource();
         $as->initLogin($returnTo, $errorURL, $params);
-        assert(false);
+        Assert::true(false);
     }
 
 
@@ -189,7 +190,7 @@ class Simple
      */
     public function logout($params = null)
     {
-        assert(is_array($params) || is_string($params) || $params === null);
+        Assert::true(is_array($params) || is_string($params) || $params === null);
 
         if ($params === null) {
             $params = Utils\HTTP::getSelfURL();
@@ -201,11 +202,11 @@ class Simple
             ];
         }
 
-        assert(is_array($params));
-        assert(isset($params['ReturnTo']) || isset($params['ReturnCallback']));
+        Assert::isArray($params);
+        Assert::true(isset($params['ReturnTo']) || isset($params['ReturnCallback']));
 
         if (isset($params['ReturnStateParam']) || isset($params['ReturnStateStage'])) {
-            assert(isset($params['ReturnStateParam'], $params['ReturnStateStage']));
+            Assert::true(isset($params['ReturnStateParam'], $params['ReturnStateStage']));
         }
 
         if ($this->session->isValid($this->authSource)) {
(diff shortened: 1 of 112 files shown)


tvdijen:
What's left using old assert()s are two deprecated classes and some old php-style templates. Not worth the effort of migrating.

jornane:
This means that the wishes from the administrator by means of [`assert_options` or `assert.`-settings in php.ini](https://www.php.net/manual/en/function.assert-options.php) no longer will be respected.

If assertions must be done using a library instead of calling the `assert` language construct of PHP, it would be preferred to use a library that uses the PHP internals instead of inventing its own way of doing things. At the very least it should respect the PHP settings.

lib/SimpleSAML/Auth/Simple.php
```diff
@@ -318,7 +319,7 @@ public function getAuthDataArray()
      */
     public function getLoginURL($returnTo = null)
     {
-        assert($returnTo === null || is_string($returnTo));
+        Assert::nulOrString($returnTo);
```
jornane: @fkooman caught an error here, I haven't checked if there are any more.

tvdijen: Wrong person tagged here, but nice catch! Thanks!

tvdijen:
Hi @jornane ! We are aware of this and do this on purpose.
It was discussed here: https://github.com/simplesamlphp/saml2/issues/90

@
```diff
@@ -0,0 +1,23 @@
+pick 6bb9a08 Migrate assertions to Webmozart
```
jornane: I don't think you really wanted to commit this?

jornane:
Thanks @tvdijen ! I've read through https://github.com/simplesamlphp/saml2/issues/90 but I think maybe I'm missing something?

> Starting in PHP 7.2 the old behavior [will be deprecated](https://wiki.php.net/rfc/deprecations_php_7_2#assert_with_string_argument).
This is about using a string as the assertion, where PHP has to run `eval()` on that string. SimpleSAMLphp doesn't do that so it doesn't apply to us.

The webmozart project uses [__callStatic](https://github.com/webmozart/assert/blob/1.4.0/src/Assert.php#L1095), which breaks static code analyses, making mistakes like ba0b826 go unnoticed. At the same time its yet another dependency to replace a one-liner with another one-liner. I already mentioned respecting the administrators wishes with regard to php.ini settings for assertions.

At the end of the day, this is a bikeshed issue, so if this has already been decided, I won't stand in the way. However I do want to point out that the original reasoning for this change (deprecation) has been based on a misunderstanding. The current way SimpleSAMLphp does things is **not** currently, or about to be, deprecated.

tvdijen:
OK, bikeshed it is! 4pm after detention :satisfied:
Maybe I've linked to the wrong issue, or it was discussed offline.. The thing is, 99% of the assertions we have are for type checking and we want this to be tested at runtime too..
@jaimeperez you have anything to add here? I don't remember the discussions too well..

jaimeperez:
The main reason for having something like _webmozart/assert_ is that we **need** to check method args and logic conditions. We haven't been using `assert()` (_only_) for development purposes, but also as sanity checks. With that in mind, keeping using `assert()` is not fine because you can (and probably _should_) turn it off in production. Given that this is security-related software granting access to resources, an error somewhere could lead to a security incident that could be exploited, and that is something we **shouldn't** allow to happen. If there is a bug in the software, and that can potentially lead to a security issue, it's **our** responsibility to avoid that, not the user's to configure the web server to stop on a failed `assert()`.

fkooman:
Using assert (any kind of assert) to do input validation is not ideal, whether it is the built in assert or a library doing "assert".

Many of the "asserts" here (in PR) seem to be about testing whether or not something is an array. Since PHP 5.1 the `array` typehint is supported in PHP, but it doesn't seem to be used.

Furtermore, a nicer approach would be to (fully) document a function's parameters using docblock and run static code analysis tools like Psalm, Phan or PHPstan to help you make sure a function taking a `string` can never get an 
`int`. Then you do not need the "assert" any longer. This works especially well for internal functions that never receive "user" input. In addition, this prepares you for a migration to PHP 7 where you can use e.g. php-cs-fixer to use your docblock and convert it to "real" PHP static types.

Of course, you can still keep "assert", but that should really be for testing for exceptional cases where (static) type hinting does not (sufficiently) work, e.g. missing array keys. Although, there are some solutions for that as well in the static code analysis tools.

Also, as this intended for 2.0, wouldn't it be possible to require PHP >= 7.2 (RHEL/CentOS 8) for 2.0? :-)

tvdijen:
This PR was (at first) never targeted for 2.0, so all your points are valid and this PR exists merely as a reminder.
I'm working on a redo of the typehints-PR, where the source will be fully typehinted.. Then only a few assertions will be left, mostly checking for the existence of array keys. Hopefully we can branch of 1.18 soon and start working on the real deal..

jaimeperez:
I agree that using any kind of assert is not ideal, but it's better than nothing. We have already started using type hints and static analysis with psalm. Tim has done an incredible job on that so far, and fixed a lot of issues. This includes having proper phpdoc blocks that can be used with automatic tools like _php-cs-fixer_.

Still, that's not enough. First, because typehints don't cover all cases. Due to PHP's flexibility, we have some places where an argument can have several types. You cannot typehint that. Second, because adding typehints breaks method signatures. And you cannot break method signatures without breaking backwards-compatibility.

Regarding the minimum version of PHP required, we shall see, although I'm not particularly optimistic. We still have plenty of users on PHP 5.x, unfortunately.

tvdijen:
> Regarding the minimum version of PHP required, we shall see, although I'm not particularly optimistic. We still have plenty of users on PHP 5.x, unfortunately.

We're gonna have to move forward at some point and the major distributions all have a possibility to support PHP 7+.. RHEL 8 comes with PHP7.2 by default, so this is the time to move if you ask me

PR #1133, opened 2019-05-31
Add logging for organizational logins

rmisev:
This adds the same logging to `UserPassOrgBase::handleLogin` function
as it is in the `UserPassBase::handleLogin` (UserPassBase.php).

Not sure if IP address logging (`$_SERVER['REMOTE_ADDR']`) is still required here.


tvdijen:
I agree that logging IP-addresses is unnecessary.. It will in most situations be the IP of a reverse proxy anyway..

jaimeperez:
I think it's nice to have consistency between the two, and I don't have a strong opinion on the IP address (it can be already logged by changing the log format, right?, so let's merge it for now and decide later on the IP. Thanks @rmisev!

PR #1134, opened 2019-05-31
Handle duplicate insertion also for SQL Server

Krisell:


tvdijen:
Thanks @Krisell! 
While your PR makes sense, I wonder if we should make error catching more specific by also testing for the correct driver.. This error codes may mean something totally different for different backend.. What do you think?

Krisell:
Yeah that sounds reasonable. This was mostly a "panic" fix on my part and I made the PR to have it documented, but I'll try to find the time to dig deeper into this and see if we can solve it in a better way. If there is no "insert or update" for SQL Server, it's perhaps better to check for existence first instead.

Krisell:
Here's the full stack trace (without this fix) for future reference:

```
 Error while saving data: SQLSTATE[23000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Violation of PRIMARY KEY co
nstraint 'PK__SSO_SAML__9AAC43D5A1C04311'. Cannot insert duplicate key in object 'dbo.SAML_kvstore'. The duplicate key value is (session, f8688675b98c53cda16877977a6761
b0).
 Unable to save session.
 PDOException: SQLSTATE[23000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Violation of PRIMARY KEY constraint 'P
K__SSO_SAML__9AAC43D5A1C04311'. Cannot insert duplicate key in object 'dbo.SAML_kvstore'. The duplicate key value is (session, f8688675b98c53cda16877977a6761b0).
 Backtrace:
 13 /var/simplesamlphp/lib/SimpleSAML/Store/SQL.php:236 (PDOStatement::execute)
 12 /var/simplesamlphp/lib/SimpleSAML/Store/SQL.php:236 (SimpleSAML\Store\SQL::insertOrUpdate)
 11 /var/simplesamlphp/lib/SimpleSAML/Store/SQL.php:363 (SimpleSAML\Store\SQL::set)
 10 /var/simplesamlphp/lib/SimpleSAML/SessionHandlerStore.php:78 (SimpleSAML\SessionHandlerStore::saveSession)
 9 /var/simplesamlphp/lib/SimpleSAML/Session.php:459 (SimpleSAML\Session::save)
 8 /var/simplesamlphp/lib/SimpleSAML/Utils/HTTP.php:189 (SimpleSAML\Utils\HTTP::redirect)
 7 /var/simplesamlphp/lib/SimpleSAML/Utils/HTTP.php:970 (SimpleSAML\Utils\HTTP::redirectTrustedURL)
 6 /var/simplesamlphp/modules/core/lib/Auth/UserPassBase.php:230 (SimpleSAML\Module\core\Auth\UserPassBase::authenticate)
 5 /var/simplesamlphp/lib/SimpleSAML/Auth/Source.php:193 (SimpleSAML\Auth\Source::initLogin)
 4 /var/simplesamlphp/lib/SimpleSAML/Auth/Simple.php:165 (SimpleSAML\Auth\Simple::login)
 3 /var/simplesamlphp/lib/SimpleSAML/Auth/Simple.php:108 (SimpleSAML\Auth\Simple::requireAuth)
 2 /var/simplesamlphp/modules/core/www/as_login.php:33 (require)
 1 /var/simplesamlphp/lib/SimpleSAML/Module.php:236 (SimpleSAML\Module::process)
 0 /var/simplesamlphp/www/module.php:9 (N/A)
```

tvdijen:
Looking at this code a bit more... It's so wrong in so many ways..
Basically this is the 'default' case of the above switch-statement...  But why do we try/catch errors here if we don't do it for the first two cases? We asume they don't fail because they support the 'replace' statement..

My first thought for a more decent fix here is that we should move this code to the 'default' of the switch, wrapping the try-catch around it and test for the combination of errorcode + driver when handling errors.. Ideally this should all be handled by an abstraction layer instead of by SimpleSAMLphp itself.. If you have any thoughts on this, that would be awesome.. I'm sure if something like `symfony/dbal` would help us out in cases like this, but we get so many DB incompatibility-issues lately that we have to start acting on it..

Krisell:
Yeah, the error handling is only to achieve the same functionality, other errors are still thrown and handled (well, logged) up the chain. 

This is perhaps a slightly better way of writing the current solution:

```php
} catch (\PDOException $e) {
    $ecode = (string) $e->getCode();

    $duplicateInsertCode = [
        'postgresql' => '23505',
        'sqlsrv' => '23000'
    ][$this->driver] ?? false;    

    if ($duplicateInsertCode && $duplicateInsertCode === $ecode) {
        return;
    }

    Logger::error('Error while saving data: '.$e->getMessage());
    throw $e;
}
```

tvdijen:
Oh good one! As a quick fix! Can you think about the bigger picture, like the external library? Or is that out of your comfort-zone?

Krisell:
Here's an attempt at making the code PHP 5.x compatible. I have not been able to test though, and as you can see this PR doesn't include any tests at all, so any help is appreciated there.

```php
} catch (\PDOException $e) {
    $duplicateInsertErrorCodes = [
        'postgresql' => '23505',
        'sqlsrv' => '23000'
    ];

    if (!array_key_exists($this->driver, $duplicateInsertErrorCodes) || $duplicateInsertErrorCodes[$this->driver] !== (string) $e->getCode()) {
        Logger::error('Error while saving data: '.$e->getMessage());
        throw $e;
    }
}

```

And I'm afraid the bigger picture is indeed out of my comfort-zone, but I'll follow any changes you make and I can help with SQL Server testing if that's necessary.

tvdijen:
Hi @Krisell! Could you please apply your last suggestion to this PR?
Then we can easily review it and see if unit tests pass.. Thanks!

Krisell:
There you go.

To clarify, this PR makes it slightly easier to add additional drivers that need to be handled the same way, but most importantly scopes the error code to the specific driver, in order to not invisibly swallow other errors.

tvdijen:
I think it makes sense to backport this to 1.17.x

tvdijen:
I took the safe route and fixed it 'the ugly way' in 1.17 and leave the 'pretty way' for a 1.18-release

Issue #1101, opened 2019-04-22
MDQ source and IDP List incompatibility.


pradtke:
The SP `authsource` supports passing a IDPList of which IdPs can be used for the authentication. The `authsource` then loads all idp metadata and performs an intersection of the idps.

https://github.com/simplesamlphp/simplesamlphp/blob/f303200ff1b63559cdee87ba6b78c30a29fb48cf/modules/saml/lib/Auth/Source/SP.php#L760-L770

The "load all idp metadata" is not compatible with MDQ, which cannot iterate metadata as part of the protocol
https://github.com/simplesamlphp/simplesamlphp/blob/d386e132e4b5a16d667a4e5ea213b56ad868827b/lib/SimpleSAML/Metadata/Sources/MDQ.php#L110-L114

In our experience the size of IDP List is a small number (usually less than 15), so instead of loading all the metadata and filtering we could load just the specified entities.

So something like:
```php
// In MetaDataStorageHandler

    /**
     * This function loads the metadata for entity IDs in $entityIds. It is returned as an associative array
     * where the key is the entity id. An empty array may be returned if no matching entities were found
     * @param array $entityIds The entity ids to load
     * @param string $set The set we want to get metadata from.
     * @return array An associative array with the metadata for the requested entities, if found.
     */
    public function getMetaDataForEntities(array $entityIds, $set)
    {
                // iterate through metadata sources and call the sources getMetaDataForEntities
               // and merges results as needed
    }
```

and in

```php
// In MetaDataStorageSource

    /**
     * This function loads the metadata for entity IDs in $entityIds. It is returned as an associative array
     * where the key is the entity id. An empty array may be returned if no matching entities were found
     * @param array $entityIds The entity ids to load
     * @param string $set The set we want to get metadata from.
     * @return array An associative array with the metadata for the requested entities, if found.
     */
    public function getMetaDataForEntities(array $entityIds, $set)
    {
                // iterate through $entityIds and call $this->getMetaData()
               // Override for PDO and FlatFile to be a more efficient implementation
    }
```

Then lastly the SP authsource would be changed to only load the list of desired entity IDs.

Thoughts?

thijskh:
I agree with you that the IDPList is likely small, so I think it is a feasible and useful approach.

tvdijen:
Was closed in #1138

PR #1138, opened 2019-06-06
Allow list of entity ids to be loaded to better support idpList

pradtke:
Fix for #1101
Certain metadata sources, like mdq, don't support loading all metadata which
prevented them from being used with idplist. This change adds support for allowing
metadata sources to load a list of entities, allowing mdq to be used for idp list.


lib/SimpleSAML/Metadata/MetaDataStorageHandler.php
```diff
@@ -256,6 +257,38 @@ public function getPreferredEntityIdFromCIDRhint($set, $ip)
...
+    /**
+     * This function loads the metadata for entity IDs in $entityIds. It is returned as an associative array
+     * where the key is the entity id. An empty array may be returned if no matching entities were found
+     * @param array $entityIds The entity ids to load
+     * @param string $set The set we want to get metadata from.
+     * @return array An associative array with the metadata for the requested entities, if found.
+     */
+    public function getMetaDataForEntities(array $entityIds, $set)
+    {
+
```
tvdijen: This newline does not belong here as per PSR-2

lib/SimpleSAML/Metadata/MetaDataStorageSource.php
```diff
@@ -53,7 +53,7 @@ public static function parseSources($sourcesConfig)
      *
      * @param array $sourceConfig Associative array with the configuration for this metadata source.
      *
-     * @return mixed An instance of a metadata source with the given configuration.
+     * @return MetaDataStorageSource An instance of a metadata source with the given configuration.
```
tvdijen: I'd like to see the fully namespaced class name in phpdocs

tvdijen:
Hi @pradtke ! Thanks for this, but unfortunately the Travis tests are failing.. I also left a few nitpicks ;)

lib/SimpleSAML/Metadata/MetaDataStorageHandler.php
pradtke: Fixed

lib/SimpleSAML/Metadata/MetaDataStorageSource.php
pradtke: Fixed

PR #1139, opened 2019-06-06
Take more advantage of namespaces / use statements

tvdijen:
This PR adds many use-statements and eliminates the use of fully namespaced classes throughout the codebase. This makes code better readable and shortens line length drastically.


tvdijen:
Since this is hard to review, Jaime approved this PR offline

PR #1153, opened 2019-06-28
Add support for RFC6265bis SameSite cookie attribute in 1.17

ghalse:
Chrome 76 [changes the behaviour of cookies](https://blog.chromium.org/2019/05/improving-privacy-and-security-on-web.html) in a way that breaks SimpleSAMLphp's session cookie. There has been [discussion of this on the REFEDS list](https://lists.refeds.org/sympa/arc/refeds/2019-06/msg00005.html), and my own testing confirms it is a problem in SSP 1.17.x.

In order to work around this problem, we need to allow people to set the SameSite=None attribute on session cookies. This patch introduces a new config option `session.cookie.samesite` that allows people to specify how this should be treated. It defaults to reproducing the existing behaviour of not setting the SameSite attribute at all, but can be set to `'None'` if people need to work around the Chrome 76 problem.

Unfortunately, PHP only introduced native support for this in the setcookie() function [in version 7.3](https://www.php.net/manual/en/function.setcookie.php#refsect1-function.setcookie-changelog). However, [I managed to find](https://stackoverflow.com/questions/39750906/php-setcookie-samesite-strict) a somewhat ugly workaround for versions < 7.3 (tested against 7.0.33 that ships with Ubuntu 16.04LTS). Thus the patch special cases for the two different versions of PHP.

The patch is against the simplesamlphp-1.17 branch, since that's what I need in our environment.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -748,6 +748,7 @@ $config = [
     'language.cookie.secure' => false,
     'language.cookie.httponly' => false,
     'language.cookie.lifetime' => (60 * 60 * 24 * 900),
+    'language.cookie.samesite' => null,
 
     /**
      * Custom getLanguage function called from SimpleSAML\Locale\Language::getLanguage().
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -545,6 +545,18 @@ $config = [
     'session.cookie.secure' => false,
 
     /*
+     * Set the SameSite attribute in the cookie.
+     *
+     * You can set this to the strings 'None', 'Lax', or 'Strict' to support
+     * the RFC6265bis SameSite cookie attribute. If set to null, no SameSite
+     * attribute will be sent.
+     *
+     * Example:
+     *  'session.cookie.samesite' => 'None',
+     */
+    'session.cookie.samesite' => null,
+
+    /*
      * Options to override the default settings for php sessions.
      */
     'session.phpsession.cookiename' => 'SimpleSAML',
--- a/lib/SimpleSAML/Locale/Language.php
+++ b/lib/SimpleSAML/Locale/Language.php
@@ -419,6 +419,7 @@ class Language
             'path'     => ($config->getString('language.cookie.path', '/')),
             'secure'   => ($config->getBoolean('language.cookie.secure', false)),
             'httponly' => ($config->getBoolean('language.cookie.httponly', false)),
+            'samesite' => ($config->getBoolean('language.cookie.samesite', null)),
         ];
 
         HTTP::setCookie($name, $language, $params, false);
--- a/lib/SimpleSAML/Locale/Language.php
+++ b/lib/SimpleSAML/Locale/Language.php
@@ -419,7 +419,7 @@ class Language
             'path'     => ($config->getString('language.cookie.path', '/')),
             'secure'   => ($config->getBoolean('language.cookie.secure', false)),
             'httponly' => ($config->getBoolean('language.cookie.httponly', false)),
-            'samesite' => ($config->getBoolean('language.cookie.samesite', null)),
+            'samesite' => ($config->getString('language.cookie.samesite', null)),
         ];
 
         HTTP::setCookie($name, $language, $params, false);
--- a/lib/SimpleSAML/SessionHandlerPHP.php
+++ b/lib/SimpleSAML/SessionHandlerPHP.php
@@ -77,13 +77,28 @@ class SessionHandlerPHP extends SessionHandler
         $params = $this->getCookieParams();
 
         if (!headers_sent()) {
-            session_set_cookie_params(
-                $params['lifetime'],
-                $params['path'],
-                $params['domain'],
-                $params['secure'],
-                $params['httponly']
-            );
+            if (\PHP_VERSION_ID >= 70300) {
+                session_set_cookie_params([
+                    'lifetime' => $params['lifetime'],
+                    'path' => $params['path'],
+                    'domain' => $params['domain'],
+                    'secure' => $params['secure'],
+                    'httponly' => $params['httponly'],
+                    'samesite' => $params['samesite'],
+                ]);
+            } else {
+                /* in older versions of PHP we need a nasty hack to set RFC6265bis SameSite attribute required by Chrome >= 76 */
+                if ($params['samesite'] !== null and !preg_match('/;\s+samesite/i', $params['path'])) {
+                    $params['path'] .= '; SameSite='.$params['samesite'];
+                }
+                session_set_cookie_params(
+                    $params['lifetime'],
+                    $params['path'],
+                    $params['domain'],
+                    $params['secure'],
+                    $params['httponly']
+                );
+            }
         }
 
         $savepath = $config->getString('session.phpsession.savepath', null);
@@ -113,13 +128,17 @@ class SessionHandlerPHP extends SessionHandler
         session_write_close();
 
         session_name($this->previous_session['name']);
-        session_set_cookie_params(
-            $this->previous_session['cookie_params']['lifetime'],
-            $this->previous_session['cookie_params']['path'],
-            $this->previous_session['cookie_params']['domain'],
-            $this->previous_session['cookie_params']['secure'],
-            $this->previous_session['cookie_params']['httponly']
-        );
+        if (\PHP_VERSION_ID >= 70300) {
+            session_set_cookie_params($this->previous_session['cookie_params']);
+        } else {
+            session_set_cookie_params(
+                $this->previous_session['cookie_params']['lifetime'],
+                $this->previous_session['cookie_params']['path'],
+                $this->previous_session['cookie_params']['domain'],
+                $this->previous_session['cookie_params']['secure'],
+                $this->previous_session['cookie_params']['httponly']
+            );
+        }
         session_id($this->previous_session['id']);
         $this->previous_session = [];
         @session_start();
@@ -330,13 +349,17 @@ class SessionHandlerPHP extends SessionHandler
             session_write_close();
         }
 
-        session_set_cookie_params(
-            $cookieParams['lifetime'],
-            $cookieParams['path'],
-            $cookieParams['domain'],
-            $cookieParams['secure'],
-            $cookieParams['httponly']
-        );
+        if (\PHP_VERSION_ID >= 70300) {
+            session_set_cookie_params($cookieParams);
+        } else {
+            session_set_cookie_params(
+                $cookieParams['lifetime'],
+                $cookieParams['path'],
+                $cookieParams['domain'],
+                $cookieParams['secure'],
+                $cookieParams['httponly']
+            );
+        }
 
         session_id($sessionID);
         @session_start();
--- a/lib/SimpleSAML/SessionHandlerPHP.php
+++ b/lib/SimpleSAML/SessionHandlerPHP.php
@@ -87,7 +87,7 @@ class SessionHandlerPHP extends SessionHandler
                     'samesite' => $params['samesite'],
                 ]);
             } else {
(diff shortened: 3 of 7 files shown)


lib/SimpleSAML/SessionHandlerPHP.php
```diff
@@ -77,13 +77,28 @@ protected function __construct()
...
-                $params['httponly']
-            );
+            if (\PHP_VERSION_ID >= 70300) {
+                session_set_cookie_params([
+                    'lifetime' => $params['lifetime'],
+                    'path' => $params['path'],
+                    'domain' => $params['domain'],
+                    'secure' => $params['secure'],
+                    'httponly' => $params['httponly'],
+                    'samesite' => $params['samesite'],
```
thijskh: This does not seem to work on PHP 7.3. According to the docs, "The value of the samesite element should be either Lax or Strict." and maybe somewhat surprisingly the docs seem to match reality: None is not accepted. So maybe also for 7.3 we should use the below strategy (also in the other invocations of this and of setcookie).

config-templates/config.php
```diff
@@ -544,6 +544,18 @@
...
+     * Set the SameSite attribute in the cookie.
+     *
+     * You can set this to the strings 'None', 'Lax', or 'Strict' to support
+     * the RFC6265bis SameSite cookie attribute. If set to null, no SameSite
+     * attribute will be sent.
+     *
+     * Example:
+     *  'session.cookie.samesite' => 'None',
+     */
+    'session.cookie.samesite' => null,
```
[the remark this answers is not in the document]
ghalse: I always try to default to not changing existing behaviour. In this case, and given the refinements in Chrome's plans, I think it should default to `null` now and should change to `None` once a clear date for their change is known.

lib/SimpleSAML/SessionHandlerPHP.php
```diff
@@ -77,13 +77,28 @@ protected function __construct()
...
+                ]);
+            } else {
+                /* in older versions of PHP we need a nasty hack to set RFC6265bis SameSite attribute required by Chrome >= 76 */
```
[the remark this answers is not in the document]
ghalse: Yes, that information was clarified after I wrote the patch :)

lib/SimpleSAML/SessionHandlerPHP.php
ghalse: We tested this against 7.3.6 and found it did work. I also looked at the PHP commit that introduced this, and there's no validation of the string in it. (On the other hand, the hacky version definitely does *not* work for PHP >= 7.3)

thijskh: I tested this also with 7.3.6 and reached my conclusion (changing it to `Lax` did cause the attribute to appear). Will investigate further then.

lib/SimpleSAML/Locale/Language.php
```diff
@@ -419,6 +419,7 @@ public static function setLanguageCookie($language)
             'path'     => ($config->getString('language.cookie.path', '/')),
             'secure'   => ($config->getBoolean('language.cookie.secure', false)),
             'httponly' => ($config->getBoolean('language.cookie.httponly', false)),
+            'samesite' => ($config->getBoolean('language.cookie.samesite', null)),
```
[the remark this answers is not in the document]
ghalse: Fixed in c4c22ab33b5f4babc2c841ebaad5ec2f55b522f6

lib/SimpleSAML/SessionHandlerPHP.php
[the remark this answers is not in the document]
ghalse: Removed references in 4d468018ea7f89e3b0c989ea8ddd7ca9277dc0b9

lib/SimpleSAML/SessionHandlerPHP.php
ghalse: My unit test still isn't quite right (it's checking an absolute string value whereas the formatting varies per version of PHP). Nevertheless, even the failing unit test shows the correct values for 7.3: https://travis-ci.org/simplesamlphp/simplesamlphp/jobs/554825414#L647

tvdijen:
Great work on the added unit tests! 🥇

lib/SimpleSAML/SessionHandlerPHP.php
ghalse: This is now explicitly unit tested. The `session_set_cookie_params` from this comment is tested at https://github.com/simplesamlphp/simplesamlphp/pull/1153/commits/4d341015e4a2f736652a08dd0685438ac7071279#diff-beeb6dbefe9cf458457456e4cb21a72dR80 and `setcookie` is tested at https://github.com/simplesamlphp/simplesamlphp/pull/1153/commits/5bab223804d7a518ae7ec95077899964d617ece9#diff-8af11dde64458ba57910e950558388ceR498

pradtke:
FWIW, iOS and Safari on mac treat a setting of None as Strict (per https://bugs.webkit.org/show_bug.cgi?id=198181). It looks like the targeted fix will be in iOS 13 and and macOS 10.15.  So enabling this setting at the current time may break things for those users.

ghalse:
> FWIW, iOS and Safari on mac treat a setting of None as Strict (per https://bugs.webkit.org/show_bug.cgi?id=198181). It looks like the targeted fix will be in iOS 13 and and macOS 10.15. So enabling this setting at the current time may break things for those users.

This is a strong argument for retaining the (backwards compatible) default of `null` for now.

The way things are currently implemented, this pull request merely introduces support for it. That means we don't need to panic if things are accelerated; people can simply turn it on when needed. However, we should probably only decide to change the default behaviour to None once it is tested with all major browsers' beta implementations.

It might also be worth looking at this architecturally in the longer term. The Shib native SP manages to avoid a dependency on a session cookie, and so isn't affected. I've not delved into how they're doing that, but it might make sense to do so. (And that doesn't negate most of this patch, as it introduces support for it in the generic cookie setting method which might be useful for other things.)

lib/SimpleSAML/SessionHandlerPHP.php
thijskh: Thanks. It does indeed work for me. It's Firefox' inspector that doesn't show the unknown value.

thijskh:
@ghalse Do you also want to propose a change for master? Otherwise we'd lose this functionality on the next release

ghalse:
> @ghalse Do you also want to propose a change for master? Otherwise we'd lose this functionality on the next release

That occurred to me. Do you want me to rebase the existing request to master, or create a new pull request based on the 465faf6 commit?

thijskh:
A new PR is probably the cleanest

tvdijen:
If I may add a nitpick for master..
I'd prefer to use `version_compare()` instead of the PHP_VERSION-constant, since we use the former throughout the rest of the codebase already.

jaimeperez:
@jornane and I were discussing this a few days ago. As @pradtke was suggesting, using this would then break Safari for now. I'm guessing it's ok to not use this feature for now, so that Safari continues to work, since this new behaviour in Chrome has not yet been rolled out. But I wonder how much time will apple-users have to update their devices (those who will or even can).

My biggest concern is that this will be a choice for SSP admins: break Chrome users or break Safari users. None of the options are good. So we were discussing the possibility of applying this change conditionally, depending on the user agent. That would be setting `SameSite=none` starting from current Chrome, and leaving it for all other user agents.

I'd really prefer **not** to go there. I really **don't** want to start messing around with user agents. But I don't see many alternatives here, unfortunately...

What do you guys think?

ghalse:
> But I wonder how much time will apple-users have to update their devices (those who will or even can).
I think we've actually got some time to wait-and-see what Safari does. It is possible that they'll respond to Google/Chromium's change well ahead of time since the problem is not unique to SSP and will affect their users in many other places. And if they don't it is likely that users will need to upgrade/workaround for reasons other than SSP. So I don't think we need a good answer to this just yet.

> My biggest concern is that this will be a choice for SSP admins: break Chrome users or break Safari users. 

It has also occurred to me that we can leave the decision to SSP admins, who (theoretically) know their user base. Since SSP's config.php is parsed PHP, the could elect to put the browser detection stuff in there themselves. For future reference of people looking for solutions, I'm suggesting something like:
```php
$config = [
    'session.cookie.samesite' => (strpos($_SERVER['HTTP_USER_AGENT'], 'Safari/') !== false ? 'None' : 'Lax'),
];
```
(This could presumably also be done with [get_browser()](https://www.php.net/manual/en/function.get-browser.php) subject to [its caveats](https://www.php.net/manual/en/misc.configuration.php#ini.browscap).)

jaimeperez:
That's an interesting approach indeed!

Still, many people use SSP for its simplicity, and this would be anything but simple. Would it make sense to have this explained in the documentation, or in the wiki, or maybe even already configured so that people can copy and paste? Maybe even distributed with the releases and enabled by default, with the possibility to remove it?

pradtke:
@ghalse In your user agent detection example, is that what you actually used? Or were you just showing that you can do such a thing? I ask because both Chrome and Safari put `Safari/` in their user agent string.

ghalse:
@pradtke it was merely an idea; we currently use the wait-and-hope-Safari-fix-this-before-Chrome-breaks-it approach in production.

pradtke:
FWIW this is our current `samesite` config logic.

```php
    // Only enable for Chrome. Safari in iOS 12 and pre-2019 MacOs Catalina are incompatible
    'session.cookie.samesite' => strpos($userAgent, 'Chrome/') === false ? null : 'None',
```

thijskh:
We use this information: https://www.chromium.org/updates/same-site/incompatible-clients
to enable it except when you have any of the mentioned browser versions.
```
    acl safari hdr(agent_internal) -m reg "Version\/.* Safari\/"
    acl chrome51to66 hdr(agent_internal) -m reg "Chrom(e|ium)\/(5[1-9]|6[0-6])"
    acl ucbrowser hdr(agent_internal) -m sub UCBrowser 
```

ghalse:
Since the deadline is looming for this, I thought to share our PHP implementation of the Chromium project's algorithm that @thijskh mentioned. You can find it at [samesitenone.inc.php](https://github.com/simplesamlphp/simplesamlphp/files/4158733/samesitenone.inc.php.txt)

YMMV, but we currently use it in `config.php` like this:
```php
<?php
require_once('samesitenone.inc.php');
$config = [
# ...
'session.cookie.samesite' => (shouldSendSameSiteNone() ? 'None' : null),
# ...
'language.cookie.samesite' => (shouldSendSameSiteNone() ? 'None' : null),
# ...
];
```
This should be "safe" in the sense that it reverts to the old behaviour of not setting SameSite on cookies for Safari and other known-incompatible clients.

jaimeperez:
Thanks for sharing Guy!

Here we took a slightly different approach, [doubling cookies](https://auth0.com/docs/sessions/concepts/cookie-attributes) with and without `SameSite`. We've been doing tests and things keep working reliably in all browsers we could find. The only thing is that it needs some changes applied to the code base, so I was wondering whether we should have that in place in SSP and have a new release. The advantage then is that people won't need to do anything in their deployments (except for updating, of course).

pradtke:
I believe [doubling cookies](https://forums.aws.amazon.com/ann.jspa?annID=7413) it was AWS rolled out, though their major concern was how the Chrome change affected CORS.   I imagine cookie doubling is more fool proof than user-agent parsing for deployers. For those of us that have deployed alternate solutions, is it something we can toggle off?

ghalse:
The doubling cookies approach looks sensible and is new since we tested. The code changes wouldn't be significant IIRC - because cookies are pretty much all abstracted.

Ultimately I believe that giving users as much flexibility as possible to control their approach to the problem is sensible though. So a works-by-default approach in a new release that allows for users to toggle it has my vote :)

PR #1156, opened 2019-07-03
Rewrite table upgrade

### LINKED ISSUE #1118: Invalid ALTER query to upgrade logout table schema in simplesamlphp/modules/saml/lib/SP/LogoutStore.php
The query used to alter the _authSource type from VARCHAR(30) to VARCHAR(255), at line 32 of simplesamlphp/modules/saml/lib/SP/LogoutStore.php (branch 1.17) is incompatible with SQLite and will produce  a syntax error. Therefore, the table schema versions is never upgraded when upgrading from an older version. This seems to have been changed in commit 1681c295210b1966462a90b0606288b83882e9c1, which was commited on 16 Apr 2017.

You can not use the ALTER TABLE statement to modify a column in SQLite. Instead you will need to rename the table, create a new table, and copy the data into the new table.

Fortunately this bug appears to be non-breaking unless your authSource hostname is longer than 30 characters and you're using SQLite as your session db engine.

tvdijen:
Closes #1118

### DIFF
--- a/modules/saml/lib/SP/LogoutStore.php
+++ b/modules/saml/lib/SP/LogoutStore.php
@@ -50,17 +50,40 @@ class LogoutStore
             switch ($store->driver) {
                 case 'pgsql':
                     // This does not affect the NOT NULL constraint
-                    $query = 'ALTER TABLE '.$store->prefix.
-                        '_saml_LogoutStore ALTER COLUMN _authSource TYPE VARCHAR(255)';
+                    $update = ['ALTER TABLE '.$store->prefix.
+                        '_saml_LogoutStore ALTER COLUMN _authSource TYPE VARCHAR(255)'];
+                    break;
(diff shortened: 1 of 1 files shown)


jaimeperez:
Should we do this only for sqlite, instead of all drivers? It's quite inefficient compared to just an `alter column` statement...

tvdijen:
Does it really matter for a one-time table upgrade?

jaimeperez:
It depends a bit on when is this executed. If it's in the middle of a transaction and the consequence is that the user has to wait a few seconds, I would say yes. On the other hand, I doubt that'll be the case... 😅

tvdijen:
It needed a rebase anyway because master changed...  How does this look?

jaimeperez:
Great, it looks great! 😉

PR #1157, opened 2019-07-04
Fix Psalm-issues batch #1

tvdijen:
Fixes all Psalm-issues in all subdirs under `lib/SimpleSAML`.
I hope it is not too big to review. I've added a few comments to the parts that are likely to raise questions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "gettext/gettext": "^4.6",
-        "jaimeperez/twig-configurable-i18n": "^2.0",
+        "jaimeperez/twig-configurable-i18n": "^2.1",
         "phpmailer/phpmailer": "^6.0",
         "robrichards/xmlseclibs": "^3.0",
         "simplesamlphp/saml2": "^3.3",
--- a/lib/SimpleSAML/Locale/Translate.php
+++ b/lib/SimpleSAML/Locale/Translate.php
@@ -108,7 +108,7 @@ class Translate
                 $fileName = substr($name, $sepPos + 1);
                 $dictDir = Module::getModuleDir($module).'/dictionaries/';
             } else {
-                $dictDir = $this->configuration->getPathValue('dictionarydir', 'dictionaries/');
+                $dictDir = $this->configuration->getPathValue('dictionarydir', 'dictionaries/') ?: 'dictionaries/';
                 $fileName = $name;
             }
 
@@ -261,7 +261,7 @@ class Translate
      * @param bool $striptags
      * @deprecated Not used in twig, gettext
      *
-     * @return string  The translated tag, or a placeholder value if the tag wasn't found.
+     * @return string|null  The translated tag, or a placeholder value if the tag wasn't found.
      */
     public function t(
         $tag,
@@ -292,6 +292,7 @@ class Translate
             );
 
             // for backwards compatibility
+            /** @psalm-suppress PossiblyInvalidArgument */
             if (!$replacements && ($this->getTag($tag) === null)) {
                 Logger::warning(
                     'Code which uses $fallbackdefault === FALSE should be updated to use the getTag() method instead.'
@@ -387,6 +388,7 @@ class Translate
         } else {
             $filebase = $this->configuration->getPathValue('dictionarydir', 'dictionaries/');
         }
+        $filebase = $filebase ?: 'dictionaries/';
 
         $lang = $this->readDictionaryFile($filebase.$file);
         Logger::debug('Translate: Merging language array. Loading ['.$file.']');
@@ -520,9 +522,9 @@ class Translate
     /**
      * Pick a translation from a given array of translations for the current language.
      *
-     * @param array $context An array of options. The current language must be specified as an ISO 639 code accessible
+     * @param array|null $context An array of options. The current language must be specified as an ISO 639 code accessible
      * with the key "currentLanguage" in the array.
-     * @param array $translations An array of translations. Each translation has an ISO 639 code as its key, identifying
+     * @param array|null $translations An array of translations. Each translation has an ISO 639 code as its key, identifying
      * the language it corresponds to.
      *
      * @return null|string The translation appropriate for the current language, or null if none found. If the
@@ -530,7 +532,7 @@ class Translate
      */
     public static function translateFromArray($context, $translations)
     {
-        if (!is_array($translations) || $translations === null) {
+        if (!is_array($translations)) {
             return null;
         }
 
--- a/lib/SimpleSAML/Metadata/SAMLParser.php
+++ b/lib/SimpleSAML/Metadata/SAMLParser.php
@@ -220,8 +220,9 @@ class SAMLParser
             }
         }
 
-        if ($entityElement->getOrganization() !== null) {
-            $this->processOrganization($entityElement->getOrganization());
+        $organization = $entityElement->getOrganization();
+        if ($organization !== null) {
+            $this->processOrganization($organization);
         }
 
         if ($entityElement->getContactPerson() !== []) {
@@ -331,10 +332,6 @@ class SAMLParser
             $doc = DOMDocumentFactory::fromString($data);
         } catch (\Exception $e) {
             throw new \Exception('Failed to read XML from file: '.$file);
-        }
-
-        if ($doc->documentElement === null) {
-            throw new \Exception('Opened file is not an XML document: '.$file);
         }
 
         return self::parseDescriptorsElement($doc->documentElement);
@@ -1118,13 +1115,12 @@ class SAMLParser
                     $ret['UIInfo']['PrivacyStatementURL'] = $e->getPrivacyStatementURL();
 
                     foreach ($e->getKeywords() as $uiItem) {
-                        if (!($uiItem instanceof \SAML2\XML\mdui\Keywords)
-                            || ($uiItem->getKeywords() === [])
-                            || ($uiItem->getLanguage() === null)
-                        ) {
+                        $keywords = $uiItem->getKeywords();
+                        $language = $uiItem->getLanguage();
+                        if (($keywords === []) || ($language === null)) {
                             continue;
                         }
-                        $ret['UIInfo']['Keywords'][$uiItem->getLanguage()] = $uiItem->getKeywords();
+                        $ret['UIInfo']['Keywords'][$language] = $keywords;
                     }
                     foreach ($e->getLogo() as $uiItem) {
                         if (!($uiItem instanceof Logo)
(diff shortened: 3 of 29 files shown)


composer.json
```diff
@@ -43,7 +43,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "gettext/gettext": "^4.6",
-        "jaimeperez/twig-configurable-i18n": "^2.0",
+        "jaimeperez/twig-configurable-i18n": "^2.1",
```
tvdijen: We have namespaced `jaimeperez/twig-configurable-i18n` to solve some issues regarding inheritance from actual Twig-classes. Psalm doesn't recognize inheritance when mixing namespaced / non-namespaced classes because class-aliases are only created at runtime and Psalm doesn't actually execute code.

lib/SimpleSAML/Locale/Translate.php
```diff
@@ -108,7 +108,7 @@ private function getDictionary($name)
                 $fileName = substr($name, $sepPos + 1);
                 $dictDir = Module::getModuleDir($module).'/dictionaries/';
             } else {
-                $dictDir = $this->configuration->getPathValue('dictionarydir', 'dictionaries/');
+                $dictDir = $this->configuration->getPathValue('dictionarydir', 'dictionaries/') ?: 'dictionaries/';
```
tvdijen: This may look weird, but if the configuration-setting is set to `null`, then Configuration::getPathValue will return `null` rather than the default value...

lib/SimpleSAML/XHTML/Template.php
```diff
@@ -261,17 +264,20 @@ private function setupTwigTemplatepaths()
...
     {
         $auto_reload = $this->configuration->getBoolean('template.auto_reload', true);
         $cache = $this->configuration->getString('template.cache', false);
+
         // set up template paths
         $loader = $this->setupTwigTemplatepaths();
+
         // abort if twig template does not exist
         if (!$loader->exists($this->twig_template)) {
-            return false;
```
tvdijen: This leads to Psalm-issues all over the place because now `$this->twig` can be false and we don't test for it..
It does however not make any sense to continue execution if the template-file does not exist, so raising an exception makes much more sense. This seems like a remnant from the days where we would fall back to php-templates if no Twig-template exists.

lib/SimpleSAML/Metadata/SAMLParser.php
```diff
@@ -333,10 +334,6 @@ public static function parseDescriptorsFile($file)
             throw new \Exception('Failed to read XML from file: '.$file);
         }
 
-        if ($doc->documentElement === null) {
-            throw new \Exception('Opened file is not an XML document: '.$file);
```
jaimeperez: Why is this removed? Can't `documentElement` be `null`?

lib/SimpleSAML/Metadata/SAMLParser.php
```diff
@@ -1118,13 +1115,12 @@ private static function processExtensions($element, $parentExtensions = [])
                     $ret['UIInfo']['PrivacyStatementURL'] = $e->getPrivacyStatementURL();
 
                     foreach ($e->getKeywords() as $uiItem) {
-                        if (!($uiItem instanceof \SAML2\XML\mdui\Keywords)
-                            || ($uiItem->getKeywords() === [])
-                            || ($uiItem->getLanguage() === null)
-                        ) {
+                        $keywords = $uiItem->getKeywords();
```
jaimeperez: Is `$uiItem` guaranteed here to be an instance of `\SAML2\XML\mdui\Keywords`?

lib/SimpleSAML/XHTML/Template.php
jaimeperez: This makes me realize of a possible issue. Is it still possible for modules using twig templates already to work, without `usenewui` being set to `true`?

lib/SimpleSAML/XML/Parser.php
```diff
@@ -41,7 +41,10 @@ public static function fromSimpleXMLElement(\SimpleXMLElement $element)
         /* Create a new parser with the xml document where the namespace definitions
          * are added.
          */
-        $parser = new Parser($element->asXML());
+
+        /** @var string $xml */
+        $xml = $element->asXML();
+        $parser = new Parser($xml);
         return $parser;
```
jaimeperez: At some point, we end up with some very ugly code due to psalm 😞 

This should be `return new Parser($element->asXML());`...

lib/SimpleSAML/XML/Signer.php
```diff
@@ -303,8 +303,10 @@ public function sign($node, $insertInto, $insertBefore = null)
             $options
         );
 
-        /** @var \RobRichards\XMLSecLibs\XMLSecurityKey $this->privateKey */
-        $objXMLSecDSig->sign($this->privateKey);
+        /** @var \RobRichards\XMLSecLibs\XMLSecurityKey $privateKey */
```
jaimeperez: Shouldn't this be fixed in the definition of the property, instead of here?

lib/SimpleSAML/Metadata/SAMLParser.php
tvdijen: Nope, it's always a DOMElement:
https://www.php.net/manual/en/class.domdocument.php#domdocument.props.documentelement

lib/SimpleSAML/Metadata/SAMLParser.php
tvdijen: Yes it is:
https://github.com/simplesamlphp/saml2/blob/release-3.x/src/SAML2/XML/mdui/UIInfo.php#L59
https://github.com/simplesamlphp/saml2/blob/release-3.x/src/SAML2/XML/mdui/UIInfo.php#L104

lib/SimpleSAML/Metadata/SAMLParser.php
jaimeperez: Great! 👍

lib/SimpleSAML/XML/Parser.php
tvdijen: I may be able to fix that using a `/** @psalm-return Parser */` on the method's docblock.. I'll give it a try

lib/SimpleSAML/XML/Signer.php
tvdijen: No, it can be `false`, but in this case we've already tested for that here:
https://github.com/simplesamlphp/simplesamlphp/blob/9f97a5dd7c353c4fde3efbf8786577ba69713382/lib/SimpleSAML/XML/Signer.php#L286

I think I can clean this up.. I'll give it a shot

lib/SimpleSAML/Metadata/SAMLParser.php
jaimeperez: In theory, it's not, as you can set the `Keywords` property there to an array of whatever objects. Look at [this](https://github.com/simplesamlphp/saml2/blob/release-3.x/src/SAML2/XML/mdui/UIInfo.php#L112-L120). However, in this case I think it's safe, since the property would only get populated with `\SAML2\XML\mdui\Keywords` objects via the constructor 👍

lib/SimpleSAML/XML/Signer.php
jaimeperez: Ah, I see! Thanks for clarifying, it's really difficult to review this without looking at the entire code...

lib/SimpleSAML/XML/Parser.php
jaimeperez: Good one! I was completely unaware of that option...

lib/SimpleSAML/XML/Parser.php
```diff
@@ -41,11 +42,7 @@ public static function fromSimpleXMLElement(\SimpleXMLElement $element)
         /* Create a new parser with the xml document where the namespace definitions
          * are added.
          */
-
-        /** @var string $xml */
-        $xml = $element->asXML();
-        $parser = new Parser($xml);
-        return $parser;
+        return new Parser($xml = $element->asXML());
```
jaimeperez: The problem here is not the return value, but the argument passed to `\SimpleSAML\XML\Parser::__construct()`, which cannot be `false`. `asXML()` can actually return false:

https://www.php.net/manual/en/simplexmlelement.asxml.php

jaimeperez: I wonder if we should actually check for error. I'm trying to see under what circumstances `false` can be returned by `asXML()`, but I'm failing to find it...

tvdijen: The return-value is also an issue.. Psalm will say that 'object' does not match the return-type 'Parser'.
I've added a test for the false-case.. it may happen if the output is an empty string, or if PHP runs out of memory, if I read the PHP source code correctly..

lib/SimpleSAML/Metadata/SAMLParser.php
tvdijen: We could add a Webmozart-assertion to the saml2-library to catch that?

lib/SimpleSAML/XHTML/Template.php
tvdijen: I think not.. We completely separated it last year..

lib/SimpleSAML/XML/Parser.php
tvdijen: I've also learned about this a couple of days ago..

tvdijen:
I think I was able to cleanup all of the forced multi-liners by using @psalm-return phpdocs..
Happy?

lib/SimpleSAML/Metadata/SAMLParser.php
jaimeperez: I think it's a good idea!

lib/SimpleSAML/XHTML/Template.php
jaimeperez: I need to check it out, because I know some third-party modules are already implementing twig templates, and it would be unfortunate for those...

lib/SimpleSAML/Locale/Translate.php
jornane: Agreed that it's a bug in `getPathValue`, but I think the solution is not good either. I'd rather throw an exception if `null === $dictDir`. Your configuration should not work when you set things to `null` that cannot be null.

tvdijen: I don't think I agree on a setting that is so fundamental..
`null` can also mean 'unset' in this case and if such a fundamental setting is unset,, a sane default should fill in.. It's just the that the second param of `getPathValue` isn't the safe default in every case..
In any case, this may not be the best fix, but it is what it is until 2.0 at least..
I guess in a future release a bool-param that specifies whether or not a value is mandatory could fix this..

jornane: The `dictionarydir` setting is indeed fundamental, that's why the software should crash if it's overridden to an invalid value. `null` is null, 'unset' is unset and a string is a string. When it's a string, use the string. When it's unset, use the default. When it's `null`, crash with a sensible error message.

PR #1162, opened 2019-07-09
Fix Psalm-issues batch #3

tvdijen:
Fixes all Psalm-issues in all if `SimpleSAML/lib/`.

### DIFF
--- a/lib/SimpleSAML/IdP.php
+++ b/lib/SimpleSAML/IdP.php
@@ -40,9 +40,9 @@ class IdP
      * We use this to support cross-protocol logout until
      * we implement a cross-protocol IdP.
      *
-     * @var string|null
+     * @var string
      */
-    private $associationGroup = null;
+    private $associationGroup;
 
     /**
      * The configuration for this IdP.
@@ -98,7 +98,7 @@ class IdP
                 // probably no SAML 2 IdP configured for this host. Ignore the error
             }
         } else {
-            assert(false);
+            throw new \Exception("Protocol not implemented.");
         }
 
         if ($this->associationGroup === null) {
@@ -435,7 +435,6 @@ class IdP
      * Find the logout handler of this IdP.
      *
      * @return IdP\LogoutHandlerInterface The logout handler class.
-     *
      * @throws Exception If we cannot find a logout handler.
      */
     public function getLogoutHandler()
@@ -453,6 +452,7 @@ class IdP
                 throw new Error\Exception('Unknown logout handler: '.var_export($logouttype, true));
         }
 
+        /** @var IdP\LogoutHandlerInterface */
         return new $handler($this);
     }
 
@@ -505,8 +505,10 @@ class IdP
 
         $this->authSource->logout($returnTo);
 
-        $handler = $this->getLogoutHandler();
-        $handler->startLogout($state, $assocId);
+        if ($assocId !== null) {
+            $handler = $this->getLogoutHandler();
+            $handler->startLogout($state, $assocId);
+        }
         assert(false);
     }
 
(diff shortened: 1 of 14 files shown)


lib/SimpleSAML/IdP.php
```diff
@@ -505,8 +505,10 @@ public function handleLogoutRequest(array &$state, $assocId)
 
         $this->authSource->logout($returnTo);
 
-        $handler = $this->getLogoutHandler();
-        $handler->startLogout($state, $assocId);
+        if ($assocId !== null) {
```
tvdijen: Twelve lines above we also test for `$assocId` being `null`, so I'm guessing this could be a valid situation

lib/SimpleSAML/IdP.php
```diff
@@ -526,8 +528,11 @@ public function handleLogoutResponse($assocId, $relayState, Error\Exception $err
         assert(is_string($assocId));
         assert(is_string($relayState) || $relayState === null);
 
+        $index = strpos($assocId, ':');
```
tvdijen: Perhaps raising an exception if `$assocId` does not meet a certain format would be better.. But what would that format be?

lib/SimpleSAML/Logger.php
```diff
@@ -13,9 +13,14 @@
...
-     * @var \SimpleSAML\Logger\LoggingHandlerInterface|false|null
+     * @var \SimpleSAML\Logger\LoggingHandlerInterface
      */
-    private static $loggingHandler = null;
+    private static $loggingHandler;
+
+    /**
+     * @var bool
+     */
+    private static $initializing = false;
```
tvdijen: By introducing a new property, we can keep the typedecl of `$loggingHandler` as simple as possible..

lib/SimpleSAML/Store.php
```diff
@@ -64,9 +64,11 @@ public static function getInstance()
                         $c
                     );
                 }
+                /** @var \SimpleSAML\Store|false */
                 self::$instance = new $className();
         }
 
+        /** @var \SimpleSAML\Store|false */
```
jaimeperez: Apparently this is not necessary.

lib/SimpleSAML/Logger.php
```diff
@@ -393,6 +398,9 @@ private static function defer($level, $message, $stats)
             register_shutdown_function([self::class, 'shutdown']);
             self::$shutdownRegistered = true;
         }
+
+        // Set initializing to true > all went well!
+        self::$initializing = true;
```
jaimeperez: There is something weird here, and it's difficult to follow the code with the current name of the property and the way it's assigned values. I think `$initializing` should default to `false`, and set to `true` when we call `createLoggingHandler()`, and back to `false` after we're done there.

lib/SimpleSAML/Logger.php
```diff
@@ -453,20 +461,20 @@ private static function createLoggingHandler($handler = null)
      */
     private static function log($level, $string, $statsLog = false)
     {
-        if (self::$loggingHandler === false) {
+        if (self::$initializing === false) {
```
jaimeperez: And then this should just be `if (self::$initializing)`, as that's what this code is aiming to capture, the case where we're trying to log something while initializing the logging handler.

lib/SimpleSAML/Memcache.php
```diff
@@ -2,6 +2,8 @@
 
 namespace SimpleSAML;
 
+use Memcache as _Memcache;
+use Memcached as _Memcached;
```
jaimeperez: These aliases are a bit confusing, and I think we can live without them.

lib/SimpleSAML/Memcache.php
```diff
@@ -304,24 +306,27 @@ private static function addMemcacheServer($memcache, $server)
...
                 'Missing Memcached implementation. You must install either the Memcache or Memcached extension.'
             );
-        } elseif (strtolower($class) === '\memcache') {
-            Logger::warning("The use of PHP-extension memcache is deprecated. Please migrate to the memcached extension.");
         }
-        self::$extension = strtolower($class);
 
-        // create the \Memcache object
-        $memcache = new $class();
+        if (self::$extension === '\memcache') {
```
jaimeperez: Is there a reason for using lowercase here? Can't we just compare with `\Memcache`?

lib/SimpleSAML/SessionHandlerPHP.php
```diff
@@ -364,7 +370,7 @@ public function setCookie($sessionName, $sessionID, array $cookieParams = null)
             );
         }
 
-        session_id($sessionID);
+        session_id(strval($sessionID));
```
jaimeperez: This sounds unnecessary?

Issue #747, opened 2017-12-12
Frontpage_federation does not display expired entities


tvdijen:
This page has logic in place to display and mark expired metadata entities on the frontpage_federation-page, but it appears that this never happens. Actually, expired metadata entities are completely hidden from the view.

A little investigation turned out that MetaDataStorageHandler::getList() filters out any entity that has expired and there really isn't an alternative to get the full unfiltered list of configured entities.

We either need a method to fetch unfiltered entities, or stop filtering them in getList and move that logic to wherever it may be desirable to filter them.

thijskh:
Is a duplicate of #183?

tvdijen:
It seems so yes.. That one doesn't clearly seem to explain the issue though... Feel free to close either one, as long as it's clear that we need to be able to pull metadata even if it's expired, then I'm fine!

tvdijen:
I think I would add a new method with a callback parameter, so that different kinds of filters can be used dynamically (or null for unfiltered).. Then we can at least fix the federation page to show expired entries

tvdijen:
I consider this an issue for the new admin module and therefore tagged it for 1.18

thijskh:
More of a wishlist thing than a requirement to release 1.18.

tvdijen:
Never leave to tomorrow that which you can do today!

PR #1173, opened 2019-07-29
Show expired entities in federation-page

tvdijen:
Closes https://github.com/simplesamlphp/simplesamlphp/issues/747


thijskh:
Great.

Although it would of course be even better if we could use our knowledge of expired entities to show better feedback at login time (not "unknown entity" but "entity expired"). But that's decidedly more involved I think.

PR #1174, opened 2019-07-30
Allow themes to inherit module templates

ghalse:
When developing themes it is common to want to make only minor cosmetic changes to the existing template, such as adding additional CSS. Under the old theming system it was possible to inherit the stock theme and alter it by simply include()ing the PHP.

Similar functionality doesn't exist in Twig, but arguably makes more sense here where we can now override a single block. This functionality is already used by themes to override the default theme, but the model doesn't extend to modules. Thus this commit introduces a special `__parent__` namespace in Twig to allow theme developers to reference the original module templates from within their own theme's template.

E.g if I wanted to inheret the entire discopower module's template save for injecting one additional script into it, rather than copying the whole `disco.twig` template to my theme and editing it, I can merely override the postload block by creating a new `mymodule/themes/mytheme/discopower/disco.twig` like this:
```
{% extends "@__parent__/disco.twig" %}
{% block postload %}
{{ parent() }}{# include the paren't postload block #}
<!-- this is all we're changing -->
<script src="/path/to/extra/javascript.js"></script>
{% endblock %}
```
That's a whole lot less maintenance for me as a theme developer when the module's template changes.

### DIFF
--- a/docs/simplesamlphp-theming.md
+++ b/docs/simplesamlphp-theming.md
@@ -131,6 +131,8 @@ If you need to make more extensive customizations to the base template, you shou
 
 Any references to `$this->data['baseurlpath']` in old-style templates can be replaced with `{{baseurlpath}}` in Twig templates. Likewise, references to `\SimpleSAML\Module::getModuleURL()` can be replaced with `{{baseurlpath}}module.php/mymodule/...`
 
+Within templates each module is defined as a separate namespace matching the module name. This allows one template to reference templates from other modules using Twig's `@namespace_name/template_path` notation. For instance, a template in `mymodule` can include the widget template from the `yourmodule` module using the notation `@yourmodule/widget.twig`.
+
 Even more advanced changes can be made by defining a theme controller in `config.php`:
 
     'theme.controller' => '\SimpleSAML\Module\mymodule\FancyThemeController',
--- a/docs/simplesamlphp-theming.md
+++ b/docs/simplesamlphp-theming.md
@@ -131,7 +131,7 @@ If you need to make more extensive customizations to the base template, you shou
 
 Any references to `$this->data['baseurlpath']` in old-style templates can be replaced with `{{baseurlpath}}` in Twig templates. Likewise, references to `\SimpleSAML\Module::getModuleURL()` can be replaced with `{{baseurlpath}}module.php/mymodule/...`
 
-Within templates each module is defined as a separate namespace matching the module name. This allows one template to reference templates from other modules using Twig's `@namespace_name/template_path` notation. For instance, a template in `mymodule` can include the widget template from the `yourmodule` module using the notation `@yourmodule/widget.twig`.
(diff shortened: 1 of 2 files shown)


tvdijen:
Hey Guy,

How is this different from the already existing functionality?
I think the exact same thing you're proposing is demonstrated here:
https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-1.17/modules/admin/templates/authsource_list.twig

ghalse:
> I think the exact same thing you're proposing is demonstrated here:
> https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-1.17/modules/admin/templates/authsource_list.twig

That's not quite the same thing - it is including content from within the same module and with a different name. That namespace, as you correctly point out, already exists as `@admin` and can be used as such.

To replicate the problem I am trying to solve with that example you would need to theme the admin module such that you had a file:

`modules/mymodule/themes/mytheme/admin/authsource_list.twig`

being a theme that replaces:

`modules/admin/authsource_list.twig`

from your theme. In your theme, you then try and include the original `modules/admin/authsource_list.twig` so that you can extend/alter a small part of it (e.g. adding a postload block) without copying the whole original file. Logically something like this:

```
{# this is modules/mymodule/themes/mytheme/admin/authsource_list.twig #} 
{% extends "@admin/authsource_list.twig" %}{# <-- this wants modules/admin/authsource_list.twig #}
{% block postload %}
{{ parent() }}
<script src="/path/to/extra/javascript.js"></script>
{% endblock %}
```

However, that doesn't work and creates a circular reference back to `modules/mymodule/themes/mytheme/admin/authsource_list.twig`. The reason is that, in this situation, the `@admin/authsource_list.twig` entry searches `modules/mymodule/themes/mytheme/admin` first and finds the `authsource_list.twig` there, and so never tries the original module's template. There is no currently-existant namespace that allows me to access `modules/admin/authsource_list.twig`.

tvdijen:
Make sense now! Thanks for the explanation!

jaimeperez:
Thanks for the further explanation Guy! I was confused as well about this. It's indeed such a useful addition!

jaimeperez:
Hi @ghalse!

We've noticed a bug that looks like it was introduced by this PR: #1312. Since I don't want to break your feature by fixing the bug, would you mind applying this patch and telling me if everything keeps working as expected?

Thanks in advance!

```patch
diff --git a/lib/SimpleSAML/XHTML/Template.php b/lib/SimpleSAML/XHTML/Template.php
index 24986d32f..25371a924 100644
--- a/lib/SimpleSAML/XHTML/Template.php
+++ b/lib/SimpleSAML/XHTML/Template.php
@@ -250,14 +250,15 @@ class Template extends Response
         $loader = new TemplateLoader();
         $templateDirs = $this->findThemeTemplateDirs();
         if ($this->module && $this->module != 'core') {
-            $templateDirs[] = [$this->module => TemplateLoader::getModuleTemplateDir($this->module)];
+            $modDir = TemplateLoader::getModuleTemplateDir($this->module);
+            $templateDirs[] = [$this->module => $modDir];
+            $templateDirs[] = ['__parent__' => $modDir];
         }
         if ($this->theme['module']) {
             try {
                 $templateDirs[] = [
                     $this->theme['module'] => TemplateLoader::getModuleTemplateDir($this->theme['module'])
                 ];
-                $templateDirs[] = ['__parent__' => TemplateLoader::getModuleTemplateDir($this->module)];
             } catch (\InvalidArgumentException $e) {
                 // either the module is not enabled or it has no "templates" directory, ignore
             }
```

PR #1178, opened 2019-08-03
Refactor routing

tvdijen:
Based on https://github.com/sgomez/simplesamlphp#refactor-routing


tvdijen:
I've managed to rebase this old branch against master.. No biggy, actually!
But running tests with with branch I'm running into:

``
Aug  3 18:13:19 webapp-1 simplesamlphp-idp[2175]: 3 [a47efda04c] Error report with id 34d6ebd0 generated.
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 7 [a47efda04c] Localization: load domain 'messages' at '/apps/installation/simplesamlphp/simplesamlphp-routing/locales'
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 7 [a47efda04c] Trying langpath for 'nl' as '/apps/installation/simplesamlphp/simplesamlphp-routing/locales/nl/LC_MESSAGES/'
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 7 [a47efda04c] Translate: Reading dictionary [/apps/installation/simplesamlphp/simplesamlphp-routing/dictionaries/errors]
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] SimpleSAML\Error\Error: UNHANDLEDEXCEPTION
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] Backtrace:
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 2 /apps/installation/simplesamlphp/simplesamlphp-routing/www/_include.php:17 (SimpleSAML_exception_handler)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 1 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/debug/ErrorHandler.php:579 (Symfony\Component\Debug\ErrorHandler::handleException)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 0 [builtin] (N/A)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] Caused by: InvalidArgumentException: Controller "SimpleSAML\Module\core\Controller\LoginController" cannot be fetched from the container because it is private. Did you forget to tag the service with "controller.service_arguments"?
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] Backtrace:
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 9 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/http-kernel/Controller/ContainerControllerResolver.php:71 (Symfony\Component\HttpKernel\Controller\ContainerControllerResolver::throwExceptionIfControllerWasRemoved)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 8 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/http-kernel/Controller/ContainerControllerResolver.php:59 (Symfony\Component\HttpKernel\Controller\ContainerControllerResolver::instantiateController)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 7 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/http-kernel/Controller/ControllerResolver.php:110 (Symfony\Component\HttpKernel\Controller\ControllerResolver::createController)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 6 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/http-kernel/Controller/ContainerControllerResolver.php:42 (Symfony\Component\HttpKernel\Controller\ContainerControllerResolver::createController)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 5 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/http-kernel/Controller/ControllerResolver.php:85 (Symfony\Component\HttpKernel\Controller\ControllerResolver::getController)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 4 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/http-kernel/HttpKernel.php:134 (Symfony\Component\HttpKernel\HttpKernel::handleRaw)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 3 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/http-kernel/HttpKernel.php:68 (Symfony\Component\HttpKernel\HttpKernel::handle)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 2 /apps/installation/simplesamlphp/simplesamlphp-routing/vendor/symfony/http-kernel/Kernel.php:198 (Symfony\Component\HttpKernel\Kernel::handle)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 1 /apps/installation/simplesamlphp/simplesamlphp-routing/lib/SimpleSAML/Module.php:174 (SimpleSAML\Module::process)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] 0 /apps/installation/simplesamlphp/simplesamlphp-routing/www/module.php:9 (N/A)
Aug  3 18:15:18 webapp-1 simplesamlphp-idp[2088]: 3 [a47efda04c] Error report with id d7a59877 generated.
``

lib/SimpleSAML/Kernel.php
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * @param $container
+     * @return void
+     */
+    private function registerModuleControllers($container)
+    {
+        try {
+            $definition = new Definition();
+            $definition->setAutowired(true);
+
```
sgomez: In my branch I was using symfony 3 and I guess you are using symfony 4. In sf4 all services are private by default. But here the message is pretty confusing, because I don't understand why say you forgot to add a tag.

Anyway try to tag the definition:

```
$definition->addTag('controller.service_arguments');
```

or do it public

```
$definition->setPublic(true);
```

tvdijen: I tried both options, but it doesn't work because `registerModuleControllers` is never called..
Could it be that you never finished this?

tvdijen: For some reason, if I set debug=true it starts working... Setting it to false breaks it again..

tvdijen: It's something in the parent-constructor of the Kernel-class that triggers this..  And I can't pinpoint it..

tvdijen: This Symfony-code is pretty hard to follow.. I can pinpoint the crappy flow up until https://github.com/symfony/http-kernel/blob/9979a7022d80fb0c9a6dcf30830e073a1bc101c7/Controller/ControllerResolver.php#L131 but there's something hidden behind that.. I don't see why it would fail, the classname is valid.. The error isn't triggered from this line, so it must somehow be hooked to other code.. Also, if I set debug=true, this particular code is never reached at all, making this very hard to pinpoint

tvdijen: It seems to have been an unfortunate case of coincidence..
Restarting httpd & php-fpm magically solved my issues...

sgomez: Good to know. Sorry, I'm on vacation and I can not follow the issue. I have the problem than I haven't a development environment to test this (I tried but it requires kerberos for some module).

A docker-compose/Dockerfile  with everything ready to test could be useful to help.

tvdijen: Hi @sgomez!  Enjoy your vacation!
If you can't instal the php-krb5 module on your development environment, you can now remove the negotiate-module from the composer.json file..
I will try and run some more tests, but I'm confident that the issue is solved and something was messed up in cache. Unfortunately I'm too unfamiliar with Docker to help you with that.

bendikrb: Just a quick note regarding docker; I'm in the midst of preparing just that. It will be pushed here; https://github.com/simplesamlphp/docker-simplesamlphp hopefully by the end of today.

lib/SimpleSAML/Kernel.php
```diff
@@ -0,0 +1,178 @@
...
+class Kernel extends BaseKernel
+{
+    use MicroKernelTrait;
+
+    const VERSION = '1.18.0-DEV';
+    const VERSION_ID = 11800;
+    const MAJOR_VERSION = 1;
+    const MINOR_VERSION = 18;
+    const RELEASE_VERSION = 00;
+    const EXTRA_VERSION = 'DEV';
```
bendikrb: I don't believe this is good practise? Shouldn't these values (which is saying something about Symfony, rather than SimpleSAML) be left alone?

lib/SimpleSAML/Resources/config/services/web.xml
```diff
@@ -0,0 +1,25 @@
...
+        <service id="argument_resolver" class="Symfony\Component\HttpKernel\Controller\ArgumentResolver">
+            <argument type="service" id="argument_metadata_factory" />
+            <argument />
+        </service>
+
+        <service id="response_listener" class="Symfony\Component\HttpKernel\EventListener\ResponseListener">
+            <argument>UTF-8</argument>
+        </service>
+    </services>
+</container>
```
bendikrb: Any reason for why some of these is defined using XML? I'll change them to yaml in my further w(f)ork.

lib/SimpleSAML/Kernel.php
tvdijen: I think you're right about that, yes

lib/SimpleSAML/Resources/config/services/web.xml
tvdijen: This branch started as a proof-of-concept, so my guess is that both ways were tested..
We should stick with one and I have no preference, so yaml is fine!

tvdijen: On second thought I think I have a slight preference for xml.. I can easily syntax-check that with the native php xml-extension, where yaml would require a PHP-library like symfony/yaml..
I think our [check-syntax.sh](https://github.com/simplesamlphp/simplesamlphp/blob/master/bin/check-syntax.sh)-script should be able to run without composer-dependencies..

jaimeperez: I know I might be shooting myself in the foot, but... I'd stay away from XML. It's harder for people to use, and yaml became quite popular. I think we should favour simplicity here. Regarding how to parse it, if we use yaml, we'll need `symfony/yaml` anyway, right? So then we could use it in the `check-syntax.sh` script...

tvdijen: If we can live with the fact that the bash-scripts are unusable without running `composer install` then I'm fine with both options..
I do however disagree with the 'easier use' of yaml.. Have you ever tried debugging a yaml-file that contains a tab instead of spaces? I wouldn't use it for end-user config files..  I guess for the routing-files it's fine

tvdijen: On third thought, it might make more to check yaml-files as part of a unit test.. I actually [wrote](https://github.com/simplesamlphp/simplesamlphp/commits/master/tests/www/RouterTest.php) one a while ago that I totally forgot about.. I little touch-up could make it more generic

bendikrb: My two cents goes to using yaml (I've been using both in previous projects, as well as annotations). As for unit testing, it doesn't really make any difference, does it? And PhpStorm's Symfony integration (which btw is awesome) understands both patterns.

bendikrb: So arguments _for_ yaml;

- It looks neater, and you don't get bad nightmares from previous tomcat/solr-encounters
- Easier to read, I find
- Easier (and again neater) way to add comments/comment blocks for documentation purposes

tvdijen: Your first argument makes zero sense in relation to this project, second is not objective, third one is a valid one.. That's a real downside of XML..
After going over the options carefully , PHP is in my favour over XML and YAML..

lib/SimpleSAML/Resources/config/routes/routes.yaml
```diff
@@ -1,6 +0,0 @@
-remove_trailing_slash:
-    path: /{url}
-    defaults: { _controller: SimpleSAML\Module:removeTrailingSlash }
-    requirements:
-        url: .*/$
-    methods: [GET]
```
tvdijen: This broke the access to `module.php/admin/` and I haven't discovered an added value yet

tvdijen:
Something that also needs fixing is the fact that `/module.php/admin` doesn't resolve to the admin main-page right now.. It doesn't recognize the `/` route in the admin-module but instead tries to find a `/admin` route in the base routing file..

lib/SimpleSAML/Resources/config/services/web.xml
tvdijen: OK, we've decided to stick with yaml.. @bendikrb have you already converted them by any chance?

tvdijen:
I think we're good, so unless anyone objects I'd like to merge & branch off 1.18

lib/SimpleSAML/Resources/config/services/web.xml
bendikrb: @tvdijen awesome. Great work! I don't see any immediate issues when skimming over it now. I was thinking I'd go through this in more depth on monday though, in a fresh environment.
Btw, as for the yaml arguments, I was only trying to mix some humour in there -- not to be taken literally :)

tvdijen: Thanks, I've been running tests with a checkout without any issues so far..

modules/admin/lib/Controller/ConfigController.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace SimpleSAML\Module\admin;
+namespace SimpleSAML\Module\admin\Controller;
```
jaimeperez: Maybe we can actually rename these controllers as `SimpleSAML\Module\admin\Controller\Config` and so on?

modules/admin/routing/routes/routes.yml
```diff
@@ -0,0 +1,24 @@
+admin-main:
```
jaimeperez: Is it a Symfony convention to have `routing/routes` and a `routing/services` directories? If not, I think it makes it slightly more complicated for users having their own modules. I think I would prefer to have these configuration files directly in the root directory of the module...

modules/core/lib/Controller/RedirectionController.php
```diff
@@ -0,0 +1,109 @@
...
+        if ($postData === null) {
+            // The post data is missing, probably because it timed out
+            throw new Exception('The POST data we should restore was lost.');
+        }
+
+        $session->deleteData('core_postdatalink', $postId);
+
+        assert(is_array($postData));
+        assert(array_key_exists('url', $postData));
+        assert(array_key_exists('post', $postData));
```
jaimeperez: We need to remember migrating to webmozart these ones after we are done merging this PR.

modules/core/lib/Controller/RedirectionController.php
```diff
@@ -0,0 +1,109 @@
...
+
+        if (!Utils\HTTP::isValidURL($postData['url'])) {
+            throw new Error\Exception('Invalid destination URL.');
+        }
+
+        $t = new Template($this->config, 'post.php');
```
jaimeperez: This made me realize... Is this applying routing only to the new user interface, or to the new one as well? If the former, this should read `post.twig`, right? If the latter... well, the latter shouldn't happen 😉

modules/core/templates/login.twig
```diff
@@ -13,7 +13,7 @@
         <ul class="pure-menu-list auth_methods">
             {% for id, config in sources -%}
                 <li class="pure-menu-item">
-                    <a href="/{{ baseurlpath }}module.php/core/login/{{ id|url_encode }}" class="pure-menu-link">
+                    <a href="{{ moduleURL('core/login/' ~ id|url_encode ) }}" class="pure-menu-link">
```
jaimeperez: This is great!! 👏👏👏

routing/services/routing.yml
```diff
@@ -0,0 +1,72 @@
+services:
```
jaimeperez: It's very confusing to have services defined in a file named `routing/services/routing.yml`.

Also, is `.yml` the proper extension or should it be `.yaml`?

routing/services/routing.yml
```diff
@@ -0,0 +1,72 @@
...
+        public: false
+
+    routing.resolver:
+        class: Symfony\Component\Config\Loader\LoaderResolver
+
+    routing.loader.xml:
+        class: Symfony\Component\Routing\Loader\XmlFileLoader
+        arguments:
+            - '@file_locator'
+        tags: ['routing.loader']
```
jaimeperez: I assume this is not necessary if we're only allowing YAML config files, right?

routing/services/services.yml
```diff
@@ -0,0 +1,33 @@
+services:
```
jaimeperez: So... even more confusing? 🤔

routing/services/simplesaml.yml
```diff
@@ -0,0 +1,16 @@
...
+        public: false
+
+    SimpleSAML\Configuration:
+        factory: ['SimpleSAML\Configuration', 'getInstance']
+
+    SimpleSAML\Session:
+        factory: ['SimpleSAML\Session', 'getSessionFromRequest']
+
+    SimpleSAML\Auth\AuthenticationFactory:
+        class: SimpleSAML\Auth\AuthenticationFactory
```
jaimeperez: This seems to be used only in tests. Should we use it in the controllers as well?

lib/SimpleSAML/XHTML/Template.php
```diff
@@ -303,6 +315,8 @@ private function setupTwig()
         $twig = new Twig_Environment($loader, $options);
         $twig->addExtension(new Twig_Extensions_Extension_I18n());
 
+        $twig->addFunction(new TwigFunction('moduleURL', [$this, 'moduleURL']));
```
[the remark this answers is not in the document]
tvdijen: Good point! It was an anonymous function at first, so this was already much cleaner...
I'll get rid of the extra layer of indirection

modules/core/lib/Controller/RedirectionController.php
tvdijen: We can actually do that as part of this PR.. It's new code, it's not gonna affect anyone.
Will take care of it as well!

modules/core/lib/Controller/RedirectionController.php
tvdijen: Good catch! Will fix!

routing/services/routing.yml
tvdijen: It can be both, but all applications I know use .yml... We used both at some point, so I picked one for consistency..
You have a valid point about the confusion, but I would have to figure out what the consequences are if we rename it

routing/services/routing.yml
tvdijen: Why would you want to restrict this? Some module may have a need for flexible PHP config-files..
What we allow is set [here](https://github.com/simplesamlphp/simplesamlphp/pull/1178/files/2d7943ac54b97aad6efa77d182bd738b73d89bf4#diff-35d461a964eb96cfff3e6212ba17ade7R23)

modules/admin/routing/routes/routes.yml
tvdijen: We load them in different ways and I don't think we have a way to distinguish them...
[here](https://github.com/simplesamlphp/simplesamlphp/pull/1178/files/2d7943ac54b97aad6efa77d182bd738b73d89bf4#diff-35d461a964eb96cfff3e6212ba17ade7R112) and [here](https://github.com/simplesamlphp/simplesamlphp/pull/1178/files/2d7943ac54b97aad6efa77d182bd738b73d89bf4#diff-35d461a964eb96cfff3e6212ba17ade7R136)
We could lose the `routing` prefix in the path though.. Most modules will only have routes, not services I think...

modules/admin/lib/Controller/ConfigController.php
tvdijen: Done

routing/services/simplesaml.yml
tvdijen: 🤷‍♂ Your guess is as good as mine

tvdijen:
As for documentation I think we just have to come up with a 2.0 version of `https://github.com/simplesamlphp/simplesamlphp-module-examplecomposer`

routing/services/routing.yml
bendikrb: > I'm also particularly interested in replacing the hooks functionality, and if I recall correctly from previous conversations, this would be possible with services. How could we use that with this setup? Would it be possible to have something analogous to what we have today, where we call a hook, and then all hooks defined by enabled modules are executed? Asking for a friend rofl
> 
> @sgomez maybe you can help us out a bit? @bendikrb do you have any suggestions?

I would recommend to implement EventDispatcher, and have the modules subscribe to the various events (i.e. "hooks").
I'm looking in to how to best do this now - while also keeping it somewhat backwards compatible. This ties together with having a service container at the core too - as far as I can grasp.

modules/admin/routing/routes/routes.yml
bendikrb: Again with the "conventions", in regards to files/structure; I would strongly suggest that we stick to the way symfony-components organizes these resources (especially for new stuff, for which backwards compability isn't an issue);
`Resources/config/services.yml`
`Resources/config/routing.yml`
Modules should also be extending Extension, and use processConfiguration (happens during building of the service container) - but I guess we need to draw the line at some point, in regards to symfonizing the application...

tvdijen:
> I believe the "Controller" suffix is a naming convention in Symfony that we should stick to.
> It can quickly become cluttered too, with such generic names as "Config" - especially if you need to use for instance "Config" the model and "Config" the controller in the same context.

I've had this discussion with Jaime a couple of days ago and it all depends on how you're namespacing things..  If you move your namespaces one level up you get `Model/Config` and `Controller/Config` .. I don't care much for conventions that don't really add anything..

modules/admin/routing/routes/routes.yml
tvdijen: Again, I don't really care for conventions that don't add anything.. We can put those files anywhere we want

bendikrb:
> If you move your namespaces one level up you get `Model/Config` and `Controller/Config` .. I don't care much for conventions that don't really add anything..

Yes, technically its of course a non issue, and separating like this is consistent with the conventions of Symfony aswell. The point I was trying to make is that then we're likely to get something like..
```
use SimpleSAML\Module\user\Controller\User as UserController;
use SimpleSAML\Module\user\Model\User as UserModel;
use SimpleSAML\Module\user\Event\User as UserEvent;
```
.. at which point I _do_ think that this naming convention provides value.
In unit testing we'll get a Test suffix on these aswell - so then it quickly can get untidy, in my opinion.

modules/admin/routing/routes/routes.yml
bendikrb: > Again, I don't really care for conventions that don't add anything.. We can put those files anywhere we want.
I agree, except with the fact that these conventions don't add anything, because they do; consistency. Which I find quite important, especially when having contributors all over the place writing code, either in core or contrib modules.
As to app/config - yes, that's where the application wide configuration is. The modules, or extensions if you may, put them in `Resources/config`.

I think the main reason for why I find reading code of Symfony projects so pleasant, is because there is consistency in where things are, and how they're named.
Not to mention the framework support from IDE's such as JetBrains - which wouldn't work (atleast not out of the box) if things are named/placed arbitrary in the codebase.

Not trying to start a flame war here, I just thought I'd point out a couple of aspects I see about this :)

tvdijen:
In our case we will only have Controllers.. The only Model we have is the Template-class and our views are Twig-templates..
Personally. if I take your example, I would `use SimpleSAML\Module\user\Controller;' and then call `new Controller\Config();'.. That leaves enough significance to distinguish classes with the same name.. Just grep the codebase, you will not find more than 1 or 2 'as ClassAlias' cases if you stick to this convention.. It's just not the Symfony-convention

tvdijen:
Part of this PR has been merged, hopefully making the remainder easier to review..

PR #1179, opened 2019-08-07
Make module.php send Etag

melanger:
for better caching and to prevent showing stale assets

- turn autoLastModified on
- turn autoEtag on
- remove manual Last-modified
- change `Cache-control: public, max-age=86400` to `must-revalidate, max-age=0`
- remove manual Content-length (Symphony handles that implicitly)

Before this, assets were cached for 24 hours regardless of their changes, then they were reloaded by browsers (even if not changed).
With this change, the resources are loaded only if changed but also they are never stale.

### DIFF
--- a/lib/SimpleSAML/Module.php
+++ b/lib/SimpleSAML/Module.php
@@ -264,12 +264,12 @@ class Module
             }
         }
 
-        $response = new BinaryFileResponse($path);
-        $response->setCache(['public' => true, 'max_age' => 86400]);
+        $response = new BinaryFileResponse($path, 200, [], true, null, true, true);
+        $response->setMaxAge(0);
+        $response->headers->addCacheControlDirective('must-revalidate', true);
+        $response->isNotModified($request);
         $response->setExpires(new \DateTime(gmdate('D, j M Y H:i:s \G\M\T', time() + 10 * 60)));
-        $response->setLastModified(new \DateTime(gmdate('D, j M Y H:i:s \G\M\T', filemtime($path))));
         $response->headers->set('Content-Type', $contentType);
-        $response->headers->set('Content-Length', sprintf('%u', filesize($path))); // force file size to an unsigned
         $response->setContentDisposition(ResponseHeaderBag::DISPOSITION_INLINE);
         $response->prepare($request);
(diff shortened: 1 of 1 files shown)


jornane:
By using Etag, the browser must revalidate (hence, `must-revalidate`) the resource every time it uses it. This means a *lot* more hits to the PHP code per page view.

`Cache-control: public, max-age=86400` means:  
The browser can cache the resource for 24 hours, then it must toss it and retrieve again.

`must-revalidate, max-age=0` means:  
The browser can cache the resource, but every time it's to be used, ask the server if the Etag still matches.

Thus, Etag is an effective tool against stale cache, but it increases the amount of communication between browser and server, and it causes a lot more PHP code to be run. I think that the amount of time saved by not re-downloading CSS, JavaScript and images every 24 hours is dwarfed by the amount of time wasted by validating the Etag multiple times on every request.

A more effective way of improving caching of static resources would be to take PHP out of the equation when serving static resources, since most webservers already have sensible defaults and don't have to run a parse and run scripting language to serve static files.

melanger:
@jornane You are right, it means a lot of extra requests (even though small). I managed to get around this for my use case (when updating CSS, I appended something like `?v2` to the URL in HTML). Would it make more sense to open another pull request and make the caching in module.php configurable? Since it uses BinaryFileResponse from Symphony, which has a general [setCache](https://github.com/symfony/symfony/blob/b8661725fff192e0deb22884d89ba62fd4abca1c/src/Symfony/Component/HttpFoundation/Response.php#L923) method which could be passed an array from configuration (with keys `'etag', 'last_modified', 'max_age', 's_maxage', 'private', 'public'`).
Because `max-age=86400` is actually not efficient either, because clients re-download the assets every day, even though they (might) change rarely.

tvdijen:
Don't we already append ?v2 to assets for the newui? That's why i thought this pr makes sense...

melanger:
@tvdijen `module.php` serves assets for all modules and custom themes, which can themselves append `?v2` etc., but cannot affect the `max-age=86400` or any other caching headers for their assets.

melanger:
Please check out #1189 instead

PR #1181, opened 2019-08-10
Add controller for the cron-module

tvdijen:


tvdijen:
I think this is done now, but I'd rather not approve my own PR's, so feedback would be appreciated!

PR #1182, opened 2019-08-10
Add controller for the multiauth-module

tvdijen:
~~This depends on #1178~~


modules/multiauth/lib/Controller/DiscoController.php
```diff
@@ -0,0 +1,137 @@
...
+                    $source = base64_decode($k[1]);
+                }
+            }
+        }
+
+        if ($source !== null) {
+            if ($as !== null) {
+                $as->setPreviousSource($source);
+            }
+            MultiAuth::delegateAuthentication($source, $state);
```
tvdijen: @jaimeperez would the StreamedResponse case apply here as well for the 'delegateAuthentication' call, like we did with the ADFS-module?

jaimeperez: Probably a `RunnableResponse` instead, in the `delegateAuthentication()` method itself. We need to call `authenticate()` on the auth source, and that's out of our control, so `RunnableResponse` would be the only alternative, I think. The entire block should then moved to a different method that we can pass to the response, since we need to consider the case where `authenticate()` actually returns, and call `Auth\Source::completeAuth($state);`.

The alternative would be to return a `RunnableResponse` passing the entire `delegateAuthentication()` method. That would be simpler, but less useful because then we can't really test how that method works.

tvdijen: @jaimeperez I'm not sure I can follow what you're saying here, so I went with the quick-and-dirty solution.. Care to help me out and deploy your even-better solution?

PR #1184, opened 2019-08-10
Replace mock-test with actual test

tvdijen:
Should increase coverage


tvdijen:
0.02%.. I definitely deserve a medal for that 😆

PR #1186, opened 2019-08-12
Remove deprecated AttributeRealm authproc

tvdijen:
It was deprecated since [1.15](https://github.com/simplesamlphp/simplesamlphp/blob/master/docs/simplesamlphp-upgrade-notes-1.15.md) and there is a proper alternative in place, so I'd like to remove this in 1.18


tvdijen:
@jaimeperez you said you wanted to review this.. Could you?

thijskh:
It's a really small thing... is it urgent to drop it or can we do that at 2.0 instead?

tvdijen:
We can have it any way we want, however, my concern is that we're putting off way too much for 2.0..
The migration for our users is going to be quite a thing, with Twig, translations, routing, event dispatching, a possible PHP upgrade if they are on  5.x, and what not more..
Should we perhaps consider an intermediate major with just Twig, translations and PHP version bump and leave the rest for 3.0? I think it would help us progress

thijskh:
The best way to progress in my opinion is to release 1.18 now and move any further work for 1.19 or 2.0

jaimeperez:
I agree with Thijs. I'm even considering having 1.17.6 released with fixes for the few bugs we have identified lately.

The entire point of 2.0 is to make all those changes that are not backwards-compatible there. Removing this authproc in a minor version (1.18) could break existing deployments, and I would therefore prefer to avoid that. Do we have a log message in place when using this authproc saying it has been deprecated and it will be removed? If so, does it point out to the alternative?

Also, the reason why we haven't released 1.18 yet is that I would like to have everything in place that people will need to effectively migrate to 2.0. A painful migration is precisely what I'm trying so hard to avoid here, and that's why we're making our own lives miserable by supporting both new and old functionalities at the same time.

The new routing mechanism is a key part for 2.0, IMO. We need to have that in place, together with a fully working new user interface, in a 1.x release. My goal was that 1.18 would be that release, but if you think it's not realistic and we should have it released already and postpone the routing refactor to 1.19, I'm fine with that too.

docs/simplesamlphp-authproc.md
```diff
@@ -129,7 +129,6 @@ The following filters are included in the SimpleSAMLphp distribution:
 - [`core:AttributeAlter`](./core:authproc_attributealter): Do search-and-replace on attributevalues.
 - [`core:AttributeLimit`](./core:authproc_attributelimit): Limit the attributes in the response.
 - [`core:AttributeMap`](./core:authproc_attributemap): Change the name of the attributes.
-- [`core:AttributeRealm`](./core:authproc_attributerealm): (deprecated) Create an attribute with the realm of the user.
```
jaimeperez: For now, I would only mark the authproc as deprecated in the documentation, pointing to the alternative. Maybe add a log message if we don't have one. But I wouldn't remove it already.

tvdijen: That all in place for years already. I've moved to milestone to 2.0

jaimeperez: Oh, jeez, I need coffee... 🤦‍♂ 

Sorry for the stupid comment Tim!

tvdijen: I'd take some caffeine straight in the veins 🤣  No worries!

PR #1189, opened 2019-08-21
config of module.php cache headers

melanger:
- caching in module.php can be configured, but defaults stay the same (as original behavior)
- new configuration options `module.php.cache`, `module.php.expires`, `module.php.etag` and `module.php.directives`
- makes it possible to configure Etags, `Cache-control: must-revalidate` etc.
- `Last-modified` and `Content-length` are added automatically (by library)


jornane:
@melanger Thanks for your update! I'm still a bit confused by this. In #1179 you stated

> @jornane You are right, it means a lot of extra requests (even though small).

When hitting module.php, most time is lost on round trip time and running the PHP engine. The actual running of the PHP code is fast, and the transfer of data is also fast (since most resources are small). Etag helps when the file is (1) very large or (2) takes a long time to generate.

Using sane defaults is good (no Etag, reasonable max-age, as you have done), but why make Etag configurable at all? Could you comment on reasons one may have for enabling Etag?

I can only see two cases:

- Production: use a long max-age and set cache to public
- Development: use max-age=0 and set cache to must-revalidate; Etag still isn't needed because time used transfering the resources is negligible

tvdijen:
I wonder though if this behaviour should be global instead of specific for module.php...

melanger:
> I think we should call `$response->setAutoLastModified();` to add the `Last-Modified` header.

In the BinaryFileResponse constructor, there is an autoLastModified parameter with `true` as default, so this already happens automatically.

melanger:
> the transfer of data is also fast (since most resources are small)

Tools like PageSpeed complain when caching is not used, because some people are still/sometimes on 3G data, low signal or slow Wi-Fi and every extra request and every 10 kB is noticable (this can be emulated in dev console by introducing simulated throttling). CSS frameworks and pictures easily have hundreds of kB, which is a pain to re-download on slow networks.
I think a reasonable production setup might also be, for example, `public, max-age=86400` with Etag (or Last-modified, but Etag is more accurate). This means that the browser does not re-download the asset every 24 hours if it does not change, it just gets `304 Not modified`.
I understand your concerns and the best caching policy may vary between deployments, that is why I changed the PR to be configurable, with the defaults staying the same.

jornane:
> > I think we should call `$response->setAutoLastModified();` to add the `Last-Modified` header.
>
> In the BinaryFileResponse constructor, there is an autoLastModified parameter with true as default, so this already happens automatically.

Ah I had overlooked that. Then it's all good :)

> CSS frameworks and pictures easily have hundreds of kB, which is a pain to re-download on slow networks.

Sad, but true.

> I think a reasonable production setup might also be, for example, `public, max-age=86400` with Etag (or Last-modified, but Etag is more accurate).

I think Last-Modified would be better so that the server doesn't have to shasum the static files all the time (this is how Symfony calculates the Etag). It can just do a stat on the file. But both solve the problem at hand. As long as it's configurable it should be okay. :)

> I wonder though if this behaviour should be global instead of specific for module.php...

@tvdijen - is there a place in SimpleSAMphp where static resources are served outside module.php?

tvdijen:
`www/assets`

jornane:
> `www/assets`

Where is the PHP code reading from there?

tvdijen:
It's not, sorry for the noise!

melanger:
> I think Last-Modified would be better so that the server doesn't have to shasum the static files all the time (this is how Symfony calculates the Etag). It can just do a stat on the file.

@jornane You're right, but in my setup (multiple instances) the timestamps are not guaranteed to be synchronized, so Etag is a better option for me.

tvdijen:
Are we happy with this now? @jaimeperez how do you feel about this?

config-templates/config.php
```diff
@@ -861,6 +861,37 @@
      */
     'production' => true,
 
+    /*
+     * Set this option to adjust basic caching directives sent by module.php.
+     *
+     * Defaults to ['public' => true, 'max_age' => 86400]
+     */
+    // 'module.php.cache' => ['public' => true, 'max_age' => 86400],
```
jaimeperez: I think the configuration could use some refactor and rewording. Naming is always hard. In this case, I don't think "`module.php.*`" tells anything of what this config options do to someone who's not looking at this PR. Also, our current practice is to try to group together configuration options that are related (e.g. the [`application` option](https://github.com/simplesamlphp/simplesamlphp/blob/master/config-templates/config.php#L36-L52) or the [`debug` option](https://github.com/simplesamlphp/simplesamlphp/blob/master/config-templates/config.php#L220-L256)). This makes it easier to configure the software, in my opinion, and to document options that are related. I would follow the same approach.

As I said, naming is hard, but since this is related to caching of assets, would it make sense to call it just `assets`, then add a `caching` option inside? In my head, this makes sense as we might want to configure other stuff about assets in the future (e.g. the base URL for them, if we want to offload assets to their own server). In this case, and as far as I can see, almost all the available options that can be passed to symfony are set in the `Cache-Control` HTTP header, and there are directives for `ETag` and `Last-Modified`. I think it makes sense to have something like this then:

```php
    /**
     * Explanation of the assets directive
     */
    'assets' => [
        'caching' => [
           /**
            * Explanation of the supported caching-related directives
            */
            'public' => true,
            'max_age' => 86400,
            'etag' => false,
            'last_modified' => ...
        ],
    ],
```

config-templates/config.php
```diff
@@ -861,6 +861,37 @@
...
+
+    /*
+     * Set this option to adjust expiration timestamp sent by module.php.
+     * The value (in seconds) is added to current time.
+     * If set to null, the Expiration header is not sent at all.
+     *
+     * Defaults to 10 * 60 (ten minutes).
+     */
+    // 'module.php.expires' => 10 * 60,
```
jaimeperez: I think this should go as well inside a general `caching` group (name again is orientative).

config-templates/config.php
```diff
@@ -861,6 +861,37 @@
...
+
+    /*
+     * Set this option to make module.php automatically send Etag.
+     *
+     * Defaults to false.
+     */
+    // 'module.php.etag' => false,
```
jaimeperez: This one would be covered already by my previous suggestion.

config-templates/config.php
```diff
@@ -861,6 +861,37 @@
...
+
+    /*
+     * Set this option to adjust more caching directives sent by module.php,
+     * for example must-revalidate. The option takes an array
+     * where keys are the caching directive names and values are the corresponding values.
+     *
+     * Defaults to [] (empty array).
+     */
+    // 'module.php.directives' => [],
```
jaimeperez: This is very confusing. I think we need to either move every directive we want to support to the general `caching` configuration group, or remove it. My main problem with this is that it forces you to know Symfony in order to configure it. Flexibility is nice, but I wouldn't depend on any Symfony-specific code.

lib/SimpleSAML/Module.php
```diff
@@ -265,11 +265,21 @@ public static function process(Request $request = null)
         }
 
         $response = new BinaryFileResponse($path);
-        $response->setCache(['public' => true, 'max_age' => 86400]);
-        $response->setExpires(new \DateTime(gmdate('D, j M Y H:i:s \G\M\T', time() + 10 * 60)));
-        $response->setLastModified(new \DateTime(gmdate('D, j M Y H:i:s \G\M\T', filemtime($path))));
+        $cacheSettings = $config->getArray('module.php.cache', ['public' => true, 'max_age' => 86400]);
```
jaimeperez: Here we could just get the `caching` config group, process it to validate the supported options, and then go ahead acting accordingly for each of them. Those that cannot be passed to `setCache()` could be processed first and removed from the array, and then the remaining could be passed to `setCache()` directly. It might be necessary to have some validation in place, although I believe Symfony would handle that for us.

lib/SimpleSAML/Module.php
```diff
@@ -265,11 +265,21 @@ public static function process(Request $request = null)
...
+        $response->setCache($cacheSettings);
+        $cacheExpiration = $config->getValue('module.php.expires', 10 * 60);
+        $expires = $cacheExpiration === null ? null : new \DateTime('@' . (time() + $cacheExpiration));
+        $response->setExpires($expires);
+        $cacheEtag = $config->getBoolean('module.php.etag', false);
+        if ($cacheEtag) {
+            $response->setAutoEtag();
```
jaimeperez: This is a nice catch. It's definitely much more convenient to leave Symfony generate the `ETag` for us, so I think this is a good idea. In this case, we should probably also process the `etag` config option prior to the call to `setCache()`.

jaimeperez:
> * If we drop making `public` configurable, we can merge `module.php.cache` and `module.php.expires` to one configuration setting, amount of seconds to cache.

That could be a good idea as well, if we don't see a good reason to give so much flexibility. If we do, though, we could still simplify it for most people like this, and still support an _advanced_ configuration format that allows you to set every detail (e.g. supporting both an integer and an array with the set of options that could be configured). This would make processing the configuration a bit more difficult, though.

jornane:
I've implemented the configuration as @jaimeperez suggested, but due to a bug in Configuration.php I had to trick around with default values. I'll make a new PR later to address those.

lib/SimpleSAML/Module.php
jornane: What do you mean "process the `etag` config option prior to the call to `setCache()`?" I think we can move these calls around freely since they are nothing but setters.

jaimeperez:
Merged! Thanks a lot @melanger and @jornane!

PR #1192, opened 2019-08-28
Fix raw template being returned by SingleLogoutResponse.php

### LINKED ISSUE #1191: SingleLogoutResponse.php returns raw twig template
**Describe the bug**
In SimpleSAMLphp version 1.17.5, when not using the new UI and with logout type iframe, when logging out of multiple services, the raw twig template is returned (which results in invalid Javascript and unsuccessful logout from "all" services). This was probably introduced by e0da40397b7a98ddf4b

**To Reproduce**
1. Configuration
```
'usenewui' => false, // or not set

'logouttype' => 'iframe',
```
2. Log in to multiple services
3. Log out of all services
4. An error occurs, in dev console you can see the raw twig template.

**Expected behavior**
The template should be processed.

**Screenshots or logs**
screenshot

**Additional context**
I was able to fix this by setting `usenewui` to `true`, but this forces me to rewrite my theme to use the new UI as well.

tvdijen:
We need a little hack to enforce the use of Twig, even if newui is set to false.
It's either this, or revert to inline HTML in our PHP-files..

Fixes #1191

### DIFF
--- a/lib/SimpleSAML/IdP/IFrameLogoutHandler.php
+++ b/lib/SimpleSAML/IdP/IFrameLogoutHandler.php
@@ -96,8 +96,17 @@ class IFrameLogoutHandler implements LogoutHandlerInterface
     {
         assert(is_string($assocId));
 
-        $config = Configuration::getInstance();
         $this->idp->terminateAssociation($assocId);
+
+        $config = Configuration::getInstance();
+        $usenewui = $config->getBoolean('usenewui', false);
+
+        // Force the use of Twig for this method. Remove if-clause in 2.0
+        if ($usenewui === false) {
+            $config = Configuration::loadFromArray([
+                'usenewui' => true,
+            ]);
+        }
 
         $t = new Template($config, 'IFrameLogoutHandler.twig');
         $t->data['assocId'] = var_export($assocId, true);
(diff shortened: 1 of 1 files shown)


jaimeperez:
I don't think this is the right way to fix this. Forcing the use of a twig template when using the old user interface could bring all kinds of unforeseen consequences. Not saying I don't trust twig templates to work properly (we're using them in production in Feide and it works incredibly well), but the mix of PHP and twig templates and forcing them when configuration says the contrary is what makes me nervous here.

It's probably ok, but I would still prefer to get rid of this hack and solve it with a PHP template.

tvdijen:
We would have to do this in several places then, because the same trick is being used in, for example, the EMail-class... I hope I can find all of them back

jaimeperez:
I can take care of this one, since I'm the one requesting it. If you remember any others, would you mind doing the same?

tvdijen:
Sure thing! Please don't forget to also backport it to the 1.17-branch

tvdijen:
The EMail-class is a bit tricky.. I don't see how I could use a tpl.php file there..
The main issue is that there is no way to retrieve the parsed content, so I already had to use a little trick to fix that for Twig-templates here.. If I would have to do that for tpl.php templates (that just echo content) we would end up buffering output.. Is that really what we want?

PR #1196, opened 2019-09-06
Clean up getConfigItem

jornane:
When using `Configuration::getConfigItem` or `Configuration::getConfigList` (only one user), the function can return a `Configuration` or `$default`. This makes it hard to handle default cases where the configuration settings are not in config.php.

This PR modifies these functions so that `Configuration::getConfigItem` only returns `Configuration` objects or `null`, and `Configuration::getConfigList` only returns arrays. This removes the need for return value checking in calling functions.

This happens at the cost of being able to require that the configuration option is set, but this is a feature; assume that `['assets']['cache']['max-age']` was required:

```php
$assets = $config->getConfigItem('assets'); // returns empty Configuration
$cache = $assets->getConfigItem('cache'); // returns empty Configuration

// Throws exception that says that ['assets']['cache']['max-age'] is not set
$maxAge = $cache->getValue('max-age', Configuration::REQUIRED_OPTION);
```

In this PR, the return checking in the callers is also removed, and the unit tests are updated to reflect the new behaviour.

### DIFF
--- a/lib/SimpleSAML/Auth/Simple.php
+++ b/lib/SimpleSAML/Auth/Simple.php
@@ -23,7 +23,7 @@ class Simple
      */
     protected $authSource;
 
-    /** @var \SimpleSAML\Configuration|null */
+    /** @var \SimpleSAML\Configuration */
     protected $app_config;
 
     /** @var \SimpleSAML\Session */
@@ -45,7 +45,7 @@ class Simple
             $config = Configuration::getInstance();
         }
         $this->authSource = $authSource;
-        $this->app_config = $config->getConfigItem('application', null);
+        $this->app_config = $config->getConfigItem('application');
 
         if ($session === null) {
             $session = Session::getSessionFromRequest();
@@ -387,11 +387,6 @@ class Simple
         $port = !empty($port) ? ':'.$port : '';
         if (($scheme === 'http' && $port === ':80') || ($scheme === 'https' && $port === ':443')) {
             $port = '';
-        }
-
-        if (is_null($this->app_config)) {
-            // nothing more we can do here
-            return $scheme.'://'.$host.$port.$path.($query ? '?'.$query : '').($fragment ? '#'.$fragment : '');
         }
 
         $base = trim($this->app_config->getString(
--- a/lib/SimpleSAML/Configuration.php
+++ b/lib/SimpleSAML/Configuration.php
@@ -1027,6 +1027,8 @@ class Configuration implements Utils\ClearableState
      * @return array The array of \SimpleSAML\Configuration objects.
      *
      * @throws \Exception If the value of this element is not an array.
+     *
+     * @deprecated Very specific function, will be removed in a future release; use getConfigItem or getArray instead
      */
     public function getConfigList($name)
     {
--- a/lib/SimpleSAML/Configuration.php
+++ b/lib/SimpleSAML/Configuration.php
@@ -980,7 +980,7 @@ class Configuration implements Utils\ClearableState
      * is given.
      *
      * @param string $name The name of the option.
-     * @param ?array $default A default value which will be used if the option isn't found. An empty Configuration
+     * @param array|null $default A default value which will be used if the option isn't found. An empty Configuration
      *                        object will be returned if this parameter isn't given and the option doesn't exist.
      *                        This function will only return null if $default is set to null and the option
      *                        doesn't exist.
--- a/lib/SimpleSAML/Configuration.php
+++ b/lib/SimpleSAML/Configuration.php
@@ -980,23 +980,25 @@ class Configuration implements Utils\ClearableState
      * is given.
      *
      * @param string $name The name of the option.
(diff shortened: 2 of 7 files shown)


lib/SimpleSAML/Auth/Simple.php
```diff
@@ -389,11 +389,6 @@ protected function getProcessedURL($url = null)
             $port = '';
         }
 
-        if (is_null($this->app_config)) {
-            // nothing more we can do here
-            return $scheme.'://'.$host.$port.$path.($query ? '?'.$query : '').($fragment ? '#'.$fragment : '');
-        }
-
         $base = trim($this->app_config->getString(
```
jornane: Should this be `rtrim`?

tvdijen: Yes

tvdijen:
This is great, but I'm hesitant to merge this until 2.0..
There are more cases like this, like Configuration::getPathValue that need work similar work..
I ended up doing horrible things like [this](https://github.com/simplesamlphp/simplesamlphp/pull/1157/files#diff-c03bd8403128f19904aff14f16a8cdcc) to make sure I got a string back...

jornane:
> PS: did you check the modules for using these methods?

* https://github.com/search?q=org%3Asimplesamlphp+getConfigItem&type=Code
* https://github.com/search?q=org%3Asimplesamlphp+getConfigList&type=Code

I don't understand much of the statistics modules, there's a lot of reflection there. But for the other modules, the behaviour should remain the same with this change.

tvdijen:
Oh you should check the mess it was around the 1.13/1.14 time 🤣 
I've already done so much work to rationalize that statistics-module! Twigifying it wasn't really a pick-nick either...

> But for the other modules, the behaviour should remain the same with this change
That's great, but it could still mean a BC-break for modules out of our sight, that's why I said I am hesitant..

jornane:
> > But for the other modules, the behaviour should remain the same with this change
>
> That's great, but it could still mean a BC-break for modules out of our sight, that's why I said I am hesitant..

We can make requiring the default until 2.0, and then remove support for it at 2.0. That way, the function will behave exactly like it did before this PR, so no invisible modules breaking. That would be a light version of this PR.

The downside is that we ourselves have to add `[]` as the second argument everywhere, to not trigger an exception if the variable isn't set. We'd have to find those calls when moving to 2.0, but since it's only one function this may be done with a grep I guess?

However, I'm not really sure this would be a problem. Typically, your module will eventually call `getString` or `getInteger` on the config, which will throw an exception if the option is missing, which it is if you earlier got an empty `Configuration` object. The only case where this would fail, is where the final value is read with `getValue`, which is not something a module should do because you don't know what type you'll get back. Modules being hurt by this change are already doing dangerous stuff that may break. Personally, I think that this change is okay for SimpleSAMLphp 1.18, but not for SimpleSAMLphp 1.17.x.

Not doing anything and keeping `getConfigItem` like it is doesn't seem like a good option to me. Every time someone uses `Configuration::getConfigItem`, they're forced to write their own special-case-handling boilerplate around it. Since @jaimeperez wants to move towards a more hierarical config file format, the amount of `Configuration::getConfigItem` calls in the codebase is likely to go up.

tvdijen:
> Not doing anything and keeping `getConfigItem` like it is doesn't seem like a good option to me.

I have many, many, many more of similar cases... I also like to move forward after talking about 2.0 for five years.. We just can't break anything in a minor release, unless Jaime feels otherwise of course..
You could help us getting 2.0 closer though! Are you coming to the NorduNet TI this month?

jornane:
@jaimeperez Some information on what behaviour changed:

## Code that will keep working after this PR, but can be written shorter

```php
// Example 1: required configuration

// BEFORE
$subconfig = $config->getConfigItem('foo');
// exception may be thrown stating that ['foo'] is missing
$val = $config->getString('bar');
// exception may be thrown stating that ['foo', 'bar'] is missing

// AFTER
$subconfig = $config->getConfigItem('foo');
$val = $config->getString('bar');
// exception may be thrown stating that ['foo', 'bar'] is missing
```

```php
// Example 2: optional configuration

// BEFORE
$subconfig = $config->getConfigItem('foo', []);
$val = 'empty';
if ($subconfig instanceof Configuration) {
	$val = $subconfig->getString('bar', 'empty');
}

// AFTER
$subconfig = $config->getConfigItem('foo');
$val = $subconfig->getString('bar', 'empty');
```

## Code that will break

```php
// BEFORE (this is strange, but works)
$subconfig = $config->getConfigItem('foo');
$subconfig = $subconfig->toArray();

// AFTER (this will work both before and after this PR)
$subconfig = $config->getArray('foo');
```

jornane:
@jaimeperez Some information on what behaviour changed:

```php
$subconfig = $config->getConfigItem('foo');
$val = $config->getString('bar');
// Before, the first line would throw an exception, now, the last line throws an exception.
```

```php
$subconfig = $config->getConfigItem('foo'

PR #1199, opened 2019-09-11
Replace trim with ltrim/rtrim in getProcessedURL

jornane:


tvdijen:
That first one is actually obsolete, because there is no way on earth that getServerPort() will return something with whitespace attached...

jornane:
The first one removes the `:` from the result from `Utils\HTTP::getServerPort()`

tvdijen:
D'oh! My bad

PR #1208, opened 2019-09-25
updated catalan translations

ifustero:


tvdijen:
Thanks @ifustero !

jaimeperez:
Thanks @ifustero! Would you mind creating catalan translations for the new user interface as well? Those are in the `locales` directories both at the root of the SSP installation and modules. Dictionaries are deprecated and will no longer work in the near future, and some of the text is actually changing, so we’ll need translators to go through it again.

Issue #549, opened 2017-01-19
Problems with custom session_set_save_handler


drjayvee:
`SessionHandlerPHP` does a neat job of detecting existing sessions, starting its own, and then restoring the previous session when `cleanup` is called (what a sweet feature!).

There is one issue that is overlooked, namely custom session handlers set through [`session_set_save_handler`](php.net/manual/en/function.session-set-save-handler.php). I could find no warning in the documentation, nor did the log contain any useful hints (even in `DEBUG` level).

Here's how I ran into this problem. Our application sets a database session handler and starts a session. When `Session::getSessionFromRequest()` is called, `SessionHandlerPHP::__construct` calls `session_write_close()` and reconfigures `session_id`, `session_name` and `cookie_params`. Then, a new session is started. This session is however stored using the custom session handler (and stored in the database).
When our application redirects the UA to SimpleSAML's own web UI, the browser sends SimpleSAML's cookie, but SimpleSAML cannot find the session (because the web UI uses the default session handler), and therefore starts a new session.
The reverse happens once the UA returns from the IdP and SimpleSAML redirects back to our application. Our application first sets the custom session handler, and SimpleSAML again cannot find the session, and creates a new one.

I'm not sure `SessionHandlerPHP` can or even should try to detect custom session handlers. As long as the SP and SimpleSAML's UI _use the same_ session handler, everything is fine. Unfortunately, there's no PHP function to determine and compare session handlers.

What would have helped though is a message in SimpleSAML's log when `SessionHandlerPHP` could not find the session even though a session cookie is set. The documentation should also contain warnings that any custom session handlers need to be unset (using `session_set_save_handler(new SessionHandler(), true);`) _before_ any calls to SimpleSAML (or more specifically, any calls that try to load a session), _unless_ the web UI uses the same one.

jaimeperez:
Hi @drjayvee!

Thanks for opening this issue! We are aware of this and we've discussed a bit about it some time ago. Unfortunately, as you have noticed, it's not easy to resolve this in a completely transparent way.

We need to take a closer look at it and see if there is a way to avoid this issue. Of course, any help would be much appreciated!

drjayvee:
Hey @jaimeperez, thanks for the quick reply.

I agree that this is going to be tricky to avoid. I think at least the documentation should contain a warning about this. Hopefully that will save other users time.

[...]

PR #1210, opened 2019-09-30
Update simplesamlphp-sp.md

drjayvee:
See #549


jaimeperez:
Thanks a lot @drjayvee!

I wonder about the example, though. It's not clear to me from it how to unset the handler. Do you see a way to improve it?

drjayvee:
I significantly expanded the example. This should be verbose enough, do you agree?

jaimeperez:
I do! Thanks a lot for helping us!

PR #1218, opened 2019-10-11
Fixes in the cs locale

melanger:
For core and consent module.


tvdijen:
Děkuji!

melanger:
Are you able to generate the JSON dictionaries from this? In those, international characters are encoded, so they are difficult to edit manually.

PR #1219, opened 2019-10-18
add store option to SQLPersistentNameID

melanger:
SQLPersistentNameID only allowed the global datastore to be used when it is of type SQL. This filter could not be used when the global datastore is for example memcache.

After this change, SQLPersistentNameID can be used with any database (specified by dsn, username, password). The change is backward compatible; without using the new option, it falls back to the SQL datastore and behaves like it did before.

This option is similar as in the Consent module.

### DIFF
--- a/modules/saml/lib/IdP/SQLNameID.php
+++ b/modules/saml/lib/IdP/SQLNameID.php
@@ -19,6 +19,13 @@ class SQLNameID
     const DEFAULT_TABLE_PREFIX = '';
     const TABLE_SUFFIX = '_saml_PersistentNameID';
 
+
+    /**
+     * @param string $query
+     * @param array $params
+     * @param array $config
+     * @return \PDOStatement
+     */
     private static function query($query, array $params = [], array $config = []) {
         if (!empty($config)) {
             $database = Database::getInstance(Configuration::loadFromArray($config));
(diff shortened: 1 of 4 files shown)


melanger:
@tvdijen Thanks!

tvdijen:
Thank you to for this PR!

modules/saml/lib/IdP/SQLNameID.php
```diff
@@ -13,40 +15,103 @@
...
      */
-    private static function createTable(Store\SQL $store)
+    private static function query($query, array $params = [], array $config = [])
     {
-        if ($store->getTableVersion('saml_PersistentNameID') === 1) {
-            return;
+        if (!empty($config)) {
+            $database = Database::getInstance(Configuration::loadFromArray($config));
+            if (stripos($query, 'SELECT') === 0) {
+                $stmt = $database->read($query, $params);
```
tvdijen: I'm a bit worried about the newly added methods... As you can see, Psalm-tests fail, because `Database::read` will not return a `PDOStatement` as you are assuming here and everywhere else in this file..

melanger: You're right, `Database::write` returns `int|bool`, but here on write operations the results are being ignored anyway, so a return type of `PDOStatement|int|bool` would fix this; anyway, I've split the `query` method into `read` and `write`, basically mimicking `Database` using `Storage`, and got rid of the `if (stripos($query, 'SELECT') === 0)`.

tvdijen:
I'll just suppress the issue in Scrutinizer..

PR #1221, opened 2019-10-28
Fix issue with multi-byte utf-8 characters in MD

ZarToK:


tvdijen:
Thanks! I think we should probably fix this in the saml2-library instead of on the SimpleSAMLphp-end..
Possibly we can set it generically in https://github.com/simplesamlphp/saml2/blob/fc7375ab08706db31132b6ce3505555b8697ad01/src/SAML2/DOMDocumentFactory.php#L106

jaimeperez:
Hi @ZarToK!

Can you provide any more information on this? Why is this necessary? What's the issue you are having, and how to reproduce?

I'm asking because UTF-8 is the default, as far as I see, and after a very quick test, it looks like it works as expected with multibyte characters. So I'm wondering if this could be an issue with your particular deployment...

tvdijen:
ping @ZarToK

tvdijen:
We can reopen this at a later time if necessary

tvdijen:
Not a problem at all! Let me know and we'll reopen the issue if necessary..  I've seen the strangest things happen in docker-containers so I'm really not surprised tbh..

tvdijen:
Of course, this has always been te case

malavolti:
On v2.0.0 I see that the metadata generate hasn't the UTF-8 encoding. 

Can I expect that on the v2.0.1 will be there?

Thank you @tvdijen

malavolti:
Without the patch proposed by @ZarToK:

ssp-without-utf8

With the patch proposed by @ZarToK:

ssp-with-utf8

I see an improvement with the patch proposed.

Is it clear now?

Thanks

thijskh:
I've checked and can reproduce indeed. The fix by @ZarToK seems to fix it just fine.

Removal of the first line seems to happen by
https://github.com/simplesamlphp/simplesamlphp/blob/09e4968963ad61c33be413d9a7208d0fcda90417/modules/saml/src/Controller/Metadata.php#L107-L113
Not sure why that was added @tvdijen? It seems newly added in f461c7b and was not present in the www-script version.

tvdijen:
So, maybe it wasn't a good idea to drop DOMDocumentFactory::create @thijskh ? Do we want to default to UTF-8 in this case?

malavolti:
Dear @tvdijen and @thijskh ,
can you tell me if this patch will be included into the next version (v2.0.1) of SimpleSAMLphp?

Thank you so much!

tvdijen:
Can't tell you that.. Need to investigate this a bit more, and there's only 48 hrs in a weekend

malavolti:
Forgive me @tvdijen, I didn't want to rush. I can wait without problems.

tvdijen:
> Forgive me @tvdijen, I didn't want to rush. I can wait without problems.

No offense taken ;)

tvdijen:
> I've checked and can reproduce indeed. The fix by @ZarToK seems to fix it just fine.
> 
> Removal of the first line seems to happen by
> 
> https://github.com/simplesamlphp/simplesamlphp/blob/09e4968963ad61c33be413d9a7208d0fcda90417/modules/saml/src/Controller/Metadata.php#L107-L113
> 
> Not sure why that was added @tvdijen? It seems newly added in [f461c7b](https://github.com/simplesamlphp/simplesamlphp/commit/f461c7bf16d223afff32cb2128cc39a13a614409) and was not present in the www-script version.

I tried to trace back my steps, but I cannot recall why this was done.. I'm fine with removing it

thijskh:
Merged in 8d847e7331f4a9b15493e3b296762763a67c3003 3c8c8d9d6461eeb4fc2f4293084ea70fe96ca0a0, thanks!

Issue #1223, opened 2019-11-04
Allow configuration of database number for Redis store


garylocke:
**Describe the solution you'd like**
I would like a way to configure a database number for my Redis store by editing `config/config.php`.

I believe this is a very straightforward change, but let me know if I've missed anything or if this is already possible through other methods.

I will submit a PR shortly with a change to include this option.

garylocke:
See PR #1224

tvdijen:
This was added only days ago in #1222?

garylocke:
Excellent. I will close this request.

PR #1224, opened 2019-11-04
Allow configuration of database number for Redis store

garylocke:
This change adds a configuration option for Redis database number in `config/config.php`. See issue #1223


garylocke:
Cancelled. This support was added by #1222

PR #1226, opened 2019-11-06
Create all folders to enable copy

RosemaryOrchard:


tvdijen:
Good catch! Thanks @RosemaryOrchard

PR #1237, opened 2019-11-14
errorreport with unfilled fields won't send an email

BaranekD:
This PR does not allow to send empty errorreports. It can help before spamming unwanted reports.

### DIFF
--- a/www/errorreport.php
+++ b/www/errorreport.php
@@ -45,12 +45,20 @@ $data['hostname'] = php_uname('n');
 $data['directory'] = dirname(dirname(__FILE__));
 
 if ($config->getBoolean('errorreporting', true)) {
-    $mail = new SimpleSAML\Utils\EMail('SimpleSAMLphp error report from ' . $email);
-    $mail->setData($data);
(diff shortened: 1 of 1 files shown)


jaimeperez:
Thanks a lot for your contribution, @BaranekD!

This one is a bit hard to decide on, because we definitely see legitimate error reports with the exception and URL not set. In the end, those fields being not set means something went really wrong, but not necessarily that the message is spam, since the data may still have been successfully recovered. Of course, lacking a backtrace means it's almost impossible for us to figure out where did that error originate, and fix it.

I'd say the most clear case here is when loading the exception data from the session fails, as that miight indeed indicate that this could be automated in any way, for whatever purpose. But if loading the data succeeds, I think then we should send the report, even if parts of it are missing.

Also, bear in mind that if someone really wants to use this to spam, it would be simple to circumvent your check by manually inducing an error (e.g. page not found).

thijskh:
I understand the issue but I agree that this might not be the best solution. What kind of reports are you getting that you want to prevent?

PR #1238, opened 2019-11-18
Add note about upgrading third party modules

ghalse:
The current upgrade instructions don't mention that people need to preserve themes or third-party modules, and this has confused some people I've tried to help. Thus I'm proposing adding a note about that.

The second patch is more controversial and had been done separately do it can be discarded if necessary. At the moment the theme on https://simplesamlphp.org/ is not rendering markdown preformatted blocks properly, and the shell script entries are being munged into one line. That means that a copy-and-paste doesn't work. We can fix this by adding semicolons, which usefully doesn't break things later if the theme works again.  (Alternatively someone could fix the theme :-))

### DIFF
--- a/docs/simplesamlphp-install.md
+++ b/docs/simplesamlphp-install.md
@@ -77,6 +77,9 @@ keep reading for other alternatives):
     cp -rv ../simplesamlphp/metadata metadata
 ```
 
+If you have installed any [third-party modules](https://simplesamlphp.org/modules) or [customised the theme](simplesamlphp-theming.md), 
+you should check whether your third-party modules need upgrading and then copy or replace those directories too.
+
 Replace the old version with the new version:
 
 ```


tvdijen:
People shouldn't be copy-pasting modules around at all... I don't think this is really helping them

jaimeperez:
The installation instructions will need some work anyway, as we need to explain more in detail how to install with composer. In the end, with the new user interface, the user will need to run `npm` to build the assets, so we need to explain that (I've been doing tests to try to automate this when installing or updating with composer, but I feel a bit nervous about running code like that during composer installation).

Regarding the semicolons, I think the proper change there would be to fix the website.

jaimeperez:
So, [the website is fixed now](https://simplesamlphp.org/docs/stable/simplesamlphp-install), meaning the documentation should display correctly again for all versions. Is it ok if I revert the change of the semicolons and merge the other one?

ghalse:
Is okay  by me - it is why I put them in separate commits in the first place :)

PR #1244, opened 2019-11-25
fix preferred translation

m0ark:
Preferred language selection is broken in discovery service. Currently it uses the last valid translation instead of the first valid.


tvdijen:
I think I've made a good start here, but I'm not really good with unit tests, so it would be great if someone can pick up from here

lib/SimpleSAML/XHTML/IdPDisco.php
```diff
@@ -600,13 +600,30 @@ public function handleRequest()
...
+        $t->data['rememberenabled'] = $this->config->getBoolean('idpdisco.enableremember', false);
+        $t->show();
+    }
+
+
+    /**
+     * @param array $idpList
+     * @return array
+     */
+    private function getPreferredTranslations(array $idpList, array $tryLanguages): array
```
jaimeperez: I think this is quite misleading. What this method is doing is returning a list of IdPs, sorted by name, where their names are translated based on the preferred translation for the given request. I don't think _getPreferrredTranslations()_ correctly describes that, but instead it sounds like it's just returning the preferred translations for some given IdPs.

I would actually suggest refactoring this code. There's a lot of repetition in there, and testing is quite hard, so we could benefit from routing to feed requests and get responses out of it.

I'm going to merge @m0ark's commit into 1.18, and we can keep working on this meanwhile.

jaimeperez:
I agree with Tim here. We should do this properly and add unit tests. However, it's quite difficult to get some tests working, so I've prioritized getting the fix into a stable release (1.18.1, out just now), and then I'm leaving this PR open as a reminder that we need to get back to this.

thijskh:
Fixed in #1512.

PR #1246, opened 2019-11-28
More support Japanese translation

kenchan0130:
Hi, I am Japanese.

I fixed it because some Japanese were strange.
Of course, there are still some Japanese expressions that are suspicious, but I tried to correct some of them.


tvdijen:
Thanks very much @kenchan0130 !

PR #1247, opened 2019-11-28
core/show_metadata: changed => to =

doedje:
The code generated: `$metadata['idp'] => array();` causing an error when I used this to copy the metadata and pasted it in the `metadata/saml20-idp-remote.php` file of another server.


tvdijen:
Wow, I can hardly believe this was undetected for 5+ years...  Good catch @doedje !

PR #1251, opened 2019-11-29
core/show-metadata: added a semicolon at the end of the metadata

doedje:
Terribly sorry I did not include this one in my pull request from yesterday. As it turned out I added the missing semicolon automatically, but to be correct and allow brainless copy-pasting of this metadata, you'll need the semicolon on the end of the metadata.

PS: Just some background information: I ran into this issue when I tried to copy-paste the metadata for my IdP from another SP that was already using it, to a new SP I was configuring. When you go straight to the IdP and ask the metadata there you use `/www/saml2/idp/metadata.php` and that one is working fine, so that might explain why it took so long to be discovered.


doedje:
This one should have been in #1247 but since it was already merged in, I had to do a new PR...

tvdijen:
Thanks again @doedje !!

PR #1260, opened 2019-12-13
Add Sesotho translation

ghalse:
This pull request adds a partial Sesotho translation.

Sesotho is one of the eleven South African official languages, covering the second of the two dominant Southern Bantu language grouping in Southern African (Tswana-Sotho).  It is spoken by about 4million South Africans. It has a very high level of mutual intelligibility with other Tswana-Sotho group languages, including Tswana and Northern Sotho (Sepedi). These are also official languages of South Africa and share a partially standardised orthography with Sesotho.

The translation was commissioned by TENET/SAFIRE, and was done by a professional translation company. It is incomplete because the intent was to cover the parts of the user interface (rather than the admin interface) that we're currently presenting to end-users, on the basis that most of the administrators of these systems will understand English. We're also aware that some of this will fall away with the advent of the new UI, and have made provision to update the four translations we've contributed {zu, xh, af, st} when this has stabilised a bit.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -737,6 +737,8 @@ $config = [
             'se' => ['nb', 'no', 'nn', 'en'],
             'nr' => ['zu', 'en'],
             'nd' => ['zu', 'en'],
+            'tw' => ['st', 'en'],
+            'nso' => ['st', 'en'],
         ],
     ],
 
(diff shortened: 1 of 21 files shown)


tvdijen:
Thanks Guy!

PR #1263, opened 2019-12-19
allow session.check_function to be any Callable

melanger:
which may be a `string`, an `array` of `[class|object, method]` or a closure

https://www.php.net/manual/en/language.types.callable.php


tvdijen:
Děkuji @melanger !

PR #1264, opened 2019-12-24
Twig Strict vars

tvdijen:
Enable twig strict vars and fix broken templates


tvdijen:
More templates may be affected..

PR #1266, opened 2020-01-03
Mention expectException instead of phpdoc annotations in TESTING.md

Krisell:
[#3d121](https://github.com/simplesamlphp/simplesamlphp/commit/3d121046c4839a99833fffd2ea78b45258b1bbae#diff-84889d0af78b35050da8d12673769eba) switched from using phpunit annotations to `expectException()` for asserting expected exceptions, but the documentation still mentions the annotation.


tvdijen:
Good catch @Krisell ! Thanks

PR #1276, opened 2020-02-11
Samesite patch

jeremyfrench:
I'm not sure if there is any ongoing work on the 1.16 branch, but I needed to get same site cookies on it and can't upgrade just yet.

I thought I'd raise a PR, so even if closed anyone else who needed this can take the patch.


jaimeperez:
Thanks @jeremyfrench!

No, there's no ongoing work on 1.16. In general, once we release a new branch, we end support for previous branches. Sometimes we might make exceptions, especially wrt security fixes, but this is not the case. Note also that there are two minor versions newer than the one you are running, and that critical security issues have been fixed since, so you should really update now.

PR #1287, opened 2020-02-25
use Symfony VarExporter in metadata converter

melanger:
Instead of `var_export`, this provides nice, PSR-12 compliant output without explicit numeric array indices.


tvdijen:
Unfortunately SSP 1.19 is still at PHP7.0 and no version of `symfony/var-exporter` supports this.
Do you mind if I change the target branch to `master` and have this end up in SSP 2.0?

tvdijen:
I had to rebase it, because your fork of SSP is pretty old

PR #1292, opened 2020-02-26
Add Metadata Directory of FlatFiles support

pkolmann:
I like to split up long config files to split config files.

The MetaData Files usually are quite long and if you need to change
something it's quite bulky to handle them. Diffing between different SPs
is also not too easy.

Thats why I added the support for Metadata config directories. Every SP
goes into it's own flatfile.

### DIFF
--- a/docs/simplesamlphp-maintenance.md
+++ b/docs/simplesamlphp-maintenance.md
@@ -165,13 +165,14 @@ By default SimpleSAMLphp will attempt to connect to Redis on the `localhost` at 
 
 ## Metadata storage
 
-Several metadata storage backends are available by default, including `flatfile`, `serialize`, `mdq` and
+Several metadata storage backends are available by default, including `flatfile`, `directory`, `serialize`, `mdq` and
 [`pdo`](https://simplesamlphp.org/docs/stable/simplesamlphp-metadata-pdostoragehandler). Here you have an
 example configuration of different metadata sources in use at the same time:
 
 ```
 'metadata.sources' => [
     ['type' => 'flatfile'],
+    ['type' => 'directory'],
     ['type' => 'flatfile', 'directory' => 'metadata/metarefresh-kalmar'],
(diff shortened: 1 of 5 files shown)


pkolmann:
@tvdijen can you please advise if this change is worth accepting?
Thanks
Philipp

tvdijen:
Thanks @pkolmann!! I am
In transit from Madrid to The Hague now.. hopefully I can get back to you tomorrow.. I like the general idea of the pr though!

pkolmann:
@tvdijen could you take another look at this please.

ebeuerle:
@tvdijen Any luck with merging this PR? This would very handy to have for larger installations...

monkeyiq:
There is also an issue submitted wanting a similar functionality for the XML files for SPs. A comment over on that issue is calling for a more generic solution to allow both flatfile and xml to be loaded on demand: https://github.com/simplesamlphp/simplesamlphp/issues/1368#issuecomment-678315386

I mention this because the comment may not have been seen on this thread. I've been looking at this PR and the code a little bit in recent days.

monkeyiq:
I had a bit of a play around with this and ended up with very similar code to what is here though trying to handle xml files as well if they are found in the metadata/subdirectory.d directory. I am not sure what I have in the below branch makes complete sense as I am very new to the SSP code base. 

https://github.com/monkeyiq/simplesamlphp/blob/c23a35a2ea33304de92331bba3962394f02feb66/src/SimpleSAML/Metadata/MetaDataStorageHandlerDirectory.php#L120

It does seem that this PR could be extended to check for php files and do something on them and something else if an xml file was found, otherwise maybe just report an ignored file?

monkeyiq:
Sorry about the wait. I have merged a PR with this functionality into the simplesamlphp-2.3 branch and master.

https://github.com/simplesamlphp/simplesamlphp/pull/2018

The patch should also back port to a 2.2 release reasonably well if you would like to take it for a test run.

Issue #1286, opened 2020-02-24
Change the module structure to a more Symfony based system


RosemaryOrchard:
During the workshop in Madrid, @sgomez @jaimeperez, @tvdijen and I discussed switching the module system to a Symfony based system, for release with the coming 2.0. This would be a breaking change (modules would have to be migrated in order to work again).

Current notes/to dos:
- Each module would be a [Symfony Bundle](https://symfony.com/doc/current/bundles.html), and follow that directory structure.
- Kernel.php needs to be updated, this will at minimum involve changes to the constructor and `registerBundles`.
- [Add scripts to composer.json](https://github.com/symfony/skeleton/blob/master/composer.json#L47)
- Base class for modules

Current proposed approach: start by restructuring a small plugin (possibly [Expiry Check](https://github.com/simplesamlphp/simplesamlphp-module-expirycheck)), and then modify simpleSAMLphp to make this work.

sgomez:
In Symfony, when a bundle has routes, a file should be added to `config/routes` directory importing this routes. So, when we create a module with routes we have two ways to import routes

* A [custom router loader](https://symfony.com/doc/current/routing/custom_route_loader.html) (need to be implemented)
* Add the routes in a `config` directory manually. This will do the installation more complicated.

First option seems the best.

sgomez:
Finally, I can load routes from Kernel and skip the creation of a custom router. Anyway there a lot issues to resolve:

* There a lot of classes that depends of `Modules` static methods. And this search the modules in 'module.enable` configuration (to check if it is enabled) and files inside `modules` directory. So, change this will require a lot of time.
* Another issue is Dependency Injection is not working (autowiring at least). I can not trace the error. Probably something in the configuration is missing.

Right now, I can have controllers but `Template` class can not found twig templates. Without DI I cannot create a template service either.

Resolve this from a small modules  instead from main modules is hard, because we need admin and other modules working and this will require a lot of extra logic in Module class. 

I will try to use a `new.php` access point to load new modules in a different way that `module.php` does.

PR #1294, opened 2020-02-26
PoC new Module system to SSP 2.0

sgomez:
This PR is related to #1286 issue.

Changes:

* `www/new.php` is created to load new kernel. Also `console` has been changed to load the new kernel.
* Old kernel renamed to `ModuleKernel.php` to indicate than only works with one module.
* New kernel to load all modules and core application.

Instructions:

1. `APP_ENV=dev` environment variable must be configured
1. Run `composer install`. This will install [a demo module](https://packagist.org/packages/sgomez/simplesamlphp-module-expirycheck.
1. Copy `modules.php` inside `config-templates` to `config` directory.

I have this authsource to check the module:

```
    'example-static' => [
        'exampleauth:StaticSource',
        'uid' => ['testuser'],
        'eduPersonAffiliation' => ['member', 'employee'],
        'cn' => ['Test User'],
        'schacExpiryDate' => '20051231235959Z'
    ],
```

And this filter:

```
        10 => array(
            'class' => SimpleSAML\Modules\ExpiryCheckModule\Auth\Process\ExpiryDate::class,
            'netid_attr' => 'eduPersonPrincipalName',
            'expirydate_attr' => 'schacExpiryDate',
            'warndaysbefore' => '60',
            'date_format' => 'd.m.Y', # php date syntax
        ),
```

The idea is to run the ExpiredController and load templates and translation from vendor without install the modules inside simplesamlphp directory.

More explanations in the code.


config-templates/modules.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+return [
```
sgomez: This file returns all modules/bundles to be loaded. The kernel supports environments, so we can indicate if they should be loaded in all or just in a few of them.

config-templates/modules.php
```diff
@@ -0,0 +1,11 @@
...
+    // Don't remove this bundles
+    Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => true],
```
sgomez: Probably FrameworkBundle is too huge to be used and we can create a custom one.

lib/SimpleSAML/Kernel.php
```diff
@@ -1,170 +1,69 @@
...
+    private const CONFIG_EXTS = '.{php,xml,yaml,yml}';
 
-    /** @var string */
-    private $module;
-
-
-    /**
-     * @param string $module
-     */
-    public function __construct(string $module)
```
sgomez: New Kernel is not just for one module.

lib/SimpleSAML/Kernel.php
```diff
@@ -1,170 +1,69 @@
...
-    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
+    protected function configureRoutes(RouteCollectionBuilder $routes)
     {
-        $configuration = Configuration::getInstance();
-        $baseDir = $configuration->getBaseDir();
-        $loader->load($baseDir . '/routing/services/*' . self::CONFIG_EXTS, 'glob');
-        $confDir = Module::getModuleDir($this->module) . '/routing/services';
-        if (is_dir($confDir)) {
-            $loader->load($confDir . '/**/*' . self::CONFIG_EXTS, 'glob');
+        foreach ($this->bundles as $bundle) {
```
sgomez: This iterator search in all bundles the ones that are SSP Modules and checks if they have a routes directory (and load the route files). The module name is used a prefix to the url.

lib/SimpleSAML/Kernel.php
```diff
@@ -1,170 +1,69 @@
...
 
-        $container->loadFromExtension('framework', [
-            'secret' => Configuration::getInstance()->getString('secretsalt'),
-        ]);
-
-        $this->registerModuleControllers($container);
-    }
+        $confDir = __DIR__ . '/Resources/config';
 
+        $routes->import($confDir.'/{routes}/'.$this->environment.'/**/*'.self::CONFIG_EXTS, '/', 'glob');
```
sgomez: Load routes required by simplesaml to work. For example the route required to have the debug toolbar.

lib/SimpleSAML/Kernel.php
```diff
@@ -1,170 +1,69 @@
...
-        }
     }
 
-
-    /**
-     * @param ContainerBuilder $container
-     * @return void
-     */
-    private function registerModuleControllers(ContainerBuilder $container): void
+    protected function configureContainer(ContainerBuilder $container, LoaderInterface $loader)
```
sgomez: Load services required by the core of simplesaml (modules services are loaded inside the Module loader class).

lib/SimpleSAML/Modules/Module.php
```diff
@@ -0,0 +1,54 @@
...
+
+
+use Symfony\Component\Config\Exception\FileLocatorFileNotFoundException;
+use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Loader\DirectoryLoader;
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+abstract class Module extends Bundle
```
sgomez: This abstract class extends from Symfony Bundle abstract class. So a module is a bundle. Probably we can create a thin abstract class because we don't require all the logic inside the base class.

lib/SimpleSAML/Modules/Module.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    abstract public function getShortName(): string;
+
+    public function build(ContainerBuilder $container)
```
sgomez: This method is called when the cache is warmed. Here we can define the services this module will use.

lib/SimpleSAML/Modules/Module.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $this->registerModuleControllers($container);
+    }
+
+    /**
+     * @param ContainerBuilder $container
+     * @return void
+     */
+    private function registerModuleControllers(ContainerBuilder $container): void
```
sgomez: We register controller services. The idea is to create similar methods for each service we want to define automatically (hooks/events, admin menus, etc.). The possibility to use a services.yml to load custom services will be available. But this methods exists to do the module development easiest.

lib/SimpleSAML/XHTML/Template.php
```diff
@@ -162,12 +162,12 @@ public function asset(string $asset, string $module = null): string
     {
         $baseDir = $this->configuration->getBaseDir();
         if (is_null($module)) {
-            $file = $baseDir . 'www/assets/' . $asset;
+            $file = $baseDir . 'www/' . $asset;
```
sgomez: I use `symfony/asset` module that implements `asset` function in twig. This modification is to have the same behaviour in the old a new way to read templates.

sgomez:
Translations in symfony supports a lot of loaders (po, yaml, xliff, csv, ...). Gettext loader (po files) does not require gettext php module (apparently, I will investigate more about this).

There are a difference. The structure of translations files change from `locales/LANGCODE/LC_MESSAGES/module.po` to `translations/module.LANGCODE.po`. With a script is easy to move files. But `trans` function in yaml files requires that we indicate the module if it is differente from default (`messages`).

tvdijen:
Oops, I did not intend to do this, but @sgomez should this be merged at some point? Let's not let this go to waste..

sgomez:
Hi @tvdijen 

This is just a PoC. This requires a lot of discussion because is a huge change in the way modules are loaded.

tvdijen (changes requested):
The RouteCollectionBuilder was deprecated by Symfony in favour of the RoutingConfigurator.
Unfortunately I can't get it to work with the new class and I don't know what to do with the `$prefix` variable.

PR #1295, opened 2020-03-02
Allows custom logging handler on the console

dlundgren:
This is useful for reducing noise during tests.

This also adds several convenience functions to the Logger class that aids in testing.


tests/Utils/ArrayLogger.php
```diff
@@ -0,0 +1,32 @@
...
+use SimpleSAML\Configuration;
+use SimpleSAML\Logger\LoggingHandlerInterface;
+
+class ArrayLogger
+    implements LoggingHandlerInterface
+{
+    /**
+     * @var array List of log entries by level
+     */
+    public $logs = [];
```
thijskh, line 16: Since master has PHP 7.4 as minimum, use typed propery instead of phpdoc annotation?

lib/SimpleSAML/Logger.php
```diff
@@ -288,6 +288,15 @@ public static function getCapturedLog(): array
     }
 
 
+    /**
+     * Clears the captured log.
+     */
+    public static function clearCapturedLog()
+    {
+        return self::$capturedLog = [];
```
thijskh, line 296: Unused return value, should probably not 'return' the empty array but just void, and also specify that as the return type

thijskh (approved):
Just some nits...

PR #1297, opened 2020-03-04
convert translation to gettext style

restena-sw:
script which shuffles .po file content from SSP-style codes to normal gettext() style msgid = EN, msgstr = XX language files, and back again

### DIFF
--- a/bin/convertTranslations.php
+++ b/bin/convertTranslations.php
@@ -1,0 +1,129 @@
+<?php
+
+function mergeWithSource($sourcePairs, $destPairs)
+{
+    $mergedPairs = [];
+
+    foreach ($sourcePairs as $msgId => $msgStr) {
+        if (array_key_exists($msgId, $destPairs)) {
+            // the index is becoming the msgId but comes as the string "msgstr" from the file, so convert it
+            $mergedPairs[preg_replace("/^msgstr/", "msgid", $msgStr)] = $destPairs[$msgId];
+        } else {
+            $mergedPairs[preg_replace("/^msgstr/", "msgid", $msgStr)] = "msgstr \"\"\n";
+        }
+    }
+    return $mergedPairs;
+}
+
+function codifyWithSource($sourcePairs, $destPairs)
+{
+    $mergedPairs = [];
+
+    foreach ($sourcePairs as $msgId => $msgStr) {
+        $modifiedKey = preg_replace("/^msgstr/", "msgid", $msgStr);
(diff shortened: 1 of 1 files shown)


tvdijen:
I'm having a hard time understanding what this script does.. Can you give an example?

restena-sw:
Hi,

the translations in the locale/ subdirs contain entries like

en/   

msgid "{logout:failed}"
msgstr "Logout failed"

fr/
msgid "{logout:failed}"
msgstr "quelque chose"

That structure is entirely unhelpful for CAT tools as they put the English source (not an abstract string) and the desired translation side-by-side, as in:

msgid "Logout failed"
msgstr "quelque chose"

As you see, this involves not much more than inverting orders of strings and changing msgstr to msgid.

One of the operation modes of the script is to do just that - take two individual language input files and generate one which has English as msgid and the translation as msgstr.

Outputs of that do not contain the original "glueing" abstract string any more - which is what translators want. But eventually, after working with the generated file, the translation needs to be brought back into SSPs native format.

That's why the inverse operation mode takes a file from original sources (abstract string+English) and generates output with the abstract string and the target translation instead. Tada!

The code is rather dense, admittedly. That's because it mostly does some s/msgstr/msgid and at other places juggles line indexes around.

restena-sw:
For context, you may also want to read my mailing list post from 04 March 2020 "Translation remarks / question".

bin/convertTranslations.php
```diff
@@ -0,0 +1,129 @@
...
+");
+    exit(1);
+}
+
+switch ($argv[1]) {
+    case "MERGE":
+        fwrite(STDERR, "Will merge two language into one .po file based on identical msgIds.\n");
+        break;
+    case "CODIFY":
+        fwrite(STDERR, "Will create .po file with codes as msgIds and dest langauge translations as msgStr.\n");
```
tvdijen, line 86: typo: language

tvdijen:
Ah ok.. I always use the dictionaryconverter script for that..

restena-sw:
"replace the tag-codes with actual sentences" sounds like manual work is needed every time? That's where an automation script helps :-)
Also, I had the impression that JSON dictionaries are deprecated/obsolete so that won't be possible for long?

tvdijen:
Yes, that's a manual action, but it's just in one file instead of every PO-file..
The JSON-dictionaries are already gone on `master`, but of course we still have them in the 1.19 branch..

thijskh (approved):
Just what we need to clean up the old translation styles. Thanks!

PR #1309, opened 2020-04-03
add the possibility to specify one AuthSource to be shown

pkolmann:
I have multiple AuthSources, but the User should be presented only one Login-Possibility. So I need to specify which authSource to use.

This change adds a pseudo authSource by the name 'showThisSource' which is used in the core-Controller to select only the one AS to be shown.

### DIFF
--- a/config-templates/authsources.php
+++ b/config-templates/authsources.php
@@ -10,6 +10,11 @@ $config = [
         'core:AdminPassword',
     ],
 
+    /* When multiple authentication sources are defined, you can specify here which one to show 
+     * and the others could be used with multiauth:MultiAuth
+     */
+    // 'showThisSource' = 'default-sp',
+
 
     // An authentication source which can authenticate against both SAML 2.0
     // and Shibboleth 1.3 IdPs.
--- a/config-templates/authsources.php
+++ b/config-templates/authsources.php
@@ -13,7 +13,7 @@ $config = [
     /* When multiple authentication sources are defined, you can specify here which one to show 
      * and the others could be used with multiauth:MultiAuth
      */
-    // 'showThisSource' = 'default-sp',
+    // 'showThisSource' = ['default-sp'],
 
 
     // An authentication source which can authenticate against both SAML 2.0
--- a/modules/core/lib/Controller.php
+++ b/modules/core/lib/Controller.php
@@ -128,9 +128,10 @@ class Controller
             $as = key($this->sources);
         }
 
-        if (array_key_exists('showThisSource', $this->sources) &&
-            array_key_exists($this->sources['showThisSource'], $this->sources)) {
-            $as = $this->sources['showThisSource'];
+        if (array_key_exists('showThisSource', $this->sources)) {
+            if (array_key_exists($this->sources['showThisSource'], $this->sources)) {
+                $as = $this->sources['showThisSource'];
+            }
             unset($this->sources['showThisSource']);
         }
 
--- a/modules/core/lib/Controller.php
+++ b/modules/core/lib/Controller.php
@@ -129,8 +129,8 @@ class Controller
         }
 
         if (array_key_exists('showThisSource', $this->sources)) {
-            if (array_key_exists($this->sources['showThisSource'], $this->sources)) {
-                $as = $this->sources['showThisSource'];
+            if (array_key_exists($this->sources['showThisSource'][0], $this->sources)) {
+                $as = $this->sources['showThisSource'][0];
             }
             unset($this->sources['showThisSource']);
         }
--- a/modules/core/lib/Controller.php
+++ b/modules/core/lib/Controller.php
@@ -128,6 +128,12 @@ class Controller
             $as = key($this->sources);
         }
 
+        if (array_key_exists('showThisSource', $this->sources) &&
+            array_key_exists($this->sources['showThisSource'], $this->sources)) {
+            $as = $this->sources['showThisSource'];
+            unset($this->sources['showThisSource']);
+        }
+
         if ($as === null) { // no authentication source specified
             $t = new Template($this->config, 'core:login.twig');
             $t->data['loginurl'] = Utils\Auth::getAdminLoginURL();


jaimeperez:
Hi @pkolmann!

May I ask why do you need to have multiple auth sources defined, if your users are only supposed to use one?

pkolmann:
Hi @jaimeperez,

we have a primary AuthSource which checks a local DB. Additinally we have also users, that need to be authenticated by a different auth-source and third we have Handysignatur (predecessor to european e-ID login) for people who have forgotten their password.

We present one page where the user can then choose, if they want to login immediately via a form, use the Handysignatur login or login as an external library user (the different auth-source from above).

2020-04-06 15_46_37-Window

Since any should end up in one session, I have created a multiAuth where I specified a preselct. But the the level before, the user was still shown all the possibilities. Thats why I had to come up with this patch.

Thanks
Philipp

tvdijen:
@sgomez once had a similar use-case:  https://github.com/simplesamlphp/simplesamlphp/pull/1005
That may be able to help you..

pkolmann:
> 
> 
> @sgomez once had a similar use-case: #1005
> That thread may be helpful you..

Thanks @tvdijen I am using already the option provided there but it still shows beforehand a list of all AuthSources. Therefore I added this option.

tvdijen:
It sounds to me as if the preselect-option is broken...

pkolmann:
@tvdijen no, preselect works. If I turn it off, I see a this list at simplesaml/module.php/multiauth/selectsource.php

grafik

without the showThisSource parameter, I see the whole list at simplesaml/module.php/core/login

grafik

tvdijen:
Now you have me confused..  You said that with the preselect-option you were given the whole list.. Now you're saying something different..  `core/login` will always give you the whole list, that's just by design... It has nothing to do with the multiauth-module

pkolmann:
>  `module.php/core/login` will always give you the whole list, that's just by design...

Sorry, I think I have confused us both. You are completly right.

The change I am proposing, is to not show the whole list at `module.php/core/login` but just continue to the one selected auth-source.

tvdijen:
That's not less confusing.. You should configure your IDP to use the right authsource;  https://github.com/simplesamlphp/simplesamlphp/blob/master/metadata-templates/saml20-idp-hosted.php#L25
Then, if that authsource happens to be one of multiAuth, the preselect should kick in..

pkolmann:
Thanks, thats all working as expected.

I have just found it might be confusing for our users if they try and access https://our.host/simplesaml/ and get the whole list presented. So I want them to get redirected to the one login form and not be confused with a lot of options.

but I understand that this might not be useful for others, so if you don't like this change, please just close it.

Thanks
Philipp

tvdijen:
Ahhh, so we're not talking about an authentication-flow, but just a single visit to the root... Now it's making sense! We should be able to detect this and handle it differently

jaimeperez:
Hi @pkolmann!

I definitely see your point. And it probably makes sense to do this, but at the same time I'd like to think it twice before adding more configuration options, as we already have tons of them. Since I see you have your own theme, wouldn't it be an option for you to override the template there? You could then choose what you want to display, which doesn't need to be the full list. Would something like that work for you?

pkolmann:
Hi @jaimeperez 

thanks for your reply. We could probably just show one line with our template, but then the user would still have the extra click.

I understand, that there are already a ton of config options. So I am also happy to keep this patch only locally for us and you close the PR.

thanks
Philipp

jaimeperez:
That's a fair point indeed. Let me rework a bit your PR if you don't mind before merging.

pkolmann:
> 
> 
> That's a fair point indeed. Let me rework a bit your PR if you don't mind before merging.

Thanks a lot. Just change whatever you like. I would be very happy to at least have contributed an idea for an improvement.

jaimeperez:
Hi Philipp!

I've taken a slightly different approach. Instead of adding a configuration option, I've used the same convention as with the admin auth source. Whatever you configure in an auth source named "default", becomes the default. If you don't use that name, then you allow the user to choose. If you already have an auth source with a different name, you can just create an alias in the `authsources.php`:

```php
$config['default'] = &$config['my_existing_auth_source'];
```

Now, this is new functionality, meaning it cannot land in 1.18 (that's frozen for new features now, and only bugfixes are applied to it), so I've applied the changes both in `master` and `simplesamlphp-1.19`. If you really want to start using it now, you can apply this patch on top of 1.18:

```patch
diff --git a/config-templates/authsources.php b/config-templates/authsources.php
index f2d562163..d909a3881 100644
--- a/config-templates/authsources.php
+++ b/config-templates/authsources.php
@@ -1,6 +1,17 @@
 <?php

 $config = [
+    /*
+     * When multiple authentication sources are defined, you can specify one to use by default
+     * in order to authenticate users. In order to do that, you just need to name it "default"
+     * here. That authentication source will be used by default then when a user reaches the
+     * SimpleSAMLphp installation from the web browser, without passing through the API.
+     *
+     * If you already have named your auth source with a different name, you don't need to change
+     * it in order to use it as a default. Just create an alias by the end of this file:
+     *
+     * $config['default'] = &$config['your_auth_source'];
+     */

     // This is a authentication source which handles admin authentication.
     'admin' => [
diff --git a/modules/core/lib/Controller/Login.php b/modules/core/lib/Controller/Login.php
index e117cdc2c..e555b1c66 100644
--- a/modules/core/lib/Controller/Login.php
+++ b/modules/core/lib/Controller/Login.php
@@ -86,7 +86,7 @@ class Login
         }

-
+
         $session = Session::getSessionFromRequest();

         $t = new Template($this->config, 'auth_status.twig', 'attributes');
@@ -132,11 +132,30 @@ class Login
             $as = key($this->sources);
         }

+        $default = false;
+        if (array_key_exists('default', $this->sources) && is_array($this->sources['default'])) {
+            $default = $this->sources['default'];
+        }
+
         if ($as === null) { // no authentication source specified
-            $t = new Template($this->config, 'core:login.twig');
-            $t->data['loginurl'] = Utils\Auth::getAdminLoginURL();
-            $t->data['sources'] = $this->sources;
-            return $t;
+            if (!$default) {
+                $t = new Template($this->config, 'core:login.twig');
+                $t->data['loginurl'] = Utils\Auth::getAdminLoginURL();
+                $t->data['sources'] = $this->sources;
+                return $t;
+            }
+
+            // we have a default, use that one
+            $as = 'default';
+            foreach ($this->sources as $id => $source) {
+                if ($id === 'default') {
+                    continue;
+                }
+                if ($source === $this->sources['default']) {
+                    $as = $id;
+                    break;
+                }
+            }
         }

         // auth source defined, check if valid
```

Thanks a lot for your contribution!

PR #1310, opened 2020-04-03
Fix the front end modules URL generation in Module.php

KalemaEdgar:
The URL was bieng generated with the folder name (module) coming as module.php hence the front end pages could not load successfully.
This changes from module.php to modules

### DIFF
--- a/lib/SimpleSAML/Module.php
+++ b/lib/SimpleSAML/Module.php
@@ -468,7 +468,7 @@ class Module
     {
         Assert::notSame($resource[0], '/');
 
-        $url = Utils\HTTP::getBaseURL() . 'module.php/' . $resource;
+        $url = Utils\HTTP::getBaseURL() . 'modules/' . $resource;
         if (!empty($parameters)) {
             $url = Utils\HTTP::addURLParameters($url, $parameters);
         }


jaimeperez:
Hi @KalemaEdgar,

This is wrong. The correct path of the URL to a module must be (for now) `.../modules.php/module_name/`, and it works fine out of the box. If you are experiencing problems with this, you probably have some sort of deployment issue.

KalemaEdgar:
> Hi @KalemaEdgar,
> 
> This is wrong. The correct path of the URL to a module must be (for now) `.../modules.php/module_name/`, and it works fine out of the box. If you are experiencing problems with this, you probably have some sort of deployment issue.

Thanks @jaimeperez, it was failing for me but when I change it to modules, it works fine. I will check with my implementation.

KalemaEdgar:
> Hi @KalemaEdgar,
> 
> This is wrong. The correct path of the URL to a module must be (for now) `.../modules.php/module_name/`, and it works fine out of the box. If you are experiencing problems with this, you probably have some sort of deployment issue.

Thanks @jaimeperez, it was failing for me but when I change it to modules, it works fine. I will check with my implementation.

PR #1324, opened 2020-05-04
 fix createLogoutTable method

### LINKED ISSUE #1323: Bug in createLogoutTable method
An error occurred while updating the simplesamlphp library

logs

>  4 [0900c986a2] Database error: array (#012  0 => '42601',#012  1 => 7,#012  2 => 'ERROR:  syntax error at or near "TIMESTAMP"#012LINE 1: ...impleSAMLphp_saml_LogoutStore ALTER COLUMN _expire TIMESTAMP#012

Marat555:
Closes https://github.com/simplesamlphp/simplesamlphp/issues/1323

### DIFF
--- a/modules/saml/lib/SP/LogoutStore.php
+++ b/modules/saml/lib/SP/LogoutStore.php
@@ -40,7 +40,7 @@ class LogoutStore
                 case 'pgsql':
                     // This does not affect the NOT NULL constraint
                     $update = [
-                        'ALTER TABLE ' . $store->prefix . '_saml_LogoutStore ALTER COLUMN type _expire TIMESTAMP'
+                        'ALTER TABLE ' . $store->prefix . '_saml_LogoutStore ALTER COLUMN TYPE _expire TIMESTAMP'
                     ];
                     break;
                 case 'sqlite':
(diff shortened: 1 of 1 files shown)


modules/saml/lib/SP/LogoutStore.php
```diff
@@ -40,7 +40,7 @@ private static function createLogoutTable(Store\SQL $store): void
                 case 'pgsql':
                     // This does not affect the NOT NULL constraint
                     $update = [
-                        'ALTER TABLE ' . $store->prefix . '_saml_LogoutStore ALTER COLUMN _expire TIMESTAMP'
+                        'ALTER TABLE ' . $store->prefix . '_saml_LogoutStore ALTER COLUMN type _expire TIMESTAMP'
```
tvdijen, line 43: Shouldn't it be the other way around? `ALTER COLUMN _expire TYPE TIMESTAMP`?

Marat555: Yes, it looks better.

Marat555:
What's wrong with tests?

tvdijen:
Nothing to worry about.. Sometimes I have to restart the Travis-tests... Scrutinizer is just a piece of junk that always fails for no reaosn

However, I think you have missed the point in my earlier comment.. I think the order of the words is wrong, unless I'm misinterpreting the docs..  The TYPE keyword should go _after_ the column name

Marat555:
Fixed)

PR #1325, opened 2020-05-05
Fix bug in saml:NameIDAttribute authproc filter

zmousm:
Override `NameID` `Format`, `NameQualifier`, `SPNameQualifier` when they **are** `null`

### DIFF
--- a/modules/saml/lib/Auth/Process/NameIDAttribute.php
+++ b/modules/saml/lib/Auth/Process/NameIDAttribute.php
@@ -123,11 +123,12 @@ class NameIDAttribute extends \SimpleSAML\Auth\ProcessingFilter
         if ($rep->getFormat() === null) {
             $rep->setFormat(Constants::NAMEID_UNSPECIFIED);
         }
-        if ($rep->getNameQualifier() === null) {
-            $rep->setNameQualifier($state['Source']['entityid']);
-        }
+
         if ($rep->getSPNameQualifier() === null) {
-            $rep->setSPNameQualifier($state['Destination']['entityid']);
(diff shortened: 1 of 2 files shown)


thijskh:
Thanks! But probably the two `if` statements below also need to be fixed then?

tvdijen:
No, those are just fine Thijs.. Ideally I'd like to see a unit test for this though

zmousm:
> Thanks! But probably the two `if` statements below also need to be fixed then?

Correct!

tvdijen:
I took care of the fix and I'll backport it.. Leaving this open for the unit test

zmousm:
> Ideally I'd like to see a unit test for this too

I will look into that...

tvdijen:
@thijskh: The tests failing seems to something that's bugged for a long time..
There's a contradiction between [here](https://github.com/simplesamlphp/simplesamlphp/blob/991415f04b931f3cd164e797d262074f423fc23a/tests/modules/saml/lib/Auth/Process/NameIDAttributeTest.php#L62) and [here](https://github.com/simplesamlphp/simplesamlphp/blob/991415f04b931f3cd164e797d262074f423fc23a/modules/saml/lib/Auth/Process/NameIDAttribute.php#L54).. Do you know if this is some kind of standardized syntax and what the correct order between IDP|SP is?

zmousm:
> Ideally I'd like to see a unit test for this too

I added a test for the overrides, when the respective NameID properties are `null`. I think the reverse is covered by the existing tests.

tvdijen:
Excellent, thanks @zmousm !  Thanks!

PR #1334, opened 2020-06-05
Allow Controllers to get the matched route

dlundgren:
This allows a controller to call `$request->attributes->get('_routeparams')` to access any available parameters.

### DIFF
--- a/lib/SimpleSAML/Module/ControllerResolver.php
+++ b/lib/SimpleSAML/Module/ControllerResolver.php
@@ -102,6 +102,7 @@ class ControllerResolver extends SymfonyControllerResolver implements ArgumentRe
         try {
             $matcher = new UrlMatcher($this->routes, $ctxt);
             $this->params = $matcher->match($ctxt->getPathInfo());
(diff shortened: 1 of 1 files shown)


tvdijen:
Hi @dlundgren ! 1.18 will only receive bugfixes, and this particular file was removed in 1.19, so I'm tempted to mark this as a wontfix..
@sgomez Does this make sense to add to the new module/routing system?

sgomez:
I don't think so. This parameter is configured by the [RouterListener service](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/EventListener/RouterListener.php#L128) and anyway the right name is `_route_params`and not `_routeparams`.

dlundgren:
I saw the dramatic migration to Symfony in master & 1.19, which is great by they way!

I've found an alternative way to do this in my Actions, by overriding the trait I am using for handling the route, and having the parameter set, then calling the trait.

Thanks for all the hard work.

tvdijen:
Yeah, we have had our challenges with Symfony (and still have), but it gonna be worth it in the long run!

PR #1335, opened 2020-06-09
Few small improvements

msalle:
- fix a Dutch translation typo (_contrab_ should be _crontab_)
- add some missing Dutch translations
- add a line break to properly align the first `# Run cron:` line in module.php/cron/croninfo.php

### DIFF
--- a/locales/nl/LC_MESSAGES/messages.po
+++ b/locales/nl/LC_MESSAGES/messages.po
@@ -44,6 +44,9 @@ msgstr "Toestandsinformatie verloren"
 msgid "{login:username}"
 msgstr "Gebruikersnaam"
 
+msgid "{login:processing}"
+msgstr "Bezig te verwerken..."
+
 msgid "Back"
 msgstr "Terug"
 
@@ -87,6 +90,9 @@ msgstr "Kan geen metadata vinden voor %ENTITYID%"
 
 msgid "{admin:metadata_metadata}"
 msgstr "Metadata"
+
+msgid "{errors:title_SSOPARAMS}"
+msgstr "Geen SAML request opgegeven"
 
 msgid "{errors:report_text}"
 msgstr ""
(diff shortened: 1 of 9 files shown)


locales/nl/LC_MESSAGES/messages.po
```diff
@@ -44,6 +44,9 @@ msgstr "Toestandsinformatie verloren"
 msgid "{login:username}"
 msgstr "Gebruikersnaam"
 
+msgid "{login:processing}"
```
tvdijen, line 47: This is exactly the opposite of what should be done.. We are getting rid of the {tags}, so the right thing to do is to fix the Twig-templates to use the msgid holding the English translation instead.

msalle: Ah, right )-: I looked at what was done for some of these strings in other languages.
By the way, this particular one (`login:processing`) and `admin:metaconv_selectfile` are not used in any template or php file, the tag is just used in the `.po` files to get the right translation, if I understand correctly.
And for some of these which are used elsewhere, it would look kind of awkward: [lib/SimpleSAML/Error/ErrorCodes.php](https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Error/ErrorCodes.php), uses lots of the `{errors:` and some are big, such as `{errors:descr_SSOPARAMS}` and `{errors:descr_ARSPARAMS}`.
In summary, I'm not sure how you suggest I should fix in particular the `{errors:` tags, the other two should be straightforward indeed, but then would need fixing in the language .po files.

tvdijen: The ones not used should be removed completely, from all languages.
The error-tags should be replaced by the english translation, also in the ErrorCodes.php file. Again, for all languages.

msalle: I've removed the two unused ones everywhere and re-fixed at least the errors:* that I had changed. Is this what you suggest, in particular for ErrorCodes.php? If so, we should also fix the rest there. The single-line ones are easy to automate, the multilines will be involved.

tvdijen: Yes, that's exactly what we're after!  It's just a job that someone started and never finished

tvdijen:
Hmm, I accepted it and then I noticed the tests failed..
Should probably use heredoc/nowdoc in the ErrorCodes-file

msalle:
> Hmm, I accepted it and then I noticed the tests failed..
> Should probably use heredoc/nowdoc in the ErrorCodes-file

Ah, sorry my bad indeed. I think a `.` at the EOL should be sufficient.

PR #1345, opened 2020-06-19
Allow additional Audiences to be specified

### LINKED ISSUE #737: Cannot Specify Audience for SAML 2.0 Remote Metadata
So I'm setting up AWS as a SP and they require that you specify the AudenceRestriction as shown here:

```
<Conditions>
  <AudienceRestriction>
    <Audience>https://signin.aws.amazon.com/saml</Audience>
    <Audience>urn:amazon:webservices</Audience>
  </AudienceRestriction>
</Conditions>
```
I see that you can override the audience from the EntityID to whatever you want on a Shib 1.3 metadata set, but not SAML 2.0.

Is there a reason behind that? or is it just something that was never added? If it's the latter I can start up a PR to do that.

tvdijen:
Closes #737

### DIFF
--- a/docs/simplesamlphp-reference-sp-remote.md
+++ b/docs/simplesamlphp-reference-sp-remote.md
@@ -160,6 +160,9 @@ The following options can be set:
 
 :   (This option was previously named `AttributeNameFormat`.)
 
+`audience`
+:   Additional entities to be added to the AudienceRestriction. By default the only audience is the SP's entityID. 
+
 `certData`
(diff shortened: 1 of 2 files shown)


docs/simplesamlphp-reference-sp-remote.md
```diff
@@ -160,6 +160,9 @@ The following options can be set:
 
 :   (This option was previously named `AttributeNameFormat`.)
 
+`audience`
```
thijskh, line 163: Maybe make explicit that it’s an array?

PR #1348, opened 2020-06-22
Provider name

### LINKED ISSUE #1346: Support ProviderName attribute
The SAML schema protocol specifies that the [AuthnRequest](https://github.com/simplesamlphp/simplesamlphp/blob/8727f0d73169aebf6a30b8e836ef4c28bfe14df9/schemas/saml-schema-protocol-2.0.xsd#L152) supports the[ ProviderName](https://github.com/simplesamlphp/simplesamlphp/blob/8727f0d73169aebf6a30b8e836ef4c28bfe14df9/schemas/saml-schema-protocol-2.0.xsd#L169) attribute. However the saml:SP configuration does not provide any way to set it.

[saml:SP](https://simplesamlphp.org/docs/stable/saml:sp) should have an option named `ProviderName` that sets the ProviderName in the AuthnRequest for every SAML request emitted for that SP.

**Additional context**
I was able to implement a solution for it, but since I dont know the code base very well its probably in the wrong place.

In `modules/saml/lib/Auth/Source/SP.php` inside the method [`startSSO2`](https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/saml/lib/Auth/Source/SP.php#L453) I've added these lines:
```php
$providerName = $this->metadata->getString("ProviderName");
if ($providerName !== null) {
  $ar->setProviderName($providerName);
}
```

Lasering:
Fixes #1346

### DIFF
--- a/modules/saml/docs/sp.md
+++ b/modules/saml/docs/sp.md
@@ -330,6 +330,11 @@ Options
 
 :   *Note*: SAML 2 specific.
 
+`ProviderName`
+:   Human readable name of the local SP sent with the authentication request.
+
+:   *Note*: SAML 2 specific.
+
 `ProtocolBinding`
 :   The binding that should be used for SAML2 authentication responses.
     This option controls the binding that is requested through the AuthnRequest message to the IdP.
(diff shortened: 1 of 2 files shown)


Lasering:
@tvdijen I cannot see what is causing Scrutinizer to fail.

tvdijen:
It just does that for fun.. No worries about that one!

PR #1349, opened 2020-06-22
Support saml:Extensions inside Authsources.php

### LINKED ISSUE #1347: Support saml:Extensions inside Authsources.php
The documentation for [saml:SP](https://simplesamlphp.org/docs/stable/saml:sp) references the `saml:Extensions` parameter, however its use seems to be limited to the \SimpleSAML\Auth\Simple class. This makes using SAML extensions in a bridge (an IdP acting as an SP to another IdP) impossible.

`saml:Extensions` should be supported for bridges.

**Additional context**
I was able to implement a solution for it, but since I dont know the code base very well its probably in the wrong place.

In `modules/saml/lib/Auth/Source/SP.php` inside the method [`startSSO2`](https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/saml/lib/Auth/Source/SP.php#L453) I've added these lines:
```php
$extensions = $this->metadata->getArray('saml:Extensions');
if ($extensions !== null) {
  $ar->setExtensions($extensions);
}
```

Lasering:
Fixes #1347

### DIFF
--- a/modules/saml/lib/Auth/Source/SP.php
+++ b/modules/saml/lib/Auth/Source/SP.php
@@ -585,17 +585,20 @@ class SP extends \SimpleSAML\Auth\Source
         );
 
         $ar->setRequesterID($requesterID);
-
-        if ($this->metadata->getArray('saml:Extensions', null) !== null) {
-            $ar->setExtensions($this->metadata->getArray('saml:Extensions'));
-        } else if (isset($state['saml:Extensions'])) {
+        
+        // If the downstream SP has set extensions then use them.
+        // Otherwise use extensions that might be defined in the local SP (only makes sense in a proxy scenario)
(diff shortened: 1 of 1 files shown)


tvdijen:
Same issue, same solution, I reckon?

Lasering:
Yes, I've just changed the order as discussed on the issue.

Lasering:
This PR makes the necessary changes so the following is true:

With the following setup `Test SP` <=> `Proxy IdP/Proxy SP` <=> `IdP A`, where the arrow means a trust relationship.
1. If `Test SP` defines extensions, and `Proxy SP` does not, then the `Test SP` extensions are correctly sent to `IdP A`,
2. If `Test SP` does not define extensions but `Proxy SP` does, then the extensions of `Proxy SP` are sent to `IdP A`.
3. If both `Test SP` and `Proxy SP` define extensions then the `Test SP` extensions are sent to `IdP A`.

PR #1351, opened 2020-07-01
Remove master-slave metaphor

ghalse:
In light of increased awareness around the use of oppressive language and its role in the #BlackLivesMatter movement, it seems appropriate to remove references to the master-slave metaphor within SSP. While there are other examples of oppressive language (notably whitelist/blacklist) within SSP, this particular one seems both the most problematic and the easiest to resolve.

Fortunately, there are very few references to master-slave within the core of SSP and their use is limited entirely to the database abstraction layer. Moreover, most of the references are entirely internal within that module with only one external dependency.

This patch follows the suggestions of [draft-knodel-terminology-00](https://tools.ietf.org/id/draft-knodel-terminology-00.html) and other language frameworks to replace the notion of master/slave with a more neutral primary/secondary.

The legacy configuration option `database.slaves` continues to work, but is marked as deprecated and issues a warning. It has been replaced by `database.secondaries` and the config template updated to reflect this.

### DIFF
--- a/lib/SimpleSAML/Database.php
+++ b/lib/SimpleSAML/Database.php
@@ -7,12 +7,13 @@ namespace SimpleSAML;
 use PDO;
 use PDOException;
 use PDOStatement;
+use SimpleSAML\Logger;
 
 /**
  * This file implements functions to read and write to a group of database servers.
  *
- * This database class supports a single database, or a master/slave configuration with as many defined slaves as a
- * user would like.
+ * This database class supports a single database, or a primary/secondary configuration with as many defined secondaries
+ * as a user would like.
  *
  * The goal of this class is to provide a single mechanism to connect to a database that can be reused by any component
  * within SimpleSAMLphp including modules. When using this class, the global configuration should be passed here, but in
@@ -31,14 +32,14 @@ class Database
     private static $instance = [];
 
     /**
-     * PDO Object for the Master database server
+     * PDO Object for the Primary database server
      */
-    private $dbMaster;
+    private $dbPrimary;
 
     /**
-     * Array of PDO Objects for configured database slaves
+     * Array of PDO Objects for configured database secondaries
      */
-    private $dbSlaves = [];
+    private $dbSecondaries = [];
 
     /**
      * Prefix to apply to the tables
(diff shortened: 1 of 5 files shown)


lib/SimpleSAML/Database.php
```diff
@@ -86,28 +87,34 @@ private function __construct(Configuration $config)
...
             $config->getString('database.dsn'),
             $config->getString('database.username', null),
             $config->getString('database.password', null),
             $driverOptions
         );
 
-        // connect to any configured slaves
-        $slaves = $config->getArray('database.slaves', []);
-        foreach ($slaves as $slave) {
+        // TODO: deprecated racism: the "database.slave" terminology is preserved here for backwards compatibility.
```
thijskh, line 98: This comment is maybe a bit harsh. Just "deprecated" will suffice.

tvdijen:
We shouldn't be spending even a minute on this excessive kind of political correctness.
It's called a master/slave architecture for a reason..

ghalse:
> It's called a master/slave architecture for a reason..

See [draft-knodel-terminology-00](https://tools.ietf.org/id/draft-knodel-terminology-00.html) for reasons why the metaphor is technically (not politically) flawed.

FWIW I adopted the primary/secondary nomenclature (rather than any of the other [suggested alternatives](https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.1.1)) because it is consistent with [MySQL](https://dev.mysql.com/doc/refman/8.0/en/group-replication-primary-secondary-replication.html), [MSSQL](https://docs.microsoft.com/en-us/sql/relational-databases/databases/secondary-database-settings?view=sql-server-ver15), [Oracle](https://docs.oracle.com/html/E16706_01/ins_multidb_unix.htm) and was thus more likely to be consistent with other documentation. PostgreSQL is the odd-one-out but still uses [primary/standby](https://www.postgresql.org/docs/12/warm-standby-failover.html) rather than master/slave. Moreover, all of these have used alternatives to the master/slave terminology for several years and any reference to that terminology in their documentation is usually in parenthesis after newer terms.

So even from the perspective of consistency with database vendors, the use of this terminology in SSP is out-of-date.

lib/SimpleSAML/Database.php
ghalse: Searching for "deprecated" brings up lots of hits, so I was trying to tag it usefully if we needed easy way to find or mark similar legacy occurrences later on, particularly if we tackle the blacklist/whitelist version too. I'll agree that "racism" is a bit harsh - I maybe could have used a reference to knodel or similar instead to create a unique search term.

tvdijen:
Well, I think the world has gone proper mad..
And apparently I have nothing to say any more in this team since this is already merged.

thijskh:
You said you didn’t want to spend time on it, that is fine. But please allow others to spend their time if they want.

Issue #1342, opened 2020-06-16
Symfony 5


bjoerne2:
I'd like to use simplesamlphp in a Laravel 7 project. Laravel 7 uses Symfony 5 libs, but simplesamlphp's composer.json references ^4.0. Now I have a version conflict I can't resolve.

Is there a plan to migrate simplesamlphp to Symfony 5 libs?

tvdijen:
Hi @bjoerne2 ! You're way ahead of us, as `^4.0` is the tag on our `master` branch, which you should never use in production.. Stable versions are still at `^3.0|^4.0`.. We're in the process of migrating to Symfony, but it's a true struggle and our resources are slim. The latter may hopefully change in the near future due to changes in the SimpleSAMLphp organization..

Even though I discourage the use of our master-branch, I don't mind changing the composer.json to require Symfony `^4.0|^5.0`, as 5.0 is probably where we're going to end once we're ready to release SSP 2.0.. Would that help you?

tvdijen:
There seem to be no real upgrade notes for Symfony 4.4 > 5.0.. At least no BC breaks.. Do you seem any harm in targeting `^4.0||^5.0` or even `^5.0` for SSP 2.0?

bjoerne2:
`^4.0||^5.0` sounds great to me! Thanks a lot for your quick reply!

tvdijen:
Ok, done that on master

bjoerne2:
@tvdijen I created a PR for the simplesamlphp-1.19 branch: https://github.com/simplesamlphp/simplesamlphp/pull/1358

PR #1358, opened 2020-07-26
Support Symfony5

bjoerne2:
According to https://github.com/simplesamlphp/simplesamlphp/issues/1342, but for simplesamlphp-1.19 branch and upcoming 1.19 release.


tvdijen:
Thanks @bjoerne2, but 1.19 is targeted to PHP 7.0 and Symfony 5 requires 7.2..
Also, the change to master has actually broken it and I'm still considering to revert it if I can't figure out what the issue is

Issue #912, opened 2018-08-23
Cannot provide default MultiAuth source from SP to IdP using method in documentation


BAGELreflex:
Per the docs located here: https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/multiauth/docs/multiauth.md

The documentation outlines two methods of defaulting the source based on the SP. The first is to pass a `GET` parameter of `?source=myPreferredSource` "to the calling URL, when accessing a service". As best I can tell, this means redirecting to a URL of this format:
```
https://mysite.com/auth/idp/saml2/idp/SSOService.php?source=myPreferredSource&SAMLRequest=…
```
when calling `$auth->login()` or `$auth->requireAuth()`.

There has been discussion about this topic: https://groups.google.com/forum/#!topic/simplesamlphp/9QAJXhZ8b1o
In this discussion, one suggestion was to modify core code, but that is just not acceptable for an enterprise-level environment, as it introduces much more work for future upgrades. The route they finally took was leveraging the `ReturnTo` parameter, and parsing that value in a custom `mymodule/themes/multiauth/selectsource.php`, then setting `$source` to the parsed value.

The second suggestion is to use a non-standard `multiauth:preselect` parameter to the `login()` or `requireAuth()` functions of `\SimpleSAML\Auth\Simple`. This seemed the most appropriate method for my use-case, however after trying to implement it, I was unsuccessful. After further review, I have found that this non-standard key is not passed on from the SP to the IdP, unless the SP and IdP are running on the same server, using the same SSP installation. This is because the `multiauth:preselect` is saved to the state on the SP, and is not passed as part of the `\SAML2\AuthnRequest`. See https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/saml/lib/Auth/Source/SP.php#L319. The state is saved here https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/saml/lib/Auth/Source/SP.php#L284 but that will only be available to the SP. When the IdP receives the request, and the state is loaded here https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/multiauth/www/selectsource.php#L18 the `multiauth:preselect` would never be available, unless the IdP and the SP are on the same server.

So, how do we make this work?

After an hour or two of tracing through the code, I have finally figured out that I can simply tack on `?source=myPreferredSource` to the **Location** index of the **SingleSignOnService** inside `metadata/saml20-idp-remote.php`.

```php
[...]

PR #1362, opened 2020-08-04
Filter multiauth authentication sources from SP using AuthnContextClassRef

Lasering:
This PR introduces the necessary changes to allow an SP to choose an authentication source of multiauth. It works even if there are redirects during the authentication process because it uses SAML2 standard AuthnContextClassRef. See #912 for more context.

### DIFF
--- a/modules/multiauth/docs/multiauth.md
+++ b/modules/multiauth/docs/multiauth.md
@@ -79,7 +79,7 @@ compatible fashion so both cases should work.
 
 Each source in the sources array has a key and a value. As
 mentioned above the key is the authsource identifier and the value
-is another array with optional keys: 'text', 'css-class', 'help', and 'ContextClassRef'.
+is another array with optional keys: 'text', 'css-class', 'help', and 'AuthnContextClassRef'.
 The text element is another array with localized strings for one
 or more languages. These texts will be shown in the selectsource.php
 view. Note that you should at least enter the text in the default
@@ -91,7 +91,7 @@ replaced by dashes. So in the previous example, the css class used
 in the 'example-admin' authentication source would be
 'core-AdminPassword'. The help element is another array with localized
 strings for one or more languages. These texts will be shown in the
-selectsource.php view. The ContextClassRef is either a string or
+selectsource.php view. The AuthnContextClassRef is either a string or
 an array of strings containing [context class ref names](https://docs.oasis-open.org/security/saml/v2.0/saml-authn-context-2.0-os.pdf).
 If an SP sets AuthnContextClassRef the list of authsources will be
 filtered to only those containing context class refs that are part of the list set by the SP.
--- a/modules/multiauth/docs/multiauth.md
+++ b/modules/multiauth/docs/multiauth.md
@@ -36,12 +36,14 @@ authentication source:
                     'es' => 'Entrar usando un SP SAML',
                 ),
                 'css-class' => 'SAML',
+                'AuthnContextClassRef' => array('urn:oasis:names:tc:SAML:2.0:ac:classes:SmartcardPKI', 'urn:oasis:names:tc:SAML:2.0:ac:classes:MobileTwoFactorContract'),
             ),
             'example-admin' => array(
                 'text' => array(
                     'en' => 'Log in using the admin password',
                     'es' => 'Entrar usando la contraseña de administrador',
                 ),
+                'AuthnContextClassRef' => 'urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport',
             ),
         ),
     ),
@@ -77,7 +79,7 @@ compatible fashion so both cases should work.
 
 Each source in the sources array has a key and a value. As
 mentioned above the key is the authsource identifier and the value
-is another array with two optional keys: 'text' and 'css-class'.
+is another array with optional keys: 'text', 'css-class', 'help', and 'ContextClassRef'.
 The text element is another array with localized strings for one
 or more languages. These texts will be shown in the selectsource.php
 view. Note that you should at least enter the text in the default
@@ -87,7 +89,14 @@ the &lt;li> element in the selectsource.php view. By default the
 authtype of the authsource is used as the css class with colons
 replaced by dashes. So in the previous example, the css class used
 in the 'example-admin' authentication source would be
-'core-AdminPassword'.
+'core-AdminPassword'. The help element is another array with localized
+strings for one or more languages. These texts will be shown in the
+selectsource.php view. The ContextClassRef is either a string or
+an array of strings containing [context class ref names](https://docs.oasis-open.org/security/saml/v2.0/saml-authn-context-2.0-os.pdf).
+If an SP sets AuthnContextClassRef the list of authsources will be
+filtered to only those containing context class refs that are part of the list set by the SP.
+If a single authsource results from this filtering the user will be taken directly to the
+authentication page for that source, and will never be shown the multiauth select page.
 
 It is possible to add the parameter `source` to the calling URL, 
 when accessing a service, to allow the user to preselect the
--- a/modules/multiauth/lib/Auth/Source/MultiAuth.php
+++ b/modules/multiauth/lib/Auth/Source/MultiAuth.php
@@ -152,6 +152,9 @@ class MultiAuth extends Auth\Source
      */
     public function authenticate(array &$state): void
     {
+        $state[self::AUTHID] = $this->authId;
+        $state[self::SOURCESID] = $this->sources;
+
         if (!is_null($state['saml:RequestedAuthnContext']) && array_key_exists('AuthnContextClassRef', $state['saml:RequestedAuthnContext'])) {
             $refs = array_values($state['saml:RequestedAuthnContext']['AuthnContextClassRef']);
             $new_sources = [];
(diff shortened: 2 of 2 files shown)


modules/multiauth/docs/multiauth.md
```diff
@@ -77,7 +79,7 @@ compatible fashion so both cases should work.
 
 Each source in the sources array has a key and a value. As
 mentioned above the key is the authsource identifier and the value
-is another array with two optional keys: 'text' and 'css-class'.
+is another array with optional keys: 'text', 'css-class', 'help', and 'ContextClassRef'.
```
tvdijen, line 82: You're missing `Authn`

Lasering: Initially I implemented it with AuthnContextClassRef. But it seemed weird to reference Authn in the multiauth config. So I changed it to ContextClassRef. I've reverted back to AuthnContextClassRef.

Lasering:
I've set AuthnContextClassRef in multiauth as optional. If the SP defines an AuthnContextClassRef but multiauth config does not, should the NoAuthnContext be returned?

How can I return a `SAML2\XML\samlp\StatusCode(SAML2\Constants::STATUS_NO_AUTHN_CONTEXT)` from inside the authenticate function (assuming that is the right thing to return)?

tvdijen:
Hmm, this is quite tricky.. I think we need a defaultAuthnContext outside the sources-array..
If the requested context doesn't match a specific source or the default one, it should definitely return an error.
I think I'd like a second opinion by @jaimeperez to make sure we don't miss anything..

To return the proper error, I think it's enough to just `throw \SimpleSAML\Module\saml\Error\NoAuthnContext(\SAML2\Constants::STATUS_RESPONDER);`..

jaimeperez:
Thanks a lot for your contribution @Lasering!

This one is indeed tricky. I think we might need to explicitly enable this in the `multiauth` configuration, unless as Tim was suggesting, we provide a default that can be applied to those sources missing it in the configuration, and then use that as an indication that the functionality is enabled. But we definitely need a clear way to enable/disable this.

Now, I see at least a couple other issues that I think we need to address:

* Should we have a way to provide, by configuration, mappings for known `AuthnContextClassRef`s? In my head, a defined auth source could match several class refs, as those are just references to an agreed set of conditions and requirements.
* We need to be extra careful with the comparison algorithm, and decide what we support. Anything other than `exact` is quite a nightmare to implement after my own experience, but it might limit a lot the usefulness of the implementation (e.g. LoA2 with comparison `minimum`).

tvdijen:
The default comparison should be `exact` because that's what SAML2INT mandates. Others could be optional.

Lasering:
If the requested context class refs only match the default value for AuthnContextClassRef which auth method should be shown? The way I see it giving the error when 0 auth methods match is already the best approach.

1. If multiauth does not define any class refs and the SP requests some refs => the error NoAuthnContext will be thrown, which is good because it signals the IdP maintainer of the problem (we could even suggest in the error message to add class refs to the multiauth config).
2. If multiauth defines refs and the SP does not => all auth methods will be shown. Which intuitively makes sense.
3. If both define refs => the list of auth methods will be filtered. If 0 methods remain after filtering the error will be thrown.

> * Should we have a way to provide, by configuration, mappings for known `AuthnContextClassRef`s? In my head, a defined auth source could match several class refs, as those are just references to an agreed set of conditions and requirements.

This is already solved. The AuthnContextClassRef of a source in multiauth can be a string or an array of strings. This way you can specify that an authentication source corresponds to multiple class refs.

> * We need to be extra careful with the comparison algorithm, and decide what we support. Anything other than `exact` is quite a nightmare to implement after my own experience, but it might limit a lot the usefulness of the implementation (e.g. LoA2 with comparison `minimum`).

Before creating the PR I went looking for that logic, since I assumed it should exist in either simplesamlphp or in the saml2 library, but I could not find it. The spec is also unhelpful since:

> If Comparison is set to _______ then the resulting authentication context in the authentication statement MUST _______.
> 
> - `exact` or omitted - be the exact match of at least one of the authentication contexts specified.
> - `minimum` - be at least as strong (as deemed by the responder) as one of the authentication contexts specified.
> - `better` - be stronger (as deemed by the responder) than any one of the authentication contexts specified.
> - `maximum` - be as strong as possible (as deemed by the responder) without exceeding the strength of at least one of the authentication contexts specified.

1. The choice of what is at least as strong/stronger/as strong is left to the responder (the IdP in this scenario).
2. Currently the field AuthnContextClassRef in simplesamlphp is an unrestricted string.

The IdP would have to somehow assign a strongness factor (for the lack of a better description) to each class ref, and then perform the comparisons based on that factor. These assignments would need to be a configuration since 1) it is very unlikely that everyone will agree on the same assignments 2) that is the only way of having a change to deal with the unrestricted string.

We could simply declare the only supported implementation for the comparison is `exact`. The use cases supported by the current implementation are already useful enough.

tvdijen:
> Before creating the PR I went looking for that logic, since I assumed it should exist in either simplesamlphp or in the saml2 library, but I could not find it.

The spec does say something in this regard on the RequestedAuthnContext:
_The set of supplied references MUST be evaluated as an ordered set, where the first element is the most preferred authentication context class or declaration._
So we start at the top and see if we can satisfy the requirement.. If not, move our way down the list and eventually if nothing works, the NoAuthnContext-exception must be thrown.

If we continue this practise or ordering the LOA's to the IDP (make sure we add it to the documentation), We should be able to handle minimum/maximum/better..

I think first we need to determine a list of possible LOA's that fit the requirement.. For example, we have two sources:
A: [LOA2, LOA1]
B: [LOA3, LOA2]
C: [LOA4, LOA3]
D: [LOA4]

Now the SP requires 'better' than LOA1.. We loop over the sources to find that A can meet this requirement.. Now we should do another loop over the sources to figure out that B can also satisfy LOA2 (or even LOA4 or LOA3, because they're higher in the ordered list). Now that we have determined that we could meet the requirement using [LOA4, LOA3, LOA2], we should be able to figure out that source D is also suitable.

I guess if we systematically peel this off for minimum/maximum/better/exact we should be able to make this work.

jaimeperez:
> 1. If multiauth does not define any class refs and the SP requests some refs => the error NoAuthnContext will be thrown, which is good because it signals the IdP maintainer of the problem (we could even suggest in the error message to add class refs to the multiauth config).

Not sure I agree with this one, as that would break existing deployments. I think this needs to be enabled explicitly in the IdP, in a way that doesn't enforce it for existing configurations.

> 2. If multiauth defines refs and the SP does not => all auth methods will be shown. Which intuitively makes sense.

Agree.

> 3. If both define refs => the list of auth methods will be filtered. If 0 methods remain after filtering the error will be thrown.

Agree as well.

> > * Should we have a way to provide, by configuration, mappings for known `AuthnContextClassRef`s? In my head, a defined auth source could match several class refs, as those are just references to an agreed set of conditions and requirements.
> 
> This is already solved. The AuthnContextClassRef of a source in multiauth can be a string or an array of strings. This way you can specify that an authentication source corresponds to multiple class refs.

Awesome! Sorry that I replied without actually having a look at the code 😅 

> The IdP would have to somehow assign a strongness factor (for the lack of a better description) to each class ref, and then perform the comparisons based on that factor. These assignments would need to be a configuration since 1) it is very unlikely that everyone will agree on the same assignments 2) that is the only way of having a change to deal with the unrestricted string.

This can be very difficult to assess in practice, so I would agree with Tim's initial suggestion and your comment to essentially support `exact`. I'd say it would be nice to support `minimum` as well, but I wouldn't focus on it now. As long as we can define a set of different class refs for each auth source, so that we can compare each of them with whatever we receive from the SP, `exact` is fairly simple and should work in most deployments.

Question: would the _default_ class ref apply also to those auth sources explicitly specifying any class ref? I'd say no, but would like your opinions as well.

tvdijen:
That's a no on the question.. Also something that we need to explicitly document

Lasering:
> The set of supplied references MUST be evaluated as an ordered set, where the first element is the most preferred authentication context class or declaration.

In my interpretation `most preferred` has a different meaning from `at least as strong/stronger/as strong as possible`.
I don't follow your logic on LOA (does that mean Level of Assurance?). How it is related?

I've been assuming the Comparison attribute can only be unset (the same as exact) or set to **one** of the possible values. That together with the `most preferred` approach makes the implementation easy.

> Not sure I agree with this one, as that would break existing deployments. I think this needs to be enabled explicitly in the IdP, in a way that doesn't enforce it for existing configurations.

I'll need to make a test to confirm this but the existing implementation (without this PR) is incorrect, if the SP is explicitly asking for ContextClassRefs the code should return the NoAuthnContext (as per the spec), and not allow the user to pick auth methods. Existing deployments will break but they were working "outside" of the spec, so in a sense they were wrong from the start. Propagating this wrongness seems the worst option of the two.

Instead of having a flag enabling this functionality in the IdP, I suggest having a flag to enable working "outside" of the spec. Something like:
```php
/**
 * Whether the user will still be allowed to choose between the different authentication methods if an
 * SP requests AuthnContextClassRefs but the multiauth configuration does not set AuthnContextClassRefs
 * for any of the methods.
 * Setting this true, will make this IdP non-compliant with the [SAML2 Protocol](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) point 3.3.2.2.1
 */
'multiauth.always-list-auth-methods' => false,
```

Lasering:
Changed implementation to invoke delegateAuthentication directly when only one authentication source remains after filtering using the ClassRefs. Save the correct list of sources in the state.

tvdijen (changes requested):
Unfortunately the tests are failing.. Please fix the code to comply with them, or adjust them accordingly to the change.

Lasering:
@tvdijen did you implement new tests? One of the failing tests is testing the preselect which I did not touch. And the tests were passing before.

tvdijen:
I think I may have flipped the order during the rebase.. It's good again

tvdijen (approved):

dbollaer:
@tvdijen in which version of simplesamlphp will this feature be released?
@Lasering thank you!

tvdijen:
Either 2.0 or 1.20, which ever comes first..

PR #1363, opened 2020-08-06
Replace simplesaml.nameidattribute with a config setting local to this authpproc

tvdijen:


tvdijen:
Hmm, I may have to take another good look at this, because I think I messed up in `modules/saml/lib/IdP/SAML2.php` and I can't remember what I was thinking

thijskh:
I'm quite confused about the PR I'm afraid. In the authproc it seems it mixes a config option "attribute" with a config option "identifyingAttribute", both seem to be present and used. The documentation is not up to date with the changes so it's quite hard to follow. Also I'm not sure what's the deal with commented out code in the PR. So I'm a bit lost.

thijskh:
Not sure, but is not just making the existing "attribute" setting of this authproc mandatory a complete solution?

tvdijen:
Coding & drinking always seems a good idea until the review 😆
I'll get back to this later this week,. I'm away for a few days

tvdijen:
You were absolutely right about this, Thijs! For some reason I missed the existing `attribute` setting..
I've done a rewrite on this and renamed the setting to identifyingAttribute as per discussion I've had with Jaime regarding the name.

PR #1365, opened 2020-08-07
Improve error handling

### LINKED ISSUE #1002: PHP 7: Stack trace on PHP fatal error does not point to actual error
The exception handler configured for SimpleSAMLphp hides the actual location of error when handling fatal errors in PHP 7. Instead we get a stack trace pointing to the error handler:

> SimpleSAML_Error_Exception: Error 1 - Call to undefined function invalidfunction()
> Backtrace:
> 2 /www/simplesamlphp/www/_include.php:58 (SimpleSAML_error_handler)
> 1 /www/simplesamlphp/www/_include.php:26 (SimpleSAML_exception_handler)
> 0 [builtin] (N/A)

This is caused by a change in the way that errors are handled in PHP 7. Before PHP 7, PHP would directly call our [error handler](https://github.com/simplesamlphp/simplesamlphp/blob/v1.16.2/www/_include.php#L34) from the location of the error. This allowed us to capture a stack trace pointing to the location the error occurred.

After PHP 7, the error is handled like a normal exception. This means that PHP will unwind the stack looking for an exception handler, before finally passing it to the [`SimpleSAML_exception_handler()`-function](https://github.com/simplesamlphp/simplesamlphp/blob/v1.16.2/www/_include.php#L10).

This function handles errors by invoking the old error handler:
https://github.com/simplesamlphp/simplesamlphp/blob/v1.16.2/www/_include.php#L20-L27

Unfortunately, that loses the information about where the error occurred, since that is only stored in the original exception.

The best way to solve this would be to not call the original error handler, but simply wrap the error in a normal unhandled exception:

```
if (class_exists('Error') && $exception instanceof Error) {
    $e = new \SimpleSAML\Error\Error('UNHANDLEDEXCEPTION', $exception);
    $e->show();
}
```

This would log the exception like any other unhandled exception:

> SimpleSAML_Error_Error: UNHANDLEDEXCEPTION
> Backtrace:
> 1 /www/simplesamlphp/www/_include.php:20 (SimpleSAML_exception_handler)
> 0 [builtin] (N/A)
> Caused by: Error: Call to undefined function invalidfunction()
> Backtrace:
> 1 /www/simplesamlphp/modules/somemodule/www/index.php:10 (require)
> 0 /www/simplesamlphp/www/module.php:135 (N/A)

The problem is that the [`Error` class](http://php.net/manual/en/class.error.php) does not inherit from the [`Exception` class](http://php.net/manual/en/class.exception.php). Thus the [constructor for `SimpleSAML\Error\Exception`](https://github.com/simplesamlphp/simplesamlphp/blob/v1.16.2/lib/SimpleSAML/Error/Exception.php#L44) does not allow us to include it as the `$cause` parameter.

To fix that will require changing the `SimpleSAML\Error\Exception` class and other related classes to accept both a `Exception` and a `Error` object as the `$cause`-parameter.

The [`Throwable`-interface](http://php.net/manual/en/class.throwable.php) is common between `Exception` and `Error`, but that only exists in PHP 7, so we cannot use that while supporting PHP 5.

tvdijen:
Closes #1002

### DIFF
--- a/lib/SimpleSAML/Error/AuthSource.php
+++ b/lib/SimpleSAML/Error/AuthSource.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace SimpleSAML\Error;
 
 use SimpleSAML\Assert\Assert;
+use Throwable;
 
 /**
  * Baseclass for auth source exceptions.
@@ -33,9 +34,9 @@ class AuthSource extends Error
      *
      * @param string $authsource  Authsource module name from where this error was thrown.
      * @param string $reason  Description of the error.
-     * @param \Exception|null $cause
+     * @param \Throwable|null $cause
      */
-    public function __construct(string $authsource, string $reason, \Exception $cause = null)
+    public function __construct(string $authsource, string $reason, Throwable $cause = null)
     {
         $this->authsource = $authsource;
         $this->reason = $reason;
--- a/lib/SimpleSAML/Error/CriticalConfigurationError.php
+++ b/lib/SimpleSAML/Error/CriticalConfigurationError.php
@@ -7,6 +7,7 @@ namespace SimpleSAML\Error;
 use SimpleSAML\Configuration;
 use SimpleSAML\Logger;
 use SimpleSAML\Utils;
+use Throwable;
 
 /**
  * This exception represents a configuration error that we cannot recover from.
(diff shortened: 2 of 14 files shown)


tvdijen:
I don't think it does?  Exception already implements Throwable...
The issue we're trying to solve is that we couldn't pass PHP's builtin Error-class as the cause-parameter..

PR #1377, opened 2020-09-05
Configurable symmetric encryption algorithm

### LINKED ISSUE #1376: Configurable symmetric key algorithm
It's related to missing AES-GCM support (https://github.com/simplesamlphp/saml2/issues/179)

For IDP's the algorithm should be configurable in the metadata, with a sane default when not configured. SP's should be able to read the algorithm from the EncryptionMethod in the response, however, since that one is optional, it should have a fallback to a preconfigured value in the metadata config.

Not supporting AES-GCM is no longer an option and this feature should land in 1.19

tvdijen:
Closes #1376 
Closes https://github.com/simplesamlphp/saml2/issues/179

### DIFF
--- a/modules/saml/lib/IdP/SAML2.php
+++ b/modules/saml/lib/IdP/SAML2.php
@@ -1294,7 +1294,12 @@ class SAML2
 
         $sharedKey = $spMetadata->getString('sharedkey', null);
         if ($sharedKey !== null) {
-            $key = new XMLSecurityKey(XMLSecurityKey::AES128_CBC);
+            $algo = $spMetadata->getString('sharedkey_algorithm', null);
+            if ($algo === null) {
+                $algo = $idpMetadata->getString('sharedkey_algorithm');
+            }
+
+            $key = new XMLSecurityKey($algo);
             $key->loadKey($sharedKey);
         } else {
             $keys = $spMetadata->getPublicKeys('encryption', true);
--- a/modules/saml/lib/IdP/SAML2.php
+++ b/modules/saml/lib/IdP/SAML2.php
@@ -1297,7 +1297,7 @@ class SAML2
             $algo = $spMetadata->getString('sharedkey_algorithm', null);
             if ($algo === null) {
                 // If no algorithm is configured, use a sane default
-                $algo = $idpMetadata->getString('sharedkey_algorithm', XMLSecurityKey::AES128_GCM);
+                $algo = $idpMetadata->getString('sharedkey_algorithm', XMLSecurityKey::AES128_CBC);
             }
 
             $key = new XMLSecurityKey($algo);
--- a/modules/saml/lib/IdP/SAML2.php
+++ b/modules/saml/lib/IdP/SAML2.php
@@ -1296,7 +1296,8 @@ class SAML2
         if ($sharedKey !== null) {
             $algo = $spMetadata->getString('sharedkey_algorithm', null);
             if ($algo === null) {
-                $algo = $idpMetadata->getString('sharedkey_algorithm');
+                // If no algorithm is configured, use a sane default
+                $algo = $idpMetadata->getString('sharedkey_algorithm', XMLSecurityKey::AES128_GCM);
             }
 
             $key = new XMLSecurityKey($algo);
--- a/docs/simplesamlphp-reference-idp-remote.md
+++ b/docs/simplesamlphp-reference-idp-remote.md
@@ -219,8 +219,17 @@ There are two modes of encryption supported by SimpleSAMLphp. One is symmetric e
 :   Note that this option overrides the option with the same name in the SP configuration.
 
 `sharedkey`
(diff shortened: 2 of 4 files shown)


thijskh (commented):
Is it useful to add even more configuration options if the used algorithm is specified in the message itself?

tvdijen:
Actually yes, because xenc:EncryptionMethod is _optional_

> EncryptionMethod is an optional element that describes the encryption algorithm applied to the cipher data. If the element is absent, the encryption algorithm must be known by the recipient or the decryption will fail.

thijskh:
Is it theoretically optional or are there actual mainstream saml implementations that omit this?

thijskh (changes requested):

tvdijen:
I wouldn't know, but I think the algorithm should be configurable on the IDP-side anyway, because sooner or later your [hardcoded default](https://github.com/simplesamlphp/simplesamlphp/pull/1377/files#diff-4ba9db222d18133821b1a07d4d4faabaL251) becomes [vulnerable](https://www.nds.ruhr-uni-bochum.de/media/nds/veroeffentlichungen/2011/10/22/HowToBreakXMLenc.pdf) and your users end up with editing code.. I can give the Shib4 IDP a go and see if it includes this, but to me it doesn't change much.. I agree on keeping the amount of config settings to a minimum, but in this case I think people need to be able to switch on the fly in case of a vulnerability..

thijskh:
For vulnerabilities we should rather have a blacklisted algorithm facility, since then you’d want to override one but still allow multiple others.

tvdijen:
The blacklist is already in place.

modules/saml/lib/IdP/SAML2.php
```diff
@@ -1300,7 +1300,13 @@ private static function encryptAssertion(
 
         $sharedKey = $spMetadata->getString('sharedkey', null);
         if ($sharedKey !== null) {
-            $key = new XMLSecurityKey(XMLSecurityKey::AES128_CBC);
+            $algo = $spMetadata->getString('sharedkey_algorithm', null);
+            if ($algo === null) {
+                // If no algorithm is configured, use a sane default
+                $algo = $idpMetadata->getString('sharedkey_algorithm', XMLSecurityKey::AES128_GCM);
```
jaimeperez, line 1306: Uhm, this breaks backwards compatibility, right?

jaimeperez (changes requested):

jaimeperez:
When decrypting messages it surely is desirable to use whatever was specified in the message. That is tricky though, not only because we're not using version 5 of the SAML2 library which implements support for that, but also because we need to make sure we don't introduce an issue that would allow for the use of a public key as a shared, symmetric key.

In any case, when encrypting messages we don't have a way to tell which algorithm we should use, and for that I'm afraid we have no other choice for now than a (yet another) config option.

The only thing in this PR that keeps me from accepting it is that it breaks backwards compatibility, making existing configurations default now to AES-GCM, and I think that shouldn't be the case. Other than that, I'm quite happy with the implementation 😄

modules/saml/lib/IdP/SAML2.php
tvdijen: I guess I did this because the old default is weak/broken... It's our legacy of not producing a new major in over a decade..
We have done these kind of BC breaks before and as long as we add proper upgrade notes I don't see an issue..
This PR is targeted at `master`, so it shouldn't be a problem.. The backport to 1.19 can be different is you like

tvdijen: @jaimeperez Please get back to me on this so I can merge this

tvdijen: @jaimeperez Pretty please?

tvdijen:
@thijskh Any objections against merging this after the above discussion?

modules/saml/lib/IdP/SAML2.php
thijskh: For 1.19 please default to what it was before. This is the algo originally supported by world+dog

thijskh (approved):
Ok to merge to master if documented in release notes. Keep defaulting to the classic saml2 algo for 1.19.

modules/saml/lib/IdP/SAML2.php
jaimeperez: Sorry so much @tvdijen! For some reason my browser logged me off github and I never got notifications about this 😞 

I agree with Thijs, we should keep what we had before. The main reason for me is that GCM support in some SAML implementations is not that common yet, so defaulting to it could potentially break interoperability with other products. For 2.0 it's totally fine to change it.

jaimeperez (approved):

tvdijen:
Reverted algo for now and backport to 1.19..
Will change algo back to AES-GCM in `master` as a new default and add notes..

thijskh (approved):

PR #1378, opened 2020-09-10
Update LDAP.php

grueneedv:
Minor typo. Frist replaced everything in my project and then saw the typo :|


jaimeperez:
Nice catch! Thanks @grueneedv!

Issue #1379, opened 2020-09-11
Incorrect MODIFY syntax for SQLServer in LogoutStore.php


Krisell:
**Describe the bug**
The correct SQLServer syntax is 
```sql
ALTER TABLE table_name ALTER COLUMN column_name datatype;
```
but the current syntax (the default case) is
```sql
ALTER TABLE table_name MODIFY COLUMN column_name datatype;
```

This happens in the switch-statement in two places, and it looks like the `sqlsrv` could be handled in the same way as `pgsql`:
https://github.com/simplesamlphp/simplesamlphp/blob/3d7d094d01bdf5c2b02ddc3ad2016027e0bec2ea/modules/saml/lib/SP/LogoutStore.php#L39
https://github.com/simplesamlphp/simplesamlphp/blob/3d7d094d01bdf5c2b02ddc3ad2016027e0bec2ea/modules/saml/lib/SP/LogoutStore.php#L116

Documentation for the SQLServer syntax:
https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver15

I have not been able to test or PR this, but might be able to so within a few weeks if no-one else beats me to it (please do).

tvdijen:
I can do the change, but if you'd be so kind to test it for us then? I don't have MSSQL available for testing

tvdijen:
This one probably needs the same change then;
https://github.com/simplesamlphp/simplesamlphp/blob/3d7d094d01bdf5c2b02ddc3ad2016027e0bec2ea/modules/saml/lib/SP/LogoutStore.php#L119

Krisell:
Sounds great, and yes I should be able to test this.
If the change is to simply use the same syntax as `pgsql`, I can test this right away and send in a PR, but there might be other changes that have to be made that I'm not seeing?

Krisell:
The`sqlsrv` syntax is not identical to `pgsql`, so I started to add a `sqlsrv` case to the switch statements. I was unable to try this though, as it turns out that the database user used in our application only has read/write-access, and not alter-access. I'm wondering if this is unusual or of the automatic table-migration might fail for more people (it only throws a warning and keeps using the old schema, but tries to migrate on every request).

I'm hoping to find the time to create a more consistent Docker-setup to test SQLServer. I'm guessing there are no automated tests against different database drivers in this project?

[...]

PR #1380, opened 2020-09-11
Fix SQLServer ALTER syntax

Krisell:
Fixes https://github.com/simplesamlphp/simplesamlphp/issues/1379

This fixes the incorrect `ALTER ... MODIFY` syntax for SQLServer (it's `ALTER ... ALTER`) which prevents the `LogoutStore` to migrate from version `1 -> 2` and from version `3 -> 4`. Previously, migrating the versions failed and the old version was kept. This caused a warning to be logged but no other effect from what I could see.

image

I used https://github.com/simplesamlphp/docker-simplesamlphp along with a SQLServer image to verify that this fixes the problem, and I was able to successfully migrate the LogoutStore from version `1 -> 2` and from version `3 -> 4` (`2 -> 3` worked before this fix).

### DIFF
--- a/modules/saml/lib/SP/LogoutStore.php
+++ b/modules/saml/lib/SP/LogoutStore.php
@@ -43,6 +43,11 @@ class LogoutStore
                         'ALTER TABLE ' . $store->prefix . '_saml_LogoutStore ALTER COLUMN _expire TYPE TIMESTAMP'
                     ];
                     break;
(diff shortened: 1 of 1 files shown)


tvdijen (approved):

tvdijen:
Note to self: Backport to 1.19 while we still can

PR #1382, opened 2020-09-25
Allow easy enabling of SameSite = 'None'

pradtke:
Set intelligent defaults for SameSite cookie automatically so new installers don't need to think.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -792,7 +792,7 @@ $config = [
     'language.cookie.secure' => false,
     'language.cookie.httponly' => false,
     'language.cookie.lifetime' => (60 * 60 * 24 * 900),
-    'language.cookie.samesite' => null,
+    'language.cookie.samesite' => \SimpleSAML\Utils\HTTP::canSetSameSiteNone() ? 'None' : null,
 
     /**
      * Custom getLanguage function called from SimpleSAML\Locale\Language::getLanguage().
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -564,7 +564,7 @@ $config = [
      * Example:
      *  'session.cookie.samesite' => 'None',
      */
-    'session.cookie.samesite' => null,
+    'session.cookie.samesite' => \SimpleSAML\Utils\HTTP::canSetSameSiteNone() ? 'None' : null,
 
     /*
      * Options to override the default settings for php sessions.
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -19,6 +19,43 @@ use SimpleSAML\XHTML\Template;
 class HTTP
 {
     /**
+     * Determine if the user agent can support cookies being sent with SameSite equal to "None".
+     * Browsers with out support may drop the cookie and or treat is a stricter setting
+     * Browsers with support may have additional requirements on setting it on non-secure websites.
+     *
+     * Based on the Azure teams experience rolling out support
+     * https://devblogs.microsoft.com/aspnet/upcoming-samesite-cookie-changes-in-asp-net-and-asp-net-core/
+     * @return bool true if user agent supports a None value for SameSite.
+     */
+    public static function canSetSameSiteNone(): bool
+    {
+        $useragent = $_SERVER['HTTP_USER_AGENT'] ?? null;
+        if (!$useragent) {
+            return true;
+        }
+        // All iOS 12 based browsers
+        if (strpos($useragent, "CPU iPhone OS 12") !== false || strpos($useragent, "iPad; CPU OS 12") !== false) {
+            return false;
+        }
+
+        // Safari Mac OS X 10.14
+        // - Safari on Mac OS X.
+        if (
+            strpos($useragent, "Macintosh; Intel Mac OS X 10_14") !== false &&
+            strpos($useragent, "Version/") !== false &&
+            strpos($useragent, "Safari") !== false
+        ) {
+            return false;
+        }
+
+        // Chrome 50-69 may have broken SameSite=None and don't require it to be set
+        if (strpos($useragent, "Chrome/5") !== false || strpos($useragent, "Chrome/6") !== false) {
+            return false;
+        }
+        return true;
+    }
+
+    /**
      * Obtain a URL where we can redirect to securely post a form with the given data to a specific destination.
      *
      * @param string $destination The destination URL.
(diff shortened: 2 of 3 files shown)


quartje:
Good work, you seem to have covered almost all browser exceptions, but two seem to be missing yet:
Versions of UC Browser on Android prior to version 12.13.2
embedded browsers on MacOS 10.14 

The complete list can be found here (with some pseudo code):
https://www.chromium.org/updates/same-site/incompatible-clients

tvdijen:
Hi Patrick,

In an earlier thread we decided to go with the 'double cookie' approach instead of keeping a list of (in)compatible user agents.
I must admit I forgot about the same-site problems completely, but I don't think it's too late to get the double-cookie fix into 1.19..

pradtke:
@quartje 👍  I used the rules that Azure team deployed (https://devblogs.microsoft.com/aspnet/upcoming-samesite-cookie-changes-in-asp-net-and-asp-net-core/) since it was based on their testing (which may not have been good enough).  UC Browser user agent is Chrome 57 so would fall into the rule for not having it set, however embedded Safari on Mac OS X probably isn't handled correctly. Unfortunately I haven't found a list of compatible and incompatible user agent strings to drive unit tests. Anyhow, it is probably worth adjusting the rules to be more specific like you suggest if user agent parsing is the way SSP project wants to go.

@tvdijen I didn't see the `double cookie` code in `master`. Could you point me to the commit? I think we should do *something* so that SSP 1.19 installs don't require installers to worry about this.  If you think putting the double cookie into 1.19 is too heavy of a lift, then I'll make adjustments to this PR per @quartje suggestions.  If we don't include something pre-enabled then I feel we'll spend a lot of time explaining how to do the config on the mailing list.

cicnavi:
Hi,
just to share our experience with SameSite attribute. Recently we updated SSP to 1.18. and set the 'session.cookie.samesite' to 'None'. We got complaints from Apple users who couldn't authn anymore... (iOS / Mac with buggy web browsers).
To cope with that, we will implement the canSetSameSiteNone() method and solution that @pradtke proposed, in our SSP instance (until you guys implement some sort of dynamic SameSite handling...). 
However, we will set the rules based on pseudocode available at https://www.chromium.org/updates/same-site/incompatible-clients

Here is our code, in case you are interested:
```php
    /**
     * Determine if the user agent can support cookies being sent with SameSite equal to "None".
     * Browsers with out support may drop the cookie and or treat is a stricter setting
     * Browsers with support may have additional requirements on setting it on non-secure websites.
     *
     * Solution based on https://github.com/simplesamlphp/simplesamlphp/pull/1382/commits/fb01fc41e99daffcad0e419455c8b7a684fef623
     * Rules based on https://www.chromium.org/updates/same-site/incompatible-clients
     *
     * @return bool true if user agent supports a None value for SameSite.
     */
    public static function canSetSameSiteNone()
    {
        $useragent = isset($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : null;

        if (!$useragent) {
            return true;
        }

        /**
         * @var array contains regex matches
         */
        $matches = [];

        // hasWebKitSameSiteBug(useragent) ||
        //           dropsUnrecognizedSameSiteCookies(useragent)

        // hasWebKitSameSiteBug(useragent):
        //      isIosVersion(major:12, useragent) ||
        //           (isMacosxVersion(major:10, minor:14, useragent) &&
        //            ((isSafari(useragent) && !isChromiumBased(useragent)) || isMacEmbeddedBrowser(useragent))

        // isIosVersion(major:12, useragent)
        if (preg_match("/\(iP.+; CPU .*OS (\d+)[_\d]*.*\) AppleWebKit\//", $useragent, $matches)) {
            if ($matches[1] == '12') {
                return false;
            }
        }
        // ... || (isMacosxVersion(major:10, minor:14, useragent)
        if (preg_match("/\(Macintosh;.*Mac OS X (\d+)_(\d+)[_\d]*.*\) AppleWebKit\//", $useragent, $matches)) {
            if ($matches[1] == '10' && $matches[2] == '14') {
                // ... && ((isSafari(useragent) && !isChromiumBased(useragent)) || isMacEmbeddedBrowser(useragent))
                if (
                    (
                        preg_match("/Version\/.* Safari\//", $useragent) &&
                        !preg_match("/Chrom(e|ium)/", $useragent)
                    ) ||
                    preg_match("/^Mozilla\/[\.\d]+ \(Macintosh;.*Mac OS X [_\d]+\) AppleWebKit\/[\.\d]+ \(KHTML, like Gecko\)$/", $useragent)
                ) {
                    return false;
                }
            }
        }

        // dropsUnrecognizedSameSiteCookies(string useragent):
        //    if isUcBrowser(useragent):
        //        return !isUcBrowserVersionAtLeast(major:12, minor:13, build:2, useragent)
        //    return isChromiumBased(useragent) &&
        //           isChromiumVersionAtLeast(major:51, useragent) &&
        //           !isChromiumVersionAtLeast(major:67, useragent)

        // if isUcBrowser(useragent)
        if (preg_match("/UCBrowser\/(\d+)\.(\d+)\.(\d+)[\.\d]* /", $useragent, $matches)) {
            // return !isUcBrowserVersionAtLeast(major:12, minor:13, build:2, useragent)
            $majorVersion = intval($matches[1]);
            $minorVersion = intval($matches[2]);
            $buildVersion = intval($matches[3]);

            if ($majorVersion != 12) {
                return $majorVersion > 12;
            }
            if ($minorVersion != 13) {
                return $minorVersion > 13;
            }

            return $buildVersion >= 2;
        }

        // isChromiumBased(useragent)
        if (preg_match("/Chrom(e|ium)/", $useragent)) {
            // isChromiumVersionAtLeast(major:51, useragent) &&
            //      !isChromiumVersionAtLeast(major:67, useragent)
            if (preg_match("/Chrom[^ \/]+\/(\d+)[\.\d]* /", $useragent, $matches)) {
                $majorVersion = intval($matches[1]);
                if ($majorVersion >= 51 && $majorVersion < 67) {
                    return false;
                }
            }
        }

        return true;
    }
```

pradtke:
I have added test cases for different versions of UC Browser and embedded versions of Safari on Mac, and adjusted the logic as needed.

thijskh (approved):

thijskh:
Good work @pradtke .

@tvdijen Indeed there's also an approach with double cookies. However we have working code here that includes unit tests even. And I can add that we've been using this approach for 1,5 years now with millions of authentications with great success.

So not to say there's no merit in the double cookie approach but it does not yet exist, this code does and we have reasonable belief that it works. I prefer working code over possible future improvements.

My idea is to merge this. If/when someone is ready to create and test the double cookie PR, we can review that and see if it can supplant this code.

tvdijen:
I agree.. Should probably add this to 1.19-RC2 because I think this can no longer wait

PR #1387, opened 2020-10-19
Removed line in last commit

zvoman:
Removed line when removing @return void in last commit


tvdijen:
I can't believe I've missed that for over a month!! Thanks @zvoman !!

Issue #371, opened 2016-04-21
PSR-3 compliant logger


jaimeperez:
We should make our logger [PSR-3](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md) compliant, and make all the classes logging something to use this. This is especially useful for unit testing because we could get rid of the static methods, and also for the SAML2 library, which could use our logger or some other instead.

In general, the steps would be:

- [ ] Add a `$context` parameter to all the methods defined by `Psr\Log\LoggerInterface`.
- [ ] Make the `\SimpleSAML\Logger::log()` method public and make it throw a `\Psr\Log\InvalidArgumentException` if the log level is unknown.
- [ ] Extract the log levels definition from `\SimpleSAML\Logger` and use the ones defined by `Psr\Log\LogLevel` instead.
- [ ] Change the formatting that we have today. This means going from `%option` to `{option}`. We also need to think how to handle dates, since those have sub-formats using the `strftime()` date and time formatting rules. We would probably need to split the log format and the format for dates, and handle replacements manually.
- [ ] Make all the classes logging something to implement the `Psr\Log\LoggerAwareInterface` interface. That is, adding a `setLogger()` method, and using the internal logger object to log the messages.
- [ ] Make all the logging methods defined in `Psr\Log\LoggerInterface` non-static (they are static today). This means we need to add a constructor for the Logger. A *Singleton* approach would probably be the way to go, with a static `getInstance()` method to retrieve a logger that can be passed to `setLogger()`. This also cleans the logger initialization we have today, and would allow us to get rid of the logging handlers and have just loggers that extend the basic logging class.
- [ ] Make the `\SimpleSAML\Logger` class extend `Psr\Log\AbstractLogger`.
- [ ] Exceptions should also be logger-aware, and pass themselves to the logger object when logging something in the context.

jaimeperez:
The new logging interface should be ready for 1.15, so that we can deprecate the old static methods and removed them for 2.0.

jaimeperez:
I don't think we've done any work on this, no 😞

juliardi:
I have a little suggestion for this issue. Instead of restructuring the logging library, why don't we use the already available solution ?
https://github.com/Seldaek/monolog IMO is a good solution because it is PSR-3 compliant and has many handler and formatter.

PR #1396, opened 2020-11-10
Replace LogLevel ints with strings

tvdijen:
This checks one of the boxes in #371:
- [ ] Extract the log levels definition from \SimpleSAML\Logger and use the ones defined by Psr\Log\LogLevel instead.


tvdijen:
Messed it up

PR #1402, opened 2020-11-17
Fix logic leading to negative expiry

ghalse:
The frontpage_federation template in 1.19 displays expiry dates in the future with a negative expiry:

`European University Cyprus (expired -95.59448483 hours ago)`

This has caused more than one brief moment of panic for metadata that still has a long way to go.

And since we're here, it also uses a precision such that we're measuring microseconds.

### DIFF
--- a/modules/core/templates/frontpage_federation.twig
+++ b/modules/core/templates/frontpage_federation.twig
@@ -48,9 +48,9 @@
             {% endif -%}</a>
 
             {%- if entity.expire is defined %}
-            {% if entity.expire > date().timestamp %}
-            <span class="entity-expired"> (expired {{ ((date().timestamp - entity.expire) / 3600) }} hours ago)</span>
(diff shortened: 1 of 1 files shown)


tvdijen:
Thanks @ghalse ! Would you mind patching the new frontpage in the admin module as well?

tvdijen:
Let me rephrase that question, because maybe that one is correct...
So, could you verify that the representation in `module.php/admin/federation` is correct and doesn't yield the same bug?

ghalse:
> Let me rephrase that question:
> Could you verify that the representation in `module.php/admin/federation` is correct and doesn't yield the same bug?

In the 1.19 branch it seems to be fine.

tvdijen:
Thanks for verifying this.. I will merge this PR then

tvdijen (approved):

PR #1403, opened 2020-11-21
Restore symfony 5 support

### LINKED ISSUE #1401: Symfony 5 support in 1.19 is gone
**Describe the bug**
Branch 1.19 used to have working symfony 5 support via 575dacf850454876ae4b51dd1705264c74ad8de7 but it got (most likely accidentally) removed with d5081ec2984bed5b4fb58c8a3a493564c0fb93cb.

**To Reproduce**
1. Have an symfony 5 app tracking SSP 1.19
2. composer up
3. `simplesamlphp/simplesamlphp dev-simplesamlphp-1.19 requires symfony/cache ^4.4 -> no matching package found.`

**Additional context**
SF5 support added in https://github.com/simplesamlphp/simplesamlphp/commit/575dacf850454876ae4b51dd1705264c74ad8de7
SF5 support removed in https://github.com/simplesamlphp/simplesamlphp/commit/d5081ec2984bed5b4fb58c8a3a493564c0fb93cb

htto:
Fixes #1401


htto:
Unfortunately there's one broken test for me locally with debian 10
```
1) SimpleSAML\Test\ModuleTest::testResolveClass
Exception: Could not resolve 'cron:Cron': no class named 'SimpleSAML\Module\cron\Cron' or 'sspmod_cron_Cron'.
```
But it's also broken without the changes.

tvdijen:
Funny that we don't see this broken test in Travis..  I can see why it's failing though, because the cron-module isn't enabled by default.

htto:
Yeah, I was wondering too and had to recheck without the patch.

Btw. any guidance on how you handle the lockfile? I wasn't sure and thus didn't commit it's changes.

tvdijen:
We need someone to run a `composer update` on a PHP 7.1 box to generate it.

tvdijen (approved):

htto:
> We need someone to run a `composer update` on a PHP 7.1 box to generate it.

OK. Would overriding composer's platform config with php: 7.1.x be enough for that?

tvdijen:
I remember either @thijskh or @jaimeperez having objections to that procedure...  Or am I mistaking guys?

thijskh:
I have no specific opinion on that matter

jaimeperez:
Ignoring the platform details would land us on the latest versions available of the dependencies, which might not be running on PHP 7.1 (I think 7.1 is safe, but just in case). The only way to make sure the dependency graph generated (and locked in the lock file) will work on the lowest possible PHP version is to run composer on a PHP 7.1 installation. The SSP docker image can be used for that (that what I've done the last times).

tvdijen:
@jaimeperez No, we can set the platform reqs to whatever PHP version we like.. I don't mind doing it your way, if you don't mind running the docker because I'm not familiar with that stuff..

htto:
Ok, regenerated the composer .lock with php 7.1.33.

Also I couldn't resist to compare the result with php 7.4.13 constrained to php-7.1.33 via config.json:
```
 $ diff -u composer.lock*
--- composer.lock	2020-11-27 00:00:04.800115599 +0100
+++ composer.lock.fake	2020-11-26 23:58:33.377283944 +0100
@@ -6834,5 +6834,8 @@
     "platform-dev": {
         "ext-curl": "*"
     },
+    "platform-overrides": {
+        "php": "7.1.31"
+    },
     "plugin-api-version": "1.1.0"
 }
```

PR #1409, opened 2020-12-09
Remove trailing space in AuthState value

sprytnyserek:
MultiAuth source selection form has a hidden field AuthState to keep a session handler when user selects the actual authentication source.

The form is created based on the template at `modules/multiauth/templates/selectsource.twig`. AuthState value is subsituted to the template via *authstate* parameter. Apparently there's a trailing space in the value of the target field in the template what propagates to the final form and request, and then may effectively prevent restoration of the session.


thijskh (approved):
Good catch!

tvdijen:
Cute how these bugs can stay hidden for three years.. Great catch!

PR #1412, opened 2020-12-15
Allow EntityExtensions in IdP metadata

melanger:
This PR adds the possibility to include `EntityExtensions` in the IdP generated metadata, for example to be able to insert a `eduidmd:RepublishRequest` for eduGAIN.

Example `saml20-idp-hosted.php` with patch applied:
```
$dom = \SAML2\DOMDocumentFactory::create();
$republishRequest = $dom->createElementNS('http://eduid.cz/schema/metadata/1.0', 'eduidmd:RepublishRequest');
$republishTarget = $dom->createElementNS('http://eduid.cz/schema/metadata/1.0', 'eduidmd:RepublishTarget', 'http://edugain.org/');
$republishRequest->appendChild($republishTarget);
$ext = [new \SAML2\XML\Chunk($republishRequest)];

$metadata['https://example.org/my/idp/entityid'] = [
    // ...
    'EntityExtensions' => $ext,
    // ...
];
```

I did not find any way to achieve this without modifying SimpleSAMLphp.


tvdijen:
I think this needs a proper implementation of the extension in the saml2-library and then we can implement it the same way we do for MDUI/MDRPI.. You have a link to the specifications of this extensions?

melanger:
`<Extensions>` are allowed in multiple places, but this PR specifically 
concerns `<Extensions>` element as a direct child of 
`<EntityDescriptor>` in the IdP metadata, which is defined in the [SAML 
2.0 metadata 
spec](https://www.oasis-open.org/committees/download.php/56786/sstc-saml-metadata-errata-2.0-wd-05-diff.pdf) 
- section 2.3.2.

```
<Extensions>[Optional]
This contains optional metadata extensions that are agreed upon between 
a metadata publisher and consumer. Extension elements MUST be 
namespace-qualified by a non-SAML-defined namespace.
```

melanger:
`<Extensions>` are allowed in multiple places, but this PR specifically concerns `<Extensions>` element as a direct child of `<EntityDescriptor>` in the IdP metadata, which is defined in the [SAML 2.0 metadata spec](https://www.oasis-open.org/committees/download.php/56786/sstc-saml-metadata-errata-2.0-wd-05-diff.pdf) - section 2.3.2.

```
<Extensions>[Optional]
This contains optional metadata extensions that are agreed upon between 
a metadata publisher and consumer. Extension elements MUST be 
namespace-qualified by a non-SAML-defined namespace.
```

tvdijen:
I know that, but I meant the specs for the `edumd` elements..

melanger:
There may be many kinds of extensions - they are specific to a metadata consumer. For example, the Czech NREN eduID.cz defined an extension like this, which expresses the request of an IdP (or an SP) to be published in eduGAIN (not only in the Czech federation):
```
<eduidmd:RepublishRequest xmlns:eduidmd="http://eduid.cz/schema/metadata/1.0">
      <eduidmd:RepublishTarget>http://edugain.org/</eduidmd:RepublishTarget>
</eduidmd:RepublishRequest>
```
and this specific metadata extension can be achieved with the PR merged and the code in https://github.com/simplesamlphp/simplesamlphp/pull/1412#issue-540337143

I don't know about any similar extensions, that's why the code in the PR just takes any XML (array of `\SAML2\XML\Chunk`s). It could also just accept a string with the raw XML.

tvdijen (changes requested):
I think I'd be more comfortable with `saml:Extensions`, similar to the saml:SP reference...
Other than that I'm OK with this change, although I'm still interested in the exact specs on this extension

melanger:
> I think I'd be more comfortable with `saml:Extensions`, similar to the saml:SP reference...

OK, I'll change the configuration entry name.

> Other than that I'm OK with this change, although I'm still interested in the exact specs on this extension

This PR allows any extension. The `<Extensions>` element is defined in the [SAML 2.0 metadata spec](https://www.oasis-open.org/committees/download.php/56786/sstc-saml-metadata-errata-2.0-wd-05-diff.pdf) - section 2.3.2.
I just included an example in the PR comment to show what may be achieved after this is merged.

The specific extension for EduID.cz, which I want to achieve via config after this PR is merged, is `RepublishRequest` (`http://eduid.cz/schema/metadata/1.0`). It does not have a dedicated page, but it is described [here](https://www.eduid.cz/cs/tech/edugain) and it probably does not need a more specific definition since it is literally just

```xml
<Extensions xmlns="urn:oasis:names:tc:SAML:2.0:metadata"> <!-- part of SAML2 standard -->
    <eduidmd:RepublishRequest xmlns:eduidmd="http://eduid.cz/schema/metadata/1.0"> <!-- the extension -->
      <eduidmd:RepublishTarget>http://edugain.org/</eduidmd:RepublishTarget> <!-- no other RepublishTargets are supported -->
    </eduidmd:RepublishRequest>
</Extensions>
```

thijskh (changes requested):
Although it's not beautiful that you need to construct an XML Chunk yourself, I can see that per definition Extensions need to be flexible and can be highly environment specific. So this will provide support for any kind of Extension required at this point which I think is good for flexibility which is what you  need with Extensions.

I do wonder if we could create some kind of "createExtensionElement" helper which makes it a bit less unwieldy. But this helper would need to be generic. Maybe just a thought or a later improvement, does not block this change.

There are some things that need to be improved though:
- Documentation. We should not add hidden features, the idp hosted reference should include this option and how to use it (maybe include the given code example even).
- Tests. Especially important for rarer used functionality like this. A test could be added to `tests/lib/SimpleSAML/Metadata/SAMLBuilderTest.php`

melanger:
> I do wonder if we could create some kind of "createExtensionElement" helper which makes it a bit less unwieldy. But this helper would need to be generic. Maybe just a thought or a later improvement, does not block this change.

@thijskh That is a good suggestion, where would this `createExtensionElement` fit? As a public static method on a class? If so, which one - `SimpleSAML\Metadata\SAMLBuilder`?

PR #1419, opened 2021-01-11
Do not allow the password hash to be used for authentication.

### LINKED ISSUE #1418: If auth.adminpassword contains a hash, the hash itself can be used to login.
Not sure if this is an actual bug or a misunderstanding on my part as to why someone would hash the password in the `config.php` file...

**Describe the bug**
If a hashed password is used in the `auth.adminpassword` value in  `config.php`, both the original password and the hash itself can be used to login.

**To Reproduce**
1. Generate a password hash with `"bin/pwgen.php"`.
For exemple, `password` is hashed as `$2y$10$ozHG4CXjZ07mN4uIu4g7NeLvmE257oi6L/DX3IbaJVR5Xay/L4iAC`.
2. Enter this hash as the value for `auth.adminpassword` in the `\config\config.php` file.
3. Also set the `admin.protectindexpage` value to `true`.
4. Go to the SimpleSAMLphp installation page.
5. You can log in with either the original password (`password`, as expected) or the hash itself (`$2y$10$ozHG4CXjZ07mN4uIu4g7NeLvmE257oi6L/DX3IbaJVR5Xay/L4iAC`).

**Expected behavior**
The hashed string itself should not allow access.
As it is, there is no difference in putting the plain text password or the hash in the `config.php` file.

thijskh:
Fixes #1418

### DIFF
--- a/lib/SimpleSAML/Utils/Crypto.php
+++ b/lib/SimpleSAML/Utils/Crypto.php
@@ -356,6 +356,9 @@ class Crypto
      */
     public static function pwValid(string $hash, string $password): bool
     {
+        if (!is_null(password_get_info($password)['algo'])) {
+            throw new Error\Exception("Cannot use a hash value for authentication.");
+        }
         if (password_verify($password, $hash)) {
             return true;
         }
--- a/tests/lib/SimpleSAML/Utils/CryptoTest.php
+++ b/tests/lib/SimpleSAML/Utils/CryptoTest.php
@@ -165,6 +165,18 @@ PHP;
         $this->assertFalse($res);
     }
 
+    /**
+     * Check that hash cannot be used to authenticate ith.
+     */
+    public function testHashAsPwInvalid(): void
+    {
+        $pw = "password";
+
+        $hash = Crypto::pwHash($pw);
+        $this->expectException(Error\Exception::class);
+        $res = Crypto::pwValid($hash, $hash);
+    }
+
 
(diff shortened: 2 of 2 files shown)


lib/SimpleSAML/Utils/Crypto.php
```diff
@@ -356,6 +356,9 @@ public static function secureCompare(string $known, string $user): bool
      */
     public static function pwValid(string $hash, string $password): bool
     {
+        if (!is_null(password_get_info($password)['algo'])) {
+            throw new Error\Exception("Cannot use a hash value for authentication.");
+        }
         if (password_verify($password, $hash)) {
```
tvdijen, line 362: Wouldn't it be sufficient to reduce this entire method to `return password_verify($password, $hash);` ?

tvdijen: Ah, I had missed your comment in the issue... So this fix would make sense in 1.19, but for master I think it should be reduced to the oneliner above

thijskh: That depends if we indeed want to force users to generate a hash for this (so disallow plain text). It does make initial setup quite a bit less straightforward.

thijskh: I will merge this because it improves the situation and we can discuss making the plain text version obsolete in a separate PR.

PR #1424, opened 2021-01-20
Ignore invalid hook names

pradtke:
Sometimes we patch or edit a hook to meet our specific needs.
Somtimes this leaves behind a file like FILENAME.orig. The hook
detection login still runs these other files, making for confusing
debugging issues. Use a stricter match on what makes a valid hook name
to avoid these issues.

### DIFF
--- a/lib/SimpleSAML/Module.php
+++ b/lib/SimpleSAML/Module.php
@@ -490,7 +490,7 @@ class Module
                 continue;
             }
 
-            if (!preg_match('/hook_(\w+)\.php/', $file, $matches)) {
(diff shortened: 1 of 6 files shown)


tvdijen (approved):

tvdijen:
Eligable for backport to 1.19

thijskh (approved):
Just wondering if we should remove the unittest module in bin/build-release.sh?

tvdijen:
That would mean someone cannot run this one unit test when using the tarball... That doesn't make much sense to me..
Either we remove the entire tests-directory from the build, or we leave it as is..

PR #1426, opened 2021-01-21
Typed class properties

tvdijen:


tvdijen:
I don't expect anyone to review this, so I've ran tests with both an SP and an IDP and nothing breaks.. Therefore merging this

PR #1435, opened 2021-03-10
Add authproc-filters to generate Subject-id & Pairwise-id

tvdijen:


modules/core/lib/Auth/Process/PairwiseID.php
```diff
@@ -0,0 +1,134 @@
...
+        Assert::keyExists(
+            $state['Attributes'],
+            $this->identifyingAttribute,
+            sprintf(
+                "core:PairwiseID: Missing attribute '%s', which is needed to generate the Pairwise ID.",
+                $this->identifyingAttribute
+            )
+        );
+
+        $userID = $state['Attributes'][$this->identifyingAttribute][0];
```
thijskh, line 100: We do need to check that this is a non-empty string to prevent a pitfall we've seen in the past with epTIDs.

modules/core/lib/Auth/Process/PairwiseID.php
```diff
@@ -0,0 +1,134 @@
...
+    {
+        parent::__construct($config, $reserved);
+
+        Assert::keyExists($config, 'identifyingAttribute', "Missing mandatory 'identifyingAttribute' config setting.");
+        Assert::keyExists($config, 'scope', "Missing mandatory 'scope' config setting.");
+        Assert::stringNotEmpty($config['identifyingAttribute']);
+        Assert::regex(
+            $config['scope'],
+            '/^[a-zA-Z0-9.-]+$/i',
+            'PairwiseID: \'scope\' contains illegal characters.'
```
thijskh, line 75: If we want to be even more sure we do not release anything spec incompliant we change the regex to `/^[a-z0-9][a-z0-9.-]{0,126}$/i` (change the exception message to match). This might be good to do since one problem with the existing identifiers is the amount of variance between implementations. So be strict in what we can send.

modules/core/lib/Auth/Process/PairwiseID.php
```diff
@@ -0,0 +1,134 @@
...
+        if (!empty($state['saml:RequesterID'])) {
+            // Proxied request - use actual SP entity ID
+            $sp_entityid = $state['saml:RequesterID'][0];
+        } else {
+            $sp_entityid = $state['core:SP'];
+        }
+
+        // Calculate hash
+        $salt = $this->configUtils::getSecretSalt();
+        $hash = hash('sha256', $salt . '|' . $userID . '|' . $idp_entityid . '|' . $sp_entityid, false);
```
thijskh, line 118: I wonder if we should add the idp_entityid into the hash. This obviously means any identifier will change when the idp entityid changes.

I think we can and/or should assume (and document) that the identifying attribute must be unique for the given scope (not per se for the given idp). So that keeping the scope the same will keep the identifiers the same.

modules/core/lib/Auth/Process/SubjectID.php
```diff
@@ -0,0 +1,102 @@
...
+    public function __construct(array &$config, $reserved)
+    {
+        parent::__construct($config, $reserved);
+
+        Assert::keyExists($config, 'identifyingAttribute', "Missing mandatory 'identifyingAttribute' config setting.");
+        Assert::keyExists($config, 'scope', "Missing mandatory 'scope' config setting.");
+        Assert::stringNotEmpty($config['identifyingAttribute']);
+        Assert::regex(
+            $config['scope'],
+            '/^[a-zA-Z0-9.-]+$/i',
```
thijskh, line 66: As above. Makes one wonder if we should factor this duplicated code out to a shared place. And/or combine the two attributes into a single AuthProc that you call with the type of identifier you want.

modules/core/lib/Auth/Process/SubjectID.php
```diff
@@ -0,0 +1,102 @@
...
+        Assert::keyExists(
+            $state['Attributes'],
+            $this->identifyingAttribute,
+            sprintf(
+                "core:SubjectID: Missing attribute '%s', which is needed to generate the subject ID.",
+                $this->identifyingAttribute
+            )
+        );
+
+        $userID = $state['Attributes'][$this->identifyingAttribute][0];
```
thijskh, line 92: As above, need to ensure it's a nonempty string.

modules/core/lib/Auth/Process/SubjectID.php
```diff
@@ -0,0 +1,102 @@
...
+        Assert::regex(
+            $userID,
+            '/^[a-zA-Z0-9=-]+$/i',
```
thijskh, line 95: As above, should be more strict in what we send and change to `/^[a-z0-9][a-z0-9=-]{0,126}$/i`

tests/modules/core/lib/Auth/Process/PairwiseIDTest.php
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Test the most basic functionality on proxied request
+     */
+    public function testBasicProxiedRequest()
+    {
+        $config = ['identifyingAttribute' => 'uid', 'scope' => 'ex-ample.org'];
+        $request = [
+            'Attributes' => ['uid' => ['u=se-r2']],
+            'IdPMetadata' => ['entityid' => 'urn:idp'],
+            'saml:RequesterID' => [0 => 'urn:sp'],
```
thijskh, line 93: Maybe slightly improve the chance that this test catches issues by adding multiple requesterid's so we guarantee the behaviour that the first one is used.

tests/modules/core/lib/Auth/Process/PairwiseIDTest.php
```diff
@@ -0,0 +1,137 @@
...
+            'core:SP' => 'urn:sp',
+        ];
+        $result = self::processFilter($config, $request);
+        $attributes = $result['Attributes'];
+        $this->assertArrayHasKey(Constants::ATTR_PAIRWISE_ID, $attributes);
+        $this->assertEquals(
+            'b9c0de5a9852b51761797af957c4b383f8e2c7586b881444af760387272b0a6b@ex-ample.org',
+            $attributes[Constants::ATTR_PAIRWISE_ID][0]
+        );
+    }
```
thijskh, line 81: I am missing tests for some i.m.o. essential parts, namely to guarantee the most important property which is that they are in fact unique for unique users, so we don't introduce accidental code changes that would violate this important property.

Maybe with a data provider check the all following combinations:
- Test that generating two Pairwise id's with different sp entityid's in the input are actually unequal.
- Test that generating two Pairwise id's with different identifyingattributes are acutally unequal.
- Test that generating two Subject id's with different identifyingattributes are acutally unequal.
- Test that generating two pairwise id's with different secretsalt are actually unequal.
- Test that generating two subject- and pairwise id's with different scopes are actually unequal. And/or check that the generated pairwise and subjectid's do end in the scope.

As a bonus, check that any of these generated id's conforms to the syntaxis.

thijskh (changes requested):

modules/core/lib/Auth/Process/PairwiseID.php
tvdijen: I've discussed adding the idp_entity with @jaimeperez and he insisted to replicate the original EPTiD-behaviour.. I agree with you though.

thijskh: Maybe I'm missing the rationale for that then. Because I don't understand it currently. For SubjectID, we find it acceptable to conclude that identifyingAttribute + scope is globally unique. For pairwiseID, we insist to add idp entityid to the mix. I don't understand why we make this difference.

tvdijen: Jaime's argument was that it's a tie between IDP and SP... I agree we can totally do without.

thijskh: I do not see a base for that in the spec.

tvdijen: I'll poke him once more on this.. See we we can settle this

modules/core/lib/Auth/Process/SubjectID.php
tvdijen: Maybe have PairwiseID extend SubjectID? Other than that, I don't think it's a good idea to merge these two classes..
I have to give this some thought, but Jaime and I agree on the fact that they should be separate filters..

modules/core/lib/Auth/Process/PairwiseID.php
```diff
@@ -0,0 +1,134 @@
...
+                "core:PairwiseID: Missing attribute '%s', which is needed to generate the Pairwise ID.",
+                $this->identifyingAttribute
+            )
+        );
+
+        $userID = $state['Attributes'][$this->identifyingAttribute][0];
+        Assert::regex(
+            $userID,
+            '/^[a-zA-Z0-9=-]+$/i',
+            'PairwiseID: \'identifyingAttribute\' contains illegal characters.'
```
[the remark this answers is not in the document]
tvdijen: I don't disagree with you, however;

3.4.1                 Syntax and Handling

The requirements for this Attribute are identical to those described in Section 3.3.1.

http://docs.oasis-open.org/security/saml-subject-id-attr/v1.0/csprd01/saml-subject-id-attr-v1.0-csprd01.html#_Toc497819495

modules/core/lib/Auth/Process/PairwiseID.php
tvdijen: The regexp on line 101 covers this.

modules/core/lib/Auth/Process/PairwiseID.php
tvdijen: I'd say we bump the specs a little but to {1,126}, because without a scope this can _never_ be a globally unique ID.. It also ensures a non-empty string

tvdijen: Ah, I guess we were both wrong on the specs..  I think I got them right now!

modules/core/lib/Auth/Process/PairwiseID.php
tvdijen: Removed the IDP entityid from the hash

modules/core/lib/Auth/Process/SubjectID.php
tvdijen: As above, the regex covered the empty string.. The new regex does this as well

tests/modules/core/lib/Auth/Process/PairwiseIDTest.php
tvdijen: Added a test with multiple RequesterIDs

modules/core/lib/Auth/Process/PairwiseID.php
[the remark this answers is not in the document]
thijskh: I don't think so. "this Attribute" here means (logically) the output/resulting attribute. Whatever information/input we use to generate it, is not relevant because our generated value always conforms to the syntax. It should e.g. be fine to use a personnel number as the identifingAttribute for which we calculate the hash.

modules/core/lib/Auth/Process/PairwiseID.php
thijskh: But it's not correct to apply the regex to the input we use.

modules/core/lib/Auth/Process/PairwiseID.php
thijskh: I'm not sure where my regex was wrong? I had at least a length of one? But agreed that probably any length < 4 is an error and cannot be globally unique.

modules/core/lib/Auth/Process/PairwiseID.php
```diff
@@ -35,6 +35,20 @@
  */
 class PairwiseID extends Auth\ProcessingFilter
 {
+    /**
+     * The regular expression to match the identifier
+     *
+     * @var string
+     */
+    public const $subject_regex = '/^[a-zA-Z0-9]{1}[a-zA-Z0-9=-]{0,126}$/i';
```
thijskh, line 43: The `{1}` does not really add anything I guess. The `A-Z` is unneeded if you add the `i` on the end.

modules/core/lib/Auth/Process/SubjectID.php
thijskh: Alright. There is some duplication also with the regexps being defined in two places etc.

modules/core/lib/Auth/Process/PairwiseID.php
tvdijen: OK agreed, so I've changed it for a non-empty string test

modules/core/lib/Auth/Process/PairwiseID.php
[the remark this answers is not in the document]
tvdijen: I agree it should be fine.. Removed the assertion in favour of a check for non-empty string

modules/core/lib/Auth/Process/PairwiseID.php
tvdijen: I misread it.. it wasn't all wrong, except for the fact that you're missing the A-Z capitals..
My suggestion would be to use the {0,126} to enforce the pattern and log a warning if it's not matching at least {3,126}..

tests/modules/core/lib/Auth/Process/PairwiseIDTest.php
tvdijen: I think all of these are now addressed (just not with a dataProvider)

modules/core/lib/Auth/Process/PairwiseID.php
thijskh: I'm not missing the capitals because the regexp uses the `i` modifier.

modules/core/lib/Auth/Process/SubjectID.php
```diff
@@ -0,0 +1,167 @@
...
+            $state['Attributes'],
+            $this->identifyingAttribute,
+            sprintf(
+                "core:" . static::NAME . ": Missing attribute '%s', which is needed to generate the ID.",
+                $this->identifyingAttribute
+            )
+        );
+
+        $userID = $state['Attributes'][$this->identifyingAttribute][0];
+        Assert::notEmpty($userID, 'core' . static::NAME . ': \'identifyingAttribute\' cannot be an empty string.');
```
thijskh, line 140: is it even better to use `Assert::stringNotEmpty`?

thijskh (approved):
Great work, I think it has really improved now and we can be sure it's very robust which is essential for identifiers.

Asked one question above.
And I'm missing a new file for `modules/core/docs/` I think so people can actually find out about it?

modules/core/lib/Auth/Process/SubjectID.php
tvdijen: Agreed on both the docs and the assertion!

tvdijen: Any remarks on the docs?

PR #1438, opened 2021-03-11
Replace LogLevel ints with strings

tvdijen:
This checks one of the boxes in #371:
- [ ] Extract the log levels definition from \SimpleSAML\Logger and use the ones defined by Psr\Log\LogLevel instead.


tvdijen:
I'm all in favour of using symfony for this!
Once we release 2.5, I want to force-push that to master and store the current master-branch in a backup-branch. Then we can start working on this on a new 3.0 branch

Issue #1407, opened 2020-12-08
SAML2INT: AuthnRequest signature ignored by default


tvdijen:
**Describe the bug**
By default, the signature of an AuthnRequest will not be verified. This breaks SAML2INT compliancy.

```
[SDP-IDP05]
If a request is signed, IdPs MUST successfully verify the signature or fail the request
```

**To Reproduce**
1. Set up basic IDP/SP with the SP signing AuthnRequests
2. Make sure to _not_ specifically set `validate.authnrequest`
3. Rollover the SP's signing certificate, do not exchange new metadata with the IDP
4. See how things remain in working order

**Expected behavior**
The IDP should refuse the AuthnRequest because the signature cannot be validated.

**Suggested behaviour**
- If nothing is set in the configuration, the signature in the message will be validated only if present
- If `false`, no validation is performed, and it doesn't matter if there's a signature
- If `true`, validation is performed and fails if the message is not signed

PR #1440, opened 2021-03-12
Conform SAML2INT; check AuthnReqsignature unless specifically disabled

tvdijen:
Closes https://github.com/simplesamlphp/simplesamlphp/issues/1407

### DIFF
--- a/docs/simplesamlphp-reference-sp-remote.md
+++ b/docs/simplesamlphp-reference-sp-remote.md
@@ -300,8 +300,8 @@ The following options can be set:
 :   Note that this option also exists in the IdP-hosted metadata.
     The value in the SP-remote metadata overrides the value in the IdP-hosted metadata.
 
-:   Note: if unset or `true`. this setting will enforce a valid signature.
-    Set to `false` to bypass (= SAML2INT incompliant)
+:   Note: SimpleSAMLphp will always try to validate signatures unless this setting is
+    specifically set to `false` (this will break SAML2INT compliancy)
 
 `validate.logout`
 :   Whether we require signatures on logout messages sent from this SP.
--- a/docs/simplesamlphp-reference-sp-remote.md
+++ b/docs/simplesamlphp-reference-sp-remote.md
@@ -296,12 +296,14 @@ The following options can be set:
 
 `validate.authnrequest`
 :   Whether we require signatures on authentication requests sent from this SP.
+    Set it to:
+
+    true: authnrequest must be signed (and signature will be validated)
+    null: authnrequest may be signed, if it is, signature will be validated
+    false: authnrequest signature is never checked
 
 :   Note that this option also exists in the IdP-hosted metadata.
     The value in the SP-remote metadata overrides the value in the IdP-hosted metadata.
-
-:   Note: SimpleSAMLphp will always try to validate signatures unless this setting is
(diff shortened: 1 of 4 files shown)


jaimeperez (approved):

tvdijen:
Needs a field-test as per linked issue to confirm this fixes the compliancy issue

thijskh (changes requested):
What I'm missing is a docs change... this behaviour must really be in the documentation of these options I think.

tvdijen:
I do not disagree, but we in general say to comply SAML2INT, right?

tvdijen:
Field-test is done; hence the fix.
I'd rather not use the global variable, but at this point I see no other way..

thijskh:
> I do not disagree, but we in general say to comply SAML2INT, right?

It is sure to break existing configs. So while this is not a problem per se, it is nice if people can then quickly find what changed so they can fix it.

modules/saml/lib/Message.php
```diff
@@ -213,7 +213,10 @@ public static function validateMessage(
             }
         }
 
-        if ($enabled === null) {
+        // If not specifically set to false, the signature must be checked to conform to SAML2INT
+        if (isset($_REQUEST['Signature']) && ($enabled !== false)) {
```
thijskh, line 217: This does not work with the POST binding, does it?

docs/simplesamlphp-reference-sp-remote.md
```diff
@@ -300,6 +300,9 @@ The following options can be set:
 :   Note that this option also exists in the IdP-hosted metadata.
     The value in the SP-remote metadata overrides the value in the IdP-hosted metadata.
 
+:   Note: if unset or `true`. this setting will enforce a valid signature.
```
thijskh, line 303: I don’t think this correcty describes the new behaviour. It is now this:

- true: authnrequest must be signed (and sig will be validated)
- null: authnrequest may be signed, if it is, sig will be validated
- false: authnrequest sig is never checked

modules/saml/lib/Message.php
```diff
@@ -221,9 +224,7 @@ public static function validateMessage(
 
         if (!$enabled) {
             return;
-        }
-
-        if (!self::checkSign($srcMetadata, $message)) {
+        } elseif (!self::checkSign($srcMetadata, $message)) {
```
thijskh, line 228: This change does not seem to do anything, why is it changed?

thijskh (changes requested):

modules/saml/lib/Message.php
tvdijen: For readability... No actual change

docs/simplesamlphp-reference-sp-remote.md
tvdijen: You are right.. I've updated it.. Better?

modules/saml/lib/Message.php
tvdijen: Good catch! Fixed it

docs/simplesamlphp-reference-sp-remote.md
thijskh: I would just make the three values and their meanings explicit in the docs for idp-hosted/sp-remote, like my suggestition, so people don't have to combine different fragments to conclude what does what.

PR #1460, opened 2021-05-06
Identifiers fix #2

### LINKED ISSUE #2: Remove "override.host" config option
This option makes it possible to override the configuration per host, but it has never been documented, and it can be replaced with simple PHP code in the configuration, so I'd remove this option for 2.0.

### LINKED ISSUE #1459: Rationalize scopeAttribute behaviour
The behaviour of `scopeAttribute` for the SubjectID & PairwiseID authproc-filters is different from the behaviour of the ScopeAttribute authproc-filter.  This should ideally work the same.

This is specifically about https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/core/lib/Auth/Process/ScopeAttribute.php#L94-L110 where the `scopeAttribute` may contain more than just a scope and will be parsed to extract the scope from it.

tvdijen:
This rationalizes the behaviour and makes it work the same as the core:ScopeAttribute authproc.
Also additional docs

Closes #1459

### DIFF
--- a/modules/core/docs/authproc_pairwiseid.md
+++ b/modules/core/docs/authproc_pairwiseid.md
@@ -7,6 +7,10 @@ http://docs.oasis-open.org/security/saml-subject-id-attr/v1.0/saml-subject-id-at
 This filter will take an attribute and a scope as input and transforms this into a anonymized and scoped
 identifier that is globally unique for a given user & service provider combination.
 
+Note:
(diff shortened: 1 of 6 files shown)


tvdijen:
@grawity Are we on the same page?

grawity:
Yeah, this looks ok.

PR #1466, opened 2021-05-26
Log which IdP issued a SAML1 assertion

francoisfreitag:
The additional information helps Service Providers identify Identity
Providers relying on SAML1 and Shibboleth 1.3.


thijskh (approved):

thijskh:
Good idea, thanks for the PR!

PR #1481, opened 2021-07-12
Now allowing Authsources to set validUntil and cacheDuration for Metadata.

tommy2d:
Now allowing generation of SP metadata XML that includes @validuntil or @cacheDuration at EntityDescriptor.

### DIFF
--- a/modules/saml/docs/sp.md
+++ b/modules/saml/docs/sp.md
@@ -245,11 +245,11 @@ Options
 :   IsPassive allows you to enable passive authentication by default for this SP.
 
 
-`maxCache`
-:   maxCache allows you to set the `validUntil` in the SP's metadata. It's an integer representing the amount of seconds.
+`validUntil`
+:   validUntil allows you to set the `validUntil` in the SP's metadata. It's an integer representing the amount of seconds since current time.
 
-`maxDuration`
-:   maxDuration allows you to set the `cacheDuration` in the SP's metadata. It's an integer representing the amount of seconds.
+`cacheDuration`
+:   cacheDuration allows you to set the `cacheDuration` in the SP's metadata. It's an integer representing the amount of seconds.
 
 `name`
 :   The name of this SP.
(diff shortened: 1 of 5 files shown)


tvdijen (approved):

modules/saml/docs/sp.md
```diff
@@ -259,6 +259,12 @@ Options
 
 :   *Note*: SAML 2 specific.
 
+`maxCache`
+:   maxCache allows you to set the `validUntil` in the SP's metadata. It's an integer representing the amount of seconds.
+
+`maxDuration`
+:   maxDuration allows you to set the `cacheDuration` in the SP's metadata. It's an integer representing the amount of seconds.
```
thijskh, line 266: I'm not so happy with the naming.

`maxCache` sets validUntil while the SAML MD spec description of `cacheDuration` is "maximum length of time a consumer should cache". So I suspect that people will be confused thinking that `maxCache` is related to the cache time of the metadata.

Probably it's most clear if the options are named according to the corresponding metadata parameters they set. So "maxCache" could be named "validUntil" and the help text coild explain that after this time an entity is expired and unusable. "maxDuration" could be named "cacheDuration" and the help text can say that this is the suggested time a remote entity should use between refresh attempts.

tvdijen: It's actually the other way around.. My added docs are wrong :/   I'll rename them and fix the docs

thijskh (changes requested):
Needs to be changed according to the (later) changes to the SP metadata refactoring + unit test update so we keep coverage of the new code up to date.

thijskh:
There's still quite a lot of code that refers to expire when talking about remote entities I think. Are you referring to that or to the setting for hosted entities (does that exist?).

tvdijen:
It does not exist for hosted entities I think, so for remote entities...  `validUntil` is translated to the `expire` setting by our metadata converter and I think it makes sense to rename it to `validUntil`..  We have to update the metarefresh-module as well if we do so

PR #1495, opened 2021-08-09
SP Metadata compliance Improvements

tommy2d:
- Now setting keyName of KeyInfo. This is derived from both the certificate fingerprint and the entityId.
- Now setting IsDefault argument for first ACS in SP metadata generation.

### DIFF
--- a/lib/SimpleSAML/Metadata/SAMLBuilder.php
+++ b/lib/SimpleSAML/Metadata/SAMLBuilder.php
@@ -698,6 +698,8 @@ class SAMLBuilder
     private function addX509KeyDescriptor(RoleDescriptor $rd, string $use, string $x509data, ?string $keyName = null): void
     {
         Assert::oneOf($use, ['encryption', 'signing']);
+        Assert::nullOrNotWhitespaceOnly($keyName);
+        Assert::regex('/^[a-zA-Z_][\w.-]*$/', $keyName, "KeyName should be valid for usage in an ID atttribute, e.g. not start with a digit, no whitespace.");
 
         $keyDescriptor = \SAML2\Utils::createKeyDescriptor($x509data, $keyName);
         $keyDescriptor->setUse($use);
--- a/lib/SimpleSAML/Metadata/SAMLBuilder.php
+++ b/lib/SimpleSAML/Metadata/SAMLBuilder.php
@@ -381,6 +381,9 @@ class SAMLBuilder
                     }
 
                     $ep['index'] = $maxIndex + 1;
+                    if (isset($ep['isDefault'])) {
+                        $t->setIsDefault($ep['isDefault']);
+                    }
                 }
 
                 $t->setIndex($ep['index']);
@@ -690,12 +693,13 @@ class SAMLBuilder
      * @param \SAML2\XML\md\RoleDescriptor $rd The RoleDescriptor the certificate should be added to.
      * @param string                      $use The value of the 'use' attribute.
      * @param string                      $x509data The certificate data.
+     * @param string|null                 $keyName The name of the key. Should be valid for usage in an ID attribute, e.g. not start with a digit
      */
-    private function addX509KeyDescriptor(RoleDescriptor $rd, string $use, string $x509data): void
+    private function addX509KeyDescriptor(RoleDescriptor $rd, string $use, string $x509data, ?string $keyName = null): void
     {
         Assert::oneOf($use, ['encryption', 'signing']);
 
-        $keyDescriptor = \SAML2\Utils::createKeyDescriptor($x509data);
+        $keyDescriptor = \SAML2\Utils::createKeyDescriptor($x509data, $keyName);
         $keyDescriptor->setUse($use);
         $rd->addKeyDescriptor($keyDescriptor);
     }
@@ -716,11 +720,15 @@ class SAMLBuilder
             if ($key['type'] !== 'X509Certificate') {
                 continue;
             }
+
+            //make sure the key is unique, even when used for multiple entities
+            $keyName = '_' . sha1( $this->entityDescriptor->getEntityID() . $key['keyName']) ?? null;
+
             if (!isset($key['signing']) || $key['signing'] === true) {
-                $this->addX509KeyDescriptor($rd, 'signing', $key['X509Certificate']);
+                $this->addX509KeyDescriptor($rd, 'signing', $key['X509Certificate'], $keyName );
             }
             if (!isset($key['encryption']) || $key['encryption'] === true) {
-                $this->addX509KeyDescriptor($rd, 'encryption', $key['X509Certificate']);
+                $this->addX509KeyDescriptor($rd, 'encryption', $key['X509Certificate'], $keyName );
             }
         }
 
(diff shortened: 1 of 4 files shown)


lib/SimpleSAML/Metadata/SAMLBuilder.php
```diff
@@ -381,6 +381,9 @@ private static function createEndpoints(array $endpoints, bool $indexed): array
                     }
 
                     $ep['index'] = $maxIndex + 1;
+                    if (isset($ep['isDefault'])) {
```
tvdijen, line 384: This should also work if no index is explicitly set, so this should go down a few lines, underneath the `$t->setIndex()`

modules/saml/www/sp/metadata.php
```diff
@@ -77,6 +77,9 @@
 $supported_protocols = [];
 foreach ($assertionsconsumerservices as $services) {
     $acsArray = ['index' => $index];
+    if ($index == 0) {
```
tvdijen, line 80: This one is problematic because not setting a default endpoint is perfectly valid in SAML.

However, if you fix my comment above, you should be able to manually set the ACS-endpoints and a set default like this;
https://simplesamlphp.org/docs/stable/simplesamlphp-metadata-endpoints#section_3

tvdijen (changes requested):

tvdijen:
Please tell Dictu to not break the specs by requiring an isDefault.. The specs are clear on this;

```
isDefault [Optional]
An optional boolean attribute used to designate the default endpoint among an indexed set. If
omitted, the value is assumed to be false.
In any such sequence of like endpoints based on this type, the default endpoint is the first such endpoint
with the isDefault attribute set to true. If no such endpoints exist, the default endpoint is the first such
endpoint without the isDefault attribute set to false. If no such endpoints exist, the default endpoint is
the first element in the sequence.
```

Taken from paragraph 2.2.3 of https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf

tvdijen (approved):

lib/SimpleSAML/Metadata/SAMLBuilder.php
```diff
@@ -716,11 +720,15 @@ private function addCertificate(RoleDescriptor $rd, Configuration $metadata): vo
             if ($key['type'] !== 'X509Certificate') {
                 continue;
             }
+
+            //make sure the key is unique, even when used for multiple entities
+            $keyName = '_' . sha1( $this->entityDescriptor->getEntityID() . $key['keyName']) ?? null;
```
tvdijen, line 725: @tommy2d I was discussing this with @jaimeperez, but some of the details you've told me this morning went down the drain during my busy workday..  Can you fill them in for us again?

You connect several SPs to your SimpleSAMLphp install and basically have one IDP being TVS.  They require you to deliver a metadata-XML with all metadata of all your SPs, stripped of all their signatures and then globally signed by you, right?  How exactly are they using the KeyName?  Would it still work for you if we change the line above to use sha256 and mix in a secret salt, to avoid collision attacks?

tommy2d: I take the metadata of several simpleSAMLphp installs and group them inside a parent document in which they all have to co-exist inside a EntitiesDescriptor element. Metadata-signing is only done on EntitiesDescriptor level; meaning that the individual EntitityDescriptor elements are not signed.

Each of the keys used by any of the EntityDescriptor elements need to have a unique name, even if use the same source-key multiple times. I need to do this because all of the (per-customer) applications that we serve in this way, have their own dedicated simpleSAMLphp install, meaning that if a customer has multiple applications that utilise TVS authentication, there will be multiple instances of the same keyName. This is something that DICTU is now allowing, so i decided to add the entityId to hash source.

What kind of collision attacks are you expecting? For the key name: common practice is using the SHA1 fingerprint stripped of colons. I simply added another factor to make sure I can re-use the same input key multiple times without getting into trouble.

tvdijen: We were thinking that a SHA-1 collision might trick the TVS IDP in selecting the wrong key to verify a signature.. That's why I asked if you know how they're using it.. Or are they just enforcing uniqueness?

tommy2d: As far as I understand, they just want it to be unique within the EntitiesDescriptor document i supply. As it seems to be common practise to simply use the SHA1 fingerprint for the specifc field, there seems to be no point in trying to prevent a theoretical collision here.

bartlangelaan: I was just looking for the functionality to provide a KeyName, and I found this issue!
I need this to connect to eToegang (specifically: eID from KPN). It requires a KeyName for each KeyInfo in the metadata.

From their documentation:
> Neem voor het certificaat / certificaten binnen de 'SPSSODescriptor' een `<KeyName>`-element op, met een unieke waarde; bij voorkeur de 'thumbprint' van het certificaat

I think that enforcing uniqueness (by hashing with an entityid or secret salt) is not needed, looking at that instruction.

tommy2d: For your use case it is indeed not be necessary to perform the additional hashing step. For mine it is, as I will need to include multiple entityDescriptor elements in the same document (LC + DV that utilise the same key and hence wil lead to the same cert name of the fingerprint is to be used directly). Salting is also not needed for my use case as I already pointed out.

tvdijen: We've just discussed this a bit more amongst devs and the general consensus is that we can offer a `key_name` setting in the SP-authsource where the name can be configured. We don't want to programmatically calculate some value, because as you see one person needs and extra level of uniqueness and the other wants the thumbprint in there (*).
It only makes sense to configure a static value when you configure a new key.

(*) We should kick KPN in the sack for that, because the thumbprint should never go in the KeyName. The X509Digest-element is meant to hold that.  Identifying a key by it's thumbprint is hazardous.

tommy2d: Yes I agree with all of this and I can very wel live with this solution.

bartlangelaan: Yeah, having that option would be enough to be compliant!

tvdijen: I'll see what I can do during the weekend, but an addition to this PR is most welcome

tommy2d: I'm happy to do it but i have no experience adding configuration settings to the SP auth source.

tvdijen: It's quite difficult since we have so many ways of specifying keys..  Including rollover-certs using the `new_` prefix.. I can't promise I have time for this in the near future..

tvdijen:
This PR has been gone unmergable and it appears this functionality was already added in https://github.com/simplesamlphp/simplesamlphp/pull/1509 (although slightly different).

PR #1504, opened 2021-08-24
Upgrade gettext

### LINKED ISSUE #1367: Migrate gettext
We currently rely on the `gettext/gettext` at v4.x, which is no longer supported.
It seems that the parts we're using do not exist anymore in v5 (BaseTranslator-class).

tvdijen:
Closes #1367


tvdijen:
TO DO:  `lib/SimpleSAML/Translate` needs the `Gettext/Translator` we create in `lib/SimpleSAML/Localization`, but there currently is no way to access it

tvdijen:
The last 2-3 commits may be debatable.. The interface of gettext was so significantly changed, that I have no clue how to properly deal with this..

tvdijen:
Unfortunately this change doesn't work at all... Nothing is being translated..

tvdijen:
I was able to make it work! Stupid mistake...

PR #1509, opened 2021-08-30
Add keys.name and key_name configuration options

bartlangelaan:
This MR adds a `key_name` configuration option, and a `name` option to the `keys` option. This way a key name can be added to the metadata, which is required for some IDPs.

This solution is based of PR #1495 and this comment: https://github.com/simplesamlphp/simplesamlphp/pull/1495#discussion_r686721690

----------

I am currently using version 1.18 because I am not yet able to upgrade to symfony 4. I created a PR that works fine for me: https://github.com/bartlangelaan/simplesamlphp/pull/1. I used that version to create a PR for the latest version of Simplesamlphp, which is the PR you're looking at. I guess this should also work, but it's hard for me to test all use-cases.


tvdijen:
Excellent @bartlangelaan !   As a nitpick, would you mind changing `NULL` too lowercase `null` ?
We try to follow PSR-12 formatting guidelines

thijskh (changes requested):
Thanks!

I wonder if we add this for IdP should we also add it for SPs? I'm not sure since I've never seen a use for key names myself.

The new option is not documented anywhere now, so people cannot know that it's available or what it's for.

Would be great to add a testcase for it.

bartlangelaan:
Hey @thijskh, thanks for your feedback.

It took some time, but I figured out how to add a complete test for the new parameter, including the new_key_name option. It should now fully work.

I have also added documentation.

thijskh (approved):

thijskh:
Excellent. Thanks!

tvdijen:
Anyone objects if I replace the certificates with the ones from `simplesamlphp/xml-security`?

PR #1510, opened 2021-08-30
When downloading metadata, add a .xml extension

bartlangelaan:
This adds a headers so when you visit a metadata url, it is donwloaded as `default-sp.xml` instead of `default-sp`. This makes it easier to open and upload in other systems.

### DIFF
--- a/modules/saml/www/sp/metadata.php
+++ b/modules/saml/www/sp/metadata.php
@@ -70,6 +70,6 @@ if (array_key_exists('output', $_REQUEST) && $_REQUEST['output'] == 'xhtml') {
     $t->send();
 } else {
     header('Content-Type: application/samlmetadata+xml');
-    header('Content-Disposition: attachment; filename="' . $sourceId . '.xml"');
+    header('Content-Disposition: attachment; filename="' . basename($sourceId) . '.xml"');
     echo($xml);
 }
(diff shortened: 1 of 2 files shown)


thijskh (commented):
Excellent idea! I do wonder if we somehow need to encode the strings though. For the `default-sp` case this yields a reasonable filename, but our default idp entityID is a long url like `https://something.example.nl/simplesamlphp/saml2/idp/metadata.php` which is not nice to have as a filename I think...

thijskh:
Maybe for the idp case just default to `metadata.xml` or `idp-metadata.xml`?

bartlangelaan:
Thanks for the feedback @thijskh! 

I have now added a basename() call, so it will never generate a long name. This means that if this PR gets merged, the metadata wil have exactly the same name as before, but just with .xml added. The browser already uses the basename by default, if no attachment filename has been provided.

thijskh (commented):
I'm afraid I'm not happy with this change either. basename() is documented as returning "trailing name component of path". This change will pass non-path things into it, like the authsource name for an SP or a URI for IdPs. I've tried it out and for an URL it does give the hostname as a result, but it seems inappropriate use of this PHP function to parse URLs with it.

For the SP (authsource) I'm pretty sure that the authsource name will be safe enough for filenames. So we can keep that (remove the basename around it).

For IdPs I see no good way to translate the unique aspect (full URL) into something that is a reasonable default filename. I therefore propose to keep it simple and just hardcode the name to `idp-metadata.xml` always. What do you think?

thijskh:
So in order to keep things rolling I've merged your SP propsal and changed the IdP filename to always be "idp-metadata.xml". Definitly an improvement!

Issue #460, opened 2016-09-06
Get rid of Translate->getPreferredTranslation()


hmpf:
Find a way to use `t()` instead. For #455

hmpf:
This method receives an array of langcode => translation pairs, for some unknown tag. We will need a new version of this somewhere, for dynamically loaded translations. It is however necessary to move all uses of this out of the template and into the calling page. Make a note where doing so would be very tricky.

An alternative is to send in the array in the context, have the currently set language as a langcode in the default context, and look up the translation directly: `{{ dynamicstatus[langcode] }}` for instance.

jaimeperez:
I'd say the best option, in this case, is to have a static method somewhere (`SimpleSAML\Locale\Translate`, maybe?) that we can add to the Twig context as a filter, so that we can do something like:

```Twig
{{ variable_holding_translations|translate }}
```

We could even tell the filter to default to some language:

```Twig
{{ variable_holding_translations|translate('en') }}
```

Or we could default ourselves to some given string:

```Twig
{{ variable_holding_translations|translate|default('Our preferred translation') }}
```

tvdijen:
Isn't `\SimpleSAML\Locale\Translate::translateFromArray()` exactly what we need? It is already available as a Twig filter..

tvdijen:
Four cases left:

simplesamlphp-module-consent - Was presumably fixed in 8f256aa3eddbc03d8be61e0657ba0825580b236d, but has not been tested yet
simplesamlphp-module-consentadmin
simplesamlphp-module-discopower
simplesamlphp-module-statistics

thijskh:
It's also used in our own codebase: lib/SimpleSAML/Locale/Translate.php

translateFromArray is not a ready-to-use alternative because it does not fall back. So if my interface is NL and I search an entity with $entity['name'] = ['en' => "Example"];, I get null back, not "Example". Probably we need some Twig convenience function for displaying an entity name, something like return the first of this in order when found:
- entity[uiinfo.displayname][curlang] 
- entity[name][curlang]
- entity[uinfo.displayname[defaultlang]
- entity[name][defaultlang]
- entity[uinfo.displayname[en]
- entity[name][en]
- entity[organizationdisplaynamename][curlang]
- entity[organizationname[curlang]
- entity[organizationdisplayname[defaultlang]
- entity[organizationname[defaultlang]
- entity[organizationdisplayname[en]
- entity[organizationname[en]
- entityid

[...]

PR #1512, opened 2021-09-03
getEntityDisplayName to get entityID's name for display

thijskh:
Reimplement getPreferredTranslation outside of IdPDisco so it's reusable and much more complete. The new method will search the current, default and fallback language for a displayable string in various metadata fields. This should be much more complete and robust than previous approches. Make it a Twig filter so it can also be used there.

For clarity:
- translateFromArray method/filter will give a single entry for an array with multiple language options. Most useful with e.g. config file options that have this format.
- getEntityPropertyTranslation method/filter will walk the three languages in priority order to find a named metadata key (useful for localized metadata fields where no other fields need to be searched)
- getEntityDisplayName (method only fornow) for the specific case for displaying the name, where besides different languages we also search different metadata fields.

If the approach is deemed OK there are some other places in the code where we can use this so we have a consistent algorithm across the code to find the name of some entity to display.

### DIFF
--- a/lib/SimpleSAML/Locale/Language.php
+++ b/lib/SimpleSAML/Locale/Language.php
@@ -52,6 +52,13 @@ class Language
     private string $defaultLanguage;
 
     /**
+     * The final fallback language to use when no current or default available
+     *
+     * @var string
+     */
+    public const FALLBACKLANGUAGE = 'en';
+
+    /**
      * An array holding a list of languages that are written from right to left.
      *
      * @var string[]
@@ -146,7 +153,7 @@ class Language
     {
         $this->configuration = $configuration;
         $this->availableLanguages = $this->getInstalledLanguages();
-        $this->defaultLanguage = $this->configuration->getString('language.default', 'en');
+        $this->defaultLanguage = $this->configuration->getString('language.default', self::FALLBACKLANGUAGE);
         $this->languageParameterName = $this->configuration->getString('language.parameter.name', 'language');
         $this->customFunction = $this->configuration->getArray('language.get_language_function', null);
         $this->rtlLanguages = $this->configuration->getArray('language.rtl', []);
@@ -166,7 +173,7 @@ class Language
      */
     private function getInstalledLanguages(): array
     {
-        $configuredAvailableLanguages = $this->configuration->getArray('language.available', ['en']);
+        $configuredAvailableLanguages = $this->configuration->getArray('language.available', [self::FALLBACKLANGUAGE]);
         $availableLanguages = [];
         foreach ($configuredAvailableLanguages as $code) {
             if (array_key_exists($code, self::$language_names) && isset(self::$language_names[$code])) {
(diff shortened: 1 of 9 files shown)


thijskh:
Would help to close #460

Used in https://github.com/simplesamlphp/simplesamlphp-module-discopower/pull/10

thijskh:
There's also some related stuff only used in IFrameLogout. Is IFrameLogout something still relevant in 2021?

tvdijen:
> 
> 
> There's also some related stuff only used in IFrameLogout. Is IFrameLogout something still relevant in 2021?

I'm not sure, I don't so SLO myself...  Maybe just leave it as is for now?

tests/lib/SimpleSAML/XHTML/TemplateTest.php
```diff
@@ -0,0 +1,114 @@
...
+        $name = $t->getEntityDisplayName($data);
+        $this->assertEquals('UIname NL', $name);
+    }
+
+    function testGetEntityPropertyTranslation(): void
+    {
+        $c = Configuration::loadFromArray([], '', 'simplesaml');
+        $t = new Template($c, self::TEMPLATE);
+
+	$prop = 'description';
```
tvdijen, line 93: indentation is off in a few places in this file

lib/SimpleSAML/Locale/Language.php
```diff
@@ -377,6 +384,16 @@ public function isLanguageRTL(): bool
...
 
+    /**
+     * Returns the list of languages in order of preference. This is useful
+     * to search e.g. an array of entity names for first the current language,
+     * if not present the default language, if not present the fallback language.
+     */
+    public function getPreferredLanguages(): array
+    {
+        $curLanguage = $this->getLanguage();
+        return array_unique([0 => $curLanguage, 1 => $this->defaultLanguage, 2 => self::FALLBACKLANGUAGE]);
```
tvdijen, line 395: Should we also consider the language priorities from the configuration here?

thijskh: Yes, that would make sense. Even better to have the code in one place so we can consider all available options

thijskh:
I've looked into `languages[priorities]`. There's an issue with it. It is _only_ used in `translateFromArray()`. If you ask me that means it does not deliver upon it's promise by far. That is, such priorities are only used in the few contexts where specifically translateFromArray is used (which is not in many places). (There's also the fact that it's the only member of the `language` config array. But that's minor.)

I see two options:
1. Actually implement language priorities across our translation system (at some point?).
2. Do away with the config option. It does make the already extensive language options quite a bit simpler. And would we really lose a lot of value?

tvdijen:
Agreed to go with 2.  The only reason to fall back to another language is when the translations for the current language are incomplete.. I'd rather put some effort in completing the translations for all languages.

tvdijen (approved):

PR #1514, opened 2021-09-06
Reimplement admin module auth test attribute presentation

### LINKED ISSUE #461: Get rid of Translate->getAttributeTranslation()
Find a way to use `t()` instead. For #455

thijskh:
This is my second attempt to fix #461. I have some trouble wrapping my head around all the present code and why this was there. But I can now confirm that the attribute names are correctly translated and that we've cleaned up some HTML generation in the controller. So without wanting to claim that I now fully understand all the (historic) intentions I think this gives us a cleaner base to start from.


tvdijen:
I think you're spot-on Thijs!!

tvdijen (approved):
I think you're spot-on Thijs!!

PR #1520, opened 2021-09-10
Deduplicate IdP hosted metadata

thijskh:
As with SP before. Duplicated code exists inside module/saml and in www/idp/metadata.php. Move into one copy and make sure this is tested. The tests are derived from the old code to ensure it covers everything that it did before.

### DIFF
--- a/tests/modules/saml/lib/IdP/SAML2Test.php
+++ b/tests/modules/saml/lib/IdP/SAML2Test.php
@@ -4,8 +4,12 @@ declare(strict_types=1);
 
 namespace SimpleSAML\Test\Module\saml\IdP;
 
+use InvalidArgumentException;
 use SimpleSAML\Configuration;
+use SimpleSAML\Error\Exception;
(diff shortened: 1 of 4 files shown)


tests/modules/saml/lib/IdP/SAML2Test.php
```diff
@@ -217,4 +230,530 @@ function ($arg) use (&$state) {
...
+
+        $_SERVER['REQUEST_URI'] = '/dummy';
+
+        return SAML2::getHostedMetadata($metadata['entityid']);
+    }
+
+    /**
+     * A minimally configured hosted IdP has all default fields with expected values.
+     */
+    function testIdPGetHostedMetadataMinimal()
```
tvdijen, line 267: needs visibility + return type

tvdijen: As well as all the other methods below

tvdijen (approved):

PR #1526, opened 2021-10-18
Use $state saml:RelayState in startSSO2

zukka77:
Hello, I had to implement this little patch because the upstream idp doesn't work without a relayState parameter.
I'm using phpsamlphp as a "proxy", when invoked by an sp no relay state is passed to the upstream idp (it does when you test the authentication).
This patch uses the saml:RelayState value already present in $state (the one from the sp request) and forwards it to the upstream idp.
I don't know if this is a violation of the SAML protocol or if this is something very bad to do, in that case please apologize me and ignore this request :)

### DIFF
--- a/modules/saml/lib/Auth/Source/SP.php
+++ b/modules/saml/lib/Auth/Source/SP.php
@@ -524,6 +524,10 @@ class SP extends \SimpleSAML\Auth\Source
         if (isset($state['\SimpleSAML\Auth\Source.ReturnURL'])) {
             $ar->setRelayState($state['\SimpleSAML\Auth\Source.ReturnURL']);
         }
+        
+        if (isset($state['saml:RelayState'])) {
+            $ar->setRelayState($state['saml:RelayState']);
+        }
 
         $accr = null;
         if ($idpMetadata->getString('AuthnContextClassRef', false)) {


thijskh:
The proxy is required to pass back the RelayState it received from the SP back to that SP.
I don't think there's anything that mandates sending that same RelayState also to upstream IdPs. I'm not in favour of doing that then unconditionally.

I understand the original problem is that the upstream IdP requires a RelayState to be present. Why? The only thing an IdP needs to do with a RelayState is passing it back to the proxy unmodified, it's not for interpretation by the IdP itself. This seems to be a bug in that IdP. I'm not in favour of adding code to any SSP using install to work around this IdPs bug. Rather, let's fix the bug in that IdP. Which software does it use?

In any case, I do not see passing the SPs relay state to the upstream IdP as a solution even to this given problematic IdP, since the SP may not pass a RelayState at all and the IdP will then again break. Rather, the proxy should be setting a RelayState for this specific IdP. This seems like a much cleaner workaround as it could be done just in local configuration and does not require modifications to the SSP core product.

zukka77:
Hi, thank you for the reply.

> Rather, the proxy should be setting a RelayState for this specific IdP. This seems like a much cleaner workaround as it could be done just in local configuration and does not require modifications to the SSP core product.

I agree this would be a much better workaround, is this already supported ? In that case I guess I could fake a relayState from SSP proxy to the idp.
Will then SSP pass the original relayState to the originating SP ?

thank you

thijskh:
To understand your problem correctly: the only requirement of this IdP is that a (any) RelayState is passed to it?

zukka77:
I think so, I didn't do any specific test, but I'm pretty confident.

thijskh:
I don't have a ready installation to verify this on, but I think an authproc along these lines should do it. But maybe you can try out for yourself a bit what works. In any case, you can always fall back to keeping your original code change as a local change of course.

```php
      90 => [
         'class' => 'core:PHP',
         'code' => '$state["\SimpleSAML\Auth\Source.ReturnURL"] = "aabbcc";',
      ],
```

tvdijen:
@zukka77 Had @thijskh suggestion work? Why was this closed?

zukka77:
@tvdijen in the end the code was already in production with that patch, I haven't had the time to try it out.
For some time I'm not going to work at that place anymore, today I closed my fork because I couldn't get it in sync.
I made another fork ( https://github.com/zukka77/simplesamlphp/tree/Forward-RelayState ) and ported the 2 line patch ( https://github.com/zukka77/simplesamlphp/commit/a5d9963518d41cad155d074f86258cc5fc637080 ) to a clean branch on top of the latest 1.19 branch.
If you think it could be of any use I could open a PR from that branch, otherwise I think we're fine :) I'll keep @thijskh suggestion under the pilow :smile:

PR #1528, opened 2021-10-21
Memcache server always seen as down when using moxi

ks3:
We're using a Couchbase cluster for our back-end, then using moxi to provide a memcache interface to it in order to make use of memcookie authentication. Unfortunately, the SimpleSAML\Memcache class is checking server status with getStats, which always seem to return false from moxi. I've locally modified our installation to use getVersion instead, and it has been working as intended for some time. I don't believe this will cause unintended consequences, as the server has to be responding for getVersion to return.


tvdijen:
Given the case, I actually don't care 😆 This change is good is at is

Issue #1273, opened 2020-01-31
Upgrade to Twig 3


simonberger:
Can you please support Twig 3? It is more or less compatible to 2. But as Twig 1 is currently supported as well there might be some things to do or to drop Twig 1.

tvdijen:
I had investigated this a while ago and there was an issue that I couldn't get around...
Twig 3 no longer relies on Twig Extensions, meaning we would have to rewrite https://github.com/simplesamlphp/twig-configurable-i18n and somehow keep it compatible with Twig 2 as well..
It's a bit of a nightmare ...

paulrwest:
I'm also in need for twig 3 to be supported as I use cakePHP framework on version 4 which has Twig package fixed to v3.1.1 I noticed even dev-master is not on v3 of twig so I'm a bit stuck. 

Any suggestions how to get around this I've not hit a composer conflict like this before?

tvdijen:
I don't understand.. CakePHP doesn't have a dependency on Twig?  Anyway, no there is no solution.. This issue needs to be resolved

paulrwest:
Thanks for the reply, sorry my mistake it is the [cakephp/bake](https://packagist.org/packages/cakephp/bake) package that relies on [cakephp/twig-view](https://packagist.org/packages/cakephp/twig-view) package which relies on [twig/twig:3](https://packagist.org/packages/twig/twig#3). 

This is a "require-dev" package so I have found a solution of removing that package. It's not ideal having to remove a feature from something I use to be able to pull down simplesamlphp. I think I'll end up removing simplesamlphp and adding bake package when I need to bake.

tvdijen:
I can eee how that's suboptimal.. Hopefully we can work out this issue in the new year if we can get a little bit more resourcing..

ajibarra:
+1 for this feature. Another CakePHP user here having issues with twig 3.

drjayvee:
Our project is relying on SimpleSAMLphp and Twig heavily, so we might be able to donate some of our tech debt time budget on this issue.

Instead of fixing `twig-configurable-i18n` - which, according to its own documentation, "is a hack" - could this project migrate to Twig's `trans`?

tvdijen:
Help is very much appreciated here @drjayvee !  For us it would make the most sense to use the `symfony/twig-bridge`, but I'm not sure it can fully replace `twig-configurable-i18n`..  Perhaps @jaimeperez can remember the exact need for the 'hack'

[...]

PR #1531, opened 2021-10-26
updated twig to version 3 and added wrapper for translation

mrbase:
This updates twig to version 3 and wrapps translations via symfony/twig-bridge.
You can register your own translation handler or fall back to gettext.

The PR addresses #1273

### DIFF
--- a/lib/SimpleSAML/Locale/TwigTranslator.php
+++ b/lib/SimpleSAML/Locale/TwigTranslator.php
@@ -1,0 +1,45 @@
+<?php
+
+/**
+ * Wrap twig trans to allow us to override the translator used.
+ *
+ * @package SimpleSAMLphp
+ */
+
+declare(strict_types=1);
+
+namespace SimpleSAML\Locale;
+
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class TwigTranslator implements TranslatorInterface
+{
+    private $translator;
+
+    /**
+     * @param callable|null $translator
+     */
+    public function __construct(callable $translator = null)
+    {
+        if (!is_callable($translator)) {
+            $translator = fn($string) => gettext($string);
+        }
+
+        $this->translator = $translator;
+    }
+
+    /**
+     * Translate message via configured translator.
+     *
+     * @param string $id
+     * @param array $parameters
+     * @param string|null $domain
+     * @param string|null $locale
+     */
+    public function trans(string $id, array $parameters = [], string $domain = null, string $locale = null)
+    {
+        $this->locale = $locale;
+
+        return call_user_func_array($this->translator, func_get_args());
+    }
+}
--- a/modules/admin/templates/config.twig
+++ b/modules/admin/templates/config.twig
@@ -18,7 +18,9 @@
     <div class="message-box">
       {% trans %}SimpleSAMLphp is installed in:{% endtrans %}
       <kbd>{{ directory }}</kbd><br/>
-      {% trans %}You are running version <kbd>%version%}</kbd>.{% endtrans %}
+      {% trans with {
+          '%version%': version
+      } %}You are running version <kbd>%version%</kbd>.{% endtrans %}
     </div>
     <h2>{% trans %}Configuration{% endtrans %}</h2>
     <div class="enablebox mini">
--- a/modules/admin/templates/config.twig
+++ b/modules/admin/templates/config.twig
@@ -18,7 +18,7 @@
     <div class="message-box">
       {% trans %}SimpleSAMLphp is installed in:{% endtrans %}
       <kbd>{{ directory }}</kbd><br/>
-      {% trans %}You are running version <kbd>{{ version }}</kbd>.{% endtrans %}
+      {% trans %}You are running version <kbd>%version%}</kbd>.{% endtrans %}
     </div>
     <h2>{% trans %}Configuration{% endtrans %}</h2>
     <div class="enablebox mini">
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "symfony/twig-bridge": "^5.3",
         "symfony/var-exporter": "^5.0",
         "symfony/yaml": "^5.0",
-        "twig/intl-extra": "3.3",
+        "twig/intl-extra": "^3.3",
         "twig/twig": "^3.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,6 @@
         "phpmailer/phpmailer": "^6.1",
         "simplesamlphp/assert": "^0.2.7",
         "simplesamlphp/saml2": "^4.2.3",
-        "simplesamlphp/twig-configurable-i18n": "^2.3",
         "symfony/cache": "^5.0",
         "symfony/config": "^5.0",
         "symfony/console": "^5.0",
@@ -62,10 +61,11 @@
         "symfony/http-foundation": "^5.0",
         "symfony/http-kernel": "^5.0",
         "symfony/routing": "^5.0",
+        "symfony/twig-bridge": "^5.3",
         "symfony/var-exporter": "^5.0",
         "symfony/yaml": "^5.0",
-        "twig/twig": "~2.0",
-        "twig/intl-extra": "^3.3"
+        "twig/intl-extra": "3.3",
+        "twig/twig": "^3.0"
     },
     "require-dev": {
         "ext-curl": "*",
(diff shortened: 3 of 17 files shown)


tvdijen:
Have you tested this yourself? Because I'm trying this right now and immediately hit Twig parser errors like;

```
Oct 30 16:29:12 webapp-10 simplesamlphp-idp[132964]: 3 [cd715205f0] Caused by: Twig\Error\SyntaxError: A message inside a trans tag must be a simple text.
Oct 30 16:29:12 webapp-10 simplesamlphp-idp[132964]: 3 [cd715205f0] Backtrace:
Oct 30 16:29:12 webapp-10 simplesamlphp-idp[132964]: 3 [cd715205f0] 16 /apps/installation/simplesamlphp/tmp/modules/core/templates/loginuserpass.twig:13 (Symfony\Bridge\Twig\TokenParser\TransTokenParser::parse)
```

I wonder if there's a way to automatically verify our templates against Twig 3 compatibility, because the unit test [we have in place right now](https://github.com/simplesamlphp/simplesamlphp-test-framework/blob/master/src/TemplateTest.php) is clearly not enough (or not working as it should, I'm not sure yet)

mrbase:
@tvdijen We did not test the full thing, the page we hit was the "error not configured" page, which worked out fine.
I can see we need some more testing - Will try to get around to doing a proper setup next week.

Looking at the block in loginuserpass.twig, I can see why the error occurs in Twig3 tho.
Why you would have space stripping "comments" in the translation is beyond me - why not just have the whole thing on one line?

Also, do you know how this works: `modules/core/templates/cardinality_error.twig` line: 17 goes:

`<dd>{% trans %}got {{ got }} values, want {{ want }}{% endtrans %}</dd>`

But the .po msgid is `"got %GOT% values, want %WANT%"` - which is more in line with how Symfony handles variables in translations..

The current translator uses `strtr()` to replace variables inside the translated text - but in the above case it would not substitute anything out of the box.

tvdijen:
Thanks @mrbase! I think the issues are minor and otherwise the PR seems solid as is.. Some more testing on the templates would help us to discover problems with the templates, not just in this repository but also for all of the modules we have outside this repository.

Regarding the loginuserpass.twig ....   Who knows what we were thinking when we were converting all our .tpl.php files to Twig..  It was a learning process for many of us and at some point I think we were glad to get things working :'D

The cardinality authproc was designed by @ghalse, so maybe he can answer this..
If I would have to rewrite this I think it would be like;  `<dd>{ 'got %GOT% values, want %WANT%'|trans({'%GOT%': got, '%WANT%': want}) }</dd>`, am I right?

mrbase:
Yeah, the dd translation example is correct.
I'm going to see if I can find a couple of hours to get some more tests in.

tvdijen:
Found another one by happy accident;

```
Nov  9 16:22:42 webapp-10 simplesamlphp-demo[148099]: 3 [4b44cb0b01] Caused by: Twig\Error\SyntaxError: A message inside a trans tag must be a simple text.
Nov  9 16:22:42 webapp-10 simplesamlphp-demo[148099]: 3 [4b44cb0b01] Backtrace:
Nov  9 16:22:42 webapp-10 simplesamlphp-demo[148099]: 3 [4b44cb0b01] 14 /apps/installation/simplesamlphp/tmp/templates/auth_status.twig:10 (Symfony\Bridge\Twig\TokenParser\TransTokenParser::parse)
```

mrbase:
Yeah spotted that one too.
I'm in the process of going through all the templates to resolve any issues and add some tests as well.

tvdijen:
It would be really great if we can spot those issues as part of a generic unit test as well, because of the modules in other repositories as well.

ghalse:
> The cardinality authproc was designed by @ghalse, so maybe he can answer this.. 
I don't recall offhand, but it is possible it was a bug in there from the beginning. I work with Twig, Jinja & Hugo templates and they're all deceivingly similar but all have their own quirks :-/.

tvdijen:
Thanks @ghalse and again thanks ffor your work @mrbase ! This is a very welcome and necessary PR

mrbase:
I have added some tests to the template translations, but could not figure out how to load modules in the tests.
Tests for the templates in the admin modules are missing tests to the changes I made...

tvdijen:
The tests are fine, thank you! I was hoping for a more generic test to check the templates for syntax errors..
Not sure that is even possible

mrbase:
Alright, just added a bunch of fixes and a new test that loops over all twig files and loads the files to see if they can parse the validation done by twig.

mrbase:
@tvdijen do let me know if there are anything else the PR needs.

tvdijen:
Ah great @mrbase ! I had missed your last update..
Let's see if I can get someone else to review this as well

modules/admin/templates/config.twig
```diff
@@ -18,7 +18,7 @@
     <div class="message-box">
       {% trans %}SimpleSAMLphp is installed in:{% endtrans %}
       <kbd>{{ directory }}</kbd><br/>
-      {% trans %}You are running version <kbd>{{ version }}</kbd>.{% endtrans %}
+      {% trans %}You are running version <kbd>%version%}</kbd>.{% endtrans %}
```
thijskh, line 21: Stray `}` here. And is %version% automatically replaced somehow?

thijskh (approved):
Very nice work. Had just one comment. Did not test it in practice but I propose to just merge it and we'll find and fix any corners where it might still have issues if they occur.

modules/admin/templates/config.twig
mrbase: No that is an oversight on my part - will push a fix in a few minutes.

lib/SimpleSAML/Locale/TwigTranslator.php
```diff
@@ -0,0 +1,45 @@
...
+class TwigTranslator implements TranslatorInterface
+{
+    private $translator;
+
+    /**
+     * @param callable|null $translator
+     */
+    public function __construct(callable $translator = null)
+    {
+        if (!is_callable($translator)) {
```
tvdijen, line 24: Help me here.. All I can find is cases where we call this with a callable, never with the null-case..
Can we drop this if-statement / the nullable-param?

mrbase: Sure, the check is here to mimic the old behaviour - but as we always provide the translator I see no reason not to drop it.

tvdijen: Great, thanks!

CheezisCrust:
Hi, I am a fairly new developer and quite unfamiliar with open-source release processes so please excuse me if this is not the proper way to ask: This fix is not currently included in any 1.9.x tags correct? Will this only be available with 2.x or will it be added to 1.9.6 as well and when might that be released? I am asking because of a critical CVE-2022-23614 having been discovered in twig 2.14.10 and older aswell as twig 3.3.7 and older. Thank you.

tvdijen:
This has been patched in the 1.19.x branch and will be part of 1.19.6..  You don't have to worry about security, because we do not rely on the broken sort-function anywhere. If we did, we would have released a new version already.

CheezisCrust:
Thank you very much for the quick response. I was fairly sure that the code in question was not in use since there was no mention of the CVE being part of simplesaml but our customers CVE reporter would/will still report the twig entry in the composer.json. So we will wait for 1.9.6. Thanks again!

tvdijen:
If you really worry about the CVE reporter, you can manually update the twig-dependency by running `composer update twig/twig`

CheezisCrust:
Maybe I am misunderstanding but that would not update the version of twig that simplesamlphp depends on, only the version of twig that we use directly. I can not simply override the simplesamlphp composer.json right?. For me wating for 1.9.6 is fine, our customer demands to be able to update to a version of our product which does not contain any critical/high/medium findings of their reporter so they will have to be patient then.

tvdijen:
Ah, you're including SimpleSAMLphp in your project as a composer dependency?
A project can only include one version of a certain package, so if you update your own twig-version, SimpleSAMLphp will use the updated version as well.

When working with open source software, you will run into vulnerabilities in secondairy dependencies more often. I have customers like this too, but you have to convince them that a known vulnerability is not a vulnerability per se. It has to be possible to say: this doesn't affect us, and therefor we accept it to exist in our software. If this isn't possible, the customer will keep itself and you as a developer in a choke hold...   You can't work like that.

CheezisCrust:
Yes correct. That is very good to know. I have not been in direct contact with this customer, I have only been told that they are very strict when it comes to CVE's but I will bring this information to their attention, thank you once again.

PR #1536, opened 2021-11-10
feat: add MongoStore

melanger:


tvdijen:
Looking good @melanger !  However, I think it makes more sense to put this in a separate module.. It's kinda specific.

melanger:
Is it more specific than `redis`, or would you separate a `redis` module as well? For sessions, I guess I can configure an external class even without changes to SimpleSAMLphp.

tvdijen:
I think `redis` and maybe even `memcache` should move to a separate module as well yes... @thijskh what do you reckon?
It is also possible to configure a kvStore without changes to SSP.. Check the `default` part of the switch-statement where you added code for mongo.

melanger:
All right, I will create a small separate SSP module, which will provide this store.

I am trying mongoDB because I am running into performance issues with SQL store for sessions combined with a mariaDB galera cluster. It does not handle the synchronization of large rows (=session data) well, when there is a lot of changes (every login updates a row).

PR #1538, opened 2021-11-11
Alter logics to support HTTP/2.0

### LINKED ISSUE #1537: HTTP 2.0 Infinite Redirects from HTTP 302 Response
An infinite redirect loop occurs on login with HTTP 2.0 enabled. 

In [lib/SimpleSAML/Utils/HTTP.php#redirect](https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Utils/HTTP.php#L233) there is a check for HTTP/1.1 that sets the header HTTP 303. With HTTP 2.0 enabled, the header HTTP 302 is set and an infinite redirect loop occurs on login. 

Is HTTP 2.0 unsupported? 
Any repercussions from patching a HTTP/2.0 check?

Configuration:
- simplesaml: 1.17.2
- PHP: 7.3.32
- nginx: 1.20.1

tvdijen:
Closes #1537

### DIFF
--- a/lib/SimpleSAML/Utils/HTTP.php
+++ b/lib/SimpleSAML/Utils/HTTP.php
@@ -225,26 +225,13 @@ class HTTP
             $url = $this->addURLParameters($url, $parameters);
         }
 
-        /* Set the HTTP result code. This is either 303 See Other or
-         * 302 Found. HTTP 303 See Other is sent if the HTTP version
-         * is >= HTTP/1.1 and the request type was a POST request.
-         */
-        if ($_SERVER['REQUEST_METHOD'] === 'POST') {
-            $code = 303;
-            if ($_SERVER['SERVER_PROTOCOL'] === 'HTTP/1.0') {
-                $code = 302;
-            }
-        } else {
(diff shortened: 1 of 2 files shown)


thijskh:
I propose to instead simplify this a lot and set the code to be 303 always, unconditionally. We have no need to support theoretical Netscape Navigators or other ancient things. If they are still around they will not be able to connect to any reasonably secure TLS configuration. All the complexity we have now just leads to bugs.

tvdijen:
Agreed on the 303, but the distinction between GET/POST is unclear to me and Jaime.. We should check it out

thijskh:
No, the whole world seems to still use the legacy 302. Probably because there's nothing actually wrong with it in practice, only in theory.

thijskh:
For what it's worth, mellon uses 303 since ages.

tvdijen:
This should be it then... I hope we don't introduce any unwanted side-effects

thijskh (approved):
Yes, great. Just KISS please.

PR #1542, opened 2021-12-01
Add module enablement tips to upgrade notes

pradtke:


thijskh:
Thanks! I just wonder if we need to have things in the manual along the lines of “in the past you could to this”. What do you think?

pradtke:
I thought a lot of admins/users won't read the upgrade notes so I wanted to indicate in main documentation that the old way was removed.  I definitely think removing mention of the old way for a 2.1 release makes sense. 

Anyhow, I'm good removing it now or leaving it in.

thijskh (approved):

PR #1543, opened 2021-12-03
Prevent error "Cannot access private property SAML2\XML\saml\AttributeValue::$element"

bartlangelaan:
I don't know exactly how and why this error is happening to us, but a simple try-catch fixed this error for now.

### DIFF
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -240,9 +240,16 @@ class Session implements Serializable, Utils\ClearableState
 
             foreach ($parameters['RawAttributes'] as $attribute => $values) {
                 foreach ($values as $idx => $value) {
-                    // this should be originally a DOMNodeList
-                    /* @var \SAML2\XML\saml\AttributeValue $value */
-                    $this->authData[$authority]['Attributes'][$attribute][$idx] = $value->getElement()->childNodes;
+                    try {
+                        // this should be originally a DOMNodeList
+                        /* @var \SAML2\XML\saml\AttributeValue $value */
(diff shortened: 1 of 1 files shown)


tvdijen:
Hi @bartlangelaan ! What version of SSP are you using and what version of PHP?
And if you've installed SSP using composer, what version of `simplesamlphp/saml2` is mentioned in the `composer.lock` file?

tvdijen:
I'm also curious to see the entire error stack, if you still have it (or are able to reproduce).. I'd rather fix the actual issue instead of ignoring the error with a try-catch

tvdijen:
Are you by any chance using SQL as a session store? I think you may have old sessions created with an earlier version of SSP/saml2-lib that are invalidated after upgrading SSP.. Could that be the case?

bartlangelaan:
Sorry, I was just posting it as a draft so I could easily patch it on an installation of us. I hotfixed it on a server, and changed the same things here.

Investigating futher, I indeed see this is already fixed with https://github.com/simplesamlphp/simplesamlphp/commit/71e781b6868741beec728c24aaff37de6d549b1c#diff-d02a8876603160d5e9bbc2934fdf4cf3892f7859ba4ba33cd55a1e93a25fff2f

We are using ssp 1.18 with saml2 4.2.4. Upgrading to 1.19 should probabily fix this.

tvdijen:
That means you have done a manual upgrade of saml2, because 1.18 shipped with saml2 v3.x.

tvdijen:
I'm not officially releasing it, but I've backported the fix and tagged 1.18.9 for your convenience.
Thanks for reporting!

PR #1551, opened 2021-12-29
Update HTTP.php

abertranb:
Check if port if forwarded too, to avoid issues behind Load Balancer

Best regards


tvdijen:
See: https://github.com/simplesamlphp/simplesamlphp/pull/376#issuecomment-220013415

Please properly configure `baseurlpath` instead.

PR #1552, opened 2022-01-05
Cleanup hooks

thijskh:
Hook 'metadata_hosted' and 'frontpage' are no longer called so do not register hooks for them. Update the docs. Remove useless call to configpage hook on admin module Test page (likely a copy-paste error).


tvdijen (approved):

tvdijen:
Uhm, I just noticed that many modules use hook_frontpage :-/

thijskh:
Probably easiest then to rename hook_configpage to hook_frontpage?

tvdijen:
Yes, I think so!

tvdijen:
Although hook_configpage also gives me plenty of hits when I grep... :(

thijskh:
I've taken a look. The new admin module is a bit different from the former 'front page' pages. The ability to add links there is provided in the `configpage` hook. However the hook is not entirely compatible with hook_frontpage. But as said it's also a bit semantically different.

I checked a few modules and they have a hook_configpage next to a hook_frontpage. My idea is that those could just function now (and possibly clean up the never called hook_frontpage). Those that only have a hook_frontpage should migrate to hook_configpage. I think this is acceptable since 2.0 already requires changes from modules. And best to make this jump now instead of breaking something more later.

So maybe it would be good to add a note to the Modules section of the upgrading notes.

tvdijen:
Thanks for checking it out!  I'll make sure to go over all the modules and fix them..

tvdijen:
I've also noticed references to `hook_frontpage` in `simplesamlphp.github.io`..

thijskh:
That's just the docs which have been fixed already.

tvdijen:
I've made an inventory of modules that need work.. Is it enough to rename the hooks from `hook_frontpage` to `hook_configpage`? And what to do with the modules that currently have both?

``
simplesamlphp-module-aggregator2
simplesamlphp-module-consentadmin
simplesamlphp-module-consentsimpleadmin
simplesamlphp-module-logpeek
simplesamlphp-module-memcachemonitor
simplesamlphp-module-metaedit
simplesamlphp-module-metarefresh
simplesamlphp-module-modinfo
simplesamlphp-module-saml2debug
simplesamlphp-module-sanitycheck
simplesamlphp-module-statistics
``

thijskh:
You cannot rename it since the interface is different. If you have both check if they do the same, then remove hook_frontpage

thijskh:
Old style:
https://github.com/simplesamlphp/simplesamlphp/blob/322ceb4d4ac6cb03ea36aec9c85c6ab1fdd969a9/modules/cron/hooks/hook_frontpage.php

New style:
https://github.com/simplesamlphp/simplesamlphp/blob/36db7a60f1650e7155cadefcfe01ff6c92e079cc/modules/cron/hooks/hook_configpage.php

tvdijen:
Is it me, is the new one broken?  `links` is a multi-dimensional array that can contain multiple links and in your new-style example it is being overwritten with a uni-dimensional array

tvdijen:
Never mind, I've figured it out and all modules have been fixed!

Issue #65, opened 2014-04-09
Single Logout should be propagated from SP to IdP


relaxnow:
SimplSAMLphp, when used as a proxy, does not properly propagate SLO requests from upstream Identity Providers. Received on it's SP endpoint they will simply terminate the SP session.

So in a given setup:
![federation](https://cloud.githubusercontent.com/assets/524386/2719251/ceac9abe-c55d-11e3-8ef6-62899d517697.png)

A SLO Request from a Our SP _will_ be propagated to the Their SPs, but a SLO Request from a Their SP _will not_ be propagated to the Our SPs.

See: https://groups.google.com/forum/#!topic/simplesamlphp/4Bm42hYfGjI

PR #1553, opened 2022-01-06
Support for IdP initiated proxy SLO

### LINKED ISSUE #1564: ADFS IdP loads metadata of disabled SAML IdP
**Describe the bug**
ADFS IdPs load available SAML IdP metadata even though they are not enabled .
In combination with the install instructions provided in the documentation, which suggests copying the whole `metadata-templates/` content to `metadata/`, probably erroneous configuration (in my case certificates not being available as `saml20-idp-hosted.php` expects) is loaded.
This will only interfere when using the created associationGroup, especially during SLO requests.

https://github.com/simplesamlphp/simplesamlphp/blob/43c1caf3fdefe7e900e6c719c7422c96c87bca42/lib/SimpleSAML/IdP.php#L94-L100

**To Reproduce**
1. Setup SimpleSAMLphp as suggested in https://simplesamlphp.org/docs/stable/simplesamlphp-install (especially copying each and every metadata template to active configuration)
2. In `config/config.php`:
```
'enable.saml20-idp' => false,
'enable.shib13-idp' => false,
'enable.adfs-idp' => true,
```
3. Configure `adfs-idp-hosted.php` (e.g. with customized certificate location)
4. Sign in using an ADFS RP
5. Logout leads to an error (unconfigured certificate path of disabled `saml20-idp-hosted.php` not found)

**Expected behavior**
SAML IdP metadata should only be loaded if it has been enabled at all (`'enable.saml20-idp' => true,` ).

m0ark:
# Problem
SimpleSAMLphp was not designed to support protocol proxy setups. Fortunately it supports a lot of those use cases.
One particular feature that has always been missing in proxy scenarios is the capability to properly propagate logout (SLO) requests coming from an upstream identity provider. This is because of SimpleSAMLphp IdPs using its SPs as authentication sources. In case of an IdP initiated logout the proxy SP is notified about the logout request and terminates its session properly without knowing about components it has been used by and thus cannot propagate the logout request to. 
The result of this behaviour is the users IdP receiving a logout response telling the SLO has properly finished but the proxy IdP(s) still having an active session.

This issue has been discussed in #65 and @relaxnow made a great job implementing a first draft to tackle it (https://github.com/relaxnow/simplesamlphp/commit/35cffcca61c02cb2a01e152e801ad3515c3e5803).
Nevertheless this implementation was way too tightly coupled to SAML components and lacked a proper differentiation between multiple identity providers.

# Solution
This patch is based on the work of @relaxnow and adds a small framework for adding IdP initiated proxy SLO capabilities. It also implements the framework for SAML SPs and SAML/ADFS IdPs to support proxy SLO flows using frontend channels.
It should be possible to add proxy SLO capabilities to other IdP/SP protocol implementations (e.g. OIDC, CAS) using this framework without much of a hassle.

To not break the current behaviour proxy SLO has to be activated by setting 
`‘proxyLogoutEnable’ => true`
in each SP configuration.

## Components of the framework
•	`registerAsConsumerLogoutCallback`
To be called by an authentication source consumer (e.g. IdP). The callback should handle everything that is needed to logout the consumer from its own perspective. In case of an IdP this could be logging out of all associated SPs; in case of a PHP application this could terminate the application local session.
•	`hasAsConsumerLogoutCallbacks`
Check if authentication source consumer logout callbacks have been registered for this source.
•	`callAsConsumerLogoutCallbacks`
To be called by a regular authentication source logout routine (e.g. SAML SP SLO handler) to trigger execution of the registered logout callbacks.
•	`handleAsConsumerLogoutCallbacks`
Only used by `callAsConsumerLogoutCallbacks` and the returnTo page that is called in case of the callback not returning immediately (e.g. redirecting, etc).


lib/SimpleSAML/IdP.php
```diff
@@ -90,13 +90,15 @@ private function __construct(string $id)
             }
             $this->config = $metadata->getMetaDataConfig(substr($id, 5), 'adfs-idp-hosted');
 
+            /*
```
thijskh, line 93: This seems misplaced

modules/saml/lib/Auth/Source/SP.php
```diff
@@ -95,13 +102,20 @@ public function __construct($info, $config)
         $this->idp = $this->metadata->getString('idp', null);
         $this->discoURL = $this->metadata->getString('discoURL', null);
         $this->disable_scoping = $this->metadata->getBoolean('disable_scoping', false);
+        $this->proxyLogoutEnabled = $this->metadata->getBoolean('proxyLogoutEnable', false);
```
thijskh, line 105: A new metadata option appears but the docs are not updated to mention it.

thijskh (commented):
Hi! Thanks for the thorough PR.

I have not reviewed it line by line (yet), did add a few observations. And I notice the base branch is simplesamlphp-1.19. It seems unlikely that we'd merge such a feature there and rather target master.

modules/saml/lib/Auth/Source/SP.php
m0ark: My bad. Thanks for pointing out.  
I just added the missing option documentation.

lib/SimpleSAML/IdP.php
m0ark: You're right. This is just a left over of a workaround of a bug we discovered.
I will remove it from this PR and raise a new issue to resolve this.

m0ark: The issue has been raised #1564 and the change been reverted.

m0ark:
Thanks for your first review.
I implemented this change using the version we used in production. So I just wanted to discuss the probability of having the PR accepted first before putting in more effort and porting this change to master.

lib/SimpleSAML/Auth/Source.php
```diff
@@ -218,6 +218,116 @@ public function initLogin($return, $errorURL = null, array $params = [])
...
+     *        Static function has to have following signature: ($param1, ..., $paramN, string $returnUrl) with
+     *        $param1-N being members of $params.
+     *        If callback does not return directly (i.e. redirect) it has to make sure that $returnUrl is called after
+     *        processing finished.
+     * @param array $params Arbitrary parameters.
+     * @throws \Exception
+     */
+    public function registerAsConsumerLogoutCallback($callback, $params = []) {
+        assert(is_callable($callback));
+        assert(is_array($params));
```
thijskh, line 234: We can use type hints here and a return type instead of the assertions. This also goes for other methods: use type hints/return types, and remove assertions that only assert the type.

lib/SimpleSAML/Auth/Source.php
```diff
@@ -218,6 +218,116 @@ public function initLogin($return, $errorURL = null, array $params = [])
...
+     * @return bool
+     * @throws \Exception
+     */
+    public function hasAsConsumerLogoutCallbacks(): bool
+    {
+        $session = Session::getSessionFromRequest();
+        $callbacks = $session->getData('\SimpleSAML\Auth\Source.AsConsumerLogoutCallbacks', $this->getAuthId());
+        if ($callbacks === null) {
+            return false;
+        }
```
thijskh, line 266: null is also empty, so this test+return is unneeded.

lib/SimpleSAML/Auth/Source.php
```diff
@@ -218,6 +218,116 @@ public function initLogin($return, $errorURL = null, array $params = [])
...
+        $session = \SimpleSAML\Session::getSessionFromRequest();
+        $callbacks = $session->getData('\SimpleSAML\Auth\Source.AsConsumerLogoutCallbacks', $authId);
+
+        if (!empty($callbacks)) {
+            // prepare updated returnToUrl in case callback redirects
+            $id = State::saveState($state, 'asConsumerLogoutCallbacks:resume');
+            $returnToUrl = Module::getModuleURL('core/source/as_logout_resume.php', ['id' => $id]);
+
+            foreach ($callbacks as $cb) { // FIXME: use loop variable or get rid of it!
+                $callback = array_shift($callbacks);
```
thijskh, line 313: Yes, please fix this

modules/saml/www/sp/saml2-logout-resume.php
```diff
@@ -0,0 +1,59 @@
...
+    $source->handleLogout($message->getIssuer());
+    $numLoggedOut = count($sessionIndexes);
+}
+
+$lr = \SimpleSAML\Module\saml\Message::buildLogoutResponse($spMetadata, $idpMetadata);
+$lr->setRelayState($message->getRelayState());
+$lr->setInResponseTo($message->getId());
+
+if ($numLoggedOut < count($sessionIndexes)) {
+    \SimpleSAML\Logger::warning('Logged out of ' . $numLoggedOut . ' of ' . count($sessionIndexes) . ' sessions.');
```
thijskh, line 42: Why is this a warning?

thijskh (commented):
I've checked the code and believe it to be OK. However, I don't have a lot of experience with the scenario in question nor do I have a representative environment. So I'd like to add the caveat that I have not verified it to actually work myself. If you use this in actual production I'd be quite confident to merge it though.

Did you give any consideration as to whether at least some basic tests could be added for this?

lib/SimpleSAML/Auth/Source.php
m0ark: Fixed in 7693d74c2d15801603ea39ab6b0a0aaf78122f45

lib/SimpleSAML/Auth/Source.php
m0ark: Fixed in f30ec019c07716d048790fd3292be6b4507db34a

lib/SimpleSAML/Auth/Source.php
m0ark: Fixed in 82f1b492009472bdb6b866beada30efada687b4b

modules/saml/www/sp/saml2-logout-resume.php
m0ark: It's just a copy of the working code of saml2-logout.php.
I think it is not expected to have any active session indexes left of the current SLO request after reaching this point.

m0ark:
Thank you for your thorough review and your helpful annotations.

Do you have any pointers on where to add tests for this code?

For an environment just create a simple SAML proxy and trigger an IdP initiated SLO request.
Without enabling this feature the logout request will not propagate to the proxy IdP component.

image

tvdijen:
Also closes #1565

lib/SimpleSAML/IdP.php
```diff
@@ -484,9 +520,11 @@ public function handleLogoutRequest(array &$state, ?string $assocId): void
         $id = Auth\State::saveState($state, 'core:Logout:afterbridge');
         $returnTo = Module::getModuleURL('core/idp/resumelogout.php', ['id' => $id]);
 
-        $this->authSource->logout($returnTo);
+        if (!$skipAsLogout) {
```
tvdijen, line 523: I'd like to see strict comparison here

lib/SimpleSAML/IdP.php
```diff
@@ -484,9 +520,11 @@ public function handleLogoutRequest(array &$state, ?string $assocId): void
...
+            $this->authSource->logout($returnTo);
+        }
 
-        if ($assocId !== null) {
+        if ($assocId !== null || $skipAsLogout) {
```
tvdijen, line 527: And here

tvdijen (approved):

tvdijen:
I hope I didn't break anything, but I've rebased the lot against `master`..

lib/SimpleSAML/IdP.php
tvdijen: Either remove the commented code or use it...   Or was this something my rebase raised?

thijskh:
I think the rebasing was not entirely accurate, at least i've spotted the reintroduction of this code: https://github.com/simplesamlphp/simplesamlphp/pull/1553/files#diff-60baa677d9f59b02cd5b3bbd371583dc05eccd0cb0bb74f2003398031fd84f8dR109-R111

tvdijen:
Good catch! I've fixed that..  Something else that needs to change is the www-script.. It should be converted to a Controller

lib/SimpleSAML/IdP.php
m0ark: This must have been raised by the rebase. I removed it earlier.

tvdijen:
I'm not sure how to fix the conflicts in `src/SimpleSAML/IdP.php` because the code has been refactored for Symfony responses

Issue #1515, opened 2021-09-07
Functional design of the front page


thijskh:
The current (in 2.0) front page of a SimpleSAMLphp installation presents just a list of available authsources which you can test-login to with no further context, explanation.

Probably this is a confusing landing page. For exising SimpleSAMLphp 1.x users who expected to find the admin interface here and do not see any reference or link to it. For new SimpleSAMLphp users because it's just a list of authsources without any way to know why they are there or what you are supposed to do with it.

I think we need to a new idea of what the root of the SimpleSAMLphp installation page should do and what to display there.

tvdijen:
As discussed elsewhere this should probably be a list of hosted IDPs..
The current situation IMHO is a design flaw that leaves developers with situations like this; https://github.com/cirrusidentity/simplesamlphp-module-ratelimit/issues/7

thijskh:
So if your installation is an SP, what will you display?

tvdijen:
I'd say in case there are no IDPs configured, we could redirect to the admin-module if enabled...
Or maybe even better, a landing page with some links to our website, github repo, etc. that also has the link to the admin module (if enabled).   There is nothing to see or to do for an end-user on an SP.

jaimeperez:
Maybe we should completely stop assuming what the right thing to display is, and just delegate that to the deployment?

I was thinking we could just provide a configuration option so that you can tell SSP where to redirect in case someone reaches the landing page, which would then open up for any custom use you may want to do of it. This could be provided with web server configuration too, though, and then we would spare adding _yet-another-configuration-option_.

I'm not too convinced on the idea of redirecting to the admin module if enabled. Not like it's security should rely on obscurity, of course, but for an end-user that accidentally lands there, there's absolutely no reason why they should see any admin-related stuff. That also has the undesired side effect that if the end user lands there and gets redirected to the admin module, they will be prompted for their username and password, and they might type their own password there without realising where they are. In any case, not a huge fan of this alternative.

The reason why we were using auth sources was precisely because in that case you cover both deployment scenarios: IdPs and SPs. It also resonated as a good idea to have a place where the end users can see their data after logging in. However, I agree that this has its drawbacks, and may lead to exposing auth sources that we don't want to expose. Again, we could just offer a configurable list of auth sources to display there, defaulting to none (and an informative message telling the user they have no business to do in there).

tvdijen:
[...]

PR #1554, opened 2022-01-06
Minimal frontpage

thijskh:
This is a first attempt at implementing the proposed solution to #1515: replace the public front page with a config setting to allow the admin to redirect people somewhere else, otherwise display a very basic message and that's it.

Screenshot of the new minimal front page

I did consider to add a "login as admin" on the top-right but decided against it for now, as per @jaimeperez' comments in 1515.

It's not per se the definitive form, I hope some concrete implementation helps us getting along to decide if this is the direction that we want and if so, how exactly.


config-templates/config.php
```diff
@@ -853,6 +851,12 @@
         ],
     ],
 
+    /**
+     * Set to a full URL if you want to rediret users that land on SimpleSAMLphp's
```
tvdijen, line 855: typo

tvdijen:
This is a no-brainer merge for me, as it takes away the vulnerability of exposing individual authsources.
Perhaps we should consider an 'OpenConext'-like landing-page where we add links to certs + metadata for all entities.

tvdijen (approved):

thijskh:
This removes all code whereby anyone from the general public can (try to) authenticate to any auth source.

You will still keep the option to test auth sources in the admin module of course.

tvdijen:
One more thing that caught my eye is that your header is in Dutch and the rest of the text is in English (screenshot).

thijskh:
That is because I have not yet added translations for the new strings.

tvdijen:
> This removes all code whereby anyone from the general public can (try to) authenticate to any auth source.
This is very very good!

> You will still keep the option to test auth sources in the admin module of course.
Yes, but you can't let a 3rd party user authenticate to get an idea on what attributes they're releasing.. One could argue that this can be done using debug-logging, but that's not something you generally use on a production environment..

I'm thinking more like an IDP-first scenario where you lack to specify the SP-entityid and instead of a redirect to the SP you get an overview of the attributes.. This would mean instead of displaying local authsources, you'd display local/remote IDPs.. I guess what I'm after is not that different from the OpenConext debug-page.

thijskh:
We could make it a module that you can enable?

tvdijen:
Yes, that'd be wonderful!

tvdijen:
It's gonna take some time cleaning up the translations for the core-module.
Some translations now belong to the admin-module (these changes can go straight to `master`),
others have become ununsed and can be removed as part of this PR.

My approach so far has been:
- Search for the fully translated English string to see if it's being used anywhere outside the .po-files; if yes, see if it needs to be moved, if no remove from all .po-files.
- Search for the 'old-style'-tag and use the same approach as above.
- Note that some translations only exist as tag in the .po-files. Try not to remove good translations in the proces.

thijskh:
Well, no one has added any translations to the admin module until now it seems...

tvdijen:
> Well, no one has added any translations to the admin module until now it seems...

I've made a start in `master`.. I found one tag that existed in the base translations and moved it to `admin`... It's a horrible job.

thijskh:
I've automatically migrated all strings that had a full string equivalent in the old files for all languages.

Only thing that is not automatically migrated is when the old files have a "namespaced" translation tag but not the full string version. If we want to port this, we need a mapping of these old tags to the new strings.

tvdijen:
Just so you know, I try and process a couple of tags every day, but I still have two weeks of work left if I continue at this pace...

tvdijen:
Finally done!  I suggest we merge this (please DO NOT squash as the commit would be HUGE).
Let's merge and tag `beta.6`.. Let's keep the issue open as a reminder we have to make a module to replace old behaviour..

Issue #493, opened 2016-10-14
Make tools to manipulate .po-files


hmpf:
- [ ] Script to clean .po-files for namespaced tags
- [ ] Script to extract strings from all files everywhere and update the relevant .po-files
- [ ] Script to interact with for instance transifex

tvdijen:
We were discussing this at the hackathon 2019/3..  There is a tool available @ https://github.com/simplesamlphp/dictionaryconverter

melanger:
Do you know about [Poedit](https://poedit.net/)?

jaimeperez:
Hi @melanger!

The idea here is more in the line of automating the processing of those files (as in, generation, push to transifex and retrieval of the translated files), rather than manipulating the files themselves. Actually, the idea is to use transifex for that, which should help translators.

PR #1558, opened 2022-01-10
Initial version of tool to extract translatable strings from templates.

thijskh:
This is likely not yet perfect or complete but it's a start to have something to extract translatable strings into PO files.

What it can do:
- For a module or the 'base' (= main code + the standard modules), walk all Twig templates for strings in the trans() function plus the PHP code under lib for the (Translate::)noop() function.
- Update the `en` po file for the module/for the global messages if not already present.

Since we currently have nothing of the sort I think this is a good step towards fixing #493.

Some things that could be improved upon (later):
- It does not yet merge the new EN strings into the language POs. This can be done but could also be done by the translator, so not sure if it's in scope. If it should be, is there a global list of what the supported languages are?
- It does not clean up strings from the PO that it not find to be present in the sources. This is probably good to add later if we're reasonably certain that what it finds is complete.

### DIFF
--- a/tests/lib/SimpleSAML/Utils/TranslateTest.php
+++ b/tests/lib/SimpleSAML/Utils/TranslateTest.php
@@ -1,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SimpleSAML\Test\Utils;
+
+use InvalidArgumentException;
+use org\bovigo\vfs\vfsStream;
+use org\bovigo\vfs\vfsStreamDirectory;
+use PHPUnit\Framework\TestCase;
+use ReflectionClass;
+use SimpleSAML\Configuration;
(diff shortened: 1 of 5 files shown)


tests/lib/SimpleSAML/Utils/TranslateTest.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SimpleSAML\Test\Utils;
+
+use InvalidArgumentException;
+use org\bovigo\vfs\vfsStream;
+use org\bovigo\vfs\vfsStreamDirectory;
```
tvdijen, line 9: You don't seem to use vfsStream or vfsStreamDirectory ?

PR #1560, opened 2022-01-17
Switch to npm lockfile v2

thijskh:
This is generated by NPM 7 or newer so if you run "npm install"
with npm 7 you get lockfile changes every time if you have v1.

v2 is [BC with v1](https://github.com/npm/cli/blob/latest/docs/content/configuring-npm/package-lock-json.md#lockfileversion) supported from NPM 5 and up (2017).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,3341 @@
 {
   "name": "simplesamlphp",
+  "lockfileVersion": 2,
   "requires": true,
-  "lockfileVersion": 1,
+  "packages": {
+    "": {
+      "name": "simplesamlphp",
+      "license": "MIT",
+      "dependencies": {
+        "@fortawesome/fontawesome-free": "^5.15.4",
+        "clipboard": "^2.0.6",
+        "es6-shim": "^0.35.6",
+        "highlight.js": "^10.7.0",
+        "jquery": "^3.6.0",
+        "jquery-ui": "^1.13.0",
+        "minimum-node-version": "^3.0.0",
+        "purecss": "^2.0.6",
+        "reset-css": "^5.0.1",
+        "selectize": "^0.12.6"
+      },
+      "devDependencies": {
+        "@babel/core": "^7.16.7",
+        "@babel/helper-compilation-targets": "^7.16.7",
+        "babel-loader": "^8.2.3",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1,7 +1,3341 @@
...
+        "mini-css-extract-plugin": "^1.6.2",
+        "sass": "^1.48.0",
+        "sass-loader": "^12.4.0",
+        "style-loader": "^2.0.0",
+        "webpack": "^5.66.0",
+        "webpack-cli": "^4.9.1"
+      },
+      "engines": {
+        "node": ">=14.0.0",
+        "npm": ">=6.0.0"
```
tvdijen: Shouldn't we raise this to `>=7.0.0` in the package.json file then?

tvdijen: I'm worried the next time I to an `npm install` with npm v6, we end up with a version 1 lockfile again...
I'm pretty sure npm v6 is the one that comes with the major OS vendors right now..

tvdijen: Yeah, so I've done some research.. It will require some workarounds to have npm 7 available on RHEL/Debian, either through using Software Collections or by manually adding their repo. I'm OK with that, since most people will use our tarball and not build from scratch..

I'm OK with raising the minimum requirements to Node v16 / NPM v7 for SSP 2.0

thijskh: But the minimum requirement is 5, not 7.

tvdijen: For the new lockfile-format, yes..  But isn't this the right moment to raise our minimum requirements to the latest LTS-version?

thijskh: Debian stable has 7.5 fwiw

thijskh: Not sure if we should break it for breaking it. We can assume that developers will use npm >= 7 so the v2 lockfile will not revert in the repo (or we will not accept such changes). I see no real reason to block someone from using the v2 lockfile with  npm 6? Or are there other benefits?

tvdijen: How can you be so sure developers will use npm >=7?   I noticed I'm still on v6 on my dev box.. The benefit to me would be that we don't have to worry about accidentally reverting to a v1 lockfile

thijskh:
This has been done now

PR #1561, opened 2022-01-18
Show overview of modules in the admin plugin

thijskh:
Previously, the admin module did not provide any information about which modules you have installed or wheter they are enabled. This seems a logical thing to show there.

This is now combined with the "SAML 2.0 IdP" enabled/disabled flag
which was a bit lonely before, as it's the only flag of its type.

Screenshot showing a table of modules with enabled or disabled icons behind them in the admin module's configuration page

### DIFF
--- a/modules/admin/lib/Controller/Config.php
+++ b/modules/admin/lib/Controller/Config.php
@@ -137,6 +137,7 @@ class Config
             ],
             'funcmatrix' => $this->getPrerequisiteChecks(),
             'logouturl' => $this->authUtils->getAdminLogoutURL(),
+            'modulelist' => $this->getModuleList(),
         ];
 
         Module::callHooks('configpage', $t);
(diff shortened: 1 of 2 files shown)


tvdijen:
I'm not sure...  You've kinda reinvented the modinfo-module here...

thijskh:
Yes. But:
- I think this information really belongs in the admin module, is actually a core part of its functionality. The config page shows various parts of your config, it really seems logical to me that which modules you've enabled is quite an important part, maybe the most important part, of your config.
- It does not make much sense to require users to fetch a separate module, enable it (!), and only then see simply what modules are installed and are enabled.
- Given it's only 10 lines of code and a few lines of template, having that hosted in its own separate module with all the overhead seems overkill.

tvdijen:
I'm fine with it, but can we archive the modinfo-module then, to reduce maintenance?

thijskh:
Definitely!

Issue #1562, opened 2022-01-18
Add option to pass IDPList as authentication parameter


bartlangelaan:
**Is your feature request related to a problem? Please describe.**
We are currently implementing with an IDP that _sometimes_ requires scoping IDPs.

**Describe the solution you'd like**
Allow the property `IDPList` to be provided to `$auth->login()`.

**Describe alternatives you've considered**
Adding the `saml:IDPList` property (undocumented) does work _if_ this statement is adjusted to it always returns false: https://github.com/simplesamlphp/simplesamlphp/blob/959aa29de0830d3f792f6b1f785158f08e67e0a9/modules/saml/lib/Auth/Source/SP.php#L731

But I don't think using `saml:IDPList` is the right option, is it?

bartlangelaan:
No, we are not. We are using an [HM](https://afsprakenstelsel.etoegang.nl/display/as/Interface+specifications+DV-HM), that implements various other IDPs.

We want to skip the HM screen where you can select the IDP you want to use. That is possible if the Scoping element is used, with only one IDPEntry.

tvdijen:
Then all you gotta do is set the `IDPList` in your SP authsource and you're set?

bartlangelaan:
But, I only want to do it _sometimes_.

In some cases, the IDP is preset, so we want to skip the HM screen.

In other cases, the IDP is not predefined, so we want to show the HM screen.

We would like to use the same SP for that. So that's why we want to provide it in the `$sp->login()` method, instead of statically in the SP authsource.

tvdijen:
Ah, I see... But that should already work if you pass it to `$sp->login()` as `saml:IDPList`...
In fact you're now adding `$state['IDPList']`, while the exact same content should already be in the `$IDPList`-array (added in line 534)...

So, the suggested fix seems wrong and we have to figure out what's going on around line 731 in the `authenticate()` method.  The phpdoc kinda already reveals that it assumes we're a proxy because we have an IDPList set, which is a false assumption.. I'll try and investigate this soon.

thijskh:
We had this issue at hand before so I've looked into it before. Have to dig a bit for the specifics though.

There are two scenarios that work with IDPList:

[...]

PR #1563, opened 2022-01-18
Add IDPList parameter

bartlangelaan:
Solves #1562


tvdijen (changes requested):
See linked issue for why we shouldn't merge this as-is.

tvdijen:
Sorry @bartlangelaan. but this doesn't seem right either..  This should go in the larger if-statement above.. We have to differentiate between `IDPList` (local SP) and `saml:IDPList` (saml proxy), so I expect a similar structure as the one for ProxyCount.. The `authenticate` method probably needs work as well

bartlangelaan:
I would expect that if you provided a `IDPList` in the ->login() function, that it would overwrite the SP and IDP metadata. Because it is provided on runtime, it should take full control.

I believe that `IDPList` is what was provided on runtime, and `saml:IDPList` and the SP/IDP metadata is the 'default' logic when nothing was provided on runtime.

thijskh:
I agree that IDPList passed as runtime is likely intended to be an override/complete set, not an addition to other things in Metadata.

I would keep the if but move the lines above it to an else. So you don't set it and overwrite, but set it either way or the other.

Also the reading of saml:IDPlist and setting of $IDPList should be removed from the code as it does nothing.

thijskh:
One could even say that you pick the first of any of these when present, instead of any merging:

1. Runtime
2. IDP config
3. SP config

thijskh:
Fixed in #1653

PR #1565, opened 2022-01-19
Fix: ADFS IdP loads metadata of disabled SAML IdP

### LINKED ISSUE #1564: ADFS IdP loads metadata of disabled SAML IdP

m0ark:
Fixes #1564

### DIFF
--- a/src/SimpleSAML/IdP.php
+++ b/src/SimpleSAML/IdP.php
@@ -91,12 +91,14 @@ class IdP
             }
             $this->config = $metadata->getMetaDataConfig(substr($id, 5), 'adfs-idp-hosted');
 
-            try {
-                // this makes the ADFS IdP use the same SP associations as the SAML 2.0 IdP
-                $saml2EntityId = $metadata->getMetaDataCurrentEntityID('saml20-idp-hosted');
-                $this->associationGroup = 'saml2:' . $saml2EntityId;
-            } catch (\Exception $e) {
-                // probably no SAML 2 IdP configured for this host. Ignore the error
+            if ($globalConfig->getBoolean('enable.saml20-idp', false)) {
+                try {
+                    // this makes the ADFS IdP use the same SP associations as the SAML 2.0 IdP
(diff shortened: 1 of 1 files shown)


tvdijen:
This shouldn't be going into `master`, because ADFS support was removed there.

m0ark:
Even though module-adfs has been moved to another repository, the master branch still contains this ADFS specific piece of code.
So what do you suggest?

m0ark:
I just checked [why](https://github.com/simplesamlphp/simplesamlphp/commit/efd37879a707ffc6912033b8c757322c5aa838c9) this part has been introduced at all.
It seems that this part is going to be obsolete as soon as my other PR (https://github.com/simplesamlphp/simplesamlphp/pull/1553) is accepted since it handles propagating logout requests properly.

Issue #1566, opened 2022-01-20
Remove selectize JS-dependency


tvdijen:
Thee selectize-package hasn't been maintained very well and contains known vulnerabilities for a long time already.
We only use this in two places;

- `modules/core/templates/loginuserpass.twig`
- `templates/_header.twig`

We should find another package to provide this functionality and remove the dependency on the selectize-package (and probably backport this to 1.19 as well).

Possible alternatives;  https://alternativeto.net/software/selectize-js/

thijskh:
Yes, it’s also not very well accessible

tvdijen:
I really like https://github.com/Choices-js/Choices as an alternative.. Mainly because it doesn't depend on jQuery.

thijskh:
But we already depend on jQuery, right?

tvdijen:
Yes, but if we need to patch that (and JQuery has security issues pretty often), we should just be able to update it without having to worry about other packages that still rely on older versions of JQuery

thijskh:
Fixed in #1567

PR #1567, opened 2022-01-21
Remove selectize

thijskh:
This is an arguably opinionated implementation of #1566: just remove Selectize and replace it with nothing.

The available alternatives come with loads of features while we essentially use none. So even "only 23 kB gzipped" is 23 kB mostly unneeded. We use it in two places, for languages in the header and organizations on the login form. The standard dropdowns work just fine there in my opinion.

We can add more complex comboboxes when we have something that actually needs them.


tvdijen (approved):

tvdijen:
Would you agree to backport this to 1.19?

PR #1577, opened 2022-01-27
Small writing error in multiauth.po

Giorgio68:
"and" was repeated twice in msgid and msgstr


tvdijen:
Good catch!  Thanks @Giorgio68

PR #1580, opened 2022-01-28
Implement the PHP 7.4  and  methods in addition to the  and  from the…

ideb:
… interface.

### DIFF
--- a/lib/SimpleSAML/Session.php
+++ b/lib/SimpleSAML/Session.php
@@ -197,17 +197,28 @@ class Session implements \Serializable, Utils\ClearableState
         self::$config = $config;
     }
 
-
     /**
      * Serialize this session object.
      *
      * This method will be invoked by any calls to serialize().
      *
      * @return string The serialized representation of this session object.
+     * @deprecated This method will be removed in SSP 2.0.
      */
-    public function serialize()
+    public function serialize() {
+        return serialize($this->__serialize());
+    }
+
+    /**
+     * Serialize this session object.
+     *
+     * This method will be invoked by any calls to serialize().
+     *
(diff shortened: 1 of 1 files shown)


lib/SimpleSAML/Session.php
```diff
@@ -197,17 +197,28 @@ public function setConfiguration(Configuration $config)
...
     }
 
-
     /**
      * Serialize this session object.
      *
      * This method will be invoked by any calls to serialize().
      *
      * @return string The serialized representation of this session object.
+     * @deprecated This method will be removed in SSP 2.0.
```
tvdijen, line 206: I'm not so sure we can remove this in SSP 2.0 as that will support PHP >= 7.4...    I think we can remove this as soon as the minimum required PHP-version is 8.1?

ideb: The __serialize and __unserialize magic methods are available on PHP7.4. Strictly speaking you could remove Serializable right now, but deprecating these non-magic methods gives other projects a chance to remove calls to these methods.

tvdijen: You seem to have some experience with this..  Would you mind checking [this PR](https://github.com/simplesamlphp/xml-common/pull/12/files) for me?  It's a generic XML-library which will be responsible for processing SAML XML-structures for SimpleSAMLphp in the near future.. Thank you!

ideb: Looks good to me at first glance. Ideally write a unit test for it, so you can check whether a refactoring breaks anything.

tvdijen:
Superseeded by #1597

PR #1581, opened 2022-01-30
Allow to refuse unsolicited responses

tvdijen:
Make it possible to refuse unsolicited responses

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -461,7 +461,7 @@ $config = [
      * In example when you are setting up a federation bridge.
      */
     'enable.saml20-idp' => false,
-    'enable.saml20-unsolicited' => false,
+    'enable.saml20-unsolicited' => true,
     'enable.adfs-idp' => false,
 
 
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -461,7 +461,6 @@ $config = [
      * In example when you are setting up a federation bridge.
      */
     'enable.saml20-idp' => false,
-    'enable.saml20-unsolicited' => true,
     'enable.adfs-idp' => false,
 
 
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -461,6 +461,7 @@ $config = [
      * In example when you are setting up a federation bridge.
      */
     'enable.saml20-idp' => false,
+    'enable.saml20-unsolicited' => false,
     'enable.adfs-idp' => false,
 
 
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -16,6 +16,7 @@ See the upgrade notes for specific information about upgrading.
     to comply with SAML2INT
   * core:PairwiseID and core:SubjectID authprocs no longer support the 'scope' config-setting.
     Use 'scopeAttribute' instead to identify the attribute holding the scope.
+  * Unsolicited responses can denied by disabling it by setting `disable_unsolicited` to `true` in the SP authsource.
 
 ## Version 1.19.1
 
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -16,7 +16,7 @@ See the upgrade notes for specific information about upgrading.
     to comply with SAML2INT
   * core:PairwiseID and core:SubjectID authprocs no longer support the 'scope' config-setting.
     Use 'scopeAttribute' instead to identify the attribute holding the scope.
-  * Unsolicited responses can denied by disabling it by setting `enable_unsolicited` to `true` in the SP authsource.
+  * Accepting unsolicited responses can be disabled by setting `enable_unsolicited` to `false` in the SP authsource.
 
 ## Version 1.19.1
 
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -16,7 +16,7 @@ See the upgrade notes for specific information about upgrading.
     to comply with SAML2INT
   * core:PairwiseID and core:SubjectID authprocs no longer support the 'scope' config-setting.
     Use 'scopeAttribute' instead to identify the attribute holding the scope.
-  * Unsolicited responses can denied by disabling it by setting `disable_unsolicited` to `true` in the SP authsource.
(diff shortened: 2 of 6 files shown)


thijskh:
Comply with which requirement exactly?

tvdijen:
My bad, I thought it was not allowed by SAML2INT or at least discouraged, but it's not...
It's still something we've discussed many times in the past, but no one ever acted on it..

thijskh:
It's good that there's a configuration option for it. I'm not sure we want to disable it by default. That breaks quite some things I think and I'm not sure it's appropriate to do that. E.g. Shibboleth supports it by default aswell, as some kind of measure of what other implementations do.

tvdijen:
Better?

modules/saml/lib/Auth/Source/SP.php
```diff
@@ -1137,6 +1137,15 @@ public static function onProcessingCompleted(array $authProcState): void
...
         $state['Attributes'] = $authProcState['Attributes'];
 
+        $config = Configuration::getInstance();
+        $allowUnsolicited = $config->getBoolean('enable.saml20-unsolicited', true);
+
+        Assert::true(
+            $allowUnsolicited,
+            'Unsolicited responsed are denied by configuration.',
+            Error\BadRequest::class,
+        );
```
thijskh, line 1147: Why is this an Assert and not an exception? And why halt processing completely if the setting is off?

config-templates/config.php
```diff
@@ -463,6 +463,7 @@
      * In example when you are setting up a federation bridge.
      */
     'enable.saml20-idp' => false,
+    'enable.saml20-unsolicited' => true,
```
thijskh, line 466: I wonder if we should add another "enable." setting like this.
For starters, the current `enable.` flags seem a bit random already: you can enable or disable the idp. But not other aspects of SSP in a similar way. But ok, this is kind of historic.

When adding a functional change for SPs, I would add it to the SAML module's SP-related config (and docs), not to the global config. As that is where other behavioural things around SP's are configured.

tvdijen: Not entirely true.. The adfs-idp is also enabled in a similar fashion.
I think I have addressed all your comments though

modules/saml/docs/sp.md
```diff
@@ -219,6 +219,9 @@ Options
     in the IdP-remote metadata overrides this the option in the SP
     configuration.
 
+`disable_unsolicited`
+:    Whether this SP will refuse to process unsolicited responses. The default value is `false`.
+
```
thijskh, line 224: I personally find it better understandable if the toggle is named "enable_unsolicited", defaulting to  "true" over "disable=false".

docs/simplesamlphp-upgrade-notes-2.0.md
```diff
@@ -27,8 +27,10 @@ Functional changes
   validated if present and requests not passing validation will be refused.
 - In the  core:TargetedID authproc-filter, the `attributename` setting has been renamed to `identifyingAttribute`.
 - The default encryption algorithm is set from `AES128_CBC` to `AES128_GCM`.
-  It is possible to switch back via the `sharedkey_algorithm`. Note however that CBC is vulnerable to the Padding oracle attack.
+  It is possible to switch back via the `sharedkey_algorithm`.
+  Note however that CBC is vulnerable to the Padding oracle attack.
 - All support for the Shibboleth 1.3 / SAML 1.1 protocol has been removed.
+- Unsolicited responses can denied by disabling it by setting `disable_unsolicited` to `true` in the SP authsource.
```
thijskh, line 33: Not really an upgrade note I think... more a changelog item?

modules/saml/www/sp/saml2-acs.php
```diff
@@ -151,12 +158,13 @@
 $attributes = [];
 $foundAuthnStatement = false;
 
-foreach ($assertions as $assertion) {
-    // check for duplicate assertion (replay attack)
-    $config = Configuration::getInstance();
-    $storeType = $config->getString('store.type', 'phpsession');
+// check for duplicate assertion (replay attack)
```
thijskh, line 161: This comment is moved to the wrong place, should be inside the foreach.

modules/saml/lib/Auth/Source/SP.php
```diff
@@ -1138,11 +1138,18 @@ public static function onProcessingCompleted(array $authProcState): void
         $state['Attributes'] = $authProcState['Attributes'];
 
         if (isset($state['saml:sp:isUnsolicited']) && (bool) $state['saml:sp:isUnsolicited']) {
+            $spMetadata = $source->getMetadata();
+            $disableUnsolicited = $spMetadata->getBoolean('disable_unsolicited', false);
+            if ($disableUnsolicited === true) {
+                throw new Error\BadRequest('Unsolicited responses are denied by configuration.');
+            }
```
thijskh, line 1145: How can this happen? Since we already refuse it in the ACS?

docs/simplesamlphp-changelog.md
```diff
@@ -16,6 +16,7 @@ See the upgrade notes for specific information about upgrading.
     to comply with SAML2INT
   * core:PairwiseID and core:SubjectID authprocs no longer support the 'scope' config-setting.
     Use 'scopeAttribute' instead to identify the attribute holding the scope.
+  * Unsolicited responses can denied by disabling it by setting `enable_unsolicited` to `true` in the SP authsource.
```
thijskh, line 19:
```suggestion
  * Accepting unsolicited responses can be disabled by setting `enable_unsolicited` to `false` in the SP authsource.
```

modules/saml/lib/Auth/Source/SP.php
```diff
@@ -1138,11 +1138,14 @@ public static function onProcessingCompleted(array $authProcState): void
         $state['Attributes'] = $authProcState['Attributes'];
 
         if (isset($state['saml:sp:isUnsolicited']) && (bool) $state['saml:sp:isUnsolicited']) {
+            $spMetadata = $source->getMetadata();
```
thijskh, line 1141: This is now unnecessary/unused

thijskh (approved):
Apart from two minor changes, good to go

thijskh (approved):

PR #1583, opened 2022-02-04
Refactor Configuration getters

tvdijen:
This slightly changes the behaviour of the getters.  If no default is passed at all, the option will be considered 'required' and an exception will be thrown if it doesn't exist. It also adds typehints to the default value, instead of accepting literally _anything_..
This means that when you're calling `getString()` you can be certain that the returned value is either `null` or a string, and not something completely different.. This also improves static analysis.

### DIFF
--- a/lib/SimpleSAML/Configuration.php
+++ b/lib/SimpleSAML/Configuration.php
@@ -710,29 +710,26 @@ class Configuration implements Utils\ClearableState
      *
      * @return mixed The option with the given name, or $default if the option isn't found and $default is given.
      *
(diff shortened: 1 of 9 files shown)


tvdijen:
Will come up with a new PR after some discussion with @jaimeperez on this subject.
The current getters will become even more strict than what this PR anticipated on.. The value _must_ exist and the return-type is guaranteed to be of the expected type.
Additionally we're adding a getOptional* variant that returns null or the expected type. The 'default' parameter will no longer be optional here.  It should be much clearer that when using getOptional*, you _must_ check the return-value for a possible `null`-case.

PR #1584, opened 2022-02-05
Change logging date formatting string from strftime() to date().

thijskh:
strftime is deprecated and will be removed in PHP 9. SSP 2.0 is our chance to change this format as it's exposed to installers of SSP.

The approch is to keep it simple and just change the format to PHP's built in date() function which remains supported, in one go. As this is SSP 2.0 we're already breaking some config options, and if we make the jump now we can keep the code simple after instead of painstakingly trying to keep compatble also with the old format.

Also we do not try to localize the logging date/timestamps anymore. This matches the syslog approach that this format tried to mimic. We are using the day without leading zeroes now also to be a bit closer to how syslog does it. All just keeping things simple, which is what logging should be.

Likely not many installs are even using custom logging formats at all.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -302,7 +302,7 @@ $config = [
      * are:
      *
      * - %date{<format>}: the date and time, with its format specified inside the brackets. See the PHP documentation
-     *   of the strftime() function for more information on the format. If the brackets are omitted, the standard
+     *   of the date() function for more information on the format. If the brackets are omitted, the standard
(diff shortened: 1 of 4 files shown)


tvdijen (approved):
The only thing that we need to check in this regard is the statistics-module.... It depends on a specific format for the logging..  Other than that, this is a go!

thijskh:
That works with `strtotime()` which I've confirmed to work equally well on the new output.

PR #1595, opened 2022-02-19
Add controller for loginuserpass/loginuserpassorg

tvdijen:


modules/core/lib/Controller/Login.php
```diff
@@ -43,11 +64,391 @@ public function __construct(
...
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return \SimpleSAML\XHTML\Template
+     */
+    public function loginuserpass(Request $request): Template
+    {
+        // Retrieve the authentication state
+        if (!$request->query->has('AuthState')) {
+            throw new Error\BadRequest('Missing AuthState parameter.');
+        }
+        $authStateId = $request->query->get('AuthState');
```
thijskh, line 113: Should we do some kind of basic validation on this input variable before we try to load a state with it?

modules/core/lib/Controller/Login.php
```diff
@@ -43,11 +64,391 @@ public function __construct(
...
+
+                $cookie = new Cookie(
+                    $source->getAuthId() . '-username',
+                    $username, // value
+                    $expire, // expire
+                    '/', // path
+                    '', // domain
+                    true, // secure
+                    true, // httponly
+                );
```
thijskh, line 181: Should these things not derive from config options we have for cookies?
Maybe these could be handled centrally in the Cookie class instead of at every point that sets a cookie?

modules/core/lib/Controller/Login.php
```diff
@@ -43,11 +64,391 @@ public function __construct(
...
+
+    /**
+     * This page shows a username/password/organization login form, and passes information from
+     * into the \SimpleSAML\Module\core\Auth\UserPassBase class, which is a generic class for
+     * username/password/organization authentication.
+     *
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return \SimpleSAML\XHTML\Template
+     */
+    public function loginuserpassorg(Request $request): Template
```
thijskh, line 265: This is extremely duplicated from above.

modules/core/templates/loginuserpass.twig
```diff
@@ -38,7 +38,7 @@
     <p>{{ 'A service has requested you to authenticate yourself. Please enter your username and password in the form below.'|trans }}</p>
     <br>
     <div class="center">
-        <form class="pure-form pure-form-aligned center-form" action="?" method="post" name="f">
+        <form class="pure-form pure-form-aligned center-form" action="{{ '?AuthState' ~ AuthState }}" method="post" name="f">
```
thijskh, line 41: This misses a `=` after ?AuthState, and likely we should be urlencoding the value of AuthState?

modules/core/lib/Controller/Login.php
tvdijen: I don't mind if we do, but we never did any of that before..

modules/core/lib/Controller/Login.php
tvdijen: We have settings for the session-cookie and the language-cookie.. Nothing else..
We _could_ use `Utils\HTTP::isHttps()` to figure out if we need to set the `secure` flag to true or false, but is there really a valid reason today to set it to false? httponly is a flag that depends on the specific cookie....

modules/core/lib/Controller/Login.php
tvdijen: Maybe we can put some common code in private methods on the controller, or make or make one controller smart enough to deal with both cases.. They already share a template together..

modules/core/lib/Controller/Login.php
thijskh: Alright, maybe not in scope now then. Maybe we could at some point make these cookie settings more generic instead of config per-cookie.

modules/core/lib/Controller/Login.php
thijskh: Yes, I think this is one reason why we'd do controllers. So we can get rid of duplicated processing easily. So if we move to controllers we should not be adding nearly the same code twice.

modules/core/templates/loginuserpass.twig
tvdijen: Good catch! I've thought about urlencoding, but we don't do that anywhere else where we put the AuthState in the url...  I wasn't sure it's really necessary..
I think the easiest way to ensure that it doesn't cause any problems is to ensure the AuthStateID is always url-safe to begin with..

tvdijen: This should be url-safe, right?
https://github.com/simplesamlphp/simplesamlphp/blob/master/lib/SimpleSAML/Utils/Random.php#L25

thijskh: It should be. But I think the really safe approach is to urlencode things at the point where we're putting them in urls. E.g. for the case someone changes the algo in random, not aware that there's an implicit requirement that it's url safe. This requirement is also not guaranteed by documentation or unit tests..

tvdijen: It's actually an  easy fix.. Our moduleURL filter can deal with parameters and ensures proper urlencoding.
I will add a task on my backlog to check and fix every single template

modules/core/lib/Controller/Login.php
thijskh: So the current code also has all this duplication. So in a sense, merging this should be OK because it improves over the current situation, albeit not solving all its problems.

Getting rid of (more parts of) the duplication would of course make it even better.

tvdijen:
@thijskh I think this is about as much reduction of code-duplication as I could get.. I'm not sure it's an improvement though..

thijskh:
Even though it does not look perfect yet, I think it's definitely an improvement: over the old code because it's now in controllers, and because there is in fact a lot less duplication. I will give it a bit of testing and then let's just merge it. And we can always improve on it later.

tests/modules/core/lib/Controller/LoginTest.php
```diff
@@ -82,6 +93,145 @@ public function testLogout(): void
...
+
+        $c = new Controller\Login($this->config);
+
+        $this->expectException(Error\BadRequest::class);
+
+        $c->loginuserpassorg($request);
+    }
+
+
+    /**
```
thijskh, line 182: Any reason this is still disabled?

tvdijen:
```
1) SimpleSAML\Test\Module\core\Controller\LoginTest::testLoginUserPassOrg
SimpleSAML\Error\NoState: NOSTATE

/apps/development/simplesamlphp/lib/SimpleSAML/Auth/State.php:282
/apps/development/simplesamlphp/modules/core/lib/Auth/UserPassOrgBase.php:334
/apps/development/simplesamlphp/modules/core/lib/Controller/Login.php:145
/apps/development/simplesamlphp/modules/core/lib/Controller/Login.php:342
/apps/development/simplesamlphp/tests/modules/core/lib/Controller/LoginTest.php:229
phpvfscomposer:///apps/development/simplesamlphp/vendor/phpunit/phpunit/phpunit:97
```

I didn't see an easy way around this..

thijskh:
Will you resolve the conflicts?

PR #1596, opened 2022-02-19
2.0:  PHP 8.1 support

tvdijen:
This PR takes care of all deprecation issues.
This is necessary because PHP 8.1 will treat deprecations as exceptions instead of warnings.


thijskh:
I think it's OK, but must say I'm not intimately familiar with all the serializing differences between the different PHP versions...

PR #1597, opened 2022-02-19
1.19: PHP 8.1 support

tvdijen:


thijskh:
I think it's OK, but must say I'm not intimately familiar with all the serializing differences between the different PHP versions...

tvdijen:
This may help in your understanding; https://php.watch/versions/8.1/serializable-deprecated
It will also make you understand the differences with the other PR.. There we drop the Serializable interface and the old interface-methods, because since PHP 7.4 the magic methods take advantage over the interface methods

tvdijen:
I've field-tested this with PHP 7.1, 7.4 and 8.1... Fairly confident this works as it's supposed to

PR #1600, opened 2022-02-23
Upgrade Gettext

### LINKED ISSUE #1367: Migrate gettext

### LINKED ISSUE #1599: 2.0: translations for modules don't work
Translations are broken for the modules, probably due to the gettext-upgrade that was done last week.

tvdijen:
TODO:  figure out why translations won't work on modules.. Probably something to do with translation domains.

Closes #1367
Closes #1599


lib/SimpleSAML/Locale/Translate.php
```diff
@@ -81,7 +81,8 @@ public static function translateSingularGettext(?string $original): string
         // This may happen if you forget to set a variable and then run undefinedVar through the trans-filter
         $original = $original ?? 'undefined variable';
 
-        $text = BaseTranslator::$current->gettext($original);
+        $loc = new Localization(Configuration::getInstance());
```
thijskh, line 84: I believe this is the main thing that breaks it. Everytime a string is translated, an new Localization object is initialized so all things done to the Localization object associated to the template are not used, 'added' domains are gone etc.

The translation is done on a 'default' initialized Localization object (plus: for every string that is translated, the main PO file is loaded and parsed...)

Not sure yet what a better solution is but pretty sure that this is at least part of the cause.

PR #1601, opened 2022-02-25
Use Symfony/filesystem

tvdijen:
Instead of making our lives difficult, we should be using the already available `symfony/filesystem`.
Added bonus is that these classes can be DI'ed for unit testing purposes.

Note: I haven't field-tested this change _yet_ and we may want to add more cases.

### DIFF
--- a/lib/SimpleSAML/Logger/FileLoggingHandler.php
+++ b/lib/SimpleSAML/Logger/FileLoggingHandler.php
@@ -7,6 +7,9 @@ namespace SimpleSAML\Logger;
 use SimpleSAML\Configuration;
 use SimpleSAML\Logger;
 use SimpleSAML\Utils;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\HttpFoundation\File\Exception\CannotWriteFileException;
+use Symfony\Component\HttpFoundation\File\File;
 
 /**
  * A logging handler that dumps logs to files.
@@ -45,6 +48,11 @@ class FileLoggingHandler implements LoggingHandlerInterface
     /** @var string */
     protected string $format = "%b %d %H:%M:%S";
 
+    /**
+     * @var \Symfony\Component\Filesystem\Filesystem;
+     */
+    protected Filesystem $fileSystem;
+
 
     /**
      * Build a new logging handler based on files.
@@ -52,6 +60,8 @@ class FileLoggingHandler implements LoggingHandlerInterface
      */
     public function __construct(Configuration $config)
     {
+        $this->fileSystem = new Filesystem();
+
         // get the metadata handler option from the configuration
         $this->logFile = $config->getPathValue('loggingdir', 'log/') .
             $config->getOptionalString('logging.logfile', 'simplesamlphp.log');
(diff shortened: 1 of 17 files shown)


lib/SimpleSAML/Logger/FileLoggingHandler.php
```diff
@@ -121,8 +133,20 @@ public function log(int $level, string $string): void
                 array_push($replacements, date($format));
             }
 
-            $string = str_replace($formats, $replacements, $string);
-            file_put_contents($this->logFile, $string . \PHP_EOL, FILE_APPEND);
+            if (preg_match('/^php:\/\//', $this->logFile)) {
+                // Dirty hack to get unit tests for Windows working.. Symfony doesn't deal well with them.
```
tvdijen, line 137: See: https://github.com/symfony/symfony/pull/29661#issuecomment-1055651509

tvdijen:
Still have to test this, but a first review wouldn't hurt

lib/SimpleSAML/Metadata/Sources/MDQ.php
```diff
@@ -130,41 +144,47 @@ private function getFromCache(string $set, string $entityId): ?array
...
-        $cachefilename = $this->getCacheFilename($set, $entityId);
-        if (!file_exists($cachefilename)) {
+        $cacheFileName = $this->getCacheFilename($set, $entityId);
+        if (!$this->fileSystem->exists($cacheFileName)) {
             return null;
         }
-        if (!is_readable($cachefilename)) {
-            throw new \Exception(__CLASS__ . ': could not read cache file for entity [' . $cachefilename . ']');
+
+        $file = new file($cacheFileName);
```
thijskh, line 152: `new File`?

thijskh, line 152: `new File`?

lib/SimpleSAML/Metadata/Sources/MDQ.php
```diff
@@ -186,12 +206,14 @@ private function writeToCache(string $set, string $entityId, array $data): void
             return;
         }
 
-        $cachefilename = $this->getCacheFilename($set, $entityId);
-        if (!is_writable(dirname($cachefilename))) {
-            throw new \Exception(__CLASS__ . ': could not write cache file for entity [' . $cachefilename . ']');
+        $cacheFileName = $this->getCacheFilename($set, $entityId);
+        $file = new File($cacheFileName);
+        if (!$file->isWritable()) {
+            throw new Exception(sprintf('%s: could not write cache file for entity [%s]', __CLASS__, strval($file)));
```
thijskh, line 212: Let exception of appendtofile handle this...?

thijskh, line 212: Let exception of appendtofile handle this...?

thijskh (commented):
Looks good in principle. It does save some code we can reuse from Symfony. I would think to depend as much as possible on this code: one advantage of changing this is that we can use the error handling that the Symfony functions provide over the builtin php functions. So I'd tend to 'just call' whatever operation is what we want to do and trust that symfony/filesystem will throw the appropriate excetion e.g. on unreadability.

lib/SimpleSAML/Logger/FileLoggingHandler.php
tvdijen: Because I didn't get a response, I've opened an issue for this: https://github.com/symfony/symfony/issues/45749

lib/SimpleSAML/Metadata/MetaDataStorageHandlerSerialize.php
```diff
@@ -215,38 +201,37 @@ public function getMetaData(string $entityId, string $set): ?array
...
+        $old = new File($this->getMetadataPath($entityId, $set), false);
+        $new = new File($old->getPathName() . '.new', false);
+
+        $loc = new File($old->getPath(), false);
+        if (!$loc->isDir()) {
+            Logger::info('Creating directory: ' . $loc);
+            try {
+                $this->fileSystem->mkdir($loc, 0777);
+            } catch (IOException $e) {
+                Logger::error('Failed to create directory ' . $loc . ': ' . $e->getMessage());
```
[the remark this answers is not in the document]
tvdijen: In this file, we catch the error and return either `null` or `false`.. If we don't do that, the exception will break execution.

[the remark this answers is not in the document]
thijskh: I don't follow: Logger:error will break execution, also?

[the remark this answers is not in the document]
tvdijen: Now I don't follow..  Logger::error only logs something.. The line underneath `false` is returned and the caller can handle that...   If we don't catch the IOException, the caller never regains control..

PR #1604, opened 2022-03-04
Migrate logout-scripts to controllers

tvdijen:


tvdijen:
I still have to add tests for the controllers, but this may be reviewed already.

The [js-file](https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-1.19/modules/core/templates/logout-iframe.tpl.php#L61) appeared to be not in use [anymore ](https://github.com/simplesamlphp/simplesamlphp/blob/72dd5a978231def2625f6ebe4f8a5d723446d54e/modules/core/templates/logout-iframe.twig#L6) since Jaime's rewrite of it. I wonder if it ever worked, because the path lacks the `idp` dir.. I've removed the file.

tvdijen:
For some reason I end up on the welcome screen when testing iframe-logout..
I have it working with `dev-master`, so it must be something in this change..

lib/SimpleSAML/IdP.php
```diff
@@ -482,7 +482,7 @@ public function handleLogoutRequest(array &$state, ?string $assocId): void
 
         // terminate the local session
         $id = Auth\State::saveState($state, 'core:Logout:afterbridge');
-        $returnTo = Module::getModuleURL('core/idp/resumelogout.php', ['id' => $id]);
+        $returnTo = Module::getModuleURL('core/logout/resume', ['id' => $id]);
 
         $this->authSource->logout($returnTo);
```
tvdijen, line 487: This method never returns during iframe-logout... (but it does on master)..

PR #1605, opened 2022-03-07
Refactor Configuration getters

tvdijen:
The current getters are too loosely typed and can return virtually anything as a default value.
This PR changes this and makes the signatures as strict as we possibly can;  The value must exist and the return-type is guaranteed to be of the expected type.
Additionally we're adding a getOptional* variant that returns null or the expected type. The 'default' parameter will no longer be optional here. It should be much clearer that when using getOptional*, you must check the return-value for a possible null-case. This also helps static analysis tools; Psalm is able to detect +4% of the var-types.


tvdijen:
No obvious issues while testing.. Some module may need an update

thijskh:
Is it (more or less) BC for modules or not?

tvdijen:
@thijskh Did you have any more concerns, or can I merge this?  I will take care of the modules

thijskh:
Alright, let's merge it

PR #1606, opened 2022-03-10
AttributeAlter:  Add option to merge target attribute values

tvdijen:
Use case for this would be to run two consecutive AttributeAlter filters using different source attributes, but the same target-attribute. It is implemented in a backwards compatible way, however, I feel that merging would be the logical default.

### DIFF
--- a/modules/core/docs/authproc_attributealter.md
+++ b/modules/core/docs/authproc_attributealter.md
@@ -39,10 +39,10 @@ Parameters
     If no other values exist, the attribute will be removed completely.
     This parameter is OPTIONAL.
 
-`merge`
-:   A boolean indicating whether the altered values must be merged with the
-    target attribute values, or if the target attribute should be overwritten.
-    Defaults to `false`.
+`%merge`
+:   Indicates whether the altered values must be merged with the target attribute values. The default
+    behaviour is to overwrite the target attribute completely.
+    This parameter is OPTIONAL.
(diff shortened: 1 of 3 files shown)


thijskh (commented):
Good addition, but also good to keep it BC. Given the many `if` branches now in the auth proc filter that this further increases, this must have unit tests to remain maintainable.

tvdijen:
Please see if my changes make sense.. I'll try to add tests later tonight

thijskh:
Yes, I think it should be OK but indeed, so many ifs, tests are welcome

tvdijen:
I kinda need this like.. yesterday..  It's not exactly semver, but would you mind if I backport this to 1.19.6?

thijskh (approved):

thijskh:
Well, isn't the easiest/cleanest way to provide it in 1.19 as a module?

tvdijen:
What do you mean? The authproc already exists in 1.19.. I only added the merge-flag..
Are you suggesting that I should make a custom module to use this functionality in 1.19?

thijskh:
Isn't that a simple solution without requiring to backport this to 1.19? Just copy the file, basically?

tvdijen:
It's not a sustainable solution, but I'll guess I'll have to deal with it (and hope I don't forget about it the next time I deploy an update)..

PR #1608, opened 2022-03-17
Never cache error responses

### LINKED ISSUE #1149: Better Caching directives
I noticed some situation whereby cache-control is not set to its best to tell browser not to cache.

In Particular I suggest this change in doRedirect() function in file  lib/SimpleSAML/Utilities.php line 265 shoudl be changed as follows:

From : 
`      // Disable caching of this response  ` 
`        header('Pragma: no-cache');`   
 `       header('Cache-Control: no-cache, must-revalidate');`   
 
To: 
`// Disable caching of this response`
      `  header('Pragma: no-cache');`
        `header('Cache-Control: no-cache, no-store, must-revalidate');`
  

I've also noticed that when an exception is thrown for unsupported passive authentication, SSOService.php will return a 200 with javascript to redirect for POST, with no Cache-Control and Pragma Headers. Unfortunately, I could not locate the right code to edit to have the headers inserted.

tvdijen:
Closes #1149

### DIFF
--- a/lib/SimpleSAML/Error/Error.php
+++ b/lib/SimpleSAML/Error/Error.php
@@ -279,6 +279,7 @@ class Error extends Exception
                 }
             }
 
+            $t->headers->set('Cache-Control', 'no-store, no-cache, must-revalidate, post-check=0, pre-check=0');
             $t->setStatusCode($this->httpCode);
             $t->data = array_merge($t->data, $data);
             $t->send();


tvdijen:
I wasn't aware that statuscodes play a role in caching.. I thought it would be good to be explicit.
I've ran a test with newui and throwing a NoPassive error and it sets a 500 error code, so we should be fine.
Given the age of the issue, and the references to _really_ old `SimpleSAML_Utilities`, I suggest we close both this PR and the issue.

PR #1609, opened 2022-03-20
Replace JQuery with native javascript

tvdijen:
I've tested clipboard / syntax highlight / logout with Chrome/Firefox/Edge..


thijskh:
Probably modules expect jQuery to be there and try to use it?

tvdijen:
Yes, likely.. I'm not sure how we can deal with that..
I've tried to make webpack/Babel to produce all ES5 compliant code, but I can't get it to work..
Too much stuff is currently loaded that isn't needed for the authentication process.

tvdijen:
I've checked and the three modules that rely on jQuery (discopower, statistics and metaedit) all three include their own copy of jQuery

tvdijen:
So, after some discussion with Jaime, we've come to the following conclusion:
1. The bundle.js should not be included in the base-template, but should be added when necessary
2. A smaller main.js should be included in the base-template, only holding the JS for the language picker
3. Modules should provide their own bundles, i.e. include a package.json / webpack config and have their assets pre-build during their build-process. This means that, for example, the admin-module should be externalized  as a git-module so we can move the highlight.js dependency there.

After this, our bundle will probably only contain the expander and clipboard. You'd have 3 js-files at most;  main, bundle, and a module-specific one.  @thijskh does that make sense to you?

thijskh:
I don't see the advantage of 3. It it's part of the main repo, we can just keep building the scripts as we do now?

thijskh:
Copy to clipboard and expander are trivally small. Why not keep all these things in one file? The overhead of fetching the second file is probably an order larger than adding it to the main file. Generically useful utilities of just a few lines can be better just always shipped so they're instantly available.

modules/core/www/assets/js/loginuserpass.js
```diff
@@ -1,7 +1,12 @@
-$(document).ready(function () {
-    $('#submit_button').on('click', function () {
-        $(this).attr('disabled', 'disabled');
-        $(this).html($(this).data('processing'));
-        $(this).parents('form').submit();
-    });
+ready(function () {
+    // your code here
```
thijskh, line 2: Copied from an example?

modules/core/www/assets/js/loginuserpass.js
```diff
@@ -1,7 +1,12 @@
...
+    var button = document.getElementById("submit_button");
+    button.onclick = function () {
+        this.innerHTML = button.getAttribute("data-processing");;
```
thijskh, line 5: Spurious `;` at end of line

src/js/main.js
```diff
@@ -0,0 +1,69 @@
+import "es5-shim";
```
thijskh, line 1: Add a `'use strict;'` to the top of each js file?

tvdijen:
> we can just keep building the scripts as we do now?

We could, but then the generated bundle is not ES5 compliant and I spent the entire weekend trying to fix that.  It always starts with the ES6-only arrow function.. I've compared this with OC Engineblock and there it doesn't have the arrow function, but I can't figure out the difference in babel config.

src/js/main.js
tvdijen: I've created several modules now.. import is always strict, so we don't have to add it everywhere..

src/js/logout/main.js
```diff
@@ -1,5 +1,7 @@
...
+
 import SimpleSAMLLogout from './logout.js';
 
-$(document).ready(function () {
-    new SimpleSAMLLogout($('body').attr('id'));
-});
\ No newline at end of file
+ready(function () {
+    new SimpleSAMLLogout(document.body.id);
+});
```
tvdijen, line 7: I'm experiencing some kind of timing-issue that causes a "partial logout" instead of a "global logout" message even though all SPs have been logged out.. When I revert this to the jQuery-equivalent things start working again..
@thijskh Any ideas here?

jaimeperez: It could be an issue with the extra level of indirection here. Before, jquery was loaded, and then when we reached this code we would use it to register the handler. Now it is kind of inverted. We need to load `ready.js` first so that `ready()` is declared at this point, but the handler is registered back there in `ready.js`. So if the state changes *before* we reach this code, it would run `handleState()`, check the `window.readyHandlers` list, and it will be empty, so it won't do anything else and our code will never be run.

This is just speculation, I'm not sure that's what's happening here, but it could be...

tvdijen: OK, so I replaced `ready.js` with event handlers all over the place and now the Office webviewer throws syntax errors again.. :(

tvdijen: The issue existed in `master` already and IframeLogout is broken in 1.19 (at least for the newui), so I don't see this as a blocking issue anymore.

tvdijen:
This PR is done now.. Two known issues still exist:
- IFrame-logout is not reliable, likely because of a timing issue / race condition.. The issue already existed in master, and IFrame-logout is broken completely in 1.19, so no reason to not merge this PR.
- The login-screen isn't properly centered in the Office Webviewer, especially if you also have a preprod-warning in place. In my fork I've fixed this by setting focus to the submit-button first and then move focus to the username-field. I'm not sure if we'd want this sort-of hack in the release.

@thijskh Would you mind going over this one more time?

tvdijen:
Regarding the second point, instead of switching focus several times, it seems the right thing to set the viewport to the actual content when a small screen is being used.

PR #1610, opened 2022-03-21
Instruct embedded IE to behave like the most modern browser they can

thijskh:
This should prevent issues with e.g. Javascript when using SSP to log into MS applications via their embedded browsers. WIth this setting it should be maximally modern.

### DIFF
--- a/templates/base.twig
+++ b/templates/base.twig
@@ -4,6 +4,7 @@
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
     <meta name="viewport" content="initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="IE=Edge">
     <title>{{ pagetitle }}</title>
     <link rel="stylesheet" href="{{ asset("css/stylesheet.css") }}">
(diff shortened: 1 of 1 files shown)


tvdijen:
I was just reading about this setting!  Running tests with it now

tvdijen:
It doesn't solve all my issues, but at least some!

tvdijen:
Do you know how you guys got rid of arrow functions in Engineblock?
Babel now outputs like [this](https://guest.stepup-tst.ssc-ict.overheid-i.nl/assets/js/bundle.js) and the Office Webviever breaks with a syntax error on line 2, character 4...  Looking at EB's bundle, you have apparently found a way around this.

thijskh:
I think [like this](https://github.com/OpenConext/OpenConext-engineblock/blob/master/theme/.browserslistrc) or [like this](https://github.com/OpenConext/Stepup-Webauthn/commit/e1c0e47aa7746ae44a92e8efab3685ad09ea8739)

tvdijen:
YES! That did the job!

PR #1623, opened 2022-04-19
WIP: Add controllers for the saml-module

tvdijen:


tvdijen:
Could perhaps use some more units tests, definitely needs *testing*..  Coverage decrease is because the old www-scripts were excluded from coverage.

tvdijen:
I've tested this with an SP and the ACS-endpoint works fine, as well as the discovery service.
I've also made sure the legacy-endpoints work, so that upgrading doesn't have to involve metadata-exchange.

PR #1627, opened 2022-04-26
base: Migrate www-scripts to Controllers

tvdijen:
Needs tests + _testing_


tvdijen:
I think it might make sense to move these controllers to the saml-module since they both deal with SAML-endpoints

thijskh:
Functionally it should be right to move these to the saml module. Whether that's easy to achieve though, I'm not certain. You could merge this now since it improves over the previous situation.

tvdijen:
It's fairly easy to do so.. It's just that we need to keep the legacy routes in the base..
I'll move them and add tests before I merge

tvdijen:
I see one issue that we need to discuss.. By migrating to controllers we've basically changed all of our endpoints. We have legacy routes in place for that for backwards compatibility, however the metadata-endpoint is also our entityID and I don't think we can solve that one without pain.  Changing the entityID would mean that upgrading to 2.0 implies exchanging metadata with your peers..

PR #1631, opened 2022-05-02
Patch 1

liedekef:
While it might no longer be needed in the twig-based 2.* version, this is an improvement for the federation-page and translated names that would display as "Array" in the twig-layout of 1.19


tvdijen:
Seems something went wrong here.. Did you change the base to `master` instead of `simplesamlphp-1.19`?

liedekef:
Seems wrong indeed, I'll cancel this

PR #1633, opened 2022-05-02
Replace remove fingerprint-check with proper signature verification

### LINKED ISSUE #1629: 2.0: restore MDQ signature verification
**Describe the bug**
See https://groups.google.com/g/simplesamlphp/c/eH8bhLokkVE
We have removed fingerprint-verification because it isn't very safe, but we didn't replace it with actual full-on signature verification.

**Expected behavior**
A signature verification should be done at all time. Can't trust any metadata blindly.

tvdijen:
Closes #1629 

Needs testing as I don't have a test-setup with MDQ in place.

### DIFF
--- a/lib/SimpleSAML/Metadata/Sources/MDQ.php
+++ b/lib/SimpleSAML/Metadata/Sources/MDQ.php
@@ -282,15 +282,15 @@ class MDQ extends MetaDataStorageSource
             $data = null;
         }
 
-        if ($data !== null && array_key_exists('expires', $data) && $data['expires'] < time()) {
-            // metadata has expired
-            $data = null;
-        }
-
         if (isset($data)) {
-            // metadata found in cache and not expired
-            Logger::debug(sprintf('%s: using cached metadata for: %s.', __CLASS__, $entityId));
-            return $data;
+            if (array_key_exists('expires', $data) && $data['expires'] < time()) {
+                // metadata has expired
+                $data = null;
+            } else {
+                // metadata found in cache and not expired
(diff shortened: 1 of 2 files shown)


lib/SimpleSAML/Metadata/Sources/MDQ.php
```diff
@@ -309,6 +320,12 @@ public function getMetaData(string $entityId, string $set): ?array
         $entity = SAMLParser::parseString($xmldata);
         Logger::debug(sprintf('%s: completed parsing of [%s]', __CLASS__, $mdq_url));
 
+        if (!empty($this->validateCertificate)) {
+            if (!$entity->validateSignature($this->validateCertificate)) {
+                throw new Exception(__CLASS__ . ': error, could not verify signature for entity: ' . $entityId . '".');
+            }
+        }
```
thijskh, line 327: Should we log that we're not validating anything in an `else` clause? At the very least for debugging, but maybe even as high as `notice` makes sense?

At least the current situation does not comply with your stated goal that verification must be done always.

tvdijen: I've been discussing this with Jaime last night.. He feels like we shouldn't do _any_ signature verification here because the integrity is already guaranteed at TLS-level..  I am no longer sure what to do here.

thijskh: I think the current code guarantees nowhere that the configured URL is TLS.

It's fine to keep it optional with me but I think we'd need to have the signature verification just like we have on other forms of metadata (because regular metadata is also routinely served over TLS). Would in any case add the else-clause with logging though, even if only at info level. Just so you can follow what it's doing.

PR #1634, opened 2022-05-03
Add extra optional header template

liedekef:
this allows people to add extra header-lines without needing to override the base.twig template completely by creating _head.twig with their own meta/css/... lines. If the template is missing nothing is added.

### DIFF
--- a/templates/base.twig
+++ b/templates/base.twig
@@ -13,6 +13,7 @@
     {% endif %}
     <meta name="robots" content="noindex, nofollow">
     <link rel="preload" href="{{ asset('js/bundle.js') }}" as="script">
+    {{ include('_head.twig', ignore_missing = true) }}
     {% block preload %}{% endblock %}
   </head>
   <body id="{{ templateId }}">


tvdijen:
This is functionality what the {% preload %} block already provides ..?

liedekef:
Well, in fact not. Using the "preload" block means you need to provide that in every single twig template you create. If you want custom CSS to be added generally, you'd need to override just about all existing templates. Currently I did it via the _header.twig template, which causes the css to be loaded in the content of the page (works, but not nice). Of course I can be wrong, but I tried extending the base.twig template itself (in a custom theme) but that didn't work, I needed to copy over the complete base.twig template and change the relevant lines (which is of course also not the intention). If there's an existing method of adding css to all pages, feel free to point me to that info and I'll just close this PR :-)

tvdijen:
I just checked my own theme and I'm also basically copying the entire base.twig..
It _should_ be possible to just extend it instead.. I think I would prefer that over the suggested change.

liedekef:
Well, I tried for over an hour and couldn't get it to work as expected. For a theme, adding CSS is however basic stuff and should thus be easy to do (without needing to copy+change a complete twig-file, which is I guess the reason the switch to twig was done in the first place). I'm all ears on something better, but couldn't find any pointers/hints at all :-)

tvdijen:
The switch to twig was completely motivated by the fact the old www-scripts were mixing PHP code & logics with markup.. Which was probably a great idea 15 yrs ago, but not so much today :)

liedekef:
Well, the current twig is still mixing logic and markup a bit too much in my opinion, but there's always a choice to be done :-)
But: do you feel we can discuss this further? Do you want to make some tests, or do you think this change is not to be accepted? While the latter is ok for me, I feel a bit conflicted in having to copy and change base.twig just to add some css. base.twig also contains other logic, so on every update I would need to validate the change again ...

tvdijen:
Ok, so our first analysis on the issue here is that if you want to extend `base.twig` from a theme, it will try and lookup `base.twig` in the `__main__` namespace, which is the /default directory in the theme.. Which is your override of `base.twig`, resulting in an infinite loop and PHP running out of memory.

tvdijen:
> Well, the current twig is still mixing logic and markup a bit too much in my opinion, but there's always a choice to be done :-) But: do you feel we can discuss this further? Do you want to make some tests, or do you think this change is not to be accepted? While the latter is ok for me, I feel a bit conflicted in having to copy and change base.twig just to add some css. base.twig also contains other logic, so on every update I would need to validate the change again ...

Yeah well, it's been an awful lot of work migrating with very limited resources.  The change _as is_ is not likely to be accepted as we feel it's hackish and we should try and solve the actual issue here

liedekef:
well, I'm happy to test whatever solution you can come up with. For now I have the codechange as option 1 or the base.twig copy+change as option 2 :-)

tvdijen:
@thijskh What do you think, before this PR goes to waste?

thijskh:
I do agree with the functional idea to make it easy to just add custom css/js globally without the need to clone entire templates. The proposal seems like a simple and not super harmful approach? So why not go with this if we do not have a working concrete alternative? We can spend lots of time to think of a “perfect” solution but maybe this is already a “good” solution?

monkeyiq:
It seems that the PR has received a good response. I am not yet fully up to speed with twig. Either merging this PR or something to allow functionality like it seems like a good candidate for me to work on.

monkeyiq:
I have taken a look at supplying a custom `favicon.ico` from the theme. 

```
     <link rel="icon" href="{{ asset("icons/favicon.ico") }}">
```

The call to `asset()` without a module will not respect the theme.use value and will fall through to the default icon at this branch

https://github.com/simplesamlphp/simplesamlphp/blob/1071b139352f6b52e39ca09567639b524018d1a7/src/SimpleSAML/XHTML/Template.php#L189

Perhaps it is best to modify the `asset()` code to first look for that file in the theme.use and then fall back to the default value? This would avoid the need to modify any of the TWIG files if you just wanted to override those existing resources in your theme. Though maybe I have also missed something.

Otherwise as mentioned in https://simplesamlphp.org/docs/stable/simplesamlphp-theming you would have to copy the TWIG files and update the `asset` calls to include your module (theme) name.

This assumes that we want to make it very easy for folks to update some icons and maybe include a custom css file to do a slight theme change without needing to maintain copies of twig files and diff/merge them each update.

monkeyiq:
It seems that this one line change is in master and simplesamlphp-2.1 (https://github.com/simplesamlphp/simplesamlphp/commit/28cf79631fa2af01841a936b74ae251e91a565a0).

https://github.com/simplesamlphp/simplesamlphp/blob/f57b7154b774ef4d5fa647a79d203e8ada5f6de1/templates/base.twig#L13

monkeyiq:
Though the finding of assets with theme.use is something I should split into another PR.

monkeyiq:
Since the code update itself was merged a while ago I am going to close the PR. It is a shame it was not mentioned here so @liedekef was aware and could enjoy the update. I'll look at my above `asset` update in a new PR from here.

PR #1635, opened 2022-05-03
Cleanup: autoloader files

tvdijen:
This removes the old autoloader-code.. Everything complies with PSR-4 now.


tvdijen:
Tested this and it works fine

PR #1638, opened 2022-05-20
Add ability to retrieve certificates or private keys from database

pbhenson:
This pull request adds the ability to load certificate and private key data from the database in addition to local files.

### DIFF
--- a/config-templates/config.php
+++ b/config-templates/config.php
@@ -56,7 +56,6 @@ $config = [
     /*
      * The following settings are *filesystem paths* which define where
      * SimpleSAMLphp can find or write the following things:
-     * - 'certdir': The base directory for certificate and key material.
      * - 'loggingdir': Where to write logs.
      * - 'datadir': Storage of general data.
      * - 'tempdir': Saving temporary files. SimpleSAMLphp will attempt to create
@@ -64,10 +63,60 @@ $config = [
      * When specified as a relative path, this is relative to the SimpleSAMLphp
      * root directory.
      */
-    'certdir' => 'cert/',
     'loggingdir' => 'log/',
     'datadir' => 'data/',
     'tempdir' => '/tmp/simplesaml',
+
+    /*
+     * Certificate and key material can be loaded from different possible
+     * locations. Currently two locations are supported, the local filesystem
+     * and the database via pdo using the global database configuration. Locations
+     * are specified by a URL-link prefix before the file name/path or database
+     * identifier.
+     */
+
+    /* To load a certificate or key from the filesystem, it should be specified
+     * as 'file://<name>' where <name> is either a relative filename or a fully
+     * qualified path to a file containing the certificate or key in PEM
+     * format, such as 'cert.pem' or '/path/to/cert.pem'. If the path is
+     * relative, it will be searched for in the directory defined by the
+     * 'certdir' parameter below. When 'certdir' is specified as a relative
+     * path, it will be interpreted as relative to the SimpleSAMLphp root
+     * directory. Note that locations with no prefix included will be treated
+     * as file locations for backwards compatibility.
+     */
+    'certdir' => 'cert/',
+
+    /* To load a certificate or key from the database, it should be specified
+     * as 'pdo://<id>' where <id> is the identifier in the database table that
+     * should be matched. While the certificate and key tables are expected to
+     * be in the simplesaml database, they are not created or managed by
+     * simplesaml. The following parameters control how the pdo location
+     * attempts to retrieve certificates and keys from the database:
+     *
+     * - 'cert.pdo.table': name of table where certificates are stored
+     * - 'cert.pdo.keytable': name of table where keys are stored
+     * - 'cert.pdo.apply_prefix': whether or not to prepend the database.prefix
+     *                            parameter to the table names; if you are using
+     *                            database.prefix to separate multiple SSP instances
+     *                            in the same database but want to share certificate/key
+     *                            data between them, set this to false
+     * - 'cert.pdo.id_column': name of column to use as identifier
+     * - 'cert.pdo.data_column': name of column where PEM data is stored
+     *
+     * Basically, the query executed will be:
+     *
+     *   SELECT cert.pdo.data_column FROM cert.pdo.table WHERE cert.pdo.id_column = :id
+     *
+     * Defaults are shown below, to change them, uncomment the line and update as
+     * needed
+     */
+
+    //'cert.pdo.table' => 'certificates',
+    //'cert.pdo.keytable' => 'private_keys',
+    //'cert.pdo.apply_prefix' => true,
+    //'cert.pdo.id_column' => 'id',
+    //'cert.pdo.data_column' => 'data',
 
     /*
      * Some information about the technical persons running this installation.
--- a/docs/simplesamlphp-advancedfeatures.md
+++ b/docs/simplesamlphp-advancedfeatures.md
@@ -90,9 +90,9 @@ Metadata signing
 SimpleSAMLphp supports signing of the metadata it generates. Metadata signing is configured by four options:
 
 - `metadata.sign.enable`: Whether metadata signing should be enabled or not. Set to `TRUE` to enable metadata signing. Defaults to `FALSE`.
-- `metadata.sign.privatekey`: Name of the file with the private key which should be used to sign the metadata. This file must exist in in the `cert` directory.
+- `metadata.sign.privatekey`: Location of the private key data which should be used to sign the metadata.
 - `metadata.sign.privatekey_pass`: Passphrase which should be used to open the private key. This parameter is optional, and should be left out if the private key is unencrypted.
-- `metadata.sign.certificate`: Name of the file with the certificate which matches the private key. This file must exist in in the `cert` directory.
+- `metadata.sign.certificate`: Location of certificate data which matches the private key.
 - `metadata.sign.algorithm`: The algorithm to use when signing metadata for this entity. Defaults to RSA-SHA256. Possible values:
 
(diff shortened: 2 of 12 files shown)


tvdijen:
HI Paul! Thanks for the PR!
Two things:
 - Does it really make sense to have the table names and column names configurable here? It's extremely flexible, but at what cost?
 - Personally I would store the PEM data without the PEM-headers

tvdijen:
Hi @pbhenson ! Thanks for your PR!
We've discussed this internally and we don't think it's a good idea to store key material in a database, because it increases attack surface for little to no gain. Can you elaborate a bit on the use-case you have for this? Maybe we can think of a better solution for it.

pbhenson:
> Hi @pbhenson ! Thanks for your PR! We've discussed this internally and we don't think it's a good idea to store key material in a database, because it increases attack surface for little to no gain. Can you elaborate a bit on the use-case you have for this? Maybe we can think of a better solution for it.

Thanks for the quick response and feedback. I received two emails from github with comments, but the ones from the first one don't seem to be there anymore? Did you decide you just didn't like the concept and delete the initial feedback :)?

Unless you are running a single persistent system with SSP and always directly log into it to generate and rotate keys, the keys are going to be coming from someplace else. In the case of dynamically instantiated instances or containers behind a load balancer, that will most likely be a network source. Possibly an EFS volume, or an S3 bucket, or a version control repository, or a database ;). Deciding that SSP itself can only access them from the local file system won't change the way a user chooses to handle them, it will only make it more inconvenient for them to do so (https://metacpan.org/pod/Fuse::DBI for example).

The specific use case that prompted the development of this feature is a deployment of SSP that will handle multiple service providers and identity providers which will be dynamically provisioned through a database using an API. I was able to get everything SSP needed to do this working completely through the database except the certificates and keys, which currently are only supported through the local file system. Being able to access them through the database puts the entire configuration in the database, such that a running instance is immediately capable of providing service for a new identity provider or service provider the moment the database is updated, and a new instance can begin serving as well without having to somehow replicate all of the certificates and keys onto itself before it would work.

There are certainly other ways of making this work with the end result of the certificates and keys ending up stored locally on each instance (or by hacking filesystem access to db data), but I'm not sure that any of them would be better, more efficient, or more secure. The attack surface is not going to be increased, as the certificates and keys will be in the database one way or another, if anything, they are now both in the database and every key is on each and every single instance at all times. The gain from my perspective is considerable, as this implementation is the most efficient and scalable with fewer moving parts.

I can understand and appreciate your concern about the security of SSP, but the existence of this feature would not in any way inherently decrease its security. Just because it's there doesn't mean someone has to use it, and anyone not using it would not be detrimentally impacted by its existence. You store metadata in the database, if someone was able to futz with that they could change ACS URLs and encryption certificates and intercept random authentications and attributes. Session data can be stored in a database, compromise thereof allowing stealing existing sessions. There could certainly be a disclaimer and discussion of security concerns involved with this feature in the documentation, but if that's the design that someone thinks best serves their purposes, I'm not sure why SSP shouldn't provide the capability of using it.

With optimism and hope you will reconsider including this feature :), I'm going to go ahead and respond to the initial feedback that doesn't seem to be present anymore.

If someone is going to store certificates and keys in a database, the assumption is that either they already have some type of infrastructure in place for doing that, or they're going to develop one, and the design for that development would probably be better driven by the management side of it than the consumption side of it. SSP allows you to store certificates and key files anywhere with any names, being able to define how to pull them out of the database is simply analogous to that.

I'm not sure what cost you are considering? Runtime processing cost or complexity of configuration cost? For the former, the cost is just a handful of configuration attributes and corresponding calls to pull them out of the configuration. For the latter, if you configure your database to match the defaults, there is really no additional complexity, it just works as is. That's pretty much the same level of complexity as if the table and column information were hardcoded. If your database doesn't match the defaults, well, that complexity allows you to make it work, as without it it wouldn't or you would have to change your database. I suppose you could always create a view of your existing database structure to provide whatever hardcoded db structure SSP expected, but that is additional complexity as well.

The intent on this was to be as plug-and-play as possible with the existing file-based implementation as well as being as unintrusive as possible in terms of code changes. The existing files contain the PEM headers, so I'm not sure why the database would not? All of the consumers of the existing files expect there to be PEM headers. Either files would need to no longer contain the headers, or everything consuming certs/keys would need to be updated to handle both header and non-header data, which would be more complex and considerably more code changes. Or the code that pulled the data from the db would need to add the headers before passing them on...

Thanks...

pbhenson:
And not that I'm endorsing all of their practices :), but I guarantee you every IaaS provider from Azure AD to Google to Ping or Okta is storing their keys in some kind of database or networked storage. You can't scale local files on every instance...

tvdijen:
I hope they're using proper HSMs for their private keys ;)  We'll reconsider this PR.

pbhenson:
> I hope they're using proper HSMs for their private keys ;) We'll reconsider this PR.

You got the money, they got the HSM ;)

https://azure.microsoft.com/en-us/services/azure-dedicated-hsm/

But these are for your own apps, I doubt they use them for each and every virtual tenant doing SAML.

Consideration much appreciated, thanks...

config-templates/config.php
```diff
@@ -56,19 +56,68 @@
...
+     *
+     * Defaults are shown below, to change them, uncomment the line and update as
+     * needed
+     */
+
+    #'cert.pdo.table' => 'certificates',
+    #'cert.pdo.keytable' => 'private_keys',
+    #'cert.pdo.apply_prefix' => true,
+    #'cert.pdo.id_column' => 'id',
+    #'cert.pdo.data_column' => 'data',
```
thijskh, line 119: Nit: I think we prefer `//` in this file to comment things out.

lib/SimpleSAML/Metadata/Signer.php
```diff
@@ -235,35 +235,31 @@ private static function getMetadataSigningAlgorithm(
...
 
-        $keyFile = $configUtils->getCertPath($keyCertFiles['privatekey']);
-        $fileSystem = new Filesystem();
-        if (!$fileSystem->exists($keyFile)) {
+        $keyLocation = $keyCertLocations['privatekey'];
+        $keyData = $cryptoUtils->retrieveKey($keyLocation);
+        if ($keyData === null) {
             throw new Exception(
-                'Could not find private key file [' . $keyFile . '], which is needed to sign the metadata'
+                'Could not find private key location [' . $keyLocation . '], which is needed to sign the metadata'
```
thijskh, line 252: If I understand it correctly, for the default use case the error will no longer show the full filesystem path where it's looking for the certificate, right? I'd prefer that it does because it's a common exception and understanding where to look is very helpful there.

lib/SimpleSAML/Metadata/Signer.php
```diff
@@ -235,35 +235,31 @@ private static function getMetadataSigningAlgorithm(
...
-
-        $certFile = $configUtils->getCertPath($keyCertFiles['certificate']);
-        $cert = new File($certFile);
-        if (!$fileSystem->exists($certFile)) {
+        $certLocation = $keyCertLocations['certificate'];
+        $certData = $cryptoUtils->retrieveCertificate($certLocation);
+        if ($certData === null) {
             throw new Exception(
-                'Could not find certificate file [' . $certFile . '], which is needed to sign the metadata'
+                'Could not find certificate location [' . $certLocation . '], which is needed to sign the metadata'
```
thijskh, line 260: As above

thijskh (commented):
I understand that there are different risks associated with different types of storage backends. I do not however think that a database is per definition less secure than a plaintext file on a filesystem. Both can have or lack guards to protect them, and for both cases you need to be careful and know what you do. But I'd be in favour of allowing people that know what they do to use this as a backend because I can see the use case.

We could make more explicit in the documentation maybe that you need to properly protect this and e.g. also ensure that your database connection is encrypted to name something.

config-templates/config.php
pbhenson: Woops, didn't notice that, updated.

lib/SimpleSAML/Metadata/Signer.php
pbhenson: The exception text in Metadata/Signer.php will only include the configured location, be it the db or file format, and yes, it won't be fully qualified if the location was configured as relative. However, the error generated in Utils/Crypto.php will contain the fully qualified path:

$location = $configUtils->getCertPath($location);
[...]
Logger::error("failed to read $data_type data from file $location");

So the information will be available. If that's not good enough, I can update retrieveCertOrKey to throw an exception on failure instead of returning null and have the full path bubble up to be caught by Metadata/Signer.php to be used in the text it generates for its exception.

pbhenson: Just touching base and seeing if you'd had a chance to consider if the current approach of logging the full path of a missing cert/key in retrieveCertOrKey and just the location in the caller is acceptable or if you want me to update it to raise an exception and have the caller be able to log the full path.

Thanks...

thijskh: Thanks. It's good enough. I'd be fine with merging this if you can resolve the conflicts.

thijskh (approved):

lib/SimpleSAML/Metadata/Signer.php
pbhenson: I rebased to latest master and it seems fine? Didn't have to manually clean up anything.

thijskh (approved):

thijskh:
@tvdijen Are you OK with merging?

pbhenson:
Thanks much guys! Very appreciated. Is there any tentative timeline yet for making a production release of 2.0?

tvdijen:
We're performing some Q&A tests as we speak .. A release candidate should be available very soon

PR #1641, opened 2022-05-25
Feature: Add auth source selector

tvdijen:
The PR introduces a new AbstractSourceSelector that can be used to delegate authentication to secondary auth sources based on a policy decision. The first implementation of this is the included IPSourceSelector which picks the secondary auth source based on the user's IP address (i.e. internal or external client).

### DIFF
--- a/modules/core/docs/authsource_selector.md
+++ b/modules/core/docs/authsource_selector.md
@@ -1,0 +1,56 @@
+# Authentication source selector
+
+The Authentication source selector is a special kind of Authentication Source
+that delegates the actual authentication to a secondary Authentication Source
+based on some form of policy decision.
+
+## AbstractSourceSelector
+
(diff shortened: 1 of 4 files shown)


tvdijen:
It's been tested and working like charm! ✅

thijskh (approved):
Looks good!

i would document that this feature exists and how to use it.


The tests could also add an IPv6 case just to confirm that this works and keeps working.

tvdijen:
Thanks for the review! I will add ipv6 tests and docs

tvdijen:
@thijskh Would you mind checking again on the last three commits?

I've kinda changed the config and opened up to configure multiple zones. Other than that, I've addressed your comments

thijskh:
Looks good. I don’t think you need to worry about how ‘good’ of a doc writer you are, it can always be improved later if need be. But undocumented features will not get doc improvements because people will not know that the feature exists.

Issue #1643, opened 2022-05-28
2.0 Q&A issues


tvdijen:
This issue is to keep track of the findings we do during our 2.0 Q&A tests that need to be taken care of before releasing a new version.

- [x]  1.  Error pages are not translated properly
- [x]  2. ~~Asset-files are generated from source to assets~~
- [x]  3.  Legacy-routes do not work
- [x]  4.  Auto-generated entityID points to legacy metadata-endpoint; relates to point 3.
- [x]  5.  Move directories following https://github.com/php-pds/skeleton
- [x]  6.  Logs are cluttered with NameID generation warnings

## 1. Error pages are not translated properly

<details>
![afbeelding](https://user-images.githubusercontent.com/841045/170819454-f4c50730-f2cb-4a98-897f-9631fc262146.png)

We need to replace the old translation-tags in `lib/SimpleSAML/Error/ErrorCodes.php` as well.
</details>

## 2. ~~Asset-files are generated directly from source~~

~~This is problematic for generating css-files with the same name as js-files and can be easily fixed by generating them in an intermittent `build` directory and then copy them from there to the `www/assets`-directory~~

Nobody seems to know how this webpack black magic works.. Not letting this block our new release

## 3. Legacy-routes do not work

<details>
The legacy-routes for the scripts that resided in the `www/` directory do not work. We need to restore the php-files and manually redirect to the new controllers in the `saml` module.

** Unless ** we can reach consensus in dropping the legacy routes & scripts completely and tell people how to configure their webserver to be able to use them
</details>

## 4 Auto-generated entityID points to legacy metadata-endpoint; relates to point 3.

<details>
Proposed solution: stop auto-generating entityIDs and make the entityID setting mandatory
</details>

## 5 Move src/ to resources/ and lib/ to src/

<details>
Move src/ to resources/ and lib/ to src/
</details>

## 6 Logs are cluttered with NameID generation warnings

<details>
The following warning show up when only setting the saml:TransientNameID authproc:
"Unable to generate NameID. Check the simplesaml.nameidattribute option."

It shouldn't do that
</details>

PR #1644, opened 2022-05-28
Remove legacy routes

tvdijen:
Ticks box 3 on #1643 

TODO: Add nginx equivalents for the `mod_rewrite` rules


thijskh:
I think we are breaking too much if we also want to impose these rewrite rules on users. This does not actually reduce complexity, it only moves it to a different point (and it takes all deployers effort to do that, and us to support them and to document it for various webservers etc).

The costs of keeping the files `www/` with just the bare content needed to make them work is simple and robust and I don't think this cost outweighs the added costs for solving it in a different point. Requiring too many changes in the end will mean that people stay longer on 1.19.

As an example, `www/admin/index.php` just contained this before it was removed.

```php
<?php

require_once('../_include.php');

$httpUtils = new \SimpleSAML\Utils\HTTP();
$httpUtils->redirectTrustedURL(\SimpleSAML\Module::getModuleURL('admin/'));
```

Other pages can simply call the relevant controller instead of performing a redirect but the amount of code is likely similarly in the order of less than a handful of lines.

tvdijen:
You really want an extra redirect on all your SAML endpoints?

jaimeperez:
Yes, we've been discussing this over the weekend and I fully agree with Tim. The whole point of 2.0 is precisely to be able to do backwards-compatibility breaking changes as the first new major in some 13-14 years. I completely see your point about breaking too much so that people refrains from updating, but at the same time I don't think it's a good idea to keep legacy forever just to avoid asking deployers to do something on their side. Consider also that for new deployments they won't care about these changes at all.

From a technical standing point there is no reason to keep multiple endpoints, in my opinion. Having a single entry point for all the endpoints means less code duplication and a single point to enforce anything that we need. It has been considered good practice for many years and I'd say most major frameworks do this. The alternative we had was simple, sure, but it was also problematic in the sense that it was enforcing an HTTP-redirect for each request, with the added roundtrip and consequential hit in performance for main endpoints.

The alternative (using `mod_rewrite`) is relatively common nowadays, I believe. The rules aren't that many and I think it's relatively simple to configure. The benefit is huge for us because it allows us to have, finally, a fresh start without legacy. If we'd rather keep the legacy scripts per endpoint, we won't have a chance to get rid of them and have a clean codebase until 3.0, which could of course happen next year (although I would argue it's worse to break bc in small changes more often than to break it more but on rare occasions), but more realistically will not happen in many years to come.

We've been putting a lot of effort on 2.0 for years and the original plan already went ashtray, but I believe in this moment, with a new maintainer hopefully in place soon, the timing is right to do this change. Just my two cents 😄

thijskh:
I just don't see the argument why the rewrite approch is better than keeping small www files approach. It seems that they are similar in complexity and amount of legacy that we keep, but one requires people to change stuff while the other doesn't and just works out of the box.

tvdijen:
I'm fine with the scripts, as long as we mark them `deprecated` and make sure the metadata generator provides us with the _new_ endpoints. This PR could be merged as soon as we branch off 2.0

PR #1645, opened 2022-05-28
Fix error messages

tvdijen:
Ticks box 1 on https://github.com/simplesamlphp/simplesamlphp/issues/1643


tvdijen:
Partial fix so far.. Bottom half is still not translated:

afbeelding

PR #1646, opened 2022-05-30
Simplify rtl css

thijskh:
Use the HTML 'dir' attribute to properly indicate text direction of this language. Browser takes care of most issues already. xml:lang is duplicated with lang and deprecated, so remove that while we're at it.

Use value of dir property in css to position things in the main layout that are explicitly right or left aligned. Uses more semantic class naming than "left" or "right" that were in use explicitly.

Probably an RTL expert can improve the UI in more specific places. But I'm no such expert and this was not done before either.


tvdijen (approved):
My only concern is that the `left` and `right` classes are widespread among the modules, but I guess we'll find out during our Q&A run

thijskh:
If they are, it's indeed better to find that out and fix it as a class named "right" is not-so-nice css anyway.

PR #1653, opened 2022-06-10
Refactor scoping code in the SP class

Paragin:
Make sure the `saml:IDPList` and `IDPList` parameters are not overloaded. From now on, saml:IDPList is being used when SSP is used as a proxy while the IDPList parameter is used when SSP is used as in SP-mode and you want to apply scoping to your AuthnRequest.

See https://github.com/simplesamlphp/simplesamlphp/pull/1563 and  for the full explanation of the changes.

Solves https://github.com/simplesamlphp/simplesamlphp/issues/1562

### DIFF
--- a/tests/modules/saml/src/Auth/Source/SPTest.php
+++ b/tests/modules/saml/src/Auth/Source/SPTest.php
@@ -468,7 +468,8 @@ class SPTest extends ClearStateTestCase
     }
 
     /**
-     * Test that IDPList with multiple valid idp and no SP config idp results in discovery redirect
+     * Test that adding IDPList to the state (of an AuthRequest)
+     * will result in that IDP being added to the scope
      */
     public function testSPIdpListScoping(): void
     {
--- a/tests/modules/saml/src/Auth/Source/SPTest.php
+++ b/tests/modules/saml/src/Auth/Source/SPTest.php
@@ -483,6 +483,44 @@ class SPTest extends ClearStateTestCase
     }
 
     /**
+     * Test that adding IDPList to the idp metadata
+     * will result in that IDP being added to the scope
+     */
+    public function testIdpMetadataScoping(): void
+    {
+        $this->idpConfigArray['IDPList'] = ['https://scope.example.com'];
+        $ar = $this->createAuthnRequest([]);
+
+        $this->assertTrue(
+            in_array('https://scope.example.com', $ar->getIDPList())
+        );
+    }
+
+    /**
+     * Test that adding IDPList to the config
+     * will result in that IDP being added to the scope
+     */
+    public function testRemoteMetadataScoping(): void
+    {
+        $info = ['AuthId' => 'default-sp'];
+        $config = [
+            'IDPList' => ['https://scope.example.com']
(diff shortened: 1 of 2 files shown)


tests/modules/saml/src/Auth/Source/SPTest.php
```diff
@@ -482,6 +482,44 @@ public function testSPIdpListScoping(): void
...
+
+        /** @var \SAML2\AuthnRequest $ar */
+        try {
+            $as->startSSO2Test($this->getIdpMetadata(), []);
+            $this->assertTrue(false, 'Expected ExitTestException');
+        } catch (ExitTestException $e) {
+            ['ar' => $ar] = $e->getTestResult();
+            $this->assertTrue(
+                in_array('https://scope.example.com', $ar->getIDPList())
+            );
```
thijskh, line 519: This can be written as `$this->assertContains('https://scope.example.com', $ar->getIDPList());`

tests/modules/saml/src/Auth/Source/SPTest.php
```diff
@@ -482,6 +482,44 @@ public function testSPIdpListScoping(): void
...
+     * will result in that IDP being added to the scope
+     */
+    public function testIdpMetadataScoping(): void
+    {
+        $this->idpConfigArray['IDPList'] = ['https://scope.example.com'];
+        $ar = $this->createAuthnRequest([]);
+
+        $this->assertTrue(
+            in_array('https://scope.example.com', $ar->getIDPList())
+        );
```
thijskh, line 496: This can be written as $this->assertContains('https://scope.example.com', $ar->getIDPList());

thijskh (changes requested):
Great, thanks! It's complex enough to wrap your head around the intended (complex) interactions in the proxy setup but I do think this is now correct and better than it was, and more straightforward.

Great to see that you've added tests! In the ideal world I'd like to see a test that also confirms the ordering as it is now. So that setting it e.g. at runtime overrides idp settings (and not vice versa or merge etc).

One must is that the documentation needs to be updated. Basically grep in docs/ for IDPList. I can do that if desired.

Paragin:
@thijskh I've applied your suggestions from the test and added a test that confirms the ordering as well.
If you can do the documentation that would be appreciated. 

I've suggested to apply these changes to the 1.19 branch as well so it can be released sooner. Is that something you are open to?

tvdijen:
I would be fine with backporting this to 1.19 and tag 1.19.6

thijskh:
It does change behaviour (for the better I think, but still...). Can we make a more targeted change for 1.19 which adds the ability to set @ runtime without 'improving' the setting behaviour?

tvdijen:
Ahh, I wasn't aware of that.. In this case it's probably better to not put any more efforts in 1.19

Paragin:
Ok, then I'll guess we have to wait. 

We have a working workaround so we'll manage, but it's not something I want to keep around for long.. (It involves selectively copy-pasting code from SSA into our own codebase so we manually call the startSSO-function with the right parameters)

PR #1654, opened 2022-06-10
Add symlink for the _autoload_modules file

Paragin:
A symlink was recently added for the traditional `_autoload.php` file, but the link to `_autoload_modules.php` was still missing. This was needed to get the modules working properly.

### DIFF
--- a/lib/_autoload_modules.php
+++ b/lib/_autoload_modules.php
@@ -1,0 +1,1 @@
+../src/_autoload_modules.php


thijskh:
I'm not sure... what loads `autoload_modules` that cannot just be changed to the new path?

We added the symlink because (as per our documentation) external code/SP application code will include the file at the old location. But the SP code usually does not need to include autoload_modules aswell?

Paragin:
I ran into this issue when using the composer-module-installer

tvdijen:
Hi @Paragin! Can you be more specific on how this became a problem? Because I don't see it.. Nothing in the `composer-module-installer` is pointing to this file. Is this reproducible for us?

thijskh:
I had the issue when I fetched updates via git but neglected to run composer install after them. Maybe that's the explanation for you aswell.

PR #1657, opened 2022-06-11
Fix NameID generation

tvdijen:
We are unable to generate persistent NameIDs based on `simplesaml.nameidattribute`, since we have removed this attribute.
The only way to generate a persistent NameID is to use the appropriate authproc-filter.

This PR cleans up old code & logs and checks point 6 of #1643


tvdijen:
Should probably also rename 'attribute' to 'identifyingAttribute' to match all the other authproc-filters

thijskh (approved):

PR #1658, opened 2022-06-11
Remove automatic entityID-generation

tvdijen:
We used to auto-generate entity IDs pointing to the metadata-url, but that goes against SAML2INT recommendations.
Instead, we let users set the entityID themselves.

Closes point 4 in #1643


src/SimpleSAML/Metadata/MetaDataStorageSource.php
```diff
@@ -349,9 +349,8 @@ private function getDynamicHostedUrl(string $set): string
 
 
     /**
-     * Updates the metadata entry's entity id and returns the modified array.  If the entity id is __DYNAMIC:*__ a
-     * the current url is assigned.  If it is explicit the entityid array key is updated to the entityId that was
-     * provided.
+     * Updates the metadata entry's entity id and returns the modified array.
+     * If it is explicit the entityid array key is updated to the entityId that was provided.
```
thijskh, line 353: It is now always explicit.

Is this method still useful now, since it doesn't do much special?

src/SimpleSAML/Metadata/MetaDataStorageHandlerPdo.php
```diff
@@ -159,16 +159,10 @@ public function getMetaData(string $entityId, string $set): ?array
...
-        // that allow the __DYNAMIC:*__ entity id.  with the current table design
-        // we need to lookup the specific metadata entry but also we need to lookup
-        // any dynamic entries to see if the dynamic hosted entity id matches
         if (substr($set, -10) == 'idp-hosted') {
             $stmt = $this->db->read(
-                "SELECT entity_id, entity_data FROM {$tableName} "
-                . "WHERE (entity_id LIKE :dynamicId OR entity_id = :entityId)",
-                ['dynamicId' => '__DYNAMIC%', 'entityId' => $entityId]
+                "SELECT entity_id, entity_data FROM {$tableName} WHERE (entity_id LIKE :entityId)",
+                ['entityId' => $entityId]
```
thijskh, line 165: Probably we should make explcit somewhere in the notes that you need to update your database and that any entities with __DYNAMIC will no longer be read?

src/SimpleSAML/Metadata/MetaDataStorageHandlerPdo.php
```diff
@@ -159,16 +159,10 @@ public function getMetaData(string $entityId, string $set): ?array
         }
 
         $tableName = $this->getTableName($set);
-
-        // according to the docs, it looks like *-idp-hosted metadata are the types
-        // that allow the __DYNAMIC:*__ entity id.  with the current table design
-        // we need to lookup the specific metadata entry but also we need to lookup
-        // any dynamic entries to see if the dynamic hosted entity id matches
         if (substr($set, -10) == 'idp-hosted') {
```
thijskh, line 162: This is no longer needed since there's no distinction in the lookup between idp-hosted and other types. Can use just one query. And that's the one with exact match, not LIKE

thijskh (changes requested):
I think it's a great thing to make people choose an explicit entityID. That will surely prevent issues in the future.

Some comments:
* I'm missing changes to the SP auth source. Is this purposefully only done for IdP entityID generation?
* I would want to add some guidance for people on what to pick/what a good value is. It's not obvious and you cannot really change it afterwards.
* Probably the default suggestion should follow this advice. Which I think is not something that's an URN.

thijskh:
"You need to set an entityID: a string that uniquely identifies this IdP or SP. It must be a URL or URN. We recommend to choose something that is likely to remain stable over time. Examples (replace example.com with your organization's domain name: `https://example.com/saml-idp` (or `https://myapp.example.com/sp`). The URL does not have to serve to any content, it's just an identifier."

tvdijen:
If that's going to be our advice, we might as well continue to auto-generate it..

"An entityID should be chosen in a manner that minimizes the likelihood of it changing **_for political or technical reasons_**, including for example a change to a different software implementation or hosting provider."

It shouldn't hold an organization's domain name for that reason.

thijskh:
Let's separate out the discussion on what the advice should be.

modules/saml/src/Auth/Source/SP.php
```diff
@@ -81,19 +81,20 @@ public function __construct(array $info, array $config)
...
             'authsources[' . var_export($this->authId, true) . ']'
         );
-        $this->entityId = $this->metadata->getString('entityID');
+
+        $entityId = $this->metadata->getString('entityID');
+        Assert::validURI($entityId);
+        Assert::maxLength(
+            $entityId,
+            Constants::ENTITYID_MAX_LENGTH,
+            'The entityID cannot be longer than 1024 characters.'
```
thijskh, line 94: Should also use the constant in the message. Otherwise quite useless to use a constant if not everywhere.

tvdijen: Smart thinking! 

- [x]

src/SimpleSAML/Metadata/MetaDataStorageHandlerPdo.php
tvdijen: Removed it

thijskh (approved):
I think only thing still missing is an update to simplesamlphp-sp.md. Other than that, good to go. Let's discuss entityID strategy in a separate issue.

PR #1662, opened 2022-06-24
Fix fallback to transient NameID when no format available

thijskh:
We only have the means to generate a transient NameID. If a different one was requested, the code would crash because nameIdValue would be used while unset. Fix that we fall back to transient in such a case.

### DIFF
--- a/modules/saml/src/IdP/SAML2.php
+++ b/modules/saml/src/IdP/SAML2.php
@@ -1214,9 +1214,32 @@ class SAML2
             $a->setAttributes($attributes);
         }
 
+        $nameId = self::generateNameId($idpMetadata, $spMetadata, $state);
+        $state['saml:idp:NameID'] = $nameId;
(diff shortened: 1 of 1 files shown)


tvdijen (approved):
I *think* the old situation would generate a transient NameID *without* SPNameQualifier.. I'm not sure if it's a problem or not, but it's a change in behaviour.
Also, we have to make sure it's still possible to use the TransientnameID authproc with NameQualifier/SPNameQualifier.. It should prevail over the fallback.

thijskh:
Although I cannot imagine what "TransientnameID authproc with NameQualifier/SPNameQualifier" would be useful for, that should still work.

Reading the old code, an SPNameQualifier was set when generating Transient ID

tvdijen:
Neither do I, but it is possible to generate one.. If it doesn't make sense, we might as well remove the TransientNameID authproc

PR #1665, opened 2022-07-04
Add job to trigger docs build

tvdijen:
Will trigger a build of the https://github.com/simplesamlphp/docs repo if any of the `.md`-files have changed.

Downsides:
- it requires a personal access token of a user-account. In this case I created one under my account and added it to the secrets-vault.
- it would require us to generate tokens not just in this repo, but for all modules.

Closes https://github.com/simplesamlphp/docs/issues/2


thijskh:
It looks good. Having the personal access token is not ideal but we can live with it I guess (would be great if repos in the same org could just share these kind of basic permissions). For every module would be nice but not a hard requirement per se I guess since these change not so often.

It should only run when md files are changed in main or any of the release branches I guess? So it does not get triggered on every PR (commit)? Maybe it's not essential.

tvdijen:
Maybe it's possible, but I haven't been able to find a way to set these permissions generically.
I'll see if I can further limit it to the master- and release-branches..

tvdijen:
I'm in doubt... The best way to run this only on specific branches is to create a separate workflow..
The downside is that it will run, even if the unit tests fail.. The upside is, that's maybe a good thing.. Because markdown !== PHP

tvdijen:
OK, so I've moved everything to a separate workflow that will only be triggered on changes to `.md` files on the `master`-branch and release-branches and also added a spell checker.
The markdown-linter and spell check are always being run, however the 'docs build' step will only run on push-events.

thijskh (approved):

PR #1670, opened 2022-07-10
Translation improvements

### LINKED ISSUE #1630: Translation, pot, po and mo files
Currently in the old layout you have the translation "definition" json files and in the new layout simplesamlphp uses po files.
However: I see no use of mo files (compiled po files) which are much faster to use, and when you combine a po-file with a pot-file (pot=po template), you get an mo-file that uses the pot-definition for the non-translated strings (which is one of the things simplesamlphp wants to accomplish).
Also the code currently uses a structure as core/locales/<LANGUAGE>/LC_MESSAGES/core.po, which is way too much in my opinion, just using core/locales/<LANGUAGE>.po is more than sufficient and drops 2 directories. Per module the pot-file should always be included in the distribution, so people can then contribute po-files using tools like poedit easily.
(btw: this is my experience as a wordpress plugin developer talking).

tvdijen:
Goal of this (WIP) PR is to:
-[ ] Start using auto-generated .pot-files;
-[ ] Automatically update .po-files when new translations are added;
-[ ] Automatically generate .mo-files.

Closes #1630


tvdijen:
@drjayvee Thanks for showing today! Would you mind sharing a screen of your 'trick' with the translation of the 'Apple' here? I'm hoping to spend some time on this PR tomorrow.

tvdijen:
One of the lines the gettext/php-scanner is currently failing on is;
https://github.com/simplesamlphp/simplesamlphp/blob/master/templates/auth_status.twig#L22

It's converted to:
```
    // line 21
    public function block_namecol($context, array $blocks = [])
    {
        $macros = $this->macros;
        // line 22
        $context["translated"] = $this->extensions['Symfony\Bridge\Twig\Extension\TranslationExtension']->trans((isset($context["name"]) || array_key_exists("name", $context) ? $context["name"] : (function () { throw new RuntimeError('Variable "name" does not exist.', 22, $this->source); })()));
        // line 23
        echo "    <td class=\"attrname\">";
        if (((isset($context["translated"]) || array_key_exists("translated", $context) ? $context["translated"] : (function () { throw new RuntimeError('Variable "translated" does not exist.', 23, $this->source); })()) != (isset($context["name"]) || array_key_exists("name", $context) ? $context["name"] : (function () { throw new RuntimeError('Variable "name" does not exist.', 23, $this->source); })()))) {
            echo " ";
            echo twig_escape_filter($this->env, (isset($context["translated"]) || array_key_exists("translated", $context) ? $context["translated"] : (function () { throw new RuntimeError('Variable "translated" does not exist.', 23, $this->source); })()), "html", null, true);
            echo " <br>";
        }
        echo " <samp>";
        echo twig_escape_filter($this->env, (isset($context["name"]) || array_key_exists("name", $context) ? $context["name"] : (function () { throw new RuntimeError('Variable "name" does not exist.', 23, $this->source); })()), "html", null, true);
        echo "</samp></td>
    ";
```

And fails with:

```
PHP Fatal error:  Uncaught Exception: Invalid gettext function in /tmp/simplesaml/temptemplatestemp/bb/bb6042f5d9ab15b0e31a996df7cc489128dc079baa771097be902a29ea0abb21.php:314. Some required arguments are not valid in /apps/development/simplesamlphp/vendor/gettext/gettext/src/Scanner/CodeScanner.php:148
```

Because it is rendered into an expression instead of a string, the gettext/php-scanner can't deal with it.

drjayvee:
Good morning Tim,

Thanks again for inviting me. I had a great time meeting y'all yesterday.

Here's our "trick" for using `trans`:
```twig
{{ (true ? 'Apple' : 'Tomato')|trans }} {# DOES NOT WORK 😱 #}
{{ true ? 'Apple'|trans : 'Tomato'|trans }} {# This works 🥳 #}
```

The difference is that, while `gettext($str)` will work in runtime, `xgettext` won't be able to extract `msgid`s in the first form.

Apparently, `name|trans` compiles to
```php
$context["translated"] = $this->extensions['Symfony\Bridge\Twig\Extension\TranslationExtension']->trans((isset($context["name"]) || array_key_exists("name", $context) ? $context["name"] : (function () { throw new RuntimeError('Variable "name" does not exist.', 22, $this->source); })()));
```
which throws an Exception in runtime?

We use [phpmyadmin/twig-i18n-extension](https://github.com/phpmyadmin/twig-i18n-extension), which compiles down to
```php
gettext((isset($context["name"]) || array_key_exists("name", $context) ? $context["name"] : (function () { throw new RuntimeError('Variable "name" does not exist.', 2, $this->source); })()));
```

The following template
```
{% set name = 'Tim' %}
{% set translated = name|trans %}

Hello, {{ translated }}
```
runs just fine in our setup, so we're not getting an Exception.

Hope that helps!

drjayvee:
By the way, we use `xgettext --language=PHP`, not `gettext/php-scanner`:

```bash
find src/ -o -regextype posix-egrep -regex ".*\.php" -print \
  |xargs xgettext -o src/locale/messages.pot -F --no-wrap --language=PHP --from-code=UTF-8
```

tvdijen:
I think the tests fail because the .mo files cannot be loaded (yet) and it's causing more/other logging.

PR #1673, opened 2022-07-21
Actually handle ETag/If-None-Match logic

dnmvisser:
Sending the `Etag` header was only one part. This MR implement the other part.
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag


dnmvisser:
https://github.com/simplesamlphp/simplesamlphp/pull/1672 was not everything that was needed. This PR completes it.

tvdijen:
Rebased it to get the tests going (was a bug in test-suite)

tvdijen (approved):

PR #1678, opened 2022-08-13
Translation contracts

Berdir:
This turned out to be slightly less straight forward than expected. I wasn't sure what to do with the new getLocale() method that 3.x requires.

TwigTranslator is a wrapper around an inner callback, the actual implementation then passes in \SimpleSAML\Locale\Translate::translateSingularGettext(), which already now ignores the passed langcode anyway and is also just a wrapper around gettext, whose interface also doesn't expose the active language and internally relies on the locale API.

Unless \Symfony\Bridge\Twig\Extension\TranslationExtension ever decides to use that method or this implementation is used in another context, it doesn't matter. And that seems very unlikely, so I just went with a hardcoded en.

For context: https://github.com/symfony/symfony/issues/40380

Like the other return type changes, this is also backwards compatible with older versions of the interface and PHP 7.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
         "symfony/http-kernel": "^5.4||^6",
         "symfony/intl": "^5.4||^6",
         "symfony/routing": "^5.4||^6",
-        "symfony/translation-contracts": "^2.5",
+        "symfony/translation-contracts": "^2.5||^3",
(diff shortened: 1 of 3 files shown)


tvdijen:
Thanks again @Berdir !

PR #1685, opened 2022-08-24
Add ETag and check for modifiedness of metadata

thijskh:
This is basically what you added fixed for 1.19 in #1672, #1673 also implemented for the new code in 2.0

### DIFF
--- a/modules/saml/src/Controller/Metadata.php
+++ b/modules/saml/src/Controller/Metadata.php
@@ -9,6 +9,7 @@ use SimpleSAML\Configuration;
 use SimpleSAML\Error;
 use SimpleSAML\HTTP\RunnableResponse;
 use SimpleSAML\Metadata as SSPMetadata;
+use SimpleSAML\Metadata\MetaDataStorageHandler;
 use SimpleSAML\Module;
 use SimpleSAML\Module\saml\IdP\SAML2 as SAML2_IdP;
 use SimpleSAML\Utils;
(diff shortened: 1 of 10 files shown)


tvdijen (approved):

thijskh:
@tvdijen Do you see what I'm missing with the testcases? The problem is that the IdP metadata controller tests cannot find any saml2-idp-hosted metadata which they obviously need to generate some metadata. I can't seem to find out how to convince the testcase to find this metadata to be able to call the controller.

tvdijen:
Wasn't the key for entityid like `entityId` or `entityID` instead of fully lowercase?

thijskh:
I suspect that the issue is in that the configuration that speifies to use this metadata source for this test is not being used in some way or the other.

tvdijen:
The underlying hidden exception is;
Exception: Could not find any default metadata entities in set [saml20-idp-hosted] for host [localhost : localhost/simplesaml]

I've tried to change the unit test to a `ClearStateTestCase`, but that didn't help..

tvdijen:
It's really strange, as if somewhere along the line the config is reset... If I copy the configuration + metadata code from setup() into the failing tests, everything passes

thijskh:
Locally if I run phpunit on just the file everything passes. If I run it on the tests/module/saml/srcController/ directory, it fails. So it seems it's relevant whether some other tests have already done something with Configuration...

tvdijen:
This is curious, because it fails for me if I run the one file locally...

thijskh:
At least I can also reproduce that part now

PR #1688, opened 2022-08-28
Add Slovak translation

kikz0rsk:
Add Slovak translation for base, and modules saml and admin.

Since I've not found any up-to-date documentation as to how to generate translation files, locale files were created manually based on English language files with headers updated partially.


tvdijen:
Thanks @kikz0rsk !

Issue #1684, opened 2022-08-23
2.0: SP reply URL not backwards compatible


drjayvee:
After upgrading from 1.19.3 to 2.0-rc1, I cannot log in to a Microsoft Azure AD IdP configured using the SP metadata obtained from 1.19.

Azure AD presents the following error message:

> AADSTS50011: The reply URL 'https://gerben.alisqi.com.test-xxx.platformsh.site/sso/module.php/saml/sp/assertionConsumerService/gerben' specified in the request does not match the reply URLs configured for the application 'https://gerben.alisqi.com.test-xxx.platformsh.site/sso/module.php/saml/sp/metadata.php/gerben'. Make sure the reply URL sent in the request matches one added to your application in the Azure portal. Navigate to https://aka.ms/urlMismatchError to learn more about how to fix this. 

The last bit of the URLs is different:
```
old: .../saml/sp/metadata.php/gerben
new: .../saml/sp/assertionConsumerService/gerben
```

Note: in 1.19.x,, we had `entityID => null` in `authsources.php`, which I set it to auto-generated value.

thijskh:
The new URL should be in the newly generated metadata. So in theory the relying party would have refreshed that metadata?

drjayvee:
> The new URL

Did the return URL did change between 1.x and 2.x? If so, doesn't that warrant a warning in the upgrade notes?

> So in theory the relying party would have refreshed that metadata?

That depends on how often the metadata is updated. I don't think we have any control over that unless we ask our customers' sys admins to manually refresh at some particular point in time.

Without any such action, users won't be able to log in when the upgrade to 2.0 is deployed until the IdP has refreshed.

Is it possible to configure the return URL? We could then either make 2.x backwards compatible or change our current 1.x setup to make it forward-compatible.

tvdijen:
You can configure your ACS-endpoints in the `authsources.php` file as per [documentation](https://simplesamlphp.org/docs/stable/simplesamlphp-metadata-endpoints.html)

drjayvee:
> The old endpoints still work.

It looks like the old ACS endpoints are equal to the previously auto-generated metadata URL (if `entityID => null`), correct? Will those keep working as ACS endpoints?

I tried `'AssertionConsumerService' => $entityID`, which fails because
```
SimpleSAML\Error\Error: UNHANDLEDEXCEPTION
Backtrace:
3 www/_include.php:28 (SimpleSAML_exception_handler)
[...]

PR #1692, opened 2022-08-31
Add new 2.0 ACS endpoints to metadata

tvdijen:
This change will start publishing the new SSP 2.0 metadata endpoints in the SP metadata to make an upgrade easier.
The behavior will be enabled by default and can be disabled using a documented setting in the SP configuration.
Partially takes away the pain of #1684

### DIFF
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -6,6 +6,12 @@ SimpleSAMLphp changelog
 This document lists the changes between versions of SimpleSAMLphp.
 See the upgrade notes for specific information about upgrading.
 
+## Version 1.19.7
+
+Released TBD
+
+  * Add new SSP 2.0 metadata endpoints to the SP metadata (#1692).
+
 ## Version 1.19.6
 
 Released 1-7-2022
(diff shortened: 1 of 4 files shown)


tvdijen:
I've discussed this with Jaime as a reasonable upgrade path.. Do you agree?

thijskh:
You propose to add endpoints to the published metadata that do not actually work? I'm not sure I agree with that..

To take one step back. What advantage is there in _changing_ the endpoint URLs in the first place. It seems that this requires migration effort and code to realise which advantage exactly?

tvdijen:
I give up on this discussion.. You and Jaime figure it out.

thijskh:
@jaimeperez can you clarify?

thijskh:
It would make the PR more acceptable since the metadata would be actually correct, but I'd prefer to know first what problem we're solving with changing the URL in the first place.

tvdijen:
Dropping www-scripts...  We don't want to keep those forever, even if they're just empty shims that call the controllers

thijskh:
Aha, great. If that is the goal then we can achieve that without changing the URL. So we solve our problem (removing www-scripts) while not inducing breakage for people.

See:
https://github.com/simplesamlphp/simplesamlphp/compare/unchanged-acs-url-no-www-script

It keeps the ACS the same while still not requiring any www-scripts to be readded.

tvdijen:
That trick doesn't work for the IDP-endpoints

thijskh:
But those do not have the problem

tvdijen:
Superseeded by #1698

PR #1695, opened 2022-09-05
Drop covers lines from tests

thijskh:
The `@covers` lines limit what is counted as coverage to exactly the classes it mentions. I can see the use case for this in some cases (where you have the time and means to explicitly cover each and every method). However, I disagree that _we_ should do that. It's in my opinion perfectly fine if we add unit tests to higher level classes that then indirectly test other classes by invoking them.

Given also that we have a long way to go re coverage I think it's useful that we focus on code that is not run by the testsuite at all, rather than trying to further optimize code that is already (indirectly covered). If lines are hit by the current testsuite, that's already a big bonus versus lines that are not hit at all.

So I propose to remove these annotations from the tests to get a better view of what % of lines/which lines we currently touch with the testsuite and which we don't yet.

### DIFF
--- a/tests/modules/admin/src/Controller/ConfigTest.php
+++ b/tests/modules/admin/src/Controller/ConfigTest.php
@@ -16,7 +16,6 @@ use Symfony\Component\HttpFoundation\Response;
 /**
  * Set of tests for the controllers in the "admin" module.
  *
- * @covers \SimpleSAML\Module\admin\Controller\Config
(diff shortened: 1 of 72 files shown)


tvdijen:
Only made 6% difference ....

thijskh:
Actually more like 20% since I dropped two in my previous PR and this brought it from 42% to 50%.

PR #1698, opened 2022-09-14
Keep the ACS url as it was in 1.19

thijskh:
This is an approach to solving #1684.

We keep all the benefits of the new structure (controllers, routes etc). The single thing that this changes is which string is used to identify the route. This means that we reap all the benefits while at the same time not cause any unnecessary migration pain on this front. Changing the ACS requires not just the deployer to do something but to coordinate with potentially many IdPs, while we can easily avoid this.

The ACS is used for machine consumption, not to be typed by a human, so which exact string is in the URL really doesn't matter. It's just a number of characters to identify it.

In my opinion this solution is simple and does not require to add more code, require people to upgrade a 1.19 install first and then exchange metadata and then exchange metadata again after upgrading to 2.0, or other complex migration instructions.

### DIFF
--- a/docs/simplesamlphp-artifact-idp.md
+++ b/docs/simplesamlphp-artifact-idp.md
@@ -70,12 +70,12 @@ In general, that should look something like:
 'AssertionConsumerService' => [
     [
         'Binding' => 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST',
-        'Location' => 'https://sp.example.org/simplesaml/module.php/saml/sp/assertionConsumerService/default-sp',
+        'Location' => 'https://sp.example.org/simplesaml/module.php/saml/sp/saml2-acs.php/default-sp',
         'index' => 0,
     ],
     [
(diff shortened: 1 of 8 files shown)


tvdijen:
I've discussed this again with Jaime once more and we do support the ease of migration.
On the other hand we want to rationalize the endpoints _at some point_ for our own convenience.

Our proposal therefor is to add a 'legacy mode' flag in 2.0 that can be used to keep the old endpoints alive. This flag would be marked @deprecated immediately, so we can make the next step in 2.1 and drop the old endpoints.

thijskh:
I would tend to agree more with the sentiment if the new URLs would not still contain "module.php/". As discussed this brings effort and complexity to the table so I propose to introduce new URLs only at the point that we can remove both `saml20-acs.php` AND `module.php` from them in one go...

tvdijen:
Sergio had already taken steps in that direction in his #1294, but unfortunately some parts of that PR are very outdated and hard to rebase.. Let's merge this and postpone the endpoint-changes

PR #1710, opened 2022-10-25
entityID change affects saml:SP too

ghalse:
Update the examples and documentation for saml:SP to be consistent with the notion that entityIDs are no longer
dynamically generated if the 'entityID' config option is null.

### DIFF
--- a/config-templates/authsources.php
+++ b/config-templates/authsources.php
@@ -27,8 +27,7 @@ $config = [
         'saml:SP',
 
         // The entity ID of this SP.
-        // Can be NULL/unset, in which case an entity ID is generated based on the metadata URL.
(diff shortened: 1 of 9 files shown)


tvdijen:
We've discussed this internally and we would like to throw an exception when people do not change the default entityID. You think you can add that to this PR?

ghalse:
> We've discussed this internally and we would like to throw an exception when people do not change the default entityID. You think you can add that to this PR?

Done.

It wasn't clear if you wanted to do that for the example in metadata-templates/saml20-idp-hosted.php as well, so I've added that as a separate commit. Feel free to leave that out if you don't want to use that :)

PR #1712, opened 2022-10-26
translateFromArray needs a FALLBACKLANGUAGE

ghalse:
Currently, the translateFromArray filter has no fallback to a default locale/language. This makes it behave differently to the trans tag/filter in Twig and the old translation system. It can result in nothing being displayed even though the array it is passed contains an entry matching the default language (hardcoded as 'en').

Because this is a static function, determining the 'language.default' from the config is expensive. But the existing trans filter (as implemented in TwigTranslator.php) already uses Language::FALLBACKLANGUAGE as the default locale and that class is already loaded. So using that constant here is consistent with the other Twig translation functions.

### DIFF
--- a/src/SimpleSAML/Locale/Translate.php
+++ b/src/SimpleSAML/Locale/Translate.php
@@ -139,6 +139,8 @@ class Translate
             return null;
         } elseif (isset($translations[$context['currentLanguage']])) {
             return $translations[$context['currentLanguage']];
+        } elseif (isset($translations[Language::FALLBACKLANGUAGE])) {
+            return $translations[Language::FALLBACKLANGUAGE];
         }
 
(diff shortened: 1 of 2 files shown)


thijskh (approved):
Yes, good improvement!

If you'd want to make use of the language default in order, we could consider to add this to $context for that purpose?

ghalse:
> If you'd want to make use of the language default in order, we could consider to add this to $context for that purpose?

That could then be used in TwigTranslator too? Ultimately I think that the way the trans tag/filter handles language fallback should be consistent with translateFromArray

thijskh:
I'll merge this since it's a significant improvement but indeed probably this can be improved even further.

PR #1717, opened 2022-11-11
Clean repo & release-script

tvdijen:
This PR does the following things:
- Removes data/ cache/ and log/ dir from the repo and/or release tarball. They shouldn't be in the installation directory. Their default values have been unset in the config template and they programmatically default to the system's temp dir.
- The config- and metadata templates have been moved to .dist files in the config- and metadata-dir. We did basically the same during the release build and it allows us to remove two dirs.
- Instead of building the release tarball ourselves we now use composer to do the job.


tvdijen:
Note: remove v2.0.0-beta99 tag when merged - also from packagist

bin/build-release.sh
```diff
@@ -31,42 +30,24 @@ fi
...
-git clone $REPOPATH $TARGET
+git clone --depth 1 --branch $TAG $REPOPATH $TARGET
 cd $TARGET
-git checkout $TAG
-cd ..
 
-if [ ! -x "$TARGET/composer.phar" ]; then
-    curl -sS https://getcomposer.org/installer | php -- --install-dir=$TARGET
+if [ ! -x "$COMPOSER" ]; then
+    curl -sS https://getcomposer.org/installer | php -- --install-dir=/tmp
```
thijskh, line 37: /tmp might not be executable, but since this is only of concern to those building releases it's probably not an issue

modules/core/src/Storage/SQLPermanentStorage.php
```diff
@@ -33,7 +33,7 @@ public function __construct(string $name, Configuration $config = null)
             $config = Configuration::getInstance();
         }
 
-        $datadir = $config->getPathValue('datadir', 'data/') ?: 'data/';
+        $datadir = $config->getPathValue('datadir', sys_get_temp_dir()) ?: sys_get_temp_dir();
```
thijskh, line 36: data/ is different from tmp (the former is expected to be persistent in ssp). So if you default to /tmp you e.g. lose your sql "permanent" storage after reboot. Maybe better to default to not working than by default storing permanent things in /tmp 'accidentally'?

bin/build-release.sh
tvdijen: To my understanding, this doesn't count for script-files that start with a she-bang, because the actual interpreter (the php binary) is not on the noexec volume.

My next step, now that this script is cleaned up, is to migrate it to a Github Action, where composer will be available out-of-the-box

modules/core/src/Storage/SQLPermanentStorage.php
tvdijen: Agreed. Will change to throw a `CriticalConfigurationException` when unset.

PR #1722, opened 2022-12-05
Psalm seems to think of "set" as a type

pradtke:
I'm am certainly no psalm expert. In my attempts to add more psalm to the ratelimit module psalm was complaining that Configuration's getOptionalString could never return null even if default is null. I think that psalm maybe doesn't understand `$default is set` since when I changed it to `$default is null` or `$default is string` (and adjust ternary expression) then psalm stopped complaining.

Open to other ideas.

### DIFF
--- a/src/SimpleSAML/Configuration.php
+++ b/src/SimpleSAML/Configuration.php
@@ -613,7 +613,7 @@ class Configuration implements Utils\ClearableState
      *                            The default value can be null or a boolean.
      *
      * @return bool|null          The option with the given name, or $default.
-     * @psalm-return              ($default is set ? bool|null : bool)
+     * @psalm-return              ($default is bool ? bool : bool|null)
      *
      * @throws \SimpleSAML\Assert\AssertionFailedException If the option is not boolean.
      */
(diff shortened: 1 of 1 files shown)


tvdijen:
Hmm, I don't understand what the issue is.. Psalm isn't complaining about it at all, and this PR only increases the amount of psalm-errors?

pradtke:
Hi @tvdijen ,  it is hard to show within the SSP project because of its psalm error levels and which things are set to just info.
I [created minimal repo to show ](https://github.com/pradtke/ssp-psalm-check)the psalm error.
There is a super simple test case

https://github.com/pradtke/ssp-psalm-check/blob/f5e8c12e2b5a80e462df0002dc84d32caf0e452d/tests/ConfigurationForPsalmStuffTest.php#L9-L17

and for phpunit it passes fine. However when running `psalm` it reports

`ERROR: DocblockTypeContradiction - tests/ConfigurationForPsalmStuffTest.php:16:16 - Cannot resolve types for $value - docblock-defined type string does not contain null (see https://psalm.dev/155)
        $this->assertNull($value);`

It is basically saying that the return type of `$config->getOptionalString` can never be null.  We know it can be though, so why is psalm confused?  I think it is because of `@psalm-return       ($default is set ? string|null : string)` on `getOptionalString`.  According to [psalms conditional types docs](https://psalm.dev/docs/annotating_code/type_syntax/conditional_types/) the format is `(<template param> is <union type> ? <union type> : <union type>)`.   SSP is using `set` as the `union type` , which isn't a [type in the docs ](https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/)(from what I can see). 

So I think it should be `($default is null ? string|null : string)` or `($default is string ? string: string|null)`

PR #1728, opened 2022-12-21
Adjusted module path from "lib" to "src"

ubiav:
It appears that the custom auth classes (i.e. UserPassBase) were moved from the modules/core/lib to modules/core/src.  Likewise, when creating a custom auth source, I found that the path in my module needs to match this path (e.g. modules/mymodule/src instead of modules/mymodule/lib).  Updated documentation page to reflect the change.


thijskh (approved):
Indeed, thanks for the PR!

PR #1729, opened 2022-12-21
getSessionHandler never returns null

pradtke:
* Fix return type
* Change how static $sessionHandler assigned so psalm can detect it is no longer null.


pradtke:
I changed the signature for  the private `createSessionHandler()` method since `psalm` was unable to detect it was creating the session handler, and then would complain about it being null in the get method.  Let me know if there is a better way to address that.

steinmb:
Drive by review, though look RTBC to me.

tvdijen (approved):

PR #1734, opened 2022-12-30
Micro optimisation and lower memory footprint

steinmb:
Avoid misuse of array_push. PHP 8.x allow us to use the faster `str_contains()`


thijskh (approved):
Thanks!

PR #1735, opened 2023-01-03
Avoid cascading calls in autoloader

steinmb:
Happy new year! 
I'll keep picking low hanging fruits while getting to know the code base, if that is OK.


tvdijen:
Absolutely @steinmb, this is more than welcome!  And I've learned something new, because I've never seen the second parameter to `dirname`.. Very useful!

tvdijen (approved):

PR #1741, opened 2023-01-05
Rationalize NameIDFormat config.

thijskh:
This was a mixture of legacy supported options, some of them deprecated, and data types, and also implemented differently in some places than others.

This changes it once more but hopefully for the better:
* The value is always either an array or null/unset.
* You can set it to a specific array to get a specific policy. (unchanged from now)
* You can leave it unset/null to keep the default policy. (unchanged from now)
* You can set it to the empty array to signal that you do not want any policy to be sent.
* The string and bool types are no longer allowed.

All in all this means that we can make code much simpler, a lot less if-branching, and also typewise make use of correct typehints.

The behaviour changes are as follows:
- We drop the already deprecated option to set it as a string (deprecated in 1.17).
- To not send the element you need to change `false` to `[]`; this was not deprecated before but I believe setting it to false was already broken in master.


modules/saml/src/Auth/Source/SP.php
```diff
@@ -160,13 +160,11 @@ public function getHostedMetadata(): array
         // add NameIDPolicy
         if ($this->metadata->hasValue('NameIDPolicy')) {
```
tvdijen, line 161: If it has a value, it must be an array... We can use `getArray` instead of `getValue`, and we can drop the `is_array` test.

tvdijen (approved):

modules/saml/src/Auth/Source/SP.php
```diff
@@ -159,14 +159,12 @@ public function getHostedMetadata(): array
 
         // add NameIDPolicy
         if ($this->metadata->hasValue('NameIDPolicy')) {
-            $format = $this->metadata->getValue('NameIDPolicy');
-            if (is_array($format)) {
+            $format = $this->metadata->getArray('NameIDPolicy');
+            if ($format !== []) {
                 $metadata['NameIDFormat'] = Configuration::loadFromArray($format)->getOptionalString(
```
tvdijen, line 164: I think we can simplify this by doing `$metadata['NameIDFormat'] = $format['Format'] ?? Constants::NAMEID_TRANSIENT;`

modules/saml/src/Auth/Source/SP.php
```diff
@@ -560,17 +558,12 @@ private function startSSO2(Configuration $idpMetadata, array $state): void
 
         if (isset($state['saml:NameIDPolicy'])) {
             $policy = null;
```
tvdijen, line 560: This should now default to `[]` I think... Psalm is now complaining about passing a possible null-value to `setNameIdPolicy()` on line 567

thijskh: I've made it even simpler now.

tvdijen: Excellent! I think it's ready to merge

tvdijen (approved):

PR #1748, opened 2023-01-24
Fix and improve multiauth module

thijskh:
- Fix several bugs to make it actually work; it was broken in several ways probably due to refactoring in SSP.
- Simplify code a lot, also make use of things we have already available now.
- Move presentation concerns to the template and use standard functionality for translation

A small change needed to be made in the name of the `css_class` setting (previously `css-class`) in the name of simplicity, it seemed better to change that now than to carry more code to convert the name.

### DIFF
--- a/modules/multiauth/templates/selectsource.twig
+++ b/modules/multiauth/templates/selectsource.twig
@@ -5,16 +5,22 @@
     <h2>{{ 'Select an authentication source'|trans }}</h2>
     <p>{{ 'The selected authentication source will be used to authenticate you and and to create a valid session.'|trans }}</p>
 
-    <form action="{{ selfUrl|escape('html') }}" method="get">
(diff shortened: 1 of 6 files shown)


modules/multiauth/templates/selectsource.twig
```diff
@@ -5,15 +5,21 @@
...
+                {%- if key == preferred %}
+                    class="pure-button pure-button-active" autofocus
+                {%- else %}
+                    class="pure-button"
+                {% endif %}
+                id="{{ button }}" value="{{ source.text is defined ? source.text|translateFromArray : key }}">
+            {% if source.help is defined %}
+              <p>{{ source.help|translateFromArray }}</p>
             {% endif %}
             </li>
```
tvdijen, line 24: nitpick: the <li> </li> are not indented correctly

tvdijen (approved):

PR #1749, opened 2023-01-27
RequestedAuthnContext source selector

tvdijen:
Select the proper authsource based on the RequestedAuthnContext in an AuthnRequest.

### DIFF
--- a/modules/core/src/Auth/Source/AbstractSourceSelector.php
+++ b/modules/core/src/Auth/Source/AbstractSourceSelector.php
@@ -59,16 +59,12 @@ abstract class AbstractSourceSelector extends Auth\Source
     public function authenticate(array &$state): void
     {
         $source = $this->selectAuthSource($state);
-        if (count($source) === 1) {
-            $as = Auth\Source::getById($source[0]);
-            if ($as === null || !in_array($source[0], $this->validSources, true)) {
-                throw new Exception('Invalid authentication source: ' . $source[0]);
-            }
-
-            static::doAuthentication($as, $state);
-            return;
+        $as = Auth\Source::getById($source);
+        if ($as === null || !in_array($source, $this->validSources, true)) {
+            throw new Exception('Invalid authentication source: ' . $source);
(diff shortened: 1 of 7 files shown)


tvdijen:
To do:  I would like to implement the comparison != exact scenarios, by calling MutliAuth dynamically..

tvdijen:
@thijskh Jaime tells me to just stick with 'exact' comparison and to run away from the rest.. What do you think? Would it be worth implementing?

modules/core/src/Auth/Source/AbstractSourceSelector.php
```diff
@@ -58,14 +58,17 @@ public function __construct(array $info, array $config)
...
+            if ($as === null || !in_array($source[0], $this->validSources, true)) {
+                throw new Exception('Invalid authentication source: ' . $source[0]);
+            }
+
+            static::doAuthentication($as, $state);
+            return;
         }
-
-        static::doAuthentication($as, $state);
+        throw new Exception('Not implemented.');
```
thijskh, line 71: Why are you going to the changes of making this return an array now if it's not implemented? Can you then just not leave it at the single value it was?

thijskh:
I agree with Jaime. We had a similar discussion at SURFconext. We know of no known use of the other variants, it seems a theoretical thing from the spec, using exact is clear and sound what you will get. I would leave it out, it keeps the code a lot simpler and no one will miss the other variants.

modules/core/src/Auth/Source/AbstractSourceSelector.php
tvdijen: It was preparation for supporting the other comparisons.. Will revert

tvdijen:
There appears to be something wrong with codecov... I noticed there wasn't an automatic comment in this PR, and now that I check on the codecov website I see only files in the src/ directory are covered and the modules/ directory seems ignored

thijskh (approved):

PR #1751, opened 2023-01-30
Tone down the error level of missing locale dirs

thijskh:
It is quite fine if e.g. a theme does not have its own locales dir. The exception is already caught, so the application does not actually stop. So, notice it but do not raise high alarms.


tvdijen:
Should be INFO-level, since NOTICE is reserved for statistics

tvdijen (approved):

PR #1754, opened 2023-02-01
Reinstate url-based authentication

### LINKED ISSUE #1682: 2.0-rc1 Login page not found
**Describe the bug**
After upgrading from 1.19.3, the URL returned by `SimpleSAML\Auth\Simple::getLoginURL()` displays a cryptic error message:

> The given page was not found. The reason was: The URL wasn't found in the module. The URL was: https://gerben.alisqi.com.test-xxx.platformsh.site/sso/module.php/core/login/gerben?AuthId=gerben&ReturnTo=https%3A%2F%2Fgerben.alisqi.com.test-xxx.platformsh.site%2F%3Fmod%3Dlogin%26ssoLogin

image

The `config.php` was changed as described in the changelog.

tvdijen:
Restores the functionality to trigger authentication by URL.
To not expose individual authsources like we did before, a check was added to test the passed authsource to be saml:SP. Those are the only ones safe to expose like this.

Closes #1682

### DIFF
--- a/modules/saml/src/Controller/ServiceProvider.php
+++ b/modules/saml/src/Controller/ServiceProvider.php
@@ -132,13 +132,6 @@ class ServiceProvider
             'ReturnTo' => $httpUtils->checkURLAllowed($returnTo),
         ];
 
-        /**
-         * Allows a saml:idp query string parameter specify the IdP entity ID to be used
-         * as used by the DiscoJuice embedded client.
-         */
-        if (!empty($request->query->has('saml:idp'))) {
-            $options['saml:idp'] = $request->query->get('saml:idp');
-        }
         $as->requireAuth($options);
 
         return new RunnableResponse([$httpUtils, 'redirectTrustedURL'], [$returnTo]);
--- a/modules/saml/src/Controller/ServiceProvider.php
+++ b/modules/saml/src/Controller/ServiceProvider.php
@@ -106,7 +106,7 @@ class ServiceProvider
 
 
     /**
-     * Start single sign-on for an SP identified with the specified entityID
+     * Start single sign-on for an SP identified with the specified Authsource ID
      *
      * @param \Symfony\Component\HttpFoundation\Request $request
      * @param string $sourceId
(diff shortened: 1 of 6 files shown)


modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -105,6 +105,46 @@ public function setAuthUtils(Utils\Auth $authUtils): void
     }
 
 
+    /**
+     * Start single sign-on for an SP identified with the specified entityID
```
thijskh, line 109: not "entityID" but "auth source ID"

modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -105,6 +105,46 @@ public function setAuthUtils(Utils\Auth $authUtils): void
...
+         * Setting up the options for the requireAuth() call later..
+         */
+        $httpUtils = new Utils\HTTP();
+        $options = [
+            'ReturnTo' => $httpUtils->checkURLAllowed($returnTo),
+        ];
+
+        /**
+         * Allows a saml:idp query string parameter specify the IdP entity ID to be used
+         * as used by the DiscoJuice embedded client.
```
thijskh, line 137: DiscoJuice is long dead... so do we need this?

thijskh (approved):

modules/saml/src/Controller/ServiceProvider.php
tvdijen: The setting itself seems not specific to DiscoJuice.. It is referenced in `modules/saml/docs/sp.md`

thijskh: Maybe just drop it from the comment then... referencing the dead product will likely make the same question arise in the future?

tvdijen: I think we can drop the entire code-block since the only references are to the SP authsource.

thijskh (approved):
Great, let's merge it, if someone misses the option no doubt we will hear about it

PR #1758, opened 2023-02-09
Registration authority consistency

ghalse:
The public documentation for the mdrpi extensions uses keys of `authority`, `identity`, and `policies`. However, SAMLBuilder creates PHP-style metadata with different keys which is inconsistent. This patch changes the behaviour to make `authority` behave consistently and adds support for the other two options.


thijskh (approved):
Yes, this is in essence a bugfix since it has been documented (also in <=1.19), and other code generates output, that has this format.

PR #1760, opened 2023-02-13
Propagate Symfony responses throughout the code

tvdijen:
This PR is far from complete, but it's big enough to review as it is already. Consider this a first batch.
I've verified that basic SP/IDP redirections still work.


modules/exampleauth/src/Auth/Source/External.php
```diff
@@ -189,21 +186,21 @@ public function authenticate(array &$state): void
...
-    public static function resume(Request $request): void
+    public static function resume(Request $request): Response
     {
         /*
          * First we need to restore the $state-array. We should have the identifier for
          * it in the 'State' request parameter.
          */
-        if (!$request->query->has('State')) {
+        if (!$request->query->has('AuthState')) {
             throw new Error\BadRequest('Missing "State" parameter.');
```
thijskh, line 196: Message is not updated for the change of parameter name

modules/exampleauth/src/Controller/ExampleAuth.php
```diff
@@ -173,22 +173,22 @@ public function authpage(Request $request)
...
-    public function redirecttest(Request $request): RunnableResponse
+    public function redirecttest(Request $request): Response
     {
         /**
          * Request handler for redirect filter test.
          */
-        $stateId = $request->query->get('StateId');
+        $stateId = $request->query->get('AuthState');
         if ($stateId === null) {
             throw new Error\BadRequest('Missing required StateId query parameter.');
```
thijskh, line 185: Message is not updated for the change of parameter name

src/SimpleSAML/Utils/HTTP.php
```diff
@@ -229,36 +230,15 @@ private function redirect(string $url, array $parameters = []): void
...
-            // set the location header
-            header('Location: ' . $url, true, 303);
+        $response = new RedirectResponse($url, 303);
 
+        if (!headers_sent()) {
             // disable caching of this response
-            header('Pragma: no-cache');
-            header('Cache-Control: no-cache, no-store, must-revalidate');
+            $response->headers->set('Pragma', 'no-cache');
+            $response->headers->set('Cache-Control', 'no-cache, no-store, must-revalidate');
```
thijskh, line 238: Better to use `$response->setCache(...)` instead

thijskh (commented):
Definitely a good cleanup. Does touch a lot of things so may require intensive testing though.

src/SimpleSAML/Utils/HTTP.php
thijskh: Or even better just drop it... the 303 already documents this is a temporary redirect.

tvdijen: You think I can drop the Pragma-header? I read it's HTTP/1.0 specific ..

tvdijen:
This is about as much as I can do right now.. If we want to take the next step, we also have to update the saml2-lib bindings to return a Symfony-response in their ::send() methods.

src/SimpleSAML/Utils/HTTP.php
thijskh: The pragma header is automatically handled by Symfony if you use the setCache() method I suggested. So better leave this to symfony if adding any caching headers at all. Or just skip it. FWIW OC EngineBlock does not add explicit caching headers to redirects.

src/SimpleSAML/Auth/Source.php
```diff
@@ -136,13 +136,13 @@ public function reauthenticate(array &$state): void
...
      *
-     * This function should be called if authentication has completed. It will never return,
+     * This function should be called if authentication has completed. It will return a Response,
      * except in the case of exceptions. Exceptions thrown from this page should not be caught,
      * but should instead be passed to the top-level exception handler.
      *
      * @param array &$state Information about the current authentication.
      */
-    public static function completeAuth(array &$state): void
+    public static function completeAuth(array &$state): Response
```
tvdijen, line 145: Side-quest: should we change the signature to non-static & protected, since all calls to this are made by child-classes?

tvdijen:
I thoroughly tested this PR in conjunction with a new version of the saml2-lib.
The lib was changed to:
- Bump PHP functional level to 8.0 and take advantage of it's features
- Alter it's namespace to `SimpleSAML\SAML2`
- Alter the binding-classes to work with PSR-17 HTTP messages as required by this PR

@thijskh OK to merge?

sinteur:
I'd recommend a 'yes' to thijskh :-D

jaimeperez:
I haven't been able to look closely into everything, but this kind of development is indeed a massive improvement! 👏

PR #1764, opened 2023-02-21
Migrate old www-script to controller

tvdijen:
This page is, as far as I could see, only used after performing a logout on the admin's 'Test authsource' page.
The parameters `link_href` and `link_text` were not used anywhere as well, so I removed them.


thijskh (approved):
Seems good indeed, I wonder if we even need this more or less empty page at all?

tvdijen:
It won't hurt for now...

PR #1769, opened 2023-02-27
Bugfix: Fix use of back-button

### LINKED ISSUE #1720: 2.0 UI back-button glitch
**Describe the bug**
When pressing the back-button after successfully login in, the login-form is shown with a disabeled submit-button.

On top of that, refreshing that page leads to a page-not-found error.

The back-button leads me to:
`https://idp.moo-archive.nl/module.php/core/loginuserpass?AuthState=_f39cfb541de5dd79f5973c2d2782e9270f7cc76504%3Ahttps%3A%2F%2Fidp.moo-archive.nl%2Fmodule.php%2Fcore%2Flogin%2FAD-LDAP%3FAuthId%3DAD-LDAP%26ReturnTo%3Dhttps%253A%252F%252Fidp.moo-archive.nl%252Fmodule.php%252Fadmin%252Ftest%252FAD-LDAP` which seems broken in the first place.. There is no query param after the AuthState one.

After refreshing I end up with a 404 without AuthState var.
https://idp.moo-archive.nl/module.php/core/login/AD-LDAP?AuthId=AD-LDAP&ReturnTo=https%3A%2F%2Fidp.moo-archive.nl%2Fmodule.php%2Fadmin%2Ftest%2FAD-LDAP

**To Reproduce**
1. Go to Admin module > Test authsources
2. Select one and log in
3. On the 'show attributes' page press the back-button
4. See error 1; the button is disabled
5. Refresh the page
6. See error 2; 404 error

**Expected behavior**
Using the back-button has always been problematic in authentication flows, so either we prevent the use of it, or we make it work.

**Screenshots or logs**

error 1

error 2

**Additional context**
Running v2.0.0-rc2 / Centos Stream (latest) / PHP 7.4 (fpm)

tvdijen:
This PR fixes two issues with the back-button;

The JavaScript-part makes sure the submit-button becomes enabled again.
Now after fixing this, a new problem arised; the restart-url would be triggered to restart the authentication, but we used `getLoginURL` to generate it, which was recently limited to only work for SP-authsources. The second fix is therefor to properly set a restart-url that works for non-SP authsources.

Closes #1720

### DIFF
--- a/modules/core/public/assets/js/loginuserpass.js
+++ b/modules/core/public/assets/js/loginuserpass.js
@@ -1,6 +1,5 @@
 ready(function () {
     var button = document.getElementById("submit_button");
-    var form = document.getElementById("f");
 
     window.onpageshow  = function () {
         var button = document.getElementById("submit_button");
@@ -13,11 +12,5 @@ ready(function () {
         button.innerHTML = button.getAttribute("data-processing");
         button.disabled = true;
     }
-
-    button.onclick = function () {
-        var form = document.getElementById("f");
-        form.submit();
-        return true;
-    };
 });
 
--- a/modules/core/public/assets/js/loginuserpass.js
+++ b/modules/core/public/assets/js/loginuserpass.js
@@ -1,9 +1,20 @@
 ready(function () {
     var button = document.getElementById("submit_button");
-    button.onclick = function () {
-        this.innerHTML = button.getAttribute("data-processing");
-        this.disabled = true;
+    var form = document.getElementById("f");
 
+    window.onpageshow  = function () {
+        var button = document.getElementById("submit_button");
+        button.innerHTML = button.getAttribute('data-default');
+        button.disabled = false;
+    }
+
+    form.onsubmit = function () {
+        var button = document.getElementById("submit_button");
+        button.innerHTML = button.getAttribute("data-processing");
+        button.disabled = true;
+    }
+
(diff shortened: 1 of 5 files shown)


modules/core/public/assets/js/loginuserpass.js
```diff
@@ -1,9 +1,20 @@
...
+    form.onsubmit = function () {
+        var button = document.getElementById("submit_button");
+        button.innerHTML = button.getAttribute("data-processing");
+        button.disabled = true;
+    }
+
+    button.onclick = function () {
         var form = document.getElementById("f");
         form.submit();
         return true;
```
thijskh, line 20: I think this can now be removed? Since the onclick event just replicates what the normal button click would do. Then also line 2 can be removed.

modules/core/public/assets/js/loginuserpass.js
```diff
@@ -1,12 +1,16 @@
 ready(function () {
     var button = document.getElementById("submit_button");
```
thijskh, line 2: Might be a bit nitpicky, but this is now unused and can also be removed

thijskh (approved):

modules/core/public/assets/js/loginuserpass.js
tvdijen: Actually, no, I removed the wrong one! We need one for `form`

Issue #1777, opened 2023-03-02
CVE-2022-24894 for symfony/http-kernel, framework-bundle


fenetikm:
**Describe the bug**
There's a CVE for symfony/http-kernel < 4.4.50 as reported here:
https://www.cve.org/CVERecord?id=CVE-2022-24894

Which affects v1 of SimpleSAMLphp (not sure about v2).

I think all that's required to fix it is to bump the dependency in composer.json and composer.lock.

Will look into creating a PR if not picked up soon by someone more familiar with all the things.

Thanks.

thijskh:
It has been fixed in v2 indeed already. Are you sure it actually affects v1? Because I'm not sure we use this functionality there. We've committed the update to the branch just in case but I'm not sure we need to trigger a 1.19 release just for this now.

fenetikm:
Apologies, but I believe it's not just http-kernel but other parts of symfony < 4.4.50 such as the framework-bundle which are referenced in the lock file in the 1.19 branch.

e.g.

https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-1.19/composer.lock#L3032

The scanner we are using (snyk) picked it up.

fenetikm:
Have a look at the PR and see what you think - there isn't a tag for framework-bundle 4.4.50 but apparently the latest version of the 4.4 branch does the trick.

fenetikm:
Have had a bit more of a look and I think there isn't actually something to "fix" i.e. the code is the same between the 4.4 branch and the 4.4.49 tag:

https://github.com/symfony/framework-bundle/compare/v4.4.49...4.4

However, when I replaced my vendor version of ssp with the one in the PR that points to 4.4.x-dev the scanner said everything was fine.

tvdijen:
From what I read, people using v1.1.9 with the new UI would be affected

tvdijen:
v1.19.8 was released

fenetikm:
Just tried out 1.19.8 - indeed fixes it. Thanks so much!

PR #1778, opened 2023-03-02
Fix for CVE 2022 24894

fenetikm:
As per https://github.com/simplesamlphp/simplesamlphp/issues/1777


tvdijen:
v1.19.8 was released

PR #1785, opened 2023-03-07
IdP logout fix case sensitive key in logout

jirinapravnik:
In my opinion there should be **entityid** in lower case.  Here is inititated with lower case: https://github.com/simplesamlphp/simplesamlphp/blob/master/src/SimpleSAML/Metadata/MetaDataStorageHandlerFlatFile.php#L127

Without this change I got "Unable to initialize logout to " in logs. And was logouted only from SP, where I click on Logout not from next SPs.

With this change I am logouted from all SPs without problems.

### DIFF
--- a/modules/saml/src/Message.php
+++ b/modules/saml/src/Message.php
@@ -534,7 +534,7 @@ class Message
     ): LogoutRequest {
         $lr = new LogoutRequest();
         $issuer = new Issuer();
-        $issuer->setValue($srcMetadata->getString('entityID'));
+        $issuer->setValue($srcMetadata->getString('entityid'));
         $issuer->setFormat(Constants::NAMEID_ENTITY);
         $lr->setIssuer($issuer);
 


thijskh:
Thanks. It took me a while to wrap my head around this but I think this is correct. For a hosted SP we may have both entityID and entityid available. For remote entities we may not. This is a confusing part of the code and when we attempted to clean it up it did not get less messy.

jirinapravnik:
@thijskh Yes, it is quite messy, I have benn looking for this bug for 3 days :-) I would like to help with cleaning out and make it better, but it is probably over my ability. This is at least easy and fixing solution for working logout.

thijskh:
That is the confusing part. This is not about SP (which is configured since always with uppercase and has lowercase variant) but about IdP metadata, which only has lowercase version.

tvdijen:
In that case it should at least be mentioned in here; https://simplesamlphp.org/docs/stable/simplesamlphp-reference-idp-hosted.html

thijskh:
I don’t think there’s a relation with config files

tvdijen:
Maybe I'm missing something, but it's pulling `entityid` from the source metadata config, right?
That item is not documented right now.

jirinapravnik:
IMHO there is nothing to update in docummentation. On this line: https://github.com/simplesamlphp/simplesamlphp/blob/master/src/SimpleSAML/Metadata/MetaDataStorageHandlerFlatFile.php#L127 it takes array key from idp-hosted metadata and put it into array with key entityid (lower case) but it is reading entityID. It is internal thing, not wrong config.

tvdijen:
My point is that `entityID` is not listed in the docs as well..

thijskh:
For the metadata the entityID comes from the outer array key (so unnamed), not from a named value inside the array.

thijskh:
Conflict resolved & merged for master and 2.0. Thanks!

PR #1813, opened 2023-05-31
Bugfix: Set destination on signed LogoutResponse for POST and Redirect bindings

### LINKED ISSUE #1812: Logout with remote Idp and signed response fails.
**Describe the bug**

Logout with remote Idp and signed response fails.

**To Reproduce**

I set up a SP with a remote Idp(keycloak).
Everything works fine as long as I turn OFF "Client signature required" in the keycloak client.

I can:
- Login 
- Logout through the SP and 
- Logout through the Idp(keycloak) with SLO configured. Both the SP and the Idp are logged out.

As soon as I turn on "Client signature required" in keycloak and configure the metadata to sign the request with 
`sign.authnrequest` and `sign.logout` set to true.

I can:
- Login 
- Logout through the SP

I cannot:
- Logout through the Idp(keycloak) with SLO configured. It FAILS 

I receive an error from keycloak invalid request. "missing_required_destination".

I checked both behaviors from "SP logout" See this method
https://github.com/simplesamlphp/simplesamlphp/blob/b6626504c8b7c5e658326e9931cb4807b5ee2ad0/modules/saml/src/Auth/Source/SP.php#L969

and I can see that the destination is set.
https://github.com/simplesamlphp/simplesamlphp/blob/b6626504c8b7c5e658326e9931cb4807b5ee2ad0/modules/saml/src/Auth/Source/SP.php#L1001

When I looked at the "Logout through the Idp(keycloak) with SLO"
Endpoint module.php/saml/sp/saml2-logout.php/default-sp

I can see that this method is called
https://github.com/simplesamlphp/simplesamlphp/blob/b6626504c8b7c5e658326e9931cb4807b5ee2ad0/modules/saml/src/Controller/ServiceProvider.php#L467

Further down I can see that only for SOAP bindings the destination is set.

https://github.com/simplesamlphp/simplesamlphp/blob/b6626504c8b7c5e658326e9931cb4807b5ee2ad0/modules/saml/src/Controller/ServiceProvider.php#L589-L599

When I modify the code and set the destination it all works fine.

```php
<?php
...

 if (!($binding instanceof SOAP)) { 
     $binding = Binding::getBinding($dst['Binding']); 
     if (isset($dst['ResponseLocation'])) { 
         $dst = $dst['ResponseLocation']; 
     } else { 
         $dst = $dst['Location']; 
     } 
     $binding->setDestination($dst);
     $lr->setDestination($dst); 
 } else { 
     $lr->setDestination($dst['Location']); 
 }
...
```

I also can solve the issue by setting the "sign.logout" key in the metadata to false which breaks the "Logout through the SP" functionality and the problem is just the other way around.

Not sure where the problem is exactly Keycloak or this library 
but this is as far as I can see wrong behavior just because I turn on client signature?

**Expected behavior**
Probably signed out of SP and Idp

tvdijen:
Closes #1812

Additionally the saml2-library has been bumped to reflect the specs requirement on incoming messages;
https://github.com/simplesamlphp/saml2/commit/b4daf21db8dba7dd6c3591286794ce5c9ccc7dc1

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "composer/ca-bundle",
-            "version": "1.3.5",
+            "version": "1.3.6",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/ca-bundle.git",
-                "reference": "74780ccf8c19d6acb8d65c5f39cd72110e132bbd"
+                "reference": "90d087e988ff194065333d16bc5cf649872d9cdb"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/74780ccf8c19d6acb8d65c5f39cd72110e132bbd",
-                "reference": "74780ccf8c19d6acb8d65c5f39cd72110e132bbd",
+                "url": "https://api.github.com/repos/composer/ca-bundle/zipball/90d087e988ff194065333d16bc5cf649872d9cdb",
+                "reference": "90d087e988ff194065333d16bc5cf649872d9cdb",
                 "shasum": ""
             },
             "require": {
@@ -64,7 +64,7 @@
             "support": {
                 "irc": "irc://irc.freenode.org/composer",
                 "issues": "https://github.com/composer/ca-bundle/issues",
-                "source": "https://github.com/composer/ca-bundle/tree/1.3.5"
+                "source": "https://github.com/composer/ca-bundle/tree/1.3.6"
             },
             "funding": [
                 {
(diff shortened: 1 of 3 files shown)


thijskh:
Why not always set a destination on the message?

And if we do that, we can remove setting it again on the binding.

tvdijen:
@thijskh What do you think? Should this be changed, or merged as-is?

thijskh:
I think it's better to keep it simpler, all these "if"'s make code complex and lead to subtle behavioural differences between different conditions, while I believe it can be quite simple.

How about you merge this as the bugfix on the release branch, and in master commit a simpler version that just sets the destination always on the message and not on the binding? Then you have the most targeted fix for bugfix release while having the most maintainable code in the longer run.

tvdijen:
OK, I just hope we don't break anything then..

tvdijen:
@thijskh Is this what you had in mind for master?
https://github.com/simplesamlphp/simplesamlphp/commit/ee4f938aaff1e01bacf544da50ac3f2000edfc2b

thijskh:
Indeed much simpler, I think `$signedMessage` is not used anymore then?

PR #1814, opened 2023-06-05
Correction of German translations 

mimkes:
Hi!
I've corrected a few German translations, mainly "Heimorganisation" to "Heimatorganisation".


tvdijen:
Vielen Dank Julika!

PR #1817, opened 2023-06-22
feat: store authsources in a database (fix #267)

### LINKED ISSUE #267: Store authsources in a database
Now that #189 has been done it is now possible to store metadata for SAML in a database. 

However SAML and a number of the other SSO protocols that SimpleSAMLPhp supports partly (or wholly) store their configuration in authsources.php and currently there is no way to store this in the database.

It would be great to improve on the work by #189 to make SimpleSAMLPhp metadata storage truly store agnostic and open the way for self-service configuration by the site user.

damikael:
If, into config/config.php, are present the parameters:

```
'authsources.storage' => 'database',
'authsources.database_table' => 'authsources',
```

then authsources are retrieved from database,
else them are retrieved from file same authsources.php, as before.

If the value of parameter `authsources.storage` is '**database**', then the authsource configuration is retrieved from database.
If the table referred by the parameter `authsources.database_table` does not exists it will be created.
The table will contain the following fields:
 - `id` VARCHAR(255) PRIMARY KEY, that is the key of authsource
 - `entity_data` JSON, that is the configuration of authsource
 - `_disabled` enum('N','Y') NOT NULL DEFAULT 'N', to enable/disable the authsource


tvdijen:
I don't think we should put a lot of code in a configuration file...

src/SimpleSAML/Configuration.php
```diff
@@ -312,8 +305,7 @@ private static function getFileConfig($filename, $configSet, $required=true)
             if ($configSet !== 'simplesaml') {
                 throw new \Exception('Configuration set \'' . $configSet . '\' not initialized.');
             } else {
-                $configUtils = new Utils\Config();
-                self::$configDirs['simplesaml'] = $configUtils->getConfigDir();
+                self::$configDirs['simplesaml'] = Utils\Config::getConfigDir();
```
tvdijen, line 308: This is not right

src/SimpleSAML/Configuration.php
```diff
@@ -245,24 +245,18 @@ public static function setPreLoadedConfig(
...
-    ) : Configuration {
-        
-        $config = array();
-
-        if($filename == 'authsources.php') {
+    public static function getConfig($filename = 'config.php', $configSet = 'simplesaml', $required = true)
+    {
+        $config = [];
 
+        if ($filename == 'authsources.php') {
```
tvdijen, line 252: Use strict comparison `===`

src/SimpleSAML/Configuration.php
damikael: fixed

src/SimpleSAML/Configuration.php
damikael: done

PR #1819, opened 2023-07-02
The setNameIdPolicy method requires the NameIDPolicy class parameter, but it passed an array.

msyk:
The fromArray static method applies to pass the setNameIdPolicy method.


tvdijen:
You must be ignoring our lock-file

PR #1829, opened 2023-07-24
Trim username

dnmvisser:
This was brought up [on slack](https://simplesamlphp.slack.com/archives/CU5E4P8PK/p1690188598460389).
It makes sense to trim the username.
But, maybe we should also trim the password?

### DIFF
--- a/modules/core/src/Auth/UserPassBase.php
+++ b/modules/core/src/Auth/UserPassBase.php
@@ -291,6 +291,11 @@ abstract class UserPassBase extends Auth\Source
             throw new Exception('Could not find authentication source with id ' . $state[self::AUTHID]);
         }
 
+        /* Do not take leading/trailing whitespace into account, this sometimes is accidentally pasted
+         * in the username field (and hard to see).
+         */
+        $username = trim($username);
+
         /*
          * $source now contains the authentication source on which authenticate()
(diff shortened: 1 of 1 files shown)


tvdijen:
I ran a test on Active Directory:

If I create a user starting with a space it will get trimmed of.
If I start my password with a space it will honor that.

Same tests performed with a trailing space, with the same outcome. We therefor **shouldn't** strip whitespace from the password.

Since this is really dependent on the backend used, I think this should be properly handled by the auth source implementing the UserPassBase.

dnmvisser:
> 



> Since this behaviour is really dependent on the backend being used, I think this should be properly handled by the auth source implementing the UserPassBase.

OK let's not touch that then and just trim the username

tvdijen:
Would it be an idea to do this in the Twig template?
This way we leave the possibility for people to override the template if they decide they need untrimmed usernames..
Adding `pattern="[\S]+"` to the username-input should do the trick, although we would have to test how this behaves when copy-pasting something that starts/ends with whitespace

jaimeperez:
I was actually convinced that this was done in the template, in fact. I'm divided as to whether we should allow users to type usernames with a space in any of the two sides or not. But I see clearly that we shouldn't rely on templates, since that check then goes away as soon as people uses their own templates.

PR #1832, opened 2023-07-25
Bugfix: do not expose cron-API if it's not properly configured

### LINKED ISSUE #1831: Web-based cron should be disabled unless a key is set
Web-based cron should always be configured to use a strong(er) key to improve security. If a (proper) key is not configured, web-based cron should be disabled.

The [default configuration](https://github.com/simplesamlphp/simplesamlphp/blob/d81adb91d3e9522a422ca8aafeed8db606441287/modules/cron/config/module_cron.php.dist) file for the cron module sets the key to `secret`. By default the cron endpoint is publicly accessible, so if the key is not changed, anyone can force running the cron tasks. The documentation does mention using a `RANDOM_KEY` as the value, so I agree that misconfiguration is human error.

The default configuration should use `null` or an empty string as the default key value. If the key is `null` or has length 0, web-based cron could be disabled, or require requests originating from localhost. This would also double as a simple way to disable web-based cron if the CLI option is used.

tvdijen:
Closes #1831

To do: backport to 2.0 branch

### DIFF
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -5,6 +5,10 @@
 This document lists the changes between versions of SimpleSAMLphp.
 See the upgrade notes for specific information about upgrading.
 
+## Version 2.0.5
+
+* Never expose the cron-API if secret is not properly configured (#1831)
+
 ## Version 2.0.4
 
 Released TBD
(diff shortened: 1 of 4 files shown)


thijskh (approved):

thijskh:
Good for master.

cannot really backport as-is to 2.0 as this will break bc. Maybe log a warning there?

PR #1836, opened 2023-08-03
Import the conditional authproc-filters

### LINKED ISSUE #1693: Conditional authproc filters
I'm successfully using **Second Factor Only (SFO)** from https://github.com/OpenConextApps/simplesamlphp-module-stepupsfo. That is implemented by configuring a few authproc filters in sequence.
Configuring the authproc filter means that the second factor functionality is enabled for everyone.
It would be good if authproc filter(s) could be made conditional, for instance on the presence/absence of specific attributes.
This would allow a phased approach, for example only apply SFO to users that are part of a specific group in our directory.
Once this is running without problem, require more groups, and eventually to everyone.

### LINKED ISSUE #1766: Conditional AuthProcFilter
**Problem**
I didn't find an easy way to support the REFEDS Assurance Framework and MFA without a third-part module developed by CIRRUS[1]. Their module allowed me to respect the requirements about the AuthnContextClassRef that an IdP and an SP have to respect for asking and answering MFA. 
The Request (SP) and the Response (IdP) have to contain the right value of AuthnContextClassRef **only if** they ask/provide a MultiFactorAuthentication. 
CIRRUS has implemented an easy way to use one or another AuthProcFilter through a condition and this made possible to address the REFEDS Assurance Framework. Their module allow me to activate an AuthProcFilter only on a specific condition: if an SP requests MFA, than the IdP answers with MFA, else... The IdP answers with a username/password login page only. 

https://github.com/cirrusidentity/simplesamlphp-module-cirrusgeneral/#phpconditionalauthprocinserter

I ask to do the same with only official SimpleSamlPhp code. I hope to find an official way to use the AuthProcFilters conditionally to be able to manage different behaviors of SPs or other. 

I didn't find any other easy solution other than CIRRUS Plugin:

https://github.com/cirrusidentity/simplesamlphp-module-cirrusgeneral/#phpconditionalauthprocinserter

tvdijen:
Taken from cirrusidentity/simplesamlphp-module-general with thanks to @pradtke 

Closes #1693
Closes #1766

### DIFF
--- a/docs/simplesamlphp-authproc.md
+++ b/docs/simplesamlphp-authproc.md
@@ -111,6 +111,20 @@ $metadata['https://example.org/saml-idp'] = [
 
 The example above is in `saml20-idp-hosted`.
 
+## Preconditional filters
+
+Any filter can be configured with a precondition that will determine whether or not a filter should run.
+The condition is represented as a string that will be evaluated, similar to the `core:PHP` filter. It also has
+the `$attributes` and `$state` variable available for use.
+The code must return either `true` to run the filter, or `false` to skip it.
+
+```php
+'authproc' => [
+    40 => 'core:TargetedID',
+    '%precondition' => 'return $attributes["displayName"] === "John Doe";',
+],
+```
+
 ## Auth Proc Filters included in the SimpleSAMLphp distribution
 
 The following filters are included in the SimpleSAMLphp distribution:
--- a/src/SimpleSAML/Auth/ProcessingChain.php
+++ b/src/SimpleSAML/Auth/ProcessingChain.php
@@ -17,6 +17,7 @@ use function call_user_func;
 use function count;
 use function is_array;
 use function is_string;
+use function sprintf;
 use function str_replace;
 use function var_export;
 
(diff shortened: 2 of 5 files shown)


tvdijen:
I'm wondering if we could or _should_ provide a better way for people to manipulate the authproc-chain, instead of using a ReflectionClass here

pradtke:
@tvdijen hmm, interesting approach.  a couple of our use cases are using the if-else capability in the cirrus module version, and being able to insert multiple authproc filters for a single condition.  For `if-else` it is often to insert different sets of filters so I don't think it would be something that could be added to the `ProcessingChain` directly.

I do like that with your alternative approach it can be enabled for any authproc without needing to wrap it. Though I would request that you use a different method name than `checkCondition()`  - solely for my selfish reason that `ConditionalProcessingFilterInserter` currently uses that name with `if-else` semantics, and it will make my life easier for module maintenance if SSP doesn't interpret that method with just `if` semantics. 

I don't know if people using `ConditionalProcessingFilterInserter` are using it more for the simpler case that `6f7b3e3` addresses, or using it for more complex use cases.

I agree with your point about not using reflection if the `ConditionalProcessingFilterInserter` is incorporated into SSP. Perhaps incorporating the methods from `ProcessingChainRuleInserter` into `ProcessingChain` would address that?

I'm out for 2 weeks, so may not respond to anything for awhile (and apologies if I missed any existing question in here).

tvdijen:
Thanks @pradtke !
To sum things up, I'll continue with this new approach, rename the `checkCondition` for your convenience and merge the two methods from your `ProcessingChainRuleInserter` into our `ProcessingChain`

marioheber (approved):

tvdijen:
Well that's uncalled for :')

marioheber:
> Well that's uncalled for :')
Hey there, sorry.

tvdijen:
> Hey there, sorry.

No worries, I just think it's strange.
I don't take reviews without any comments from random persons.
I need a review from my fellow devs and no one else.

tvdijen:
@pradtke I think your holidays are over by now.. How does this look to you?

pradtke:
@tvdijen I think this looks really good. Thank you for taking this forward and getting it incorporated.

PR #1843, opened 2023-08-11
Integrate sanitycheck hook into admin module

thijskh:
The sanitycheck module overlaps the config check functionality in the admin module. Make it one place where the results of all sanity checks are shown. This is done in admin so the user does not need to install a separate module for this - that will no doubt lead to most checks never being ran.

The change kept the existing sanitycheck hook format, so all modules implementing this hook will work out of the box without changes required.

The 'core' module hooks have been removed since the checks duplicate already present checks in the config page.

As an example, the saml-specific checks have been moved from the admin to saml module.

More changes are possible, e.g. the check for LDAP could be moved to the ldap module so it's only shown if you actually use LDAP. Etc.

An alternative approach would be to keep the sanitycheck module, make it part of the core module set, and move all existing checks in the admin module into sanity check hooks.


tvdijen (approved):
Excellent work! Just a few sniffer-issues to fix and then merge!

tvdijen:
This btw is something that should be well documented in upgrade notes.

tvdijen:
That this functionality has become implicit and it's no longer required to install the sanitycheck-module?

thijskh:
Yes. Although it doesn’t hurt if you keep it.

tvdijen:
Should I add this to the 2.1 release?

thijskh:
I don't see why not. It's simple and indeed does not break anything existing.

PR #1854, opened 2023-08-25
Default for entity nameformat when not set should be uri

thijskh:
This matches both the code, and the code comments, [the documentation](https://simplesamlphp.org/docs/stable/simplesamlphp-metadata-extensions-attributes.html) and the behaviour of 1.19 so it seems to be a bug due to refactoring that it was changed to Unspecified in 2.x.


tvdijen:
Good catch, thanks!

PR #1857, opened 2023-08-31
Replace composer-dependency with phar-file

### LINKED ISSUE #1855: Move composer/composer dependency to require-dev
We are attempting to update to simplesamlphp 2.x from 1.19.x and are struggling with the composer/composer dependency, because one of our other package dependencies explicitly blocks composer/composer with:

```json
"conflict": {
  "composer/composer": "*"
}
```

The reason this conflict is defined is because there were a number of issues using composer within the DDEV local environment alongside composer/composer inside the vendor directory. We ended up with two versions of composer which were out of sync and it was just a mess.

Having looked through the simplesamlphp code I was struggling to see why composer/composer would need to be a required dependency, I noticed composer being used in the github actions, but presumably if it was necessary that could be added as a step/task in the github action. 

Is there any way that the dependency on composer/composer could be moved to being a dev dependency, or have I missed where it is used in this package?

We are looking into seeing if we can remove the conflict on composer/composer elsewhere, but because we had a lot of trouble with it in the past, it is unlikely that we will want to reintroduce composer/composer into our dependency chain.

tvdijen:
Closes #1855

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           # Should match the minimum required version for SimpleSAMLphp
           php-version: '8.0'
-          tools: composer:v2
+          tools: composer:v2, phive
           extensions: mbstring, xml
           coverage: none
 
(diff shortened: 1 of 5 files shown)


davereid:
Thank you for this!

tvdijen:
Much welcome Dave!

PR #1859, opened 2023-09-06
[PATCH] Add IDPList parameter that overrides SP and IDP metadata

ralphvandenhoudt:


tvdijen:
Hi @ralphvandenhoudt !
Thanks for this pull request. I'll merge it, but please note that v1.19 ended active support half a year ago, so there won't be a new release unless a security bug pop up.

PR #1861, opened 2023-09-09
Duplicate use statement for Request class.

fadonascimento:
> Fatal Error: Cannot use 'Request' because the name is already in use.

I know that is already fixed in main branch, but the Drupal module "Simplesamlphp Auth" will be use the next release 2.1 in their code as dependency, so it would be nice if this bug could be fixed.

See: [Drupal Simplesamlphp Auth issue](https://www.drupal.org/project/simplesamlphp_auth/issues/3349278)


thijskh:
Thanks! But there are more. I will also fix the others.

Note that these endpoints are legacy redirects provided for remote entities that may still have them in their metadata. You are advised to update to the modern endpoints.

PR #1862, opened 2023-09-15
Changed the UNIQUE key to a PRIMARY key for the tableVersion and saml_Logo…

folsomke:
…utStore tables for better compatibility with Drupal.

----

Some of us who are testing simplesamlphp-2.1 for use with Drupal 10 have found that the lack of primary keys for two database tables created by SimpleSAMLphp is causing an error message to be generated by the Drupal site status report function. It is recommended that Drupal website databases have their transaction isolation level set to READ-COMMITTED. When Drupal detects tables without primary keys, it reports that READ-COMMITTED won't work correctly without all tables having primary keys. Because SimpleSAMLphp is using a different database connection, the transaction isolation level is not truly compromised, but the status error could be confusing for website maintainers.

I was able to eliminate this error by changing two UNIQUE keys to PRIMARY keys in the noted tables. These are

Table: <prefix>_saml_LogoutStore
Key fields: _authSource + _nameId + _sessionIndex

Table: <prefix>_tableVersion
Key field: _name

None of the key fields allow for NULL values, so this change should not affect operation, and my testing shows this is the case.

### DIFF
--- a/modules/saml/src/SP/LogoutStore.php
+++ b/modules/saml/src/SP/LogoutStore.php
@@ -32,7 +32,82 @@ class LogoutStore
     private static function createLogoutTable(Store\SQLStore $store): void
     {
         $tableVer = $store->getTableVersion('saml_LogoutStore');
-        if ($tableVer === 4) {
+        if ($tableVer === 5) {
+            return;
+        } elseif ($tableVer === 4) {
+            // The _authSource index is being changed from UNIQUE to PRIMARY KEY for table version 5.
+            switch ($store->driver) {
+                case 'pgsql':
+                    // Drop old index and add primary key
+                    $update = [
+                        'ALTER TABLE ' . $store->prefix . '_saml_LogoutStore DROP CONSTRAINT IF EXISTS ' .
+                          $store->prefix . '_saml_LogoutStore___authSource__nameId__sessionIndex_key',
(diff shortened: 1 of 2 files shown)


folsomke:
Thanks for triggering checks on my pull request.  Is there anything I can do to fix the failing check?  I'm not clear if it's an expected failure or some problem with how I submitted the request.

tvdijen:
It's safe to ignore.. I think it has something to do with the source of the PR being a fork.

folsomke:
> It's safe to ignore.. I think it has something to do with the source of the PR being a fork.

That's good to hear.  Thanks for the follow-up.

folsomke:
@tvdijen I will add the update to the PR.  Thanks for the reference to how this was handled for past table updates.

tvdijen:
Ideally we'd migrate this to Doctrine at some point, however we have so few updates that I didn't think it worth doing it... Should you by accident have experience with Doctrine, we would be happy to take your PR :)

folsomke:
@tvdijen No such luck...I don't have any experience with Doctrine.  :)

tvdijen:
Old style is fine!

folsomke:
@tvdijen I have added a commit that revs the table version numbers and updates existing tables.  I was only able to test the MySQL code...I don't have a testing set-up that will work for PostgreSQL, SQL Server, or SQLite.

PR #1866, opened 2023-09-27
Update authproc config syntax

pmeulen:
Use config format per https://simplesamlphp.org/docs/2.0/saml/nameid.html


tvdijen:
Thanks a lot @pmeulen !

PR #1870, opened 2023-10-04
Make error codes easier to extend

### LINKED ISSUE #1867: Custom error message for SimpleSAML\Error\Error() method
I'm trying to display a custom error message to my Authentication Module. Some operation of authenticacion in my module could be another message instead of WRONGUSERPASS

I would like to use SimpleSAML\Error\Error() method to display a custom error for authentication

An example is:
throw new Error\Error(["CUSTOM","%TITLE%"=>"Error title","%DESCRIPTION%"=>"Error description"]);

I tried to use locales without success

I could make a proof of concept that works.  It involves a change in the file Error\ErrorCodes.php

Added the following to defaultGetAllErrorCodeTitles method
'CUSTOM' => Translate::noop('%TITLE%'),

and to defaultGetAllErrorCodeDescriptions method

'CUSTOM' => Translate::noop('%DESCRIPTION%'),

Another Alternative that I have considered is to duplicate the Error Class and use in my Module, and do some code changes to adapt.

**Additional context**
None

cicnavi:
This PR
- makes ErrorCodes easier to extend and to be used in custom Error classes
- introduces unit tests for ErrorCodes and Error classes
- moves to error code constants usage

Closes #1867

### DIFF
--- a/docs/simplesamlphp-authproc.md
+++ b/docs/simplesamlphp-authproc.md
@@ -31,15 +31,15 @@ The configuration of *Auth Proc Filters* is a list of filters with priority as *
 ```php
 'authproc.idp' => [
     10 => [
-        'class' => 'core:AttributeMap', 
+        'class' => 'core:AttributeMap',
         'addurnprefix'
     ],
     20 => 'core:TargetedID',
     50 => 'core:AttributeLimit',
     90 => [
-        'class' => 'consent:Consent', 
-        'store' => 'consent:Cookie', 
-        'focus' => 'yes', 
+        'class' => 'consent:Consent',
+        'store' => 'consent:Cookie',
+        'focus' => 'yes',
         'checked' => true
     ],
 ],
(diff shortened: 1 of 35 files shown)


tvdijen:
Constants sound great! Go ahead!
While you're at it.. Could you perhaps see if you have a clever solution for https://github.com/simplesamlphp/simplesamlphp/issues/1867 ?

cicnavi:
Ok, great... 

For custom codes, I think the way would be to create a custom OwnError class which extends SimpleSAML\Error\Error, and there use custom OwnErrorCodes class (which extends SimpleSAML\Error\ErrorCodes) to resolve title and description. In custom OwnErrorCodes we would have to do some overrides since there are some props and usage of 'self'... Than in userland code we then throw OwnError instead of SimpleSAML\Error\Error. I think I'll also suggest some modifications to SimpleSAML\Error\Error and SimpleSAML\Error\ErrorCodes to make them more easy to extend... I'll try this out if it sounds OK.

tvdijen:
Yes, please go ahead! It's been on my to-do list for too long already!

cicnavi:
Whilst at this state it is easier to extend ErrorCodes to be used in custom Errors (as show in unit tests), personally I don't like the usage of static classes / methods, meaning it is not possible to mock them (without introducing another tool like mockery / that is - adding another level of testing complexity with aliasing static classes / methods...). Would it be OK if I suggest using ErrorCodes as an instance dependency? I'll do it in a non-breakable way (will mark static methods as deprecated for now). It won't make any difference regarding unit test usefulness in this case, but just to show a strategy with easily mockable dependencies... If you don't like it, I'll just revert it.

tvdijen:
You mean that you would have to instantiate ErrorCodes before using it? Just like our utilities?
Sure, that's what we're trying to do anywhere we can!

tvdijen (approved):

tvdijen:
Awesome, thanks @cicnavi !

tvdijen:
Since this is backwards-compatible, I'd like to add it to the 2.1 milestone.
I don't think anyone has tried the RC1 yet, since we have zero feedback...

PR #833, opened 2018-04-17
SAML: Support for authnContextClassRef in proxy mode

tauceti2:
If SSP works as a proxy, this code retain requested authnContextClassRef from
SP behind the proxy and pass it to the upper IdP. It also pass the
authnContextClassRef sent by the upper IdP back to the SP behind the proxy. By this change SPs behind the proxy can request different authn methods, especially useful for support multi-factor authn.

Please review the code, if it is acceptable also for mode when SSP doesn't work as a proxy.

PR #1874, opened 2023-10-16
Handle AuthnContextClassRef changes when proxying

ghalse:
We introduced the ability to proxy RequestedAuthnContext in #833. However, if we're doing that, we must also consider what happens during authentication when the RequestedAuthnContext doesn't match the AuthnContext we have in a cached session.

If we don't compare what we have against what's requested, we may miss a request to step up authentication (e.g. when MFA is required by a specific service). This results in a poor user experience where they've got an IdP behind the proxy that is capable of meeting the request, but have to log out/terminate their SSO session to trigger it.

This patch tries to fix that by comparing whats in our session with what was requested, and triggering authentication again if they don't match. It only uses an exact comparison since we've no way of comparing relative weights of different AuthnContextClassRefs without also maintaining a lookup table. However, the exact match should be good enough for the majority of use cases.

This patch was written for 2.0.x but should apply cleanly to 2.1.x. It'll need refactoring for master.

### DIFF
--- a/modules/saml/src/Auth/Source/SP.php
+++ b/modules/saml/src/Auth/Source/SP.php
@@ -862,6 +862,30 @@ class SP extends \SimpleSAML\Auth\Source
             $state['saml:sp:AuthId'] = $this->authId;
             self::askForIdPChange($state);
         }
+
+        /*
+         * When proxying AuthnContextClassRef, we also need to check if the
+         * original IdP used a compatible AuthnContext. If not, we may need
+         * need to do step-up authentication (e.g. for requesting MFA). At
+         * the moment this only handles exact comparisons, since handling
+         * better/minimum/maximum would require a prioritised list.
+         */
+        if (
+            $this->passAuthnContextClassRef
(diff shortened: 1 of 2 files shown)


modules/saml/src/Auth/Source/SP.php
```diff
@@ -855,6 +855,30 @@ public function reauthenticate(array &$state): void
...
+         * When proxying AuthnContextClassRef, we also need to check if the
+         * original IdP used a compatible AuthnContext. If not, we may need
+         * need to do step-up authentication (e.g. for requesting MFA). At
+         * the moment this only handles exact comparisons, since handling
+         * better/minimum/maximum would require a prioritised list.
+         */
+        if (
+            $this->passAuthnContextClassRef
+            && isset($state['saml:RequestedAuthnContext'])
+            && $state['saml:RequestedAuthnContext']['Comparison'] == Constants::COMPARISON_EXACT
```
tvdijen, line 869: Please make these comparisons strict e.g. `===` and `!==`

PR #1887, opened 2023-10-30
Update docs to recommend use a tag when installing through git checkout

pradtke:
I hope this cuts down on support questions for people checking out master branch for their deploys.
Probably worth apply to SSP 2.1 and 2.2 branches as well, for when those make the next stable documentation.


tvdijen:
Thanks @pradtke 

P.S. SSP 2.1 has already been released and is the current stable

PR #1893, opened 2023-11-12
Clarify enabled by default modules

RoSk0:
Only 'admin' is enabled additionally, 'core' and 'saml' are enabled by default in \SimpleSAML\Module::isModuleEnabled().


RoSk0:
I believe current configuration is misleading - it looks like `core` and `saml` modules must be enabled, but they are actually enabled by default even when 'module.enable' is empty.

PR #1895, opened 2023-11-14
Fix minor code error in documentation for setting up google apps

mwnciau:
Two of the code examples on the google apps documentation page contain invalid PHP code. This is a documentation only change to fix the examples.


tvdijen:
Nice catch @mwnciau ! Thank you!

Issue #1917, opened 2023-12-14
Undefined variable $this_prefix


esetnik:
# **Help us to keep this issue-tracker clean!  For questions or support, please refer to our [mailing lists](https://simplesamlphp.org/lists)**

**Specifics of your environment**
1. Are you acting as SP/IdP/proxy?
SP
3. SimpleSAMLphp: What version are you using?
2.1.1
5. PHP: What version are you using?
8.3.0
7. Platform: unix or Windows?
linux docker
9. Webserver: Apache/Nginx/ISS?
nginx

**Describe the bug**
When using the `sqlsrv` driver in the function `initTableVersionTable()` there is an undefined variable `$this_prefix`.

```php
                    $update = [
                        'ALTER TABLE ' . $this->prefix . '_tableVersion DROP INDEX IF EXISTS SELECT CONSTRAINT_NAME ' .
                          'FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS WHERE TABLE_NAME=' . $this_prefix . '_tableVersion',
                        'ALTER TABLE ' . $this->prefix . '_tableVersion ADD CONSTRAINT _name PRIMARY KEY CLUSTERED (_name)'
                    ];
                    break;
```
https://github.com/simplesamlphp/simplesamlphp/blob/master/src/SimpleSAML/Store/SQLStore.php#L137

**Expected behavior**
No undefined variable should be present. It appears that `prefix` is a property on `$this`. I fixed it in https://github.com/simplesamlphp/simplesamlphp/pull/1918/commits/f18f180c3b50a8a84b0a557607c321b3e22b4f1d

**Screenshots or logs**
<img width="921" alt="image" src="https://github.com/simplesamlphp/simplesamlphp/assets/664434/12e8247d-8b19-47b9-98b7-0308de145710">

PR #1918, opened 2023-12-14
[sqlsrv] fix prefix

esetnik:
Resolves https://github.com/simplesamlphp/simplesamlphp/issues/1917


tvdijen:
Good catch @esetnik ! Thank you!

PR #1925, opened 2024-01-11
Automatically set the secure flag on session cookie

### LINKED ISSUE #1924: Test local http domain in Chrome browser result in "State Information Lost" errors
I use 2.1.1 for a local test. So my test domain is http, since I can't have a https domain locally. 
After installing and configuring, the welcome page is showing as expected. But when I access admin/ path, I came across the "State Information Lost" error in a Chrome browser. It is working fine in other browser, such as Firefox.

I inspect the request header and figure out that the 'SimpleSAMLSessionID' cookie wasn't sent with the request if I use Chrome browser, but Firefox did.

As the [document](https://simplesamlphp.org/docs/stable/simplesamlphp-nostate.html) advised, cookies will not be sent when a POST request is performed between websites, which is typical for the SAML WebSSO flow. The lack of cookies will cause SimpleSAMLphp's session to be lost when receiving an assertion via the HTTP-POST binding.

To resolve this, you can set the session.cookie.samesite attribute in config.php to None . You also need to enable the session.cookie.secure setting.

But the 'session.cookie.secure' can't be enabled for a http domain. So the suggest solution is not working for Chrome browser.

Is there a workaround for that?

Thanks.

tvdijen:
Closes #1924

### DIFF
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -621,8 +621,10 @@ $config = [
      * Set this to TRUE if the user only accesses your service
      * through https. If the user can access the service through
      * both http and https, this must be set to FALSE.
+     *
+     * If unset, SimpleSAMLphp will try to automatically determine the right value
      */
-    'session.cookie.secure' => true,
+    //'session.cookie.secure' => true,
 
     /*
      * Set the SameSite attribute in the cookie.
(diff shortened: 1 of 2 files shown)


thijskh (approved):

tvdijen:
Closed in 7bcf0d0d4f847dc0cf2d01471bbff951382e0b86

PR #1927, opened 2024-01-17
Never cache metadata when protect.metadata is set to true

### LINKED ISSUE #1926: Cached SP Metadata XML file accessible for all
**Environment**
1. Are you acting as SP/IdP/proxy?
SP
3. SimpleSAMLphp: What version are you using?
2.1.1
5. PHP: What version are you using?
8.1.18
7. Platform: unix or Windows?
Unix
9. Webserver: Apache/Nginx/ISS?
Nginx
10. Web cache server
Varnish reverse proxy
11. Configuration
'production' => false,
'admin.protectmetadata' => true,

**Describe the bug**
We are using Varnish reverse proxy as the web cache service and set the 'admin.protectmetadata' as true in the config.php file to prevent non-admin user from accessing our SP metadata.
But once an admin user (after logging) access the SP metadata via the url (../module.php/saml/sp/metadata/{sourceId}), that metadata XML file will be cached by the web cache service, in our case is the Varnish, as a public resource and the caching time (max-age) is set to 86400 (1 day) as default. Until that time is over, anyone can access/download the SP metadata XML file from anywhere (any IP) as a public cached file.

**Expected behavior**
Only admin user can access/download the SP metadata XML file when 'admin.protectmetadata'  is set to true.

tvdijen:
Closes #1926

### DIFF
--- a/modules/saml/src/Controller/Metadata.php
+++ b/modules/saml/src/Controller/Metadata.php
@@ -73,7 +73,8 @@ class Metadata
         }
 
         // check if valid local session exists
-        if ($this->config->getOptionalBoolean('admin.protectmetadata', false)) {
+        $protectedMetadata = $this->config->getOptionalBoolean('admin.protectmetadata', false);
+        if ($protectedMetadata === true) {
             $response = $this->authUtils->requireAdmin();
             if ($response instanceof Response) {
                 return $response;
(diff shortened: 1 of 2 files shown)


thijskh (approved):

tvdijen:
Closed in 79ce63763bff5bac0496d63ac042a025a20efcb0

PR #1932, opened 2024-01-29
Add cachedir-setting and use it for Symfony cache

### LINKED ISSUE #1931: 2.1: Cache being written to temp directory
**Describe the bug**

```
SimpleSAML\Error\Error: UNHANDLEDEXCEPTION
Backtrace:
1 public/_include.php:28 (SimpleSAML_exception_handler)
0 [builtin] (N/A)
Caused by: RuntimeException: Unable to write in the "cache" directory (/tmp/simplesaml/cache/saml).
Backtrace:
5 vendor/symfony/http-kernel/Kernel.php:641 (Symfony\Component\HttpKernel\Kernel::buildContainer)
4 vendor/symfony/http-kernel/Kernel.php:545 (Symfony\Component\HttpKernel\Kernel::initializeContainer)
3 vendor/symfony/http-kernel/Kernel.php:787 (Symfony\Component\HttpKernel\Kernel::preBoot)
2 vendor/symfony/http-kernel/Kernel.php:190 (Symfony\Component\HttpKernel\Kernel::handle)
1 src/SimpleSAML/Module.php:234 (SimpleSAML\Module::process)
0 public/module.php:17 (N/A)
```

The /tmp directory is being emptied on most OS'es, so we should be writing Symfony cache elsewhere (like /var/cache)

tvdijen:
Closes #1931

### DIFF
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -58,6 +58,7 @@ $config = [
     /*
      * The following settings are *filesystem paths* which define where
      * SimpleSAMLphp can find or write the following things:
+     * - 'cachedir': Where Symfony can write it's cache. 
      * - 'loggingdir': Where to write logs. MUST be set to NULL when using a logging
      *                 handler other than `file`.
      * - 'datadir': Storage of general data.
@@ -66,9 +67,10 @@ $config = [
      * When specified as a relative path, this is relative to the SimpleSAMLphp
      * root directory.
      */
+    'cachedir' => '/var/cache/simplesamlphp',
     //'loggingdir' => '/var/log/',
     //'datadir' => '/var/data/',
-    'tempdir' => '/tmp/simplesaml',
+    'tempdir' => '/tmp/simplesamlphp',
 
     /*
      * Certificate and key material can be loaded from different possible
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -58,7 +58,7 @@ $config = [
     /*
      * The following settings are *filesystem paths* which define where
      * SimpleSAMLphp can find or write the following things:
-     * - 'cachedir': Where Symfony can write it's cache. 
+     * - 'cachedir': Where SimpleSAMLphp can write its cache. 
      * - 'loggingdir': Where to write logs. MUST be set to NULL when using a logging
      *                 handler other than `file`.
      * - 'datadir': Storage of general data.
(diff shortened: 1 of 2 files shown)


config/config.php.dist
```diff
@@ -58,6 +58,7 @@ $config = [
     /*
      * The following settings are *filesystem paths* which define where
      * SimpleSAMLphp can find or write the following things:
+     * - 'cachedir': Where Symfony can write it's cache.
```
thijskh, line 61:
```suggestion
     * - 'cachedir': Where SimpleSAMLphp can write its cache. 
```

config/config.php.dist
```diff
@@ -66,9 +67,10 @@ $config = [
      * When specified as a relative path, this is relative to the SimpleSAMLphp
      * root directory.
      */
+    'cachedir' => '/var/cache/simplesamlphp',
     //'loggingdir' => '/var/log/',
     //'datadir' => '/var/data/',
-    'tempdir' => '/tmp/simplesaml',
+    'tempdir' => '/tmp/simplesamlphp',
```
thijskh, line 73: Begs the question whether this is still used - will file a separate issue to discuss that.

thijskh (commented):
Probably we need to add install docs that you need to create this dir and make it writable for the webserver...

thijskh:
What is the behaviour if the configured dir is not present and/or not writable for the web server? Does it provide a clear error message?

tvdijen:
The error is as described in the attached issue. It's pretty clear.

I have a setup with php-fpm (running on a dedicated service account per pool) and when the OS cleans up the /tmp-directory I get this error. The strange thing is that when I restart php-fpm, the directory is automatically re-created. I can't explain, but I _do_ know that static cache shouldn't be written to /tmp

config/config.php.dist
thijskh: My comment is more that for the user/deployer the term "Symfony" does not necessarily mean anything. It's just cache for SimpleSAML.

cicnavi:
> The error is as described in the attached issue. It's pretty clear.
> 
> I have a setup with php-fpm (running on a dedicated service account per pool) and when the OS cleans up the /tmp-directory I get this error. The strange thing is that when I restart php-fpm, the directory is automatically re-created. I can't explain, but I _do_ know that static cache shouldn't be written to /tmp

Hi, maybe this is related to having SimpleSAMLphp related cron jobs set up, but which run under different user(s). The first cron job which runs after the reboot, sets the tmp permission... Recently this was the actual case for us.

tvdijen:
Thanks Marco! I suspect is has something to do with opcache. I have no cron jobs set up on my IdPs

PR #1934, opened 2024-01-30
Fix atomically writing to a file

thijskh:
For the trick to work, the file needs to be on the same filesystem. The tempdir is likely not on the same filesystem so the rename will not be instantaneous. This probably worked in the past when the tempdir was expected to be a subdir of the SimpleSAMLphp installation.

Fix it by writing a file to the same directory as the target file, which will always be on the same FS. Also do not just use a `rand()` number but something more random as we do in other places in the code.

### DIFF
--- a/src/SimpleSAML/Utils/System.php
+++ b/src/SimpleSAML/Utils/System.php
@@ -187,7 +187,7 @@ class System
      */
     public function writeFile(string $filename, string $data, int $mode = 0600): void
     {
-        $tmpFile = $this->getTempDir() . DIRECTORY_SEPARATOR . rand();
+        $tmpFile = $filename . '.' . bin2hex(random_bytes(4));
 
         $res = @file_put_contents($tmpFile, $data);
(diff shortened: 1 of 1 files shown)


src/SimpleSAML/Utils/System.php
```diff
@@ -187,7 +187,7 @@ public function resolvePath(string $path, string $base = null): string
      */
     public function writeFile(string $filename, string $data, int $mode = 0600): void
     {
-        $tmpFile = $this->getTempDir() . DIRECTORY_SEPARATOR . rand();
+        $tmpFile = $filename . '.' . bin2hex(openssl_random_pseudo_bytes(4));;
```
tvdijen, line 190: It's not necessarily wrong, but I wouldn't use an `openssl_` function for non-cryptographic actions. Using `random_bytes(4)` seems more appropriate to use here.

tvdijen (approved):

PR #1961, opened 2024-02-13
A first attempt at storing generated po files in sorted order. 

monkeyiq:
This should hopefully minimize the number of changes that git sees when storing these files in a repo

Note that some of the header metadata is not cloned at the moment. If this ordering fixes the issue that was causing git chatter then the metadata clone can be added before merge.

Tested using
```
cd ./bin
./translations translations:update:translatable
vi ../locales/en/LC_MESSAGES/messages.po
```

### DIFF
--- a/src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php
+++ b/src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php
@@ -32,7 +32,7 @@ use function ksort;
 use function sprintf;
 
 
-function cloneIteratorToTranslations( $ret, $iterator )
+function cloneIteratorToTranslations($ret, $iterator)
 {
     while ($iterator->valid()) {
         $ret->addOrMerge(
--- a/src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php
+++ b/src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php
@@ -31,6 +31,22 @@ use function in_array;
 use function ksort;
 use function sprintf;
 
+
+function cloneIteratorToTranslations( $ret, $iterator )
+{
+    while($iterator->valid()) {
+        $ret->addOrMerge(
+            $iterator->current(),
+            Merge::TRANSLATIONS_THEIRS | Merge::COMMENTS_OURS | Merge::HEADERS_OURS | Merge::REFERENCES_OURS,
+        );
+        
+        $iterator->next();
+    }
+
+    return $ret;
+}
+
+
 class UpdateTranslatableStringsCommand extends Command
 {
     /**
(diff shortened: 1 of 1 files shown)


src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php
```diff
@@ -31,6 +31,22 @@
 use function ksort;
 use function sprintf;
 
+
+function cloneIteratorToTranslations($ret, $iterator)
```
tvdijen, line 35: This should probably be private method inside the class.. The parameters should be type-hinted.

monkeyiq:
I guess the big question is whether having the keys in the generated po files in sorted order is the main issue that was causing the git chatter or not. Hopefully I understood the issue enough.

src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php
monkeyiq: Agreed. I did mention that this was a quick and dirty PoC :)

tvdijen:
> I guess the big question is whether having the keys in the generated po files in sorted order is the main issue that was causing the git chatter or not. Hopefully I understood the issue enough.

I think you did! This PR won't hurt anyone if it doesn't make it's promise, so feel free to fix the review comments and merge!

monkeyiq:
I should probably also look at cloning the metadata too before the merge

src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php
tvdijen: You did ;)  We're strict like this! :goberserk:

tvdijen:
What metadata are we talking about? I think we keep comments, headers and references.. Is there more we would want/need? I feel like the translations-files should not become a source of documentation..

tvdijen:
> I think there are some in the attributes.po
> [..]
> Though it seems that file is not processed by `translations translations:update:translatable`.

Correct. The `attributes.po` file is not processed at all... It's a leftover that still needs work. I believe there is also an open issue to fix the `bin/translateAttributes.sh` script

src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php
monkeyiq: Updated, https://github.com/simplesamlphp/simplesamlphp/pull/1961/commits/9a09e5e89415fa8d56525e0b1c408f9582956c59

monkeyiq:
So this should be ok now, sans the awaiting workflow. I seem to not have write access at the moment so can not self merge.

monkeyiq:
translateAttributes.sh might be something I can take a look at after this to see if there is something I can do there too.

tvdijen:
I've enabled the workflow and sent you an invitation to the developers-team which should grant you access to all repositories.
The linter-workflow is known to fail, so you can forget about that one.
Since this file is not covered at all by unit tests, I'd suggest you fix the quality-workflow and just merge..

monkeyiq:
I'll double check this tomorrow and merge.

monkeyiq:
I reverified that locales/en/LC_MESSAGES/messages.po and the original messages.po are the same data when both are sorted.

Issue #1914, opened 2023-12-07
SimpleSAMLPHP 2.1.1 language selection fails


kakopedreros:
**Specifics of your environment**
1. Are you acting as SP/IdP/proxy?
2. SimpleSAMLphp: What version are you using?
3. PHP: What version are you using?
4. Platform: unix or Windows?
5. Webserver: Apache/Nginx/ISS?

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots or logs**
If applicable, add screenshots and/or log files to help explain your problem.

**Additional context**

# **Help us to keep this issue-tracker clean!  For questions or support, please refer to our [mailing lists]

-----------------------

**### Environment**
SimpleSAMLPHP: 2.1.1 as IdP but just working with the Admin module as well.
PHP: 8.2.13 Remi Repo
Platform: Linux Centos 7.9 (Using a dockers image in a MacBook Pro 2015) see attached dockerfile.
Webserver: Apache 2.4.6

[docker-compose.zip](https://github.com/simplesamlphp/simplesamlphp/files/13602991/docker-compose.zip)

![Captura de Pantalla 2023-12-07 a la(s) 14 04 55](https://github.com/simplesamlphp/simplesamlphp/assets/4590300/dad3d5fe-83f1-4515-945a-b4d978bf8731)

### **Bug description**
Trying to select a language, it's set the language cookie, but it not change the content of simplesamlphp, official modules or external. Does not work even changing the default language in the configuration.
It was review file permission, path of the files, proper configuration of PHP dependencies.
Also tested in RockyLinux8 with the same PHP Remi Repo.
SimpleSAMLPHP installed using composer 2.6.5.

### **Screenshots**
Spanish - Español
![Captura de Pantalla 2023-12-07 a la(s) 13 04 26](https://github.com/simplesamlphp/simplesamlphp/assets/4590300/93b18051-05ac-420f-8f87-01269b69fffa)
English (Default)
![Captura de Pantalla 2023-12-07 a la(s) 13 04 19](https://github.com/simplesamlphp/simplesamlphp/assets/4590300/7283442d-489f-411e-bd5a-042cf12c52ad)
(Brazilian Portuguese)
![Captura de Pantalla 2023-12-07 a la(s) 13 04 14](https://github.com/simplesamlphp/simplesamlphp/assets/4590300/4e2a68f4-af27-4dbf-9a1a-e64f47c563b7)

**### Expected Bed**
[...]

PR #1967, opened 2024-02-20
composer: install the deps so that the translation script can run

monkeyiq:
The `php bin/translations translations:update:binary` command relies on the vendor directory which is not installed at the time the script is called when SSP is installed via composer.  Perhaps we should (or need to) add this to the `post-update-cmd` clause as well. 

I have found this update to work when a composer tree is made (where the script fails) and one then runs the following. 
```
composer run-script post-install-cmd
```

Hopefully with this update the below command will also generate the .mo files during composer installation.
```
composer require simplesamlphp/simplesamlphp
```

This relates to https://github.com/simplesamlphp/simplesamlphp/issues/1914 and is a solution I proposed in https://github.com/simplesamlphp/simplesamlphp/issues/1914#issuecomment-1951475336

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -127,6 +127,7 @@
         "translations:update:binary": "php bin/translations translations:update:binary",
         "translations:update:translatable": "php bin/translations translations:update:translatable",
         "post-install-cmd": [
+            "@composer update",
             "php bin/translations translations:update:binary"
         ],
         "post-update-cmd": [


tvdijen:
I'm sorry, didn't we agree to revert to .po files on Slack?

monkeyiq:
The inability to run scripts in a secondary library is really unfortunate. I'll have a look at how we can revert to using .po files if there are no .mo files available.

tvdijen:
> The inability to run scripts in a secondary library is really unfortunate. I'll have a look at how we can revert to using .po files if there are no .mo files available.

I'm pretty sure this fallback-mechanism is already in place

tvdijen:
I was wrong.. The code is here:
https://github.com/simplesamlphp/simplesamlphp/blob/master/src/SimpleSAML/Locale/Localization.php#L240

Reverting the change from .mo > .po should be as easy as changing the file extension, replace `MoLoader` with `PoLoader` and drop the script that generates the binaries.

The changes we made between 2.0 > 2.1 are covered by this PR:
https://github.com/simplesamlphp/simplesamlphp/pull/1670

monkeyiq:
I left the first attempt at the mo file for now. At least until a full performance analysis is done to see the impact of not using the binary `.mo` compiled files. It may be a small impact, but it may also add up if used on a busy server and folks might like to run a script to make the mo files and recoup that performance.

There was some talk about having a configuration validation script which could also mention that (or offer to) running a script to make the `.mo` files might be useful for an installation.

Now that we have a fallback to po files things should just work in more places. I will have a further dig to see if there are other translation pain points in 2.1.x which I might be able to help with.

PR #1971, opened 2024-02-26
Add regex support for attribute names in AttributeLimit authproc filter

nathanjrobertson:
Up until now, AttributeLimit allows regexp matching on the values of an attribute, but the attribute name itself must be specified as a single value of string. This PR adds a new configuration parameter (`nameIsRegex`) to support attribute names in the configuration being regular expressions, rather than just static strings.

When `nameIsRegex => true`, the key is treated as a regexp. When it is false or not specified, it is treated as a string (ie. evaluated in the traditional way).

The implementation allows using both traditional plain strings and regexps side by side:

```
    'authproc' => [
        50 => [
            'class' => 'core:AttributeLimit',
            'cn', 'uid',
            '/^eduPerson' => [ 'nameIsRegex' => true ]
        ],
    ],
```

and is fully backward compatible. You can even use regex support for both attribute name and value simultaneously:

```
    'authproc' => [
        50 => [
            'class' => 'core:AttributeLimit',
            'cn', 'uid',
            '/^eduPerson/' => [
                'nameIsRegex'=>true,
                'regex'=>true,
                '/@example.org$/'
            ],
        ],
    ],
```

There is support for the metadata method of providing configuration as well. There is a new key available in the metadata `attributesRegex`, which works as the existing `attributes` key does, but for regular expression entries. Again, they can be used side by side. In the case where only one of the two is specified, it is assumed the other is an empty list (which is better than taking defaults independently - that would be confusing and inconsistent).

```
$metadata['https://saml2sp.example.org'] = [
    'AssertionConsumerService' => 'https://saml2sp.example.org/simplesaml/module.php/saml/sp/saml2-acs.php/default-sp',
    'SingleLogoutService' => 'https://saml2sp.example.org/simplesaml/module.php/saml/sp/saml2-logout.php/default-sp',
    ...
    'attributes' => ['cn', ... ],
    'attributesRegex' => [ '/^mail$/', ... ],
    ...
];
```

If `nameIsRegex` (authsources.php method) and `attributesRegex` (metadata method) aren't specified, the preexisting behaviour continues. No breaking changes.


nathanjrobertson:
Ok. I don't think I can make Scrutinizer complain less than it is without making the code less readable. I think it's now best to have somebody review and give feedback.

thijskh:
Thanks, I like the idea and implementation! Also good that you provide documentation and test coverage. I haven’t yet reviewed it line by line.

nathanjrobertson:
@thijskh Thanks. I think I've fixed the phpcs errors and markdown linting warnings. Could you please trigger a re-test for those?

tvdijen:
Thank you @nathanjrobertson ! I will backport this to the 2.2 release-branch

nathanjrobertson:
> @nathanjrobertson You think you can help us fixing the warnings in the unit tests? Something's still off there, even though the tests pass; https://github.com/simplesamlphp/simplesamlphp/actions/runs/9028875799/job/24810192242#step:11:19

Yep - I'll take a look at it today or tomorrow.

nathanjrobertson:
> @nathanjrobertson You think you can help us fixing the warnings in the unit tests? Something's still off there, even though the tests pass; https://github.com/simplesamlphp/simplesamlphp/actions/runs/9028875799/job/24810192242#step:11:19

@tvdijen Proposed fix is in https://github.com/simplesamlphp/simplesamlphp/pull/2095 - it's a harmless warning though, because it's all handled in the function it calls with the null value that results from the warning.

Looking through the rest of that file - the remaining warning is [this test](https://github.com/simplesamlphp/simplesamlphp/blob/34c082ae12c6a4bd311c7c4e7dd4f0a69cfcee55/tests/modules/core/src/Auth/Process/AttributeLimitTest.php#L501). I disagree with the comment there - surely an invalid regular expression in the configuration is an error condition, and it should error out, not silently (except for the warning) ignore the error and continue. That could leave people with a potentially insecure configuration, which they won't pick up on as easily as if it errors and bails out. Anyway, your call - just noticed it and that irked me.

thijskh:
@nathanjrobertson Indeed, probably the tests should be split out and the no-match case should be correctly skipped and the invalid case should be handled in a different test with an "expectException" for the PHP warning that's emitted in that case?

PR #1983, opened 2024-03-03
docs: update the upgrade notes for 2.0 with some information from slack about entityIDs

monkeyiq:
I also moved the entityID point to the top as it seems to be the most common question about upgrading from 1.x to 2.x.

Most of the other information added at the end of document was obtained from slack so might not be 100% the opinion of core developers. Some extra eyes are appreciated, if we can get this info extended and updated in one place that should help everybody.

It also might be useful to mention / link to the 2.0 upgrade notes in the 2.1 upgrade notes. Something like the following: If you are upgrading from a 1.x series release please see the 2.0 upgrade notes (linked) to ensure a smooth transition.

### DIFF
--- a/docs/simplesamlphp-upgrade-notes-2.0.md
+++ b/docs/simplesamlphp-upgrade-notes-2.0.md
@@ -153,10 +153,8 @@ are using for your SP and IdP should be available in
 module.php/core/frontpage_federation.php location on your
 SimpleSAMLphp server.
 
-An EntityID is a unique identifier for your SP or IdP. It has to be
-unique, like an IP-address has to be unique in your network. You might
-like to use your host name and some additional path information for
-the EntityID.
+For a service provider, if it was set as auto-generated in 1.19, it
+will likely have the form of (<https://yourhostname/simplesaml/module.php/saml/sp/metadata.php/default-sp>).
 
 The EntityID is set in two locations, as the property 'entityID' for
 an SP and as the index in the $metadata array for an IdP. Examples of
@@ -192,6 +190,5 @@ $metadata['https://example.com/the-service/'] = [
 ...
 ```
 
-If you are using Azure and if you are configuring a hosted sp in ssp,
-the entityID will be a relying party you added in azure. It will have
-an entityid in azure like <https://your.ssp.hostname/default-sp>.
+If you use SimpleSAMLphp as an SP, the IdP you are using will have
+your correct entityID configured.
--- a/docs/simplesamlphp-upgrade-notes-2.0.md
+++ b/docs/simplesamlphp-upgrade-notes-2.0.md
@@ -21,6 +21,12 @@ composer require simplesamlphp/simplesamlphp-module-ldap --update-no-dev
 
 ## Functional changes
 
+- EntityIDs are no longer auto-generated. Make sure to set something sensible in the array-keys in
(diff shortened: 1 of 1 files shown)


docs/simplesamlphp-upgrade-notes-2.0.md
```diff
@@ -144,3 +145,53 @@ $x = \SimpleSAML\Utils\Arrays::arrayize($someVar)
...
+
+```php
+...
+$metadata['https://example.com/the-service/'] = [
+...
+```
+
+If you are using Azure and if you are configuring a hosted sp in ssp,
+the entityID will be a relying party you added in azure. It will have
+an entityid in azure like <https://your.ssp.hostname/default-sp>.
```
thijskh, line 197: This last paragraph seems a bit out of place. I assume it's about using Azure (now called Entra) as an IdP? So this could genericallt be said as "If you use SimpleSAMLphp as an SP, the IdP you are using will have your correct entityID confifgured"? There seems nothing special about Azure (Entra) in this regard.

docs/simplesamlphp-upgrade-notes-2.0.md
```diff
@@ -144,3 +145,53 @@ $x = \SimpleSAML\Utils\Arrays::arrayize($someVar)
 
 [1]: https://github.com/simplesamlphp/simplesamlphp/wiki/Migrating-translations-(pre-migration)
 [2]: https://github.com/simplesamlphp/simplesamlphp/wiki/Twig:-Migrating-templates
+
+## Upgrading and EntityIDs
+
+If you still have your 1.x installation available, the entityID you
+are using for your SP and IdP should be available in
+module.php/core/frontpage_federation.php location on your
+SimpleSAMLphp server.
```
thijskh, line 154: I would add a new paragraph:

"For a service provider, if it was set as auto-generated in 1.19, it will likely have the form of `https://<your hostname>/simplesaml/module.php/saml/sp/metadata.php/default-sp`."

docs/simplesamlphp-upgrade-notes-2.0.md
```diff
@@ -144,3 +145,53 @@ $x = \SimpleSAML\Utils\Arrays::arrayize($someVar)
...
+
+An EntityID is a unique identifier for your SP or IdP. It has to be
+unique, like an IP-address has to be unique in your network. You might
+like to use your host name and some additional path information for
+the EntityID.
```
thijskh, line 159: We have some instructions in the docs on how to pick one. Maybe keep this section focussed on the changes between 1.19 and 2.x, as the document is about upgrading. People should not land here that will need to pick a new entityID from scratch - they are helped in the other parts of the docs (and if they are not we should fix that there).

docs/simplesamlphp-upgrade-notes-2.0.md
monkeyiq: ok, I will change the last paragraph to the suggested sentence.

docs/simplesamlphp-upgrade-notes-2.0.md
monkeyiq: ok removed this paragraph. It is a bad idea if this happens to be favoured by the search engines over the existing part of the docs discussing EntityIDs.

docs/simplesamlphp-upgrade-notes-2.0.md
monkeyiq: adding.

Issue #1368, opened 2020-08-21
metadata.sources xml should accept a directory


Lasering:
**Is your feature request related to a problem? Please describe.**
When an IDP has a lot of SPs registered it makes more sense to have each SP metadata in an individual file. Currently this type of layout is not directly supported.

**Describe the solution you'd like**
`metadata.sources` with type xml should support a config named `directory` which would search for all xml files inside that directory, and add all of those metadatas.

**Describe alternatives you've considered**
Writing a piece of code in the config.php which lists all XML files inside the pretended directory and for each one adds an entry to `metadata.sources` with `['type' => 'xml', 'file' => 'path-to-the-file']`

**Additional context**
An even better approach would be:
`['type' => 'directory', 'directory' => 'the-directory', 'set' => 'remote']`
All php and xml metadata files inside `the-directory` would be added.

A XML metadata can be easily distinguished from IDP vs SP because of `md:IDPSSODescriptor` vs `md:SPSSODescriptor`.
A simplesamlphp metadata can be easily distinguished from IDP vs SP because of the `metadata-set` field. In this case it is even possible to distinguish between remote vs hosted.

monkeyiq:
I have merged a PR https://github.com/simplesamlphp/simplesamlphp/pull/2018 with this functionality into the simplesamlphp-2.3 branch and master.

The patch should also back port to a 2.2 release reasonably well if you would like to take it for a test run.

PR #1984, opened 2024-03-04
A new MetaDataStorageHandlerDirectory class to load metadata from subdirectories

monkeyiq:
This is another attempt at loading metadata from a subdirectory containing either php or xml files (or both). This specifically allows SP files to be broken out into one file per SP if there are a number of SPs and you wish to maintain a list using one file per SP. 

It is quite possible that some code cleanups to this are desired before merging. It seems from the below links that the feature is desirable for a few users and having an implementation based on simplesamlphp-2.1 may allow a merge and closing of a long standing issue and long standing PR.

To be generic this also adds a `file` property to the config of the MetaDataStorageHandlerFlatFile class so it can be used to load the file instead of directly including it in the new MetaDataStorageHandlerDirectory class. This has the added advantage of a single code path to load those php files so other checks etc in the future can be all in one place.

This relates to https://github.com/simplesamlphp/simplesamlphp/pull/1292 and https://github.com/simplesamlphp/simplesamlphp/issues/1368.

### DIFF
--- a/docs/simplesamlphp-maintenance.md
+++ b/docs/simplesamlphp-maintenance.md
@@ -198,7 +198,7 @@ Configure your master group by setting `store.redis.mastergroup` (`mymaster` by 
 
 ## Metadata storage
 
-Several metadata storage backends are available by default, including `flatfile`, `serialize`, `mdq` and
+Several metadata storage backends are available by default, including `flatfile`, `directory`, `serialize`, `mdq` and
 [`pdo`](https://simplesamlphp.org/docs/stable/simplesamlphp-metadata-pdostoragehandler). Here you have an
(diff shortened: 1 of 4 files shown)


monkeyiq:
I should redirect this PR to 2.2 now.

tvdijen:
It's a new feature, not a bugfix, so this should target 2.3

monkeyiq:
I am guessing normally this would be against master instead. Though it seems master is a little more of a longer term release target at this stage :/ I might raise with the ssp board if they would like me to work a bit on master.

monkeyiq:
This is superseded by https://github.com/simplesamlphp/simplesamlphp/pull/2018

PR #1985, opened 2024-03-04
translations: do not make mo files anymore

monkeyiq:
The code is already falling back to the po files when the mo files are not found. (assuming the update from simplesaml-2.1 is ported forward: https://github.com/simplesamlphp/simplesamlphp/blob/09056cbb49b1883f33eb30c6d8e3f5a274039611/src/SimpleSAML/Locale/Localization.php#L244 Since the translation files are loaded en masse the performance difference between the binary and text format might not be as substantial.

The motivation for moving back to using the po files is correct handling of things when SimpleSAMLphp is installed as a dependency using composer. If a project has a composer.json file that lists a dependency on SimpleSAMLphp then composer does not run the install hooks for SimpleSAMLphp when installing it. In that case (which seems common) there are no `.mo` binary translation files generated and things do not work as expected. This may also be the case for modules that are installed if the binary translation script is not executed.

I am not 100% sure we want to remove the command itself. Easy enough to grab them from there or the 2.1 branch if needed though. The other options is to just remove the calls in composer.json hooks so that the default tarball files are made without the `.mo` files.


tvdijen:
Yes, we want to remove the commands and their references from `composer.json`

> I am not 100% sure we want to remove the command itself. Easy enough to grab them from there or the 2.1 branch if needed though. The other options is to just remove the calls in composer.json hooks so that the default tarball files are made without the `.mo` files.

Yes, we want to remove the commands and their references from `composer.json`

monkeyiq:
I am not sure what is causing the CI / Syntax and unit tests to fail here. I'll keep digging.

monkeyiq:
After some tinkering I see the same patch passing CI in https://github.com/simplesamlphp/simplesamlphp/pull/1994 apart from the codecov.

tvdijen:
> I am not sure what is causing the CI / Syntax and unit tests to fail here. I'll keep digging.

I'm pretty sure that we've fixed a bug related to these tests failing recently.. Perhaps your fork is old?

tvdijen:
This one was causing it, I think:
https://github.com/simplesamlphp/simplesamlphp/commit/3cc46a0ee9cbd56fc6347ee2f9d9d7e387059ae7

PR #1997, opened 2024-03-11
Azure2Name Attribute Mapping for SAML Claims

Whats-A-MattR:
Adding mappings for Azure/Entra Enterprise Application SAML Sign-on, including a commonly included option of 'groups'. 

Without these transformations, Azure / Entra can be very painful to use as an IdP.


thijskh:
Thanks! Would it make sense to map the attributes to the same target "names" as the other `2name` maps use. E.g. `email` and `sn`?

I also wonder if the name `azure` is not better replaced by `entra` now.

Whats-A-MattR:
Definitely agree on the topic of Entra vs Azure as that is the correct nomenclature. 

As for mapping to match other 2name maps I can definitely see value in doing that, though I think there is also something to be said for the simplicity of a close-to-vendor mapping as it's an easier mental model. Transformations would then take place on callback etc. 

Is it worth having two attributemaps? One for a 1:1 mapping where the schema gets cleaned up and made usable (like the example I provided) and one that fits the 2name standard?

Totally up to maintainers as to what the preferred method is, just my two cents :)

Whats-A-MattR:
I've updated the map file name to Entra, and the mappings to match other 2name mappings. 
Additionally, I've added a name2 mapping for casting back to SAML. 

Let me know your thoughts.

Whats-A-MattR:
@thijskh just bumping :)

monkeyiq:
I don't have any Entra to verify things against. Looks good having mappings from those attributes to the same target "names" so things can coexist.

monkeyiq:
I have cherry picked this back to the 2.3 branch as well. a8d0018bfb71d55a94f3103a48d339efdbad3094 c2e7fbfa91992832bf9e1fd45dafbfe3fffc2be1 87a6078afa22f3d0640b510f58fd602d6021e8b2

PR #2004, opened 2024-03-13
Fix URL in error page redirection in ExpectedAuthContextClassRef auth…

9oOzv:
…proc

ExpectedAuthContextClassRef authproc seems to redirect to an invalid error page. I think this is a bug.

Could not find an issue of PR addressing this.

I hope this fix is correct.


thijskh (approved):

thijskh:
This is indeed a correct fix, it will be part of the next release. Thanks!

Issue #2009, opened 2024-03-15
Translations in v2.2 seem broken


cicnavi:
In admin area of SimpleSAMLphp v2.2.0 instance, if I change language using the language selector - nothing happens, that is, only 'msgid' is shown in UI for each string that should be translated.

It think it has something to do with domains. When parsing po files, translations are groped by domains (in contrast to mo files, when that was not the case). If I hardcode domain to null in vendor/gettext/translator/src/Generator/ArrayGenerator.php::generateArray() (excerpt below), then it works. 

```php
        return [
            //'domain' => $translations->getDomain(),
            'domain' => null,
            'plural-forms' => $translations->getHeaders()->get(Headers::HEADER_PLURAL),
            'messages' => $messages,
        ];
```

tvdijen:
Confirmed!  I'm not sure what you mean with the domains though.. We group translations by domain (= module name) and I can still see the domains being loaded in the debug logs.

cicnavi:
When it works, If I dump property 'dictionary' from vendor/gettext/translator/src/Translator.php, it looks like this:

```
array (size=1)
  '' => 
    array (size=1)
      '' => 
        array (size=233)
          'A service has requested you to authenticate yourself. Please enter your username and password in the form below.' => string 'Un servicio solicita que se autentique. Esto significa que debe indicar su nombre de usuario y su clave de acceso en el siguiente formulario.' (length=141)
          'ADFS IdP Metadata' => string 'Metadatos IdP ADFS' (length=18)
...
```

When it doesn't work, it looks like this (note the 'messages' key / domain):

```
array (size=1)
  'messages' => 
    array (size=1)
      '' => 
        array (size=233)
          'A service has requested you to authenticate yourself. Please enter your username and password in the form below.' => string 'Un servicio solicita que se autentique. Esto significa que debe indicar su nombre de usuario y su clave de acceso en el siguiente formulario.' (length=141)
          'ADFS IdP Metadata' => string 'Metadatos IdP ADFS' (length=18)
...
```

tvdijen:
You mean 'when it works' if you hardcode the domain to `null`?
`messages` is the default domain.. In case of the admin module, I would expect the key to be `admin`.

tvdijen:
@monkeyiq I don't understand what you did to revert the translations-patch.. I still see .mo file related code and I still see scripts in the composer.json file..

monkeyiq:
[...]

PR #2010, opened 2024-03-16
i18n: when loading from po file the domain from the file has to be wiped

monkeyiq:
Loading from po files is slightly different than the existing path for mo files. The domain from the po file has to be removed in order for the loaded translation to be found by SSP.

This relates to https://github.com/simplesamlphp/simplesamlphp/issues/2009.

### DIFF
--- a/src/SimpleSAML/Locale/Localization.php
+++ b/src/SimpleSAML/Locale/Localization.php
@@ -244,6 +244,7 @@ class Localization
             $file = new File($langPath . $domain . '.po', false);
             if ($file->getRealPath() !== false && $file->isReadable()) {
                 $translations = (new PoLoader())->loadFile($file->getRealPath());
+                $translations->setDomain('');
                 $arrayGenerator = new ArrayGenerator();
                 $this->translator->addTranslations(
                     $arrayGenerator->generateArray($translations)


tvdijen:
I think this should be fixed the other way around.. This change will load every single translation into the default domain. Instead, the lookup has to be fixed so that the translations are searched for in the proper domain.

monkeyiq:
Looking at a slightly older release, simplesamlphp-2.1.1-full, the admin.po file has an `X-Domain: admin` but it seems that running `msgunfmt admin.mo` does not show a domain.

If I copy admin.po to test.po and msgfmt that test.po to get test.mo and msgunfmt that test.mo then I see a domain. So the gettext commands are happy to preserve the domain.

```
$ cp admin.po test.po
$msgfmt test.po -o test.mo
$ msgunfmt test.mo > test.unf
$ grep -i domain test.unf
msgstr "X-Domain: admin\n"
```

So it seems maybe that the processing of the admin.po file into admin.mo in ssp 2.1 might have been removing the x-domain. Other code that then used the translations were relying on the domain not being set in the binary.

If we again ignore the domain in this context like 2.1 releases did then a 2.2.1 release would have working translations again while we work out the code to place these in domains and have the reading code seek them in the correct location.

tvdijen (approved):

monkeyiq:
I cherry picked these to 2.3 and master as well. Should help make things easier to pick when/if the domain handling code is updated.

tvdijen:
I'll tag the new release

PR #2019, opened 2024-03-24
Fixes to the custom username/password auth docs

abirkill:
When getting started with SimpleSAMLphp and creating my first custom authentication module, following [the docs](https://simplesamlphp.org/docs/devel/simplesamlphp-customauth.html), I hit a couple of minor issues where the docs seem to be slightly outdated when running against version 2.2.1:

**Outdated method signature for login method:**
The existing docs suggest creating a method with signature `protected function login($username, $password)`.  Doing so will result in the following error when attempting to test or use the module:

```
SimpleSAML\Error\Error: UNHANDLEDEXCEPTION
Backtrace:
3 src/SimpleSAML/Error/ExceptionHandler.php:35 (SimpleSAML\Error\ExceptionHandler::customExceptionHandler)
2 vendor/symfony/error-handler/ErrorHandler.php:535 (Symfony\Component\ErrorHandler\ErrorHandler::handleException)
1 vendor/symfony/error-handler/ErrorHandler.php:627 (Symfony\Component\ErrorHandler\ErrorHandler::handleFatalError)
0 [builtin] (N/A)
Caused by: Symfony\Component\ErrorHandler\Error\FatalError: Compile Error: Declaration of SimpleSAML\Module\mymodule\Auth\Source\MyAuth::login($username, $password) must be compatible with SimpleSAML\Module\core\Auth\UserPassBase::login(string $username, string $password): array
Backtrace:
0 modules/mymodule/src/Auth/Source/MyAuth.php:7 (N/A)
```

I have updated the docs to use the current method signature, `protected function login(string $username, string $password): array`, and confirmed this resolves the issue.

**Required steps to enable the module are missing:**
The existing docs suggest that there is no explicit action needed to enable the module, stating:

> Since this is a custom module, it should always be enabled in the configuration.

However, this no longer appears to be the case, possibly due to changes in version 2.0?  Without explicitly enabling the new module, the following error occurs when attempting to test or use the module:

```
SimpleSAML\Error\Error: UNHANDLEDEXCEPTION
Backtrace:
2 src/SimpleSAML/Error/ExceptionHandler.php:32 (SimpleSAML\Error\ExceptionHandler::customExceptionHandler)
1 vendor/symfony/error-handler/ErrorHandler.php:535 (Symfony\Component\ErrorHandler\ErrorHandler::handleException)
0 [builtin] (N/A)
Caused by: Exception: The module 'mymodule' is not enabled.
Backtrace:
8 src/SimpleSAML/Module.php:449 (SimpleSAML\Module::resolveClass)
7 src/SimpleSAML/Auth/Source.php:313 (SimpleSAML\Auth\Source::parseAuthSource)
6 src/SimpleSAML/Auth/Source.php:356 (SimpleSAML\Auth\Source::getById)
5 src/SimpleSAML/Auth/Simple.php:62 (SimpleSAML\Auth\Simple::getAuthSource)
4 src/SimpleSAML/Auth/Simple.php:151 (SimpleSAML\Auth\Simple::login)
3 [builtin] (call_user_func_array)
2 src/SimpleSAML/HTTP/RunnableResponse.php:68 (SimpleSAML\HTTP\RunnableResponse::sendContent)
1 vendor/symfony/http-foundation/Response.php:423 (Symfony\Component\HttpFoundation\Response::send)
0 public/module.php:24 (N/A)
```

I have updated the docs to include the requirement to enable the new module after creating it.

### DIFF
--- a/docs/simplesamlphp-customauth.md
+++ b/docs/simplesamlphp-customauth.md
@@ -21,7 +21,6 @@ cd modules
 mkdir mymodule
 ```
 
-Since this is a custom module, it should always be enabled in the configuration.
 Now that we have our own module, we can move on to creating an authentication source.
 
 Creating a basic authentication source
@@ -41,7 +40,7 @@ namespace SimpleSAML\Module\mymodule\Auth\Source;
 
 class MyAuth extends \SimpleSAML\Module\core\Auth\UserPassBase
 {
-    protected function login($username, $password)
+    protected function login(string $username, string $password): array
     {
         if ($username !== 'theusername' || $password !== 'thepassword') {
             throw new \SimpleSAML\Error\Error(\SimpleSAML\Error\ErrorCodes::WRONGUSERPASS);
@@ -76,7 +75,7 @@ Configuring our authentication source
 -------------------------------------
 
 Before we can test our authentication source, we must add an entry for it in `config/authsources.php`.
-`config/authsources.php` contains an list of enabled authentication sources.
+`config/authsources.php` contains a list of enabled authentication sources.
 
 The entry looks like this:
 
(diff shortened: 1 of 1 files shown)


tvdijen:
> Since this is a custom module, it should always be enabled in the configuration.

I think you are misinterpreting this line. It means that because it's a custom module, it _has_ to be manually enabled.. We simply don't know the name of your custom module, so we cannot enable it by default.

Issue #2013, opened 2024-03-19
Perhaps respect the gettext domain for modules


monkeyiq:
**Specifics of your environment**
1. Are you acting as SP/IdP/proxy: n/a
2. SimpleSAMLphp: 2.2.1
3. PHP: 8.2.16
4. Platform: Fedora 38
5. Webserver: Apache httpd-2.4.58

**Describe the bug**
The gettext po files contain an `x-domain` setting, for example, the admin module has `X-Domain: admin` in modules/admin/locales/fr/LC_MESSAGES/admin.po. The 2.1 series used to compile these po files to mo binary files and would drop the domain in that process. This would put all translation strings in the top level domain.

It may be useful to retain and respect the domain for modules. This would give each module it's own domain and translation strings could not clash and existing translations would not be overriden by loading new .po files from different domains.

The update would need to retain the domain and update the code that is looking for a translation to look in the correct domain when inside code for a module.

**To Reproduce**
Take a 2.2.1 release and update to revert this PR https://github.com/simplesamlphp/simplesamlphp/pull/2010 so that the domain is retained. Translations will no longer be found.

**Expected behavior**
Perhaps domains should be retained and a module can indicate which domain it is in and translations applicable to that module will be found in that domain.

The commands in https://github.com/simplesamlphp/simplesamlphp/pull/2010#issuecomment-2002165916 can be useful for inspecting the gettext domain in binary files and general manipulation on the command line.

monkeyiq:
I now have a prototype that respects the domains. 

It makes sense to fallback to the messages domain here if the selected domain does not have a translation. For example the admin domain might wish to get a translation from the `locales/en/LC_MESSAGES/messages.po` file. It seems reasonable to allow that to happen so that the admin domain gets native access to it's own explicit translations from `modules/admin/locales/en/LC_MESSAGES/admin.po` first and then can fallback to those offered already in `locales/en/LC_MESSAGES/messages.po` as well.

The domain name spacing would however keep the items in `modules/admin/locales/en/LC_MESSAGES/admin.po` from being interacted with by `modules/examplemodule/locales/en/LC_MESSAGES/examplemodule.po`. These two domains would be explicitly separate. Either of them could access strings from `locales/en/LC_MESSAGES/messages.po` from their own modules though. If they overwrite a translation in their module then it would be taken in preference to the `locales/en/LC_MESSAGES/messages.po` string.

[...]

PR #2037, opened 2024-04-03
respect the gettext domain for modules

monkeyiq:
This is a cleaned up version of the rough code to solve https://github.com/simplesamlphp/simplesamlphp/issues/2013.

The core module has only a few translations (25 ish) and seems like it should be available to all modules for translation. The current fallback translation ordering is module, core, then messages. I have been tinkering with the code at https://github.com/monkeyiq/simplesamlphp/tree/2024/march/i18n-domains2 which includes some translations in each of admin, core, and messages and an explicit callout to these in the federaion admin twig page. That let me see the three files are reachable from that page.

### DIFF
--- a/src/SimpleSAML/Locale/Localization.php
+++ b/src/SimpleSAML/Locale/Localization.php
@@ -26,6 +26,7 @@ class Localization
      * @var string
      */
     public const DEFAULT_DOMAIN = 'messages';
+    public const CORE_DOMAIN = 'core';
 
     /**
      * The default locale directory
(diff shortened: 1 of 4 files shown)


monkeyiq:
Doing another round of final testing on this I noticed that on the admin/test page the Your attributes uid row is slightly different. In `simplesamlphp-2.2` it shows extra information near the uid such as "Identificador de usuario" for lang=es. That is missing with this PR applied. That comes from the `locales/es/LC_MESSAGES/attributes.po` file which has no `X-Domain` setting. Making that `"X-Domain: messages\n"` makes the translation appear on the admin/test page just as it does in 2.2 head.

Issue #1211, opened 2019-09-30
Validate post requests via the errorreport.php


roland-d:
**Is your feature request related to a problem? Please describe.**
The problem is that you can create a spam attack via the errorreport.php

**Describe the solution you'd like**
Add a form token to the error report form. This form token can be checked on submission, if it is invalid no email is send out.

**Describe alternatives you've considered**
The only real alternative is to turn off the error report form.

**Additional context**
I posted this first in the Google group, here is my post for reference:
https://groups.google.com/forum/#!topic/simplesamlphp/ccy6fN0PslI

jaimeperez:
Hi @roland-d!

Although I see your point, I'm afraid there's little we can do here. We can of course add a nonce to the form to make sure emails are only sent after effectively triggernig an error, but what's the advantage there? I mean, the same way you can automate it in a loop to send a thousand post requests to the page, you can also automate it to trigger an error (e.g. page not found), grab the nonce and then send the email.

Also, the attack surface is quite unclear here. You cannot control the recipient of the emails, and you can't include HTML in that email either, so the only viable attack that I can see is flood the mailbox with error reports. Of course a denial of service is not good, but a nonce wouldn't really be effective against an attacker with minimum knowledge...

roland-d:
Hello @jaimeperez 

The other issue I guess I see here is not only flooding a mailbox but also hitting rate-limits on your mailserver causing actual mails not to go out.

Yes, I agree, it won't help against any kind of DDOS as that will still happen.

I am also not sure if we should do something about this, just I ran into it and wanted to share this.

jaimeperez:
Yes, I was essentially assuming that (rate limiting) as part of the denial of service scenario.

I honestly don't know. Adding a nonce to the form should be trivial, but I really don't see that offering much. I think essentially the only way to avoid this is to have proper DoS protection measures in front of the server.

tvdijen:
I'm not a fan of captcha's, but in this case they could offer a solution.
If we could implement them in a generic way, people could easily add one to their templates..

BAGELreflex:
This is often handled by IPS on the server, rather than within the application itself. Using something like [fail2ban](https://www.fail2ban.org/wiki/index.php/Main_Page), you could prevent this flooding attack vector for any request going through your web server, not just these email requests.

https://samnicholls.net/2016/06/06/fail2ban-apache-dos/

jaimeperez:
[...]

PR #2040, opened 2024-04-05
Add error report spam protection honeypot

Zipixx:
I have been experiencing an increase in spam mails via the error report form.

I agree with the comments and closing of #1211. However, bot farm requests from different origins are now creating almost daily Tickets across our SSP instances.

This patch adds a simple Honeypot with the common input name "name".
I have added it to some of our prod-instances and have not received any spam error reports and no false positives from them ever since.

It is easy to test by setting the "name" input value to a non-empty string via the browsers Inspect-function before submitting the report and does not effect real error reports due to the input element being invisible to human users.


tvdijen:
If the bots are indeed targeting SSP, this will not put them off for long.
Maybe it makes more sense to add a captcha to this page?

monkeyiq:
I agree that if bots are being made to attack a feedback form then they will probably evolve inline with small changes on that page. They have the source code after all ;)

I recalled some drop on the limits in recent months for Google Recaptcha which seems to be down to 10k/month on the free tier now. I imagine that even at that reduced rate it is likely to cover error feedback forms.

Zipixx:
I just wanted to test the waters on the topic with this harmless merge request first.
However, a modularized captcha based setup would surely be the most future proof. I agree.
Preferably with an easy to overwrite endpoint. We already use paid captcha services in some of our modules and need high accessibility for fully blind users because of government requirements.
If you decide in favor of it, I'd be happy to offer my help in any form. <3

tvdijen:
> I recalled some drop on the limits in recent months for Google Recaptcha which seems to be down to 10k/month on the free tier

Unless their documentation is outdated, according to [this](https://www.google.com/recaptcha/about/) page it's free up to 1 million Assessments / Month .

I've already done some work locally, so I will try and finish my concept this coming weekend.

monkeyiq:
There were a bunch of sites reporting back in Jan/Feb that the tier was moving from 1 million to 10k per month. And mentions of "reCAPTCHA Lite" as the free tier. Though Google would be the ones to know rather than these other sites.

monkeyiq:
It might be handy to use the "captcha" support from symfony
https://symfony.com/doc/6.4/the-fast-track/en/16-spam.html

tvdijen:
I tried that, but that requires us to migrate our forms to Symfony as well, and I couldn't get that to work because our controllers do not extend Symfony's AbstractController.. It a rabbit hole to get even started.

PR #2046, opened 2024-04-09
Add support for SP request initiation protocol

ioigoume:


tvdijen:
Hi, just to be sure, this is to handle the actual protocol messages, right? Not the metadata..

modules/saml/src/Auth/Source/SP.php
```diff
@@ -868,7 +888,7 @@ public function reauthenticate(array &$state): void
          * original IdP used a compatible AuthnContext. If not, we may need
          * need to do step-up authentication (e.g. for requesting MFA). At
          * the moment this only handles exact comparisons, since handling
-         * better/minimum/maximum would require a prioritised list.
+         * better/minimum/maximum would require a prioritized list.
```
pradtke, line 891: I think this is just US spelling vs UK spelling. I'm not sure if it is worth changing.

modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -126,6 +135,20 @@ public function login(Request $request, string $sourceId): RunnableResponse
...
         ];
 
+        // Set options provided by the Initiator Request
+        // configuration.
+        if($spSource->isRequestInitiation()) {
+            if($request->query->has('entityID')) {
+                $options['saml:idp'] = $request->query->get('entityID');
+            }
+            if($request->query->has('isPassive')) {
+                $options['isPassive'] = $request->query->get('isPassive');
```
pradtke, line 145: Does this (and the forceauthn) need to be `->getBoolean()` so we have the correct type in `$options`?

tests/modules/saml/src/Controller/ServiceProviderTest.php
```diff
@@ -141,6 +165,34 @@ public function testLogin(): void
...
+    /**
+     * @TODO: This cannot be tested until we are PSR-7 compliant
+     *
+     * Test that the target query parameter is mapped to the ReturnTo
+     * Test that the entityID query parameter is mapped to the saml:Idp
+     *
+     *
+     * @return void
+     *
+    public function testLoginRequestInitiation(): void
```
pradtke, line 177: I see what you mean about it being hard to write a unit test.

modules/saml/src/Auth/Source/SP.php
```diff
@@ -76,6 +76,13 @@ class SP extends \SimpleSAML\Auth\Source
      */
     private array $protocols = [Constants::NS_SAMLP];
 
+    /**
+     * Flag to indicate whether to disable sending the Scoping element.
+     *
+     * @var bool
+     */
+    private bool $request_initiation;
```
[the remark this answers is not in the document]
tvdijen: camelCase! Our codesniffer will complain about this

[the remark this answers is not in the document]
ioigoume: Done

modules/saml/src/Auth/Source/SP.php
ioigoume: done

modules/saml/src/Controller/ServiceProvider.php
ioigoume: We could. The reason why i did it this way is because when the actual value is used it is type casted:
```php
        if (isset($state['isPassive'])) {
            $ar->setIsPassive((bool) $state['isPassive']);
        }
```
Better be strict though.

ioigoume:
> Hi, just to be sure, this is to handle the actual protocol messages, right? Not the metadata.. the code-style issues

Hi @tvdijen , yes exactly. Please provide your feedback when possible.

> It would be appreciated if you can fix 

I fixed all the code style issues reported by the code-sniffer

pradtke:
@ioigoume  I think there is a use case that we don't handle correctly.  If a user has an existing session from idp A, and then there is a request initiation request to use idp B then with the current implementation the existing session is found rather than forcing a login from idp B.

I think to handle the above the controller (or something beside the `Simple` class, since this change is only saml SP related and not applicable to all authsources) would need to check if user is currently authenticated and for which idp, and then call `authenticate` to switch. I'd think move the `ForceAuthn` logic to the same place.

I think there is enough logic in here that we should figure out how to do a unit test, perhaps a helper method in the controller that can take the dependencies (`Auth\Simple` and the $request) as arguments, and then you can mock the `Auth\Simple`?

Lastly,  there are a large number of changes in the PR unrelated to the request initiation, such as the changes in `src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php` . I think those make it harder to review the PR. I think if the are changes to fix php code sniffer or psalm findings, then that should be in a separate PR so this one can stay focussed on request initiation.

modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -112,42 +113,68 @@ public function login(Request $request, string $sourceId): RunnableResponse
...
+        // Check if we are authenticated and under which Authenticating Authority
+        if(
+            $as->isAuthenticated()
+            && $spSource->isRequestInitiation()
+            &&  $options['saml:idp']
+        ) {
+            // Check the IdP we are currently authenticated to
+            $authData = $as->getAuthDataArray();
+            if(
+                (isset($authData['saml:AuthenticatingAuthority'])
```
pradtke, line 180: I think you can look in  key `saml:sp:IdP`

tvdijen:
Would it make sense to create a separate endpoint+controller method for this, instead of the added complexity to the existing endpoint-controllers?

ioigoume:
> Would it make sense to create a separate endpoint+controller method for this, instead of the added complexity to the existing endpoint-controllers? Also, shouldn't we automatically add the endpoint to the SP-metadata if the feature is enabled?

HI @tvdijen ,
to my opinion the context of the request is 

> Would it make sense to create a separate endpoint+controller method for this, instead of the added complexity to the existing endpoint-controllers? Also, shouldn't we automatically add the endpoint to the SP-metadata if the feature is enabled?

Hi @tvdijen ,
to my opinion the `login` action makes sense since we are handling a special use case of a login. Also, the protocol itself only provides a list of query parameters that dictate what should happen if they exist in the request url. This gives us the flexibility to extend the current functionality and create DRY code. Instead of duplicated existing functionality over different endpoints that handle the same action but under a different conditions.

Thank you,
Ioannis

tvdijen:
Why did you remove this? :(

ioigoume:
I will shortly open a new one. Making a cleanup.

PR #2058, opened 2024-04-20
translations: copy non empty translation for very core modules

monkeyiq:
If a translation in messages.po has no msgstr but one is available in the core, saml, or admin modules then copy it to the messages.po file. The opposite is also included as well to avoid empty transation strings if possible.


monkeyiq:
It turns out there are not so many of these cases. Perhaps few enough that the complexity of this action is not justified.

PR #2065, opened 2024-04-26
when migrating from an old saml to a new one the session sometimes desn't deserialize, and it should just be thrown away at that point

davidbuzz:
pretty obvious I hope.

### DIFF
--- a/src/SimpleSAML/SessionHandlerPHP.php
+++ b/src/SimpleSAML/SessionHandlerPHP.php
@@ -246,8 +246,11 @@ class SessionHandlerPHP extends SessionHandler
         $session = $_SESSION['SimpleSAMLphp_SESSION'];
         Assert::string($session);
 
-        $session = unserialize($session);
-
+        try {
+            $session = unserialize($session);
+        } catch (\Throwable $e) {
+            $session = null;  # sometimes deserializing fails, so we throw it away
+        }
         return ($session !== false) ? $session : null;
     }
 


davidbuzz:
I got this when migrating from 1.19 to 2.3 , so i did it as a PR on 2.3, but it probably relevant to any version change, and should go into master as well

monkeyiq:
I think this PR could be handy to allow version migration in SSP. 

On the other hand for day to day use I would assume that a failure to `unserialize` might be something that we might not like to silently ignore?

A method to explicitly drop all stored sessions during an upgrade might make more sense?

tvdijen:
Agreed. A `bin/console` command to purge old sessions makes more sense.

davidbuzz:
[Decades of php programming]day-to-day use of unserialize its very reliable unless the schema is changed on it.. ie the type/structure of data being stored is changed. I've never seen a failure without the code changing. .. so this PR feels entirely safe to me for day-to-day use as its throwing away the 'trust' element in the session data whenever it has and reason not to trust the integrity of it, and effectively logging that user out.

thijskh:
I think it's a good approach, it avoids additional hoop jumping on upgrades and just does the right thing for the user. Maybe it helps if a notice or warning is logged when this happens? So if it happens without you just upgrading you have something to hold on to when debugging.

monkeyiq:
I was thinking about maybe having a config key to set to ignore it if you just upgraded. Maybe just logging it for now and doing a `session = null` as in the PR. If you just upgraded you will notice a few logs about session loads falling over but you will expect to see it after an upgrade.

davidbuzz:
If u wanted to do that u could  popup a 'u have been logged out' msg too, but such a rare thing

monkeyiq:
Since it seems there is support for a merge I'll merge this and add a new PR to log when the session load fails.

PR #2067, opened 2024-04-29
start to allow adfs metadata in the admin ui again

monkeyiq:
This goes hand in hand with https://github.com/simplesamlphp/simplesamlphp-module-adfs/pull/18

With these and https://github.com/simplesamlphp/simplesamlphp-module-adfs/pull/18 I can again see the ADFS metadata in the admin/federation page.

The next logical move is getting `getSecurityTokenService` to work again too. Perhaps phase one might be to directly reach out to it and then move to using a hook to call the function.

### DIFF
--- a/modules/saml/src/MetadataBuilder.php
+++ b/modules/saml/src/MetadataBuilder.php
@@ -222,10 +222,10 @@ class MetadataBuilder
 //            case 'attributeautority-remote':
 //                $descriptors[] = $this->getAttributeAuthority();
 //                break;
-//            case 'adfs-idp-hosted':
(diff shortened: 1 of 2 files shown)


monkeyiq:
There are likely to be issues as I go along. I am still coming up to speed with SAML itself and didn't know about ADFS until very recently. That said my first move was to try to get something in the admin / federation page and then start working to refine things to be as expected. 

I will dig into an update for `$baseurl` to prp-endpoint as suggested.

I will also look at how to shift all code that references adfs out of src/SimpleSAML/Metadata (and src/SimpleSAML/*) and into the adfs module as a follow up.

Issue #174, opened 2015-01-29
Add support for the SP Request Initiation Protocol


trscavo:
Please add support for the [Service Provider Request Initiation Protocol](https://wiki.oasis-open.org/security/RequestInitProtProf), an extension protocol for SAML2 request initiation at the SP.

A sample `<init:RequestInitiator>` element in SP metadata might look like:

    <init:RequestInitiator 
        xmlns:init="urn:oasis:names:tc:SAML:profiles:SSO:request-init" 
        Binding="urn:oasis:names:tc:SAML:profiles:SSO:request-init" 
        Location="https://app.example.com/saml2/sp/Login"/>

For additional reference, see the [Shibboleth SessionInitiator](https://wiki.shibboleth.net/confluence/x/doFC) documentation.

alexstuart:
RequestInitiators provide functionality to start login flows parameterised by IdP `entityID`, which makes it straightforward for librarians to set a link on a library portal that bypasses the IdP discovery step at login. They're referred to as WAYFless URLs in the UK federation. IdP discovery is performed once at onboarding rather than at every login.

RequestInitiators can be expressed in SAML metadata which means that applications can extract the RequestInitator and generate a WAYFless URL. The UK federation has [WUGEN (the WAYFless URL Generator)](https://wugen.ukfederation.org.uk/) to do this.

In SAML 1, there's an IdP-initiated form of WAYFless URL, but in SAML 2 there's no standards-compliant way of doing this, so RequestInitiators are the only way to automatically build WAYFless URLs. This will assume more importance when [SAML 1 is removed from simpleSAMLphp](https://simplesamlphp.org/docs/stable/simplesamlphp-upgrade-notes-1.18)

[I see there wasn't a use case for RequestInitiators in 2015](https://groups.google.com/forum/#!searchin/simplesamlphp/RequestInitiator/simplesamlphp/vSU9TTmBYBQ/V8n6nYaArtMJ) although I hope I've convinced you that there is a valid use case now. Is it therefore possible to raise the importance of this issue?

I might be able to work on a pull request. Could you give some pointers where in the code base to add this functionality? Presumably it's in the code that reads the configuration; in the metadata generation module; and in the session initiation code.

jaimeperez:
Hi Alex,

I understand the most important part would be the actual support for this feature, over metadata stating support for it.

[...]

PR #2073, opened 2024-04-29
sp_request_initiation_protocol

ioigoume:
Addresses issue #174

### DIFF
--- a/modules/saml/src/Controller/ServiceProvider.php
+++ b/modules/saml/src/Controller/ServiceProvider.php
@@ -193,9 +193,7 @@ class ServiceProvider
         /**
          * Try to authenticate
          */
-        if (!$authSource->isAuthenticated()) {
-            $authSource->login($options);
-        }
+        $authSource->requireAuth($options);
 
         // Return the redirect target
         return $options['ReturnTo'];
--- a/modules/saml/src/Controller/ServiceProvider.php
+++ b/modules/saml/src/Controller/ServiceProvider.php
@@ -65,8 +65,8 @@ class ServiceProvider
      *
      * It initializes the global configuration for the controllers implemented here.
      *
-     * @param \SimpleSAML\Configuration $config The configuration to use by the controllers.
-     * @param \SimpleSAML\Session $session The Session to use by the controllers.
+     * @param   Configuration  $config   The configuration to use by the controllers.
+     * @param   Session        $session  The Session to use by the controllers.
      */
     public function __construct(
         protected Configuration $config,
@@ -98,36 +98,107 @@ class ServiceProvider
     }
 
 
-    /**
-     * Start single sign-on for an SP identified with the specified Authsource ID
-     *
-     * @param \Symfony\Component\HttpFoundation\Request $request
-     * @param string $sourceId
-     * @return \SimpleSAML\HTTP\RunnableResponse
-     */
+  /**
+   * Start single sign-on for an SP identified with the specified Authsource ID
+   *
+   * @param   \Symfony\Component\HttpFoundation\Request  $request
+   * @param   string                                     $sourceId
+   *
+   * @return \SimpleSAML\HTTP\RunnableResponse
+   * @throws Error\Exception
+   */
     public function login(Request $request, string $sourceId): RunnableResponse
     {
-        $as = new Auth\Simple($sourceId);
-        if (!($as->getAuthSource() instanceof SP)) {
+        // Initialize all the dependencies
+        $authSource = new Auth\Simple($sourceId);
+        $spSource = $authSource->getAuthSource();
+
+        if (!($spSource instanceof SP)) {
             throw new Error\Exception('Authsource must be of type saml:SP.');
         }
 
-        if (!$request->query->has('ReturnTo')) {
+        $httpUtils = new Utils\HTTP();
+
+        $returnTo = $this->loginHandler($request, $authSource, $spSource, $httpUtils);
+
+        // Redirect to the returnTo destination
+        return new RunnableResponse([$httpUtils, 'redirectTrustedURL'], [$returnTo]);
+    }
+
+    /**
+     * @param   Request       $request
+     * @param   Auth\Simple   $authSource
+     * @param   Auth\Source   $spSource
+     * @param   Utils\HTTP    $httpUtils
+     *
+     * @return string
+     * @throws BadRequest
+     * @throws Error\Exception
+     */
+    protected function loginHandler(
+        Request $request,
+        Auth\Simple $authSource,
+        Auth\Source $spSource,
+        Utils\HTTP $httpUtils
+    ): string {
+        $options = [];
+
+        if ($spSource->isRequestInitiation()) {
+            if ($request->query->has('target')) {
+                $options['ReturnTo'] = $httpUtils->checkURLAllowed($request->query->get('target'));
+            }
+            if ($request->query->has('forceAuthn')) {
+                $options['ForceAuthn'] = $request->query->getBoolean('forceAuthn');
+            }
+            if ($request->query->has('entityID')) {
+                $options['saml:idp'] = $request->query->get('entityID');
+            }
+            if ($request->query->has('isPassive')) {
+                $options['isPassive'] = $request->query->getBoolean('isPassive');
+            }
+        }
+
+        if (
+            !isset($options['ReturnTo'])
+            && !$request->query->has('ReturnTo')
+            && !$spSource->getMetadata()->hasValue('RelayState')
+        ) {
             throw new Error\BadRequest('Missing ReturnTo parameter.');
         }
-        $returnTo = $request->query->get('ReturnTo');
+        if (!isset($options['ReturnTo'])) {
+            $options['ReturnTo'] = $httpUtils->checkURLAllowed(
+                $request->query->get('ReturnTo') ?? $spSource->getMetadata()->getString('RelayState')
+            );
+        }
+
+        $authData = $authSource->getAuthDataArray();
+
+        if (
+            $authSource->isAuthenticated()
+            && $spSource->isRequestInitiation()
+        ) {
+            if (
+                // Check the IdP we are currently authenticated to
+                (isset($authData['saml:sp:IdP'], $options['saml:idp'])
+                 && $options['saml:idp'] !== $authData['saml:sp:IdP'])
+                ||
+                (isset($options['ForceAuthn']) && $options['ForceAuthn'])
+            ) {
+                // Force a re-authentication
+                $authSource->login($options);
+            }
+            // We are already authenticated, do nothing
+        }
 
         /**
-         * Setting up the options for the requireAuth() call later..
+         * Try to authenticate
          */
-        $httpUtils = new Utils\HTTP();
-        $options = [
-            'ReturnTo' => $httpUtils->checkURLAllowed($returnTo),
-        ];
+        if (!$authSource->isAuthenticated()) {
+            $authSource->login($options);
+        }
 
-        $as->requireAuth($options);
-
-        return new RunnableResponse([$httpUtils, 'redirectTrustedURL'], [$returnTo]);
+        // Return the redirect target
+        return $options['ReturnTo'];
     }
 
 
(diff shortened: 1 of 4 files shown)


modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -98,36 +102,110 @@ public function setAuthUtils(Utils\Auth $authUtils): void
...
-        if (!($as->getAuthSource() instanceof SP)) {
-            throw new Error\Exception('Authsource must be of type saml:SP.');
+        $options = [];
+
+        if ($spSource->isRequestInitiation()) {
+            if ($request->query->has('target')) {
+                $options['ReturnTo'] = $httpUtils->checkURLAllowed($request->query->get('target'));
+            }
+            if ($request->query->has('forceAuthn')) {
+                $options['ForceAuthn'] = $request->query->get('forceAuthn');
```
pradtke, line 154: I think this and the passive one should bet getBoolean

modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -98,36 +102,110 @@ public function setAuthUtils(Utils\Auth $authUtils): void
...
+            if ($request->query->has('isPassive')) {
+                $options['isPassive'] = $request->query->get('isPassive');
+            }
         }
 
-        if (!$request->query->has('ReturnTo')) {
+        if (
+            !isset($options['ReturnTo'])
+            && !$request->query->has('ReturnTo')
+            && $spSource->getMetadata()->getOptionalString('RelayState', null) === null
```
pradtke, line 167: I think `&& !$spSource->getMetadata()->hasValue('RelayState')` is more readable.

modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -98,36 +102,110 @@ public function setAuthUtils(Utils\Auth $authUtils): void
...
+        ) {
             throw new Error\BadRequest('Missing ReturnTo parameter.');
         }
-        $returnTo = $request->query->get('ReturnTo');
+        if (!isset($options['ReturnTo'])) {
+            $options['ReturnTo'] = $httpUtils->checkURLAllowed(
+                $request->query->get('ReturnTo') ?? $spSource->getMetadata()->getOptionalString('RelayState', null)
```
pradtke, line 173: I think if the RelayState is being loaded here it is not optional.
`$spSource->getMetadata()->getString('RelayState')`

modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -98,36 +102,110 @@ public function setAuthUtils(Utils\Auth $authUtils): void
...
+            );
+        }
+
+        $authData = $authSource->getAuthDataArray();
+        // We are copying the current IDP to the targeted one in case we want to ForceAuthn
+        if(empty($options['saml:idp']) && !empty($authData['saml:sp:IdP'])) {
+            $options['saml:idp'] = $authData['saml:sp:IdP'];
```
pradtke, line 180: I don't think this is true. We may be using ForceAuthn to send the user to discovery. I'd remove this part.

modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -98,36 +102,110 @@ public function setAuthUtils(Utils\Auth $authUtils): void
...
+        }
+
+        if (
+            $authSource->isAuthenticated()
+            && $spSource->isRequestInitiation()
+            && !empty($options['saml:idp'])
```
pradtke, line 186: I'm not sure why the options are being checked.

tests/modules/saml/src/Controller/ServiceProviderTest.php
```diff
@@ -184,6 +218,436 @@ public function testWithAuthIdWithoutEntity(): void
         $c->discoResponse($request);
     }
 
+    /**
+     * Test Login without Session
+     *
+     * @return void
+     */
+    public function testLoginHandleNoQueryNoSession(): void
```
pradtke, line 226: I'll give you an example of how to use dataproviders in the unit tests. Most of the tests have the same structure and are just checking that for given query parameters , that login is called with expected options.

modules/saml/src/Controller/ServiceProvider.php
ioigoume: done

modules/saml/src/Controller/ServiceProvider.php
ioigoume: done

modules/saml/src/Controller/ServiceProvider.php
ioigoume: Done

modules/saml/src/Controller/ServiceProvider.php
ioigoume: Removed

tests/modules/saml/src/Controller/ServiceProviderTest.php
```diff
@@ -185,6 +208,247 @@ public function testWithAuthIdWithoutEntity(): void
...
+        $info = ['AuthId' => 'phpunit'];
+        $config =
+            [
+                'saml:SP',
+                'entityID' => 'urn:x-simplesamlphp:example-sp',
+                'RequestInitiation' => true,
+                'RelayState' => self::RELAY_STATE,
+            ];
+
+        $spSource = new class ($info, $config) extends Source\SP {
```
pradtke, line 423: I'm curious about the need for using an anonymous subclass of `Source\SP` rather than using it directly

tests/modules/saml/src/Controller/ServiceProviderTest.php
```diff
@@ -185,6 +208,247 @@ public function testWithAuthIdWithoutEntity(): void
...
+                ],
+                [
+                    'ReturnTo' =>  self::RELAY_STATE,
+                ],
+                false
+            ],
+            'entityID=example.edu&forceAuthn=false&isPassive=false' => [
+                [
+                    'entityID' => 'example.edu',
+                    'forceAuthn' => false,
```
pradtke, line 350: I'm wondering if these should all be strings for the query parameters values.  e.g `'forceAtuhn' => 'false'` since in a real scenarios everything in `$_GET` is a string.

pradtke (commented):
@ioigoume It is looking pretty good. I'll test it out in our of our environments.

tests/modules/saml/src/Controller/ServiceProviderTest.php
ioigoume: done

pradtke:
I ran some tests in our dev environment and things behaved like I expected.   Thank you @ioigoume

ioigoume:
Hi @tvdijen,
this pull request is ready for review. We introduced a new method, the loginHandler, in order to facilitate testing.

Thank you,
Ioannis Igoumenos

modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -98,36 +98,107 @@ public function setAuthUtils(Utils\Auth $authUtils): void
...
-         * Setting up the options for the requireAuth() call later..
+         * Try to authenticate
          */
-        $httpUtils = new Utils\HTTP();
-        $options = [
-            'ReturnTo' => $httpUtils->checkURLAllowed($returnTo),
-        ];
-
-        $as->requireAuth($options);
+        if (!$authSource->isAuthenticated()) {
```
monkeyiq, line 196: This change is doing the same thing. `requireAuth` will check is the session is valid and return if we are already logged in. The same check on the session is what `isAuthenticated` does. IMHO we should just call `requireAuth` to signify that we require auth here. The less that surrounds a call to require login is probably a good thing in the code.

ioigoume: > This change is doing the same thing. `requireAuth` will check is the session is valid and return if we are already logged in. The same check on the session is what `isAuthenticated` does. IMHO we should just call `requireAuth` to signify that we require auth here. The less that surrounds a call to require login is probably a good thing in the code.

Hi @monkeyiq . I applied the change as you suggested. Thank you for the feedback.

monkeyiq: Cool thanks :)

monkeyiq:
I should probably also include a link here to the previous PR https://github.com/simplesamlphp/simplesamlphp/pull/2046

monkeyiq:
One thing I notice is that in general this will accept `RelayState` and turn it into `ReturnTo` if there is no existing returnTo. If that is ok for a general case in SAML then all is well, otherwise we are introducing a potential change. 

From my reading it seems RelayState might be a deep pointer the SP wants mirrored back from the IdP. Does the RelayState have potential other semantics with SP Request Initiation Protocol interactions?

It may also make sense to hint to the reader that they might like to investigate the `trusted.url.domains` configuration option to best validate the `target`. I imagine sites will already have trusted domains set up but mentioning it in the docs seems like a good idea to follow section 2.5 of the spec.

tests/modules/saml/src/Controller/ServiceProviderTest.php
```diff
@@ -185,6 +208,246 @@ public function testWithAuthIdWithoutEntity(): void
...
+                [],
+                true
+            ]
+        ];
+    }
+
+    /**
+     * Test Login without Session/ Not Authenticated
+     *
+     * @dataProvider loginNotAuthenticatedDataProvider
```
tvdijen, line 278: Please use PHPUnit [attributes](https://docs.phpunit.de/en/10.5/writing-tests-for-phpunit.html#data-providers)-syntax instead

ioigoume: Done

ioigoume:
@monkeyiq 

> One thing I notice is that in general this will accept `RelayState` and turn it into `ReturnTo` if there is no existing returnTo. If that is ok for a general case in SAML then all is well, otherwise we are introducing a potential change.

According to the [SAML specification](http://docs.oasis-open.org/security/saml/v2.0/saml-bindings-2.0-os.pdf)

```
Some bindings define a "RelayState" mechanism for preserving and conveying state information. When such a mechanism is used in conveying a request message as the initial step of a SAML protocol, it places requirements on the selection and use of the binding subsequently used to convey the response. Namely, if a SAML request message is accompanied by RelayState data, then the SAML responder MUST return its SAML protocol response using a binding that also supports a RelayState mechanism, and it MUST place the exact RelayState data it received with the request into the corresponding RelayState parameter in the response.
```
and 
```
Sometimes a binding-specific field called RelayState is used to coordinate messages and actions of IdPs and SPs, for example, to allow an IdP (with which SSO was initiated) to indicate the URL of a desired resource when communicating with an SP."
```

Based on above descriptions from the specification, the RelayState is just a pointer that the Service Provider will use to find out the final TARGET or ReturnTo URL.

> From my reading it seems RelayState might be a deep pointer the SP wants mirrored back from the IdP. Does the RelayState have potential other semantics with SP Request Initiation Protocol interactions?

According to the protocol if the `target` is omitted then the SP MUST use a default value, which it unilaterally determines. In other words, the RelayState is the fallback in case the target URL is omitted.

> It may also make sense to hint to the reader that they might like to investigate the `trusted.url.domains` configuration option to best validate the `target`. I imagine sites will already have trusted domains set up but mentioning it in the docs seems like a good idea to follow section 2.5 of the spec.

I added a sentence in the md file

monkeyiq:
I am happy with the RelayState being used for the `target` if that is not set in a Sstc-request-initiation setting. As mentioned in 2.3.1 Defined Parameters in Sstc-request-initiation-cs-01 the SP must provide a default if nothing is given and has the choice of what to provide there.

monkeyiq:
Actually that seems ok in general. The returnto is coming from the request first in both the old and new code and only from the sp metadata as a fallback in the PR. There isn't any call facing change here and if the sp->md has a relaystate it should come from the configuration and so be chosen by the admin anyway.

--- $returnTo = $request->query->get('ReturnTo');
+++ $request->query->get('ReturnTo') ?? $spSource->getMetadata()->getString('RelayState')

monkeyiq:
It also helps the review that this has been tested in a dev environment as well :)

PR #2074, opened 2024-04-30
Add missing changelog entry for sqlauth adding username_regex option

nathanjrobertson:
I'm guessing this is manually maintained, and that the latest version of `sqlauth` will be pulled through in the simplesamlphp release. If so, the update in `sqlauth` was missed.


tvdijen:
That wasn't supposed to be released in a bugfix-release :-/
I will try and fix that, and then merge this since it's released now.

tvdijen:
Fixed in bf2bfd189

PR #2086, opened 2024-05-09
Deprecate Net Utils and use Symfony IpUtils instead

tvdijen:


monkeyiq:
I have played around with a small test command using the asserts from the test suite and doing both the original and new `ipCIDRcheck` methods and the asserts match up as expected.

modules/core/src/Auth/Source/SourceIPSelector.php
```diff
@@ -91,18 +90,18 @@ public function __construct(array $info, array $config)
...
         foreach ($this->zones as $name => $zone) {
             foreach ($zone['subnet'] as $subnet) {
-                if ($netUtils->ipCIDRcheck($subnet, $ip)) {
+                if (IpUtils::checkIp($ip, $subnet)) {
                     // Client's IP is in one of the ranges for the secondary auth source
                     Logger::info(sprintf(
                         "core:SourceIPSelector:  Selecting zone `%s` based on client IP %s",
                         $name,
-                        $ip
+                        $ip,
```
monkeyiq, line 104: small thing, both this and https://github.com/simplesamlphp/simplesamlphp/pull/2087/files have a trailing comma after the $ip. It is handy that php -a doesnt seem to care about this and maybe it is not so much of an issue for sprintf()s that might have trailing args anyway.

monkeyiq:
I like this PR. If we are relying on Symfony we might as well lean into it and use it in place of local code.

monkeyiq (approved):
looks good to me. I notice that if `$ip = Request::createFromGlobals()->getClientIp()` is run in a local command $ip == null after and this results in the following. Though the old code looks like it wouldn't work to well in a local command without explicitly supplying an $ip value as well. So it provides the same outcome in both cases though probably more verbose in the new code...

```
Fatal error: Uncaught TypeError: 
Symfony\Component\HttpFoundation\IpUtils::checkIp(): Argument #1 ($requestIp) 
must be of type string, null given,
```

monkeyiq:
I need to brush up my symfony knowledge a bit. There are likely other places we can lighten to core ssp code and leverage sym a bit more.

modules/core/src/Auth/Source/SourceIPSelector.php
tvdijen: Trailing comma's are perfectly fine in PHP since 7.2. It reduces diff-output if you would add another parameter.

tvdijen:
> I notice that if `$ip = Request::createFromGlobals()->getClientIp()` is run in a local command $ip == null after and this results in the following.

I've made it fall back to `127.0.0.1` if getClientIp() would return `null`.

PR #2089, opened 2024-05-09
Remove Languages::$languages_names and use symfony instead

tvdijen:


monkeyiq:
This is the removal of deprecated features from https://github.com/simplesamlphp/simplesamlphp/pull/2088

bin/translateAttributes.php
```diff
@@ -32,7 +32,7 @@
 
 $attributes = [];
 
-$languages = SimpleSAML\Locale\Language::$language_names;
+$languages = Symfony\Component\Intl\Languages::getNames();
```
monkeyiq, line 35: without a `use` above wouldn't this need a leading `\`, ie `\Symfony\Component....`

tvdijen: Nope, not unless the file starts with a `namespace` statement, which it doesn't.

monkeyiq: ah ok. cool :)

monkeyiq:
I started digging into what changes there might be in the arrays. For Norwegian it looks like be a move in the right direction having `no` as the macrolanguage now  https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes#no

The old $language_names
```
        'no'    => 'Bokmål', // Norwegian Bokmål
        'nn'    => 'Nynorsk', // Norwegian Nynorsk
```
in symfony `Languages::getNames()`:
```
    [no] => Norwegian
    [nn] => Norwegian Nynorsk
    [nb] => Norwegian Bokmål
```

I am not a lang code expert... just doing a little sifting to see what might have changed :)

tvdijen:
I think for `nb` we're already mapping it to `no` in the `$defaultLanguageMap` array.. `nn` would have to be added to that array.

PR #2090, opened 2024-05-10
Deprecate some of the crypto-utils and use Symfony's password hasher instead

tvdijen:


monkeyiq:
FWIW I also played around with pwgen making a password hash and verifying it with an out of band pwverify.php command just to be pedantic. Made sense to me to play with it a little since AdminPassword.php is updated in the PR.
```
$pw = $argv[1];
echo "$pw\n";
$hasher = new Symfony\Component\PasswordHasher\Hasher\NativePasswordHasher();
if (!$hasher->verify('$argon2id...known hash string here', $pw)) {
    echo "not right\n";
} else {
    echo "ok\n";
}
```

PR #2092, opened 2024-05-10
Deprecate AES-methods in Crypto-utils

tvdijen:
They are only used once and can be replaced with a call to the `xml-security`-library


monkeyiq:
This looks good! My main reservation was that the old code was using CBC and the new GCM mode. 

https://github.com/simplesamlphp/simplesamlphp/blob/061fc1773969f1c24397c74cd726084a5dc5f92e/src/SimpleSAML/Utils/Crypto.php#L58

But as these calls to aesEncrypt() etc are only used once and only to create a redirection I am thinking that the change might be fine. There might be some redirections that are in flight right as the code is updated but I don't see then as being long running? An existing CBC payload would fail to decode as GCM but should only happen within a very limited timeframe.

tvdijen:
Yes, this was for a reason.. The old utility-methods did some form of integrity validation with hashing that I couldn't reproduce. GCM doesn't have this requirement, because it has integrity validation built in.

Issue #2105, opened 2024-05-24
Hash session ID in the database


drupal-spider:
**Is your feature request related to a problem? Please describe.**
The session ID is a sensitive credential information as important as user's password, which need to be hashed in the database.

**Describe the solution you'd like**
In the SimpleSAML\Store\SQLStore class, check whether the key is a session id, if so then hash the key before get or set from the database.

drupal-spider:
Created a pull request for this new feature.
#2106

monkeyiq:
This has been merged and backported to the 2.3 branch. Thanks @drupal-spider & @tvdijen!

PR #2106, opened 2024-05-24
[#2105] Hash session ID in the database

drupal-spider:
This pull request makes the following changes:

- Add a new private function to hash string.
- Hash session ID when set, get and delete session from the database.

### DIFF
--- a/src/SimpleSAML/Store/SQLStore.php
+++ b/src/SimpleSAML/Store/SQLStore.php
@@ -347,6 +347,10 @@ class SQLStore implements StoreInterface
      */
     public function get(string $type, string $key): mixed
     {
+        if ($type == 'session') {
+            $key = static::hashBase64($key);
+        }
+
         if (strlen($key) > 50) {
             $key = sha1($key);
         }
@@ -393,6 +397,10 @@ class SQLStore implements StoreInterface
             $this->cleanKVStore();
         }
 
+        if ($type == 'session') {
+            $key = static::hashBase64($key);
+        }
+
         if (strlen($key) > 50) {
             $key = sha1($key);
         }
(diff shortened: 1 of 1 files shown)


src/SimpleSAML/Store/SQLStore.php
```diff
@@ -436,4 +448,25 @@ public function delete(string $type, string $key): void
...
+
+
+    /**
+     * Calculates a base-64 encoded, URL-safe sha-256 hash.
+     *
+     * @param string $data
+     * @return string The hashed data.
+     */
+    public static function hashBase64(string $data) {
+        $hash = base64_encode(hash('sha256', $data, TRUE));
```
tvdijen, line 460: May I suggest using the following instead?

```
$secretSalt = (new Utils\Config())->getSecretSalt();
return hash_hmac('sha256', $data, $secretSalt);
```

hash_hmac() will always produce a url-safe string.

tvdijen: I also think this can be solved with a non-static private method, which is easier to unit-test.

monkeyiq: Using the salt will also provide more privacy

drupal-spider: Yes, that is a good idea.

tvdijen:
Please fix the coding style and we're a go

drupal-spider:
> Please fix the coding style and we're a go

Style has been fixed, please review.
Thanks.

drupal-spider:
Rebase simplesamlphp:master

tvdijen:
Merged and backported to the 2.3.x branch. Release is yet to be determined.

PR #2110, opened 2024-05-27
allow again admin module login with raw password stored in config.php

monkeyiq:
The old pwValid method would fall back to a direct `===` compare if the `password_verify()` failed. 

https://github.com/simplesamlphp/simplesamlphp/blob/e2cd10b10e8cab40d534f938058911e94b32d66b/src/SimpleSAML/Utils/Crypto.php#L379

If we want to allow login as admin to the admin module again without storing the bin/pwgen for the admin password in the config we need to also check if the stored admin password is exactly what was supplied and allow that again.

### DIFF
--- a/modules/core/src/Auth/Source/AdminPassword.php
+++ b/modules/core/src/Auth/Source/AdminPassword.php
@@ -59,6 +59,9 @@ class AdminPassword extends UserPassBase
             throw new Error\Error(Error\ErrorCodes::WRONGUSERPASS);
         }
 
+        if( $password === $adminPassword ) {
+            return ['user' => ['admin']];
+        }
         $hasher = new NativePasswordHasher();
         if (!$hasher->verify($adminPassword, $password)) {
             throw new Error\Error(Error\ErrorCodes::WRONGUSERPASS);
--- a/modules/core/src/Auth/Source/AdminPassword.php
+++ b/modules/core/src/Auth/Source/AdminPassword.php
@@ -59,7 +59,7 @@ class AdminPassword extends UserPassBase
             throw new Error\Error(Error\ErrorCodes::WRONGUSERPASS);
         }
 
-        if( $password === $adminPassword ) {
+        if ($password === $adminPassword) {
             return ['user' => ['admin']];
         }
         $hasher = new NativePasswordHasher();


jaimeperez:
Ben, I'm not sure this is a good idea. What is the context for allowing passwords in clear text in the configuration? SSP should discourage insecure practice.

In any case, the proposed fix also introduces a security issue, since `===` does not evaluate in constant time. That leads to a timing vulnerability that would allow an attacker to recover the actual password.

monkeyiq:
I am happy to go either way. As of 2 weeks ago the above was allowed and used the `===` and may have been changed as a side effect. 

Storing hashed passwords is a more secure choice. If we want to only allow that then we need to update the docs to indicate that folks must update their config.php when upgrading from 2.2 to 2.3 to update their old config/`auth.adminpassword` value with the output from `bin/pwgen.php`.

jaimeperez:
I wasn't aware that this was already in place. I would update the changelog and upgrade notes to reflect the change, but if that's not enough because we have an avalanche of support cases opened by people locked out of the admin module, then I would at least not use `===` to check the password. We used to have a `secureCompare()` function in `Utils\Crypto` that implemented a secure comparison in constant time, but it seems that one is no longer available:

https://github.com/simplesamlphp/simplesamlphp/commit/b72c79e3070f930d758f5c269333d63ed7509e2e#diff-cca3439eded50221779137e4bb134e1889adf893e08651a31833b4f8888ee565R352-R381

tvdijen:
> We used to have a `secureCompare()` function in `Utils\Crypto` that implemented a secure comparison in constant time, but it seems that one is no longer available:
> 
> [b72c79e#diff-cca3439eded50221779137e4bb134e1889adf893e08651a31833b4f8888ee565R352-R381](https://github.com/simplesamlphp/simplesamlphp/commit/b72c79e3070f930d758f5c269333d63ed7509e2e#diff-cca3439eded50221779137e4bb134e1889adf893e08651a31833b4f8888ee565R352-R381)


I recently removed that because it was no longer in use. The `hash_equal()` method is available since PHP 5.6 to take over this functionality

jaimeperez:
Yep, I noticed. It makes sense to me, although it's also true that `hash_equals()` forces you to hash the password, and therefore plaintext passwords are no longer usable. As I said previously, I would just update the changelog and upgrade notes and keep the more secure behaviour you implemented Tim, but if this becomes a big support issue, then at least we should recover the old manual implementation of `secureCompare()` to perform the comparison here.

monkeyiq:
Adding something to the logs to help folks with exact steps for `bin/pwgen.php` seems reasonable. It is a very quick process when you know what to do but having a log line to an FAQ with the exact steps might help things for folks who upgrade and discover afterwards that the admin is not working. 

It might even make sense to link this from the login page to show on login failure if username==admin it is fairly easy to see if the password is raw or hashed so quick to verify.... Or better yet display a specific error if the password is not hashed in the config.php like the "Password not set" error to catch it before the login fails. This relies on a little heuristic sniffing for the algo name and/or format of the `auth.adminpassword` string but might help avoid frustration.

monkeyiq:
The quicker a user sees why it failed and gets direct information on how to fix it the better the update.

tvdijen:
No heuristics needed.. See https://www.php.net/manual/en/function.password-get-info.php

monkeyiq:
As discussed, I will make a new PR that informs the user of why the login has failed when not using a hashed password to help with migration to 2.3+ for those who may skip part of the release notes.

PR #2113, opened 2024-05-28
23: if the auth.adminpassword is not hashed provide some insight

monkeyiq:
In 2.3 and above non hashed admin passwords are not accepted any more. This warns the user and provides some insight to what went wrong and a link with more detail on the exact steps to fix it.

This is the opposite to https://github.com/simplesamlphp/simplesamlphp/pull/2110

This references https://github.com/simplesamlphp/simplesamlphp/wiki/Frequently-Asked-Questions-(FAQ)#failed-to-login-to-the-admin-page-with-and-error-message-admin-password-not-set-to-a-hashed-value for the solution so we can potentially make it more interesting if need be.

### DIFF
--- a/modules/core/src/Auth/Source/AdminPassword.php
+++ b/modules/core/src/Auth/Source/AdminPassword.php
@@ -63,7 +63,7 @@ class AdminPassword extends UserPassBase
         if (!$pwinfo['algo']) {
             throw new Error\Error(Error\ErrorCodes::ADMINNOTHASHED);
         }
-        
+
         $hasher = new NativePasswordHasher();
         if (!$hasher->verify($adminPassword, $password)) {
             throw new Error\Error(Error\ErrorCodes::WRONGUSERPASS);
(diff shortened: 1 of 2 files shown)


modules/core/src/Auth/Source/AdminPassword.php
```diff
@@ -59,6 +59,11 @@ protected function login(string $username, string $password): array
             throw new Error\Error(Error\ErrorCodes::WRONGUSERPASS);
         }
 
+        $pwinfo = password_get_info($adminPassword);
+        if (!$pwinfo['algo']) {
```
tvdijen, line 63: Let's make this a bit more specific `=== null`

tvdijen:
Works for me! Don't forget to port this to `master` too

modules/core/src/Auth/Source/AdminPassword.php
monkeyiq: updated in 804b42bc

monkeyiq:
cherry picked over to `master`

PR #2129, opened 2024-06-13
Fixes and completes luxembourgish (LB) Translation

PitWenkin:
Fixes existing translatiion according to official rulesets (lod.lu / spellchecker.lu)
Adds (most) missing translations


tvdijen:
Thanks @PitWenkin !
We have a lot of modules that probably also need improvements if you're up for it? 😇

PR #2131, opened 2024-06-13
BC-break: remove deprecated single-string format for endpoints

### LINKED ISSUE #2130: Undefined array key
When connecting to the IDP the errorlog throws:

`PHP Warning:  Undefined array key "index" in /var/www/saml/modules/saml/src/IdP/SAML2.php on line 228`
https://github.com/simplesamlphp/simplesamlphp/blob/d7a7357c842552b5385d67873fa929aab0f83caf/modules/saml/src/IdP/SAML2.php#L246

Dumping `$ep` shows only 'Location' and 'Binding', with:
'Location' consisting of 'AssertionConsumerService' of the given SP
'Binding'  beeing 'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST'

I'm not sure if this is due to a configuration mishap on my part, or an oversight, or intended?

`saml20-sp-remote.php`  has multiple simple entries in form of: 
```
$metadata['https://domain.tld'] = [
        'AssertionConsumerService' => 'https://domain.tld/endpoint1',
        'SingleLogoutService' => 'https://domain.tld/endpoint2'
];
```

```
Version 2.2.2. 
Linux Ubuntu 22.04.4
PHP 8.1.2
```

tvdijen:
Fixes #2130

Since this is broken anyway, I'd like to backport this to the 2.3 branch and add an upgrade note about it.

### DIFF
--- a/docs/simplesamlphp-metadata-endpoints.md
+++ b/docs/simplesamlphp-metadata-endpoints.md
@@ -12,32 +12,7 @@ Endpoint                       | Indexed | Default binding
 `SingleLogoutService`          | N       | HTTP-Redirect
 `SingleSignOnService`          | N       | HTTP-Redirect
 
-The various endpoints can be specified in three different ways:
-
-* A single string.
-* Array of strings.
-* Array of arrays.
-
-A single string
----------------
-
-    'AssertionConsumerService' => 'https://sp.example.org/ACS',
-
-This is the simplest endpoint format.
-It can be used when there is only a single endpoint that uses the default binding.
-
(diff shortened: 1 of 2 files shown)


thijskh (approved):

monkeyiq:
Hey, I missed out due to being sleeping. 

I seems that in simplesamlphp-metadata-endpoints.md from the part that was removed it didn't mention that the string/array styles were deprecated. Still, if the string method doesn't work then not too many folks can be using it.

PR #2132, opened 2024-06-14
Update test suite for no single string endpoints

monkeyiq:
This is an update after single string and single array endpoints were removed. See https://github.com/simplesamlphp/simplesamlphp/pull/2131

Note that there are some heuristics that seem to have been implicit.

This will effect all endpoints ArtifactResolutionService, SingleSignOnService, and SingleLogoutService.

If there is no binding set then one is set for you. This matches the default value table from
https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-2.2/docs/simplesamlphp-metadata-endpoints.md

If there is only one end point in the array and the binding is not valid then it is refreshed for you to something that is valid.

On the other hand, for a list of end points no such refreshing is done so that the case of a non default Location with a valid Binding can be the one that is used.

### DIFF
--- a/src/SimpleSAML/Configuration.php
+++ b/src/SimpleSAML/Configuration.php
@@ -1187,6 +1187,7 @@ class Configuration implements Utils\ClearableState
             case 'saml20-sp-remote:AssertionConsumerService':
                 return Constants::BINDING_HTTP_POST;
             case 'saml20-idp-remote:ArtifactResolutionService':
+            case 'attributeauthority-remote:AttributeService':
                 return Constants::BINDING_SOAP;
             default:
                 throw new Exception('Missing default binding for ' . $endpointType . ' in ' . $set);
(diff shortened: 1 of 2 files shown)


tvdijen:
I think, if you agree with the changes I made, this is ready to be merged.
Don't forget to cherry-pick this into master as well!

monkeyiq:
Your updates look good. I knew there was a better way to write `isValidBinding()`. I added it along the way while experimenting and didn't get back to levering it out for the proper solution. Thus the lack of method comments etc, I guess a "FIXME: make this real code" might have been a nicer metadata.

monkeyiq:
This has been picked over to master with commits starting at 
https://github.com/simplesamlphp/simplesamlphp/commit/2d0b4a7b71238cb4ff754258a23cad1523a90a25

tvdijen:
Next time maybe squash & merge?.. It will save you from having to cherry-pick every single commit

monkeyiq:
This is a good idea since I think the semantics of the individual 10 or so commits is not super interesting in this case.

PR #2140, opened 2024-06-18
Fixes and completes luxembourgish (LB) Translation

PitWenkin:
Fixes existing translation according to official rulesets (lod.lu / spellchecker.lu)
Adds (most) missing translations

BTW: I'm wondering of the format of lines 85-87 is correct …


tvdijen:
I've also fixed the format.. It wasn't exactly broken, but the newline is just not necessary

PR #2141, opened 2024-06-19
fix for: ‘autofill username’ that resulted in a WRONGUSERPASS

doedje:
The commit [autofill username when it's hinted by the SP in a request parameter](https://github.com/simplesamlphp/simplesamlphp/commit/2eaa634f095143f43d320e8b305378efd8d61431) resulted in a WRONGUSERPASS error when an SP (outlook.office.com for instance) also supplies a 'username' attribute in it's request.

The code from file `modules/core/src/Controller/Login.php` at the `private function handleLogin` would trigger the `handleLogin` functions of either `UserPassOrgBase` or `UserPassBase` without a `$password` being set. Changing the `if` statement to only check for `!empty($password)` fixes this.

I also moved up the `public function loginuserpassorg` so it is now together with the other `public function loginuserpass`. I also think the `private function handleLogin` could use a bit more cleaning up, which I am willing to do and add to this PR, but rather discuss these with some of you before I start... ;)

### DIFF
--- a/modules/core/src/Controller/Login.php
+++ b/modules/core/src/Controller/Login.php
@@ -116,6 +116,37 @@ class Login
 
 
     /**
+     * This page shows a username/password/organization login form, and passes information from
+     * into the \SimpleSAML\Module\core\Auth\UserPassBase class, which is a generic class for
+     * username/password/organization authentication.
+     *
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function loginuserpassorg(Request $request): Response
+    {
+        // Retrieve the authentication state
+        if (!$request->query->has('AuthState')) {
+            throw new Error\BadRequest('Missing AuthState parameter.');
+        }
+        $authStateId = $request->query->get('AuthState');
+        $this->authState::validateStateId($authStateId);
+
+        $state = $this->authState::loadState($authStateId, UserPassOrgBase::STAGEID);
+
+        /** @var \SimpleSAML\Module\core\Auth\UserPassOrgBase $source */
+        $source = $this->authSource::getById($state[UserPassOrgBase::AUTHID]);
+        if ($source === null) {
+            throw new BuiltinException(
+                'Could not find authentication source with id ' . $state[UserPassOrgBase::AUTHID],
+            );
+        }
+
+        return $this->handleLogin($request, $source, $state);
+    }
+
+
+    /**
      * This method handles the generic part for both loginuserpass and loginuserpassorg
      *
      * @param \Symfony\Component\HttpFoundation\Request $request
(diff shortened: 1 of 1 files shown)


tvdijen:
> "[..]" resulted in a WRONGUSERPASS error when an SP [..] also supplies a 'username' attribute in it's request

I'm not sure I follow.. What would a request look like?

doedje:
As copied from SAML-tracer, this is the request sent to our SSP-IdP when trying to login at outlook.office.com with my mail address: (shortened SAMLRequest and RelayState for readability)
```
POST
SAMLRequest: PHNh...VzdD4=
RelayState: estsredirect=2&estsrequest=rQQIARAA...k2_CXp-Aw2
username: remy.blom@hku.nl
```
On our IdP I already implemented the change from this PR. So you will not get an error. But with the code in 2.2.2 it will try to login (using LDAP in our case) right away since $username is filled.

tvdijen:
I'm worried now that an empty username and non-empty password will now throw an exception. Could you verify this?

doedje:
It will show the WRONGUSERPASS error, which is quite as expected...

tvdijen:
No, no, it shouldn't error.. It should just show the login-screen again.. That's what the `!empty($username)` did..

doedje:
Sorry, I should be more precise. When I try to login with an empty username and non-empty password, it will show BOTH the "Incorrect username or password" WRONGUSERPASS error and the login form again.

doedje:
okay, but maybe we should get down on what that function should do:

- only username, just show form
- only password, just show form
- both username and password, try to login
- only show WRONGUSERPASS when login failed.

right? because then I know what to aim for when rewriting that function.

tvdijen:
Could it be that you're not using the latest version of the ldap-module? Because I cannot reproduce the issue (and the prefilling of the username isn't working either)

tvdijen:
But then why does my setup not give the same issue you are facing? :/

doedje:
Hmmm, if we only knew... I also tried disabling my theme, but the result is the same...

(surf to outlook.office.com, enter my mail address, this is what I get...)

image

tvdijen:
I've reverted the autofill commit and took a different approach in 1139c8623

tvdijen:
I'm closing this because the issue has been resolved. Please feel free to discuss the improvements to `handleLogin` here.

doedje:
There is something with all the nested `if`s that is bothering me. I like to give it a shot to try and iron out the folds ;)

- only username, just show form
- only password, just show form
- both username and password, try to login
- only show WRONGUSERPASS when login failed.

This still stands right?

tvdijen:
Yeah, it's messy.. Feel free to improve it!
Just bear in mind that this code is also used for the username/pass/org flow..


>
> * only username, just show form
> * only password, just show form

I don't think I agree. We cannot process the form unless both username and password are present.. If anyone tries to post it without username or password, it should show the "Wrong username or password" error.

Issue #2133, opened 2024-06-14
Custom Error not working using Error class extension


rcalpha:
**Specifics of your environment**
1. Are you acting as SP/IdP/proxy?: IdP
2. SimpleSAMLphp: What version are you using?: 2.2.2
3. PHP: What version are you using?: 8.2
4. Platform: unix or Windows?: linux
5. Webserver: Apache/Nginx/ISS?: Apache

**Describe the bug**

I was trying to migrate our Auth Source module which was using custom error message by my own with version 2.0.6.

I tried to use the proposed solution of issue #1867 by extending the Error class done in 2.1 and 2.2 as proposed in the pull #1870

I couldn't tried before due to limited time.

I did two class extension as proposed in our Auth Module (I'm attaching both file):

[CustomErroModuler.zip](https://github.com/user-attachments/files/15840207/CustomErroModuler.zip)

- ErrorUc extends from Error\Error
- ErrorUcCodes extends from Error\ErrorCodes

In my auth source module, which is an extension of UserPassBase class , to do a simple test, in the login method, at the beginning invoke the error with:
---------------
            throw new ErrorUc(ErrorUcCodes::BIND_SEARCH_CONNECT_ERROR);
---------------

Instead the usual
---------------
            throw new Error\Error('WRONGUSERPASS');
---------------

And I got an exception: (I attached a txt file with debug information) 
-------------------------------------------------------------
Caused by: SimpleSAML\Error\Exception: Key "BINDSEARCHCONNECTERROR" for array with keys "ACSPARAMS, ARSPARAMS, AUTHSOURCEERROR, BADREQUEST, CASERROR, CONFIG, CREATEREQUEST, DISCOPARAMS, GENERATEAUTHNRESPONSE, INVALIDCERT, LDAPERROR, LOGOUTINFOLOST, LOGOUTREQUEST, MEMCACHEDOWN, METADATA, METADATANOTFOUND, NOACCESS, NOCERT, NORELAYSTATE, NOSTATE, NOTFOUND, NOTFOUNDREASON, NOTSET, NOTVALIDCERT, PROCESSASSERTION, PROCESSAUTHNREQUEST, RESPONSESTATUSNOSUCCESS, SLOSERVICEPARAMS, SSOPARAMS, UNHANDLEDEXCEPTION, UNKNOWNCERT, USERABORTED, WRONGUSERPASS" does not exist in core:loginuserpass.twig
-------------------------------------------------------------

**To Reproduce**
[...]

PR #2142, opened 2024-06-19
NO MERGE: Example for testing of custom error codes in templates

monkeyiq:
This is not intended to be merged.

This is a PR mainly to show the diff for testing of the use of a custom error code in templates. The PR needs both a better name for and some security around the use of the `codeclass` state variable. By storing the `ErrorCodes` class in the state we can recreate it later and thus support both normal and custom error codes in templates. 

By storing the errorcodes class both of the code paths tested in `UserPass.php` work. A regular `Error` can be thrown and will show up correctly as well as a custom `myError` object. If the class is instead fixed then your subclass will need to supply it's own codes as well as compatible ones with those in the normal ErrorCodes.

This relates to the issue https://github.com/simplesamlphp/simplesamlphp/issues/2133


monkeyiq:
since this is a no merge I am closing it to avoid cluttering things. It is mainly to allow the diff to be easily seen.

PR #2152, opened 2024-06-26
docs: upgrade command should leave user on a tag checkout

monkeyiq:
I don't think the user wants to go from a tag checkout to master here if they return to this doc file to perform an update.

I don't know if the composer command needs the windows additional options for the upgrade too. It seems that way.


tvdijen:
I don't think I need my email address propagated to the website tbh..

PR #2161, opened 2024-07-03
Fix bug in post.twig

Athanasius2:
If a posted value is an array, this template iterates through it as 'index' and 'item' and then passes 'value' to the template where it should be 'item'. This is causing post values that are supposed to be the elements of arrays to be the string "Array" instead.


tvdijen:
Nice catch! And kind of incredible this got undetected for five years!

PR #2162, opened 2024-07-04
Use the appropriate way to set the charset in HTML5

tvdijen:
See https://webhint.io/docs/user-guide/hints/hint-meta-charset-utf-8/?source=devtools for the reasoning behind this change. It's fully backwards compatible.


monkeyiq:
Looking at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#browser_compatibility it seems that charset is supported in older browsers than http-equiv.content-type so we shouldn't get any regressions on older browsers.

monkeyiq (approved):

PR #2163, opened 2024-07-04
Possibly use the HSTS response header

monkeyiq:
I would imagine that almost all interaction with web servers in the SAML process is going to be via https. Given that perhaps we want to remove the possibility of a user somehow being sent to a non https page?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#preloading_strict_transport_security

### DIFF
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -287,6 +287,7 @@ $config = [
             "base-uri 'none'",
         'Referrer-Policy' => 'origin-when-cross-origin',
         'X-Content-Type-Options' => 'nosniff',
+        'Strict-Transport-Security: max-age=63072000; includeSubDomains',
     ],
      */
 
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -269,6 +269,9 @@ $config = [
 
     /*
      * Set custom security headers. The defaults can be found in \SimpleSAML\Configuration::DEFAULT_SECURITY_HEADERS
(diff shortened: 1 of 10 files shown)


tvdijen:
If we do this, we should at least make it conditional based on whether \SimleSAML\Utils\HTTP::isHttps() is true or not.

thijskh:
I agree for production but I think the problem with adding the header by default makes development immediately very hard. So might increase startup costs for people just trying it out? Maybe Tim's suggestion might mitigate this.

monkeyiq:
It might be an idea to update or merge this sometime this week

monkeyiq:
The new `getHeadersSecurity` method only adds HSTS on https requests and is based on the configuration so can optionally not do that as well.

monkeyiq:
For example if you have an empty `headers.security.https.merge` config then no HSTS is added on https requests.

tvdijen:
I'm not sure @monkeyiq ...   We didn't add HSTS to the default headers, because it's usually something a webserver would do.  If people really want to do this inside SSP, they can just add it to the config.

monkeyiq:
Sounds good. If there was previous thoughts on HSTS and it was left as something up to the admin to consider then we should probably leave it that way for now. I probably should have thrown the idea out there on slack or as a gh issue for open discussion first :)

PR #2172, opened 2024-07-09
i18n: allow a theme and module to work together

monkeyiq:
If you have a module in use such as `multiauth:MultiAuth` and use are using a theme at the same time then you have (or should have) two default translation domains in action: the multiauth and the theme. This PR allows both of these to be in use and will allow a theme to see the module translation domain as well as it's own.

This bug was reported on the users mailing list.

### DIFF
--- a/src/SimpleSAML/Locale/Localization.php
+++ b/src/SimpleSAML/Locale/Localization.php
@@ -17,6 +17,7 @@ use Gettext\Loader\PoLoader;
 use Gettext\{Translations, Translator, TranslatorFunctions};
 use SimpleSAML\{Configuration, Logger};
 use Symfony\Component\HttpFoundation\File\File;
+use SimpleSAML\Locale\Translate;
 
 class Localization
 {
@@ -137,6 +138,7 @@ class Localization
     public function defaultDomain(string $domain): self
     {
         $this->translator->defaultDomain($domain);
+        Translate::addDefaultDomain($domain);
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


monkeyiq:
Considering the setupTwig method, I think the problem is with the use of `localization->defaultDomain()`. In `defaultDomain()` things are passed on to `this->translator->defaultDomain`. this->translator is a `Gettext::Translator`.

Taking a look at ./vendor/gettext/translator/src/Translator.php the `defaultDomain` method just sets a local instance variable. So there is only one (whatever the last one was) that is the default domain. 

In `setupTwig()` the localization->defaultDomain is called once first for the module and then subsequently for the theme if there is a theme in use. So if there is a theme in use the last defaultDomain wins and the domain for the module is no longer a default in `Gettext::Translator`.

This means that with the current setup a theme would have to import the translations for a module into its
own repo and change the domain of those translations to the domain of the theme. I found this by trying to expand the login twig file for the multiauth module in my own custom but very much based on the "fancytheme" module/themne directions from the website.

I think in the past this may have been unnoticed as the domains from the binary mo files was ignored and so a theme could use a module translation without any explicit action on domains.

The PR might not be the cleanest solution to the problem. But it does work in an A/B test when theming a multiauth login page.

The update allows Translate to collect more than one default domain in `addDefaultDomain` and use many defaults in translateSingularGettext to find a translation. So if the theme doesn't provide a translation in it's own theme domain then, in the specific example above, the `X-Domain: multiauth` domain is then tried.

tvdijen (approved):

monkeyiq:
squash + merged and cherry picked. I squashed as the separate style commit doesn't really add semantics to the history.

PR #2181, opened 2024-07-18
Allow custom ErrorCodes subclasses through Login::handleLogin

monkeyiq:
The goal is to allow a custom ErrorCodes subclasses to be used and have Login keep that association and be able to present information from those custom error codes on the web page.

This is an update to
https://github.com/simplesamlphp/simplesamlphp/pull/2142

This relates to and should resolve issue
https://github.com/simplesamlphp/simplesamlphp/issues/2133

The ErrorCodes subclass name needs to be saved in the state information in order to set data['errorcodes'] to the custom error code class later in the login method.

To restrict which classes can be used this way
Login::registerErrorCodeClass() explicitly tracks each errorcodes subclass. This is automatic in the constructor of ErrorCodes but it does mean that any subclass will need to be instantiated at least once before use or an explicit call to Login::registerErrorCodeClass made.

To test this I created a new authsource module in
modules/testsauthsource. Perhaps there are better ways or places to do this. The tests/modules/core/src/Controller/LoginTest.php enables this module and in setupPrivateAuthSource() it sets up the environment to allow a subsequent call to loginuserpass to execute the custom testing auth source with a supplied configuration.

I noticed that testClearDiscoChoicesReturnToDisallowedUrlRejected causes strange things to happen when running the test suite locally with the following command:

 ./vendor/bin/phpunit --filter LoginTest

Although it is not strictly Module related I have put the resolveNonModuleClass() method in the module class for now to compliment the resolveClass() method. One needs the have the NonModule version to allow inline classes and the like to resolve as they contain the class name followed by a colon and number/offset information. That colon will make the resolveClass() assume we are dealing with a module:class specification.


monkeyiq:
I might play with another PR to see what is triggering the "CI / Security checks" issue here.

monkeyiq:
Seems happy now.

monkeyiq:
I think I see the issue now, the new modules/testsauthsource should be in a different project and installed during setup of the CI. One of those things I sort of left out until I got things working ok locally.

monkeyiq:
I think this will be in reasonable shape once the new testing module is split into another project which is pulled during CI setup. As my original testing was done by playing around in the example password login it seemed logical to include a new login path that could throw normal and custom error codes and use that in the test suite. The PR does this locally with the below but that module is not found on the remote CI env. It is likely much cleaner for that authsource to be a different repo on github anyway as it is only useful to the testing. I guess the testauthsource module could also live in the simplesamlphp-test-framework repo.

```
./vendor/bin/phpunit --filter LoginTest
```

monkeyiq:
I might start adding some of this to a new PR in increments and try to see what is breaking the other CI stuff

monkeyiq:
I will create a new PR later and merge the 2 new tests into the test suite at that stage. The core code has been merged.

PR #2184, opened 2024-07-25
allow phpstan/extension-installer

monkeyiq:
Hopefully this will get the CI system happy again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -116,7 +116,8 @@
             "composer/package-versions-deprecated": true,
             "simplesamlphp/composer-module-installer": true,
             "muglug/package-versions-56": true,
-            "dealerdirect/phpcodesniffer-composer-installer": true
+            "dealerdirect/phpcodesniffer-composer-installer": true,
+            "phpstan/extension-installer": true
         }
     },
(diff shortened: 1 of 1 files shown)


monkeyiq:
I think the phpstan/extension-installer is being pulled in at this line...

https://github.com/simplesamlphp/simplesamlphp/blob/7cb00541c9ed925bbfe4b86fef48334ef60c7b00/.github/workflows/php.yml#L142

monkeyiq:
It seems to be from simplesamlphp-test-framework-1.7.9 / require / phpstan/extension-installer

tvdijen:
Yes, this needs to happen for all of our modules too.
I added the phpstan mockery-extension so it knows what to do with mocked objects

monkeyiq:
It seems 2.2 and master are using an older version of simplesamlphp/simplesamlphp-test-framework in their composer.json. Maybe I should cherry pick this into master tomorrow so it is ready when the test framework is updated.

tvdijen:
Yes, only master should be fine

monkeyiq:
seems already picked

PR #2185, opened 2024-07-25
In Module::resolveClass throw a different error for class not found

monkeyiq:
This might help to debug when a class is renamed or moved to a new module.

This is a follow on to the thread "Unhandled exeception with v2.2.2" on google groups from today. I put the exists test inside the subclass one so it only runs when things are failing. 

Looking at the other side we should consider if we want this because although it is handy for debugging it does potentially allow probing for classes to see if they exist and perhaps leak information about what version is installed or what modules are available.

### DIFF
--- a/src/SimpleSAML/Module.php
+++ b/src/SimpleSAML/Module.php
@@ -455,9 +455,9 @@ class Module
             $className = 'SimpleSAML\\Module\\' . $tmp[0] . $type . $tmp[1];
         }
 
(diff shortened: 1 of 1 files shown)


tvdijen:
I would lift this up one level and make it an if-elseif

monkeyiq:
I lifted the test up a level for general use. I left it as just an if() as the exception is a fairly small line and it avoids needing to worry about elseif lines for normal code flow.

PR #2190, opened 2024-07-31
Allow custom ErrorCodes subclasses through Login::handleLogin v2

monkeyiq:
I am building this up from a fresh branch to see where the CI system becomes unhappy.

It should eventually have the same changes as
https://github.com/simplesamlphp/simplesamlphp/pull/2181

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -84,7 +84,8 @@
         "symfony/var-exporter": "^6.4",
         "symfony/yaml": "^6.4",
         "twig/intl-extra": "^3.7",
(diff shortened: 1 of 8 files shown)


monkeyiq:
I think the actual new test methods are never called. The module debug shows that the CI doesn't think it can find the testauthsource module, so it shouldn't work if it is running.

monkeyiq:
When `testClearDiscoChoicesReturnToDisallowedUrlRejected` is enabled then `testLoginTestAuthSourceCustomError` is never called.

PR #2192, opened 2024-07-31
Allow custom ErrorCodes subclasses through Login::handleLogin no test suite version

monkeyiq:
This is a version of https://github.com/simplesamlphp/simplesamlphp/pull/2181 without the test suite update.

I found a few side issues while trying to get the test suite update going so thought I would split out the code code here for consideration.

The goal is to allow a custom ErrorCodes subclasses to be used and have Login keep that association and be able to present information from those custom error codes on the web page.

This is an update to https://github.com/simplesamlphp/simplesamlphp/pull/2181

This relates to and should resolve issue https://github.com/simplesamlphp/simplesamlphp/issues/2133

The ErrorCodes subclass name needs to be saved in the state information in order to set data['errorcodes'] to the custom error code class later in the login method.

To restrict which classes can be used this way
Login::registerErrorCodeClass() explicitly tracks each errorcodes subclass. This is automatic in the constructor of ErrorCodes but it does mean that any subclass will need to be instantiated at least once before use or an explicit call to Login::registerErrorCodeClass made.

Although it is not strictly Module related I have put the resolveNonModuleClass() method in the module class for now to compliment the resolveClass() method. One needs the have the NonModule version to allow inline classes and the like to resolve as they contain the class name followed by a colon and number/offset information. That colon will make the resolveClass() assume we are dealing with a module:class specification.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -23,7 +23,7 @@ jobs:
           fetch-depth: 0
 
       - name: Lint Code Base
-        uses: super-linter/super-linter/slim@v6
+        uses: super-linter/super-linter/slim@v6.7.0
         env:
           # To report GitHub Actions status checks
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/modules/core/src/Controller/Login.php
+++ b/modules/core/src/Controller/Login.php
@@ -38,6 +38,12 @@ class Login
      */
     protected $authState = Auth\State::class;
 
+    /**
+     * These are all the subclass instances of ErrorCodes which have been created
+     */
+    protected static array $registeredErrorCodeClasses = [];
+
+
 
     /**
      * Controller constructor.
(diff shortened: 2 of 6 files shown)


tvdijen:
I don't agree with the `resolveNonModuleClass` .. We don't need it.

monkeyiq:
Cool, I'll update to move the body of resolveNonModuleClass to the one place that calls it and move to those `use` statements tomorrow.

tvdijen:
But why is it used in the first place?
If you want to extend our software, you have to do it through a custom module.. period.

monkeyiq:
Hmm the js linter is now unhappy "ESLint couldn't find the config "json:recommended" to extend from"

monkeyiq:
Looking at an A/B with the last PR (or push to this PR) that linter worked on it was using super-linter/suepr-linter:slim-v6.7.0 and the latest run uses super-linter/suepr-linter:slim-v6.8.0.

tvdijen:
I was able to fix half of the super-linter issue..  The eslint-config is now good again, but it still breaks on the stylelint-config and I cannot figure out why..

However, while I was working on this, I noticed that your unit test in this PR is breaking phpunit (it just kills phpunit) and therefore the coverage-job is also failing... Can you try to fix that?

monkeyiq:
This PR doesn't include the unit tests for the custom error codes. I changed the super linter to the explicit 6.7.0 version and the only tests change is the access for a method in `Error/ErrorTest.php` going from protected to public so that it can be used in the Login code.

While I was testing I noticed that if `testClearDiscoChoicesReturnToDisallowedUrlRejected()` in `LoginTest.php` is enabled then it seems to halt the rest of the tests in that file. Perhaps this method also stops the phpunit from completing?

tvdijen:
My bad.. It seems https://github.com/simplesamlphp/simplesamlphp/commit/f469bd27adebff59c11dc00f4bde2c46d42d9498 is where the tests started to fail.. Will have to investigate this.

PR #2200, opened 2024-08-09
Fix MDQ cache file bugs

### LINKED ISSUE #2199: MDQ: Metadata cache broken
# **Help us to keep this issue-tracker clean!  For questions or support, please refer to our [mailing lists](https://simplesamlphp.org/lists)**

**Specifics of your environment**
1. Are you acting as SP/IdP/proxy? **both**
3. SimpleSAMLphp: What version are you using? **2.2.2**
4. PHP: What version are you using? **8.2**
5. Platform: unix or Windows? **Linux**
6. Webserver: Apache/Nginx/ISS? **Apache**

**Describe the bug**
When the MDQ class calls `json_decode` on the cached file, it returns an object not an array. A few lines down, a check for `!is_array` causes the cached data to be considered invalid, and the metadata is re-queried from the MDQ source on every request.

This issue is compounded by the `writeToCache` function appending to the cache file instead of replacing the content. Since a JSON file cannot simply be appended to, this results in a corrupt file that grows endlessly until finally `$file->getContent` exceeds the PHP memory limit and the application crashes.

**To Reproduce**
1. Setup an SP for testing
2. Setup an MDQ metadata source
3. Login to an IdP that needs metadata from the MDQ source
4. Look at the cache file (it's valid at this point)
5. Login to the IdP again
6. Look at the cache file (it is now invalid JSON, with two instances of the metadata appended one after the other)
7. Notice in the logs that the cache file was considered invalid because it wasn't an array

**Expected behavior**
MDQ metadata is read from and written to a cache file correctly.

**Screenshots or logs**
```
Aug 8 17:50:24 simplesamlphp ERROR [a3b0c82f69] SimpleSAML\Metadata\Sources\MDQ: Cached metadata from "/REDACTED_PATH/mdq/saml20-idp-remote-587674389711beef0bcee98f4fdf32d5469965b9.cached.xml" wasn't an array.
Aug 8 17:52:43 simplesamlphp ERROR [a3b0c82f69] SimpleSAML\Metadata\Sources\MDQ: Cached metadata from "/REDACTED_PATH/mdq/saml20-idp-remote-587674389711beef0bcee98f4fdf32d5469965b9.cached.xml" wasn't an array.
```

**Additional context**
I will provide a pull request with fixes for this bug soon.

m0n1ker:
Fixes #2199 

* The MDQ cache file is now explicitly decoded as an associative array to match subsequent code expectations.
* The MDQ cache file is overwritten instead of appended to, to prevent an ever-growing corrupt JSON file.
* The cache file extension was updated to `.json` instead of `.xml` to better reflect the content of the file.
  * If this change isn't wanted, please let me know and I can revert it.
  * I believe this change will also help anyone impacted by the growing cache file since the application will immediately stop trying to read a large file into memory.

The steps to reproduce from the linked issue verifies that the application is no longer re-querying MDQ metadata on every request, and the `<FILE> wasn't an array` errors are no longer appearing in the logs.

### DIFF
--- a/src/SimpleSAML/Metadata/Sources/MDQ.php
+++ b/src/SimpleSAML/Metadata/Sources/MDQ.php
@@ -130,7 +130,7 @@ class MDQ extends MetaDataStorageSource
         }
 
         $cachekey = sha1($entityId);
-        return $this->cacheDir . '/' . $set . '-' . $cachekey . '.cached.xml';
+        return $this->cacheDir . '/' . $set . '-' . $cachekey . '.cached.json';
     }
 
 
--- a/src/SimpleSAML/Metadata/Sources/MDQ.php
+++ b/src/SimpleSAML/Metadata/Sources/MDQ.php
@@ -183,7 +183,8 @@ class MDQ extends MetaDataStorageSource
             ));
         }
 
-        $data = json_decode($rawData);
+        // ensure json is decoded as an associative array not an object
+        $data = json_decode($rawData, true);
         if ($data === false) {
             throw new Exception(
                 sprintf('%s: error unserializing cached data from file "%s".', __CLASS__, strval($file)),
--- a/src/SimpleSAML/Metadata/Sources/MDQ.php
+++ b/src/SimpleSAML/Metadata/Sources/MDQ.php
@@ -218,8 +218,8 @@ class MDQ extends MetaDataStorageSource
 
         Logger::debug(sprintf('%s: Writing cache [%s] => [%s]', __CLASS__, $entityId, $cacheFileName));
 
-        /** @psalm-suppress TooManyArguments */
-        $this->fileSystem->appendToFile($cacheFileName, json_encode($data), true);
+        // using dumpFile instead of appendToFile here to replace any existing cache data
+        $this->fileSystem->dumpFile($cacheFileName, json_encode($data));
     }
 
 


monkeyiq:
This looks good. I think the `json_decode` part would block the code as mentioned. 

I will have to setup an MDQ environment to have a play around with things before I do a merge.

monkeyiq:
I notice that the original PR was using `serialize`/`unserialize` which got updated to using json at https://github.com/simplesamlphp/simplesamlphp/commit/ce43aa901a5f84ae490694d6ad2df798a33b8c17

I thought maybe that the move to json might have left something behind. For example that two serialize() calls could be appended to the same file and a single `unserialize` could read them both back as a single object.

That is not the case. The appending of serialized data was always a bug it seems.
```
<?php

$fn = "/tmp/datafile.txt";

$d["hi"] = "there";
file_put_contents( $fn, serialize($d), FILE_APPEND );
$d["new"] = "thing";
file_put_contents( $fn, serialize($d), FILE_APPEND );
//file_put_contents( $fn, serialize($d) );


$data = file_get_contents( $fn );
echo "data: $data \n";
$d2 = unserialize($data);
echo "--- original ----------------\n";
var_dump($d);
echo "--- reread   ----------------\n";
var_dump($d2);
```
Output contains
```
PHP Warning:  unserialize(): Extra data starting at offset 49 of 125 bytes in /.../src/php/serializertest.php on line 14
--- original ----------------
array(2) {
  ["hi"]=>
  string(5) "there"
  ["new"]=>
  string(5) "thing"
}
--- reread   ----------------
array(2) {
  ["hi"]=>
  string(5) "there"
  ["new"]=>
  string(5) "thing"
}

```

The example code does return an array, it is just that any appended data is ignored.

monkeyiq:
The code is the same in 2.2 so it a candidate for a cherry pick

src/SimpleSAML/Metadata/Sources/MDQ.php
```diff
@@ -183,7 +183,8 @@ private function getFromCache(string $set, string $entityId): ?array
             ));
         }
 
-        $data = json_decode($rawData);
+        // ensure json is decoded as an associative array not an object
+        $data = json_decode($rawData, true);
         if ($data === false) {
```
thijskh, line 188: Maybe replace this block with the `JSON_THROW_ON_ERROR` flag added to json_decode.

We could consider adding our own json util function that does 'correctly' process any json and handle errors as we use json_decode in several places around the code. This is not in scope for this PR though, it can be merged as it is now..

src/SimpleSAML/Metadata/Sources/MDQ.php
```diff
@@ -183,7 +183,8 @@ private function getFromCache(string $set, string $entityId): ?array
             ));
         }
```
thijskh, line 185: Boyscout: above is several exception handling about the file not existing or the read failing. This can be removed since the Symfony FIle class handles that already.

thijskh (approved):
Seems like very useful changes. You can tell that MDQ is not used a lot yet with SSP. I think the append does not make sense at all also from a functional perspective (there's only one latest version of any specific entityid).

src/SimpleSAML/Metadata/Sources/MDQ.php
```diff
@@ -217,8 +218,8 @@ private function writeToCache(string $set, string $entityId, array $data): void
 
         Logger::debug(sprintf('%s: Writing cache [%s] => [%s]', __CLASS__, $entityId, $cacheFileName));
 
-        /** @psalm-suppress TooManyArguments */
-        $this->fileSystem->appendToFile($cacheFileName, json_encode($data), true);
+        // using dumpFile instead of appendToFile here to replace any existing cache data
```
thijskh, line 221: Not sure this comment means anything when the previous code is gone. Probably it belongs in the commit message.

src/SimpleSAML/Metadata/Sources/MDQ.php
```diff
@@ -217,8 +218,8 @@ private function writeToCache(string $set, string $entityId, array $data): void
...
+        $this->fileSystem->dumpFile($cacheFileName, json_encode($data));
```
thijskh, line 222: You could add the JSON_THROW_ON_ERROR json constant here as well. If for ever reason `$data` can not be properly encoded, we throw an exception instead of stubling on with writing an empty cacheFile which will fail reading lateron.

thijskh (approved):

monkeyiq:
I like the idea of separated json encode/decode functions. Perhaps to live in src/SimpleSAML/Utils/JSON.php. The less calling code has to remember to add `JSON_THROW_ON_ERROR` around the place the more chance there is of issues being caught by the `JSON::json_decode` method.

It seems like we could merge this now and then update the comment that was above `appendToFile` and look at the THROW_ON_ERROR or just start moving to new Util json functions directly?

tvdijen:
> eems like we could merge this now and then upda

tvdijen:
> I like the idea of separated json encode/decode functions. Perhaps to live in src/SimpleSAML/Utils/JSON.php.

Can we first investigate if `symfony/serializer` can do this for us, instead of introducing a new utility-class?

monkeyiq:
I agree with investigating `symfony/serializer`. In this case we don't really even need it to be JSON for the cache file and the more we can lean into symfony the better.

monkeyiq:
@m0n1ker Thank you for the issue clearly describing the problem and PR!

monkeyiq:
I have picked this back to 2.3 and 2.2 as well.

PR #2203, opened 2024-08-13
An update to the MDQ PR #2200

monkeyiq:
As mentioned in some of the review comments for PR #2200.

This is a small PR to update a comment and sprinkle around `JSON_THROW_ON_ERROR`.

### DIFF
--- a/src/SimpleSAML/Metadata/Sources/MDQ.php
+++ b/src/SimpleSAML/Metadata/Sources/MDQ.php
@@ -187,7 +187,7 @@ class MDQ extends MetaDataStorageSource
         }
 
         // ensure json is decoded as an associative array not an object
-        $data = json_decode($rawData, true);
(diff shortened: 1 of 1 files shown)


thijskh:
If you add this you can remove the explicit exception checking since it will never be run now.

monkeyiq:
A good point :)

monkeyiq:
I was digging into other things that might go off the rails in the json encode/decode. It seems if you have numbers with reasonable decimal places then rounding can occur on the way to JSON.  JSON_NUMERIC_CHECK tries to convert things that appear as numbers into them in JSON but there is no inverse, ie to quote php numbers into strings to completely capture them.

It is a reasonable amount of decimal places to trigger this, `55.6666666666666666` is rounded as a number but not a string.

This might be not so interesting if we move to Symphony serialization on this. And also since we don't actually care if it is json here anyway.

PR #2204, opened 2024-08-13
Bugfix: Continue unencrypted if no encryption key is available

tvdijen:
See:  https://groups.google.com/g/simplesamlphp/c/PVX0f99OJqg

We can set `assertion.encrypted` all we want on our IdP-side, but if the SP doesn't provide a key it's useless. Instead of failing the hard way, see if the SP is willing to accept an unencrypted assertion.

Alternative: add another configuration flag that enforces the use of encryption. This could be a more generic flag to enforce SAML2INT compliance.

### DIFF
--- a/modules/saml/src/IdP/SAML2.php
+++ b/modules/saml/src/IdP/SAML2.php
@@ -1453,11 +1453,8 @@ class SAML2
                 $key = new XMLSecurityKey(XMLSecurityKey::RSA_OAEP_MGF1P, ['type' => 'public']);
                 $key->loadKey($pemKey);
             } else {
-                throw new Error\ConfigurationError(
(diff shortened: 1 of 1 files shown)


thijskh:
I'm not such a fan of silent fallthrough to unencrypted. If the user explicitly set encryption on in the config, it would violate the principle of least surprise that attribute values are revealed if metadata is missing a required cert. I think the fail safe approach is that if encryption is explicitly requested but not possible, to give a clear error and not _not_ encrypt things the user wanted encrypted.

The suggested optionalEncryption approach in the linked mail thread sounds like a much safer approach and keeps the ability for admins to indicate whether they always want their data to be encrypted, or if they just want opportunistic encryption.

PR #2207, opened 2024-08-19
Bugfix/preloaded resource not used

tvdijen:
Fixes console warning in Edge+Chrome:

The resource https://example.org/assets/base/fonts/fa-solid-900.woff2?tag=0656d was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.

I'm not sure why Firefox is still complaining, but at least this is an improvement.

### DIFF
--- a/src/SimpleSAML/XHTML/Template.php
+++ b/src/SimpleSAML/XHTML/Template.php
@@ -192,7 +192,7 @@ class Template extends Response
             // don't be too harsh if an asset is missing, just pretend it's there...
             return $path;
         }
-        
+
         if ($tag === false) {
             // The asset is requested without a tag
             return $path;
--- a/src/SimpleSAML/XHTML/Template.php
+++ b/src/SimpleSAML/XHTML/Template.php
@@ -191,7 +191,10 @@ class Template extends Response
         if (!$this->fileSystem->exists($file)) {
             // don't be too harsh if an asset is missing, just pretend it's there...
             return $path;
-        } elseif ($tag === false) {
+        }
+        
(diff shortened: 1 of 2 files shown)


monkeyiq:
It looks like the `stylesheet.css` is referencing `src:url("../fonts//fa-solid-900.woff2")` I guess the tag version and non tag version are seen as two assets so Chrome is complaining about the wasted load of the tag version.

monkeyiq:
Without the tag I see only one line in the browser network traffic for the woff. The console message in Firefox also disappears for me with the no tag (this PR) version installed. Firefox 127.0.2 (64-bit).

monkeyiq:
This comment is more a matter of personal style. Feel free to ignore. 

I would just use an `if` instead of `elseif` for the `$tag` check. Where an `if()` has a short `return` path I like to assume that it has knocked out the case it is intending to act on and not bother with an else around something that has already returned. 

Here the file exists and `$tag` check as fairly unrelated, then the rest of the function is just in the block and not in a final `else`.

monkeyiq:
I might also put the comment "preloading must be done after referencing the resource" in the base twig.  It might not be obvious to a reader that the preload line needs to remain at that location.

tvdijen:
> It looks like the `stylesheet.css` is referencing `src:url("../fonts//fa-solid-900.woff2")` I guess the tag version and non tag version are seen as two assets [..].

It's both the tag and the duplicate forward-slash, which I have also removed in an update of the `simplesamlphp-assets-base` package.

PR #2208, opened 2024-08-21
Add the possibility to continue if no encryption cert is available

tvdijen:
Continuation of #2204

### DIFF
--- a/docs/simplesamlphp-reference-idp-hosted.md
+++ b/docs/simplesamlphp-reference-idp-hosted.md
@@ -142,9 +142,12 @@ The following SAML 2.0 options are available:
 
 `assertion.encryption`
 :   Whether assertions sent from this IdP should be encrypted. The default
-    value is `FALSE`.
+    value is `FALSE`. When set to `TRUE` encryption will be enforced for all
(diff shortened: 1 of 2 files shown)


docs/simplesamlphp-reference-idp-hosted.md
```diff
@@ -190,6 +190,10 @@ The following SAML 2.0 options are available:
     any value in the SP-remote metadata overrides the one configured
     in the IdP metadata.
 
+`encryption.optional`
+:   Whether or not we may continue to sent an unencrypted assertion if the SP has no encryption certificate.
```
monkeyiq, line 194: s/sent/send

monkeyiq:
Would it make sense to mention `encryption.optional` from the `assertion.encryption` documentation in a Note that encryption may optionally, not done by default, be lifted for an SP that does not supply a certificate? 

I imagine the person who is setting up a config will notice all of these things but somebody who inherits the maintenance of an SSP install and is reading the docs might be in a better position to know about `encryption.optional` if it is mentioned from the main `assertion.encryption` documentation.

monkeyiq (approved):

PR #2216, opened 2024-08-26
i18n in 2.3 allow Brazilian to be selected again

monkeyiq:
Changing this line allows Portuguese (Brazil) to be selectable from the language menu again in 2.3.

Note that setLanguageCookie() and getLanguageCookie() also contain forced lower casing which is worth investigating too in order to make sure things do not fall through the cracks there.

### DIFF
--- a/src/SimpleSAML/Locale/Language.php
+++ b/src/SimpleSAML/Locale/Language.php
@@ -224,17 +224,6 @@ class Language
      */
     public function setLanguage(string $language, bool $setLanguageCookie = true): void
     {
-        $language = strtolower($language);
-
-        // @deprecated support these older names for now.
-        // @deprecated - remove entire if-block in a new major release
-        if ($language == 'pt_br') {
-            $language = 'pt_BR';
-        }
-        if ($language == 'zh_tw') {
-            $language = 'zh_TW';
-        }
-
         if (in_array($language, $this->availableLanguages, true)) {
             $this->language = $language;
             if ($setLanguageCookie === true) {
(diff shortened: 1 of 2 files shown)


tvdijen:
Thanks @monkeyiq for diving into this.. I think in this case we could just remove the strotolower-line..

monkeyiq:
That CI test that is failing is checking if language.parameter.name value is transformed to lower case. So if folks were using things like `'Es'` to select a language then that will no longer work.

tvdijen:
I guess that's OK.. I really doubt people are adding values to the list, they only remove what is not needed.

tvdijen (approved):

tvdijen:
I wonder if the Brazilian translations ever worked, because it is now just using `pt`..
It's being reduced from `pt_BR` to `to` in Localization::getLangPath..

monkeyiq:
I see what you mean in `getLangPath` with the `explode('_', $this->langcode);`. I suspect in this case it is still working because going from `pt_BR` to `pt` still gives Portuguese. 

I think in this case we might want to rename the pt-br directory to pt_BR and try that full path first before falling back to the first two chars only. IIRC there are locale choices where the language is the same but the customs of the locale might want a different translation or other arangement.

tvdijen:
Yeah, I did that (renaming pt-br directory to pt_BR and test that one before falling back to `pt`) and that gave me completely different translations.. I'll create yet another PR to fix this too, and remain backwards compatible.

Can you see if you can restore BC on the admin-password issue? Maybe I can release 2.3.1 by the end of this week then

tvdijen:
This should not have been merged with broken tests by the way..

Issue #2212, opened 2024-08-22
Missing mention to breaking changes in 2.3 upgrade notes


jaakristioja:
**Is your feature request related to a problem? Please describe.**

The [Upgrade notes for SimpleSAMLphp 2.3](https://github.com/simplesamlphp/simplesamlphp/blob/74c4130659d1574f44aa2606090a8d87ffc95ebb/docs/simplesamlphp-upgrade-notes-2.3.md) fail to mention a breaking change: 
  * The `auth.adminpassword` configuration value no longer allows raw passwords (see [FAQ](https://github.com/simplesamlphp/simplesamlphp/wiki/Frequently-Asked-Questions-(FAQ)#failed-to-login-to-the-admin-page-with-and-error-message-admin-password-not-set-to-a-hashed-value)).

This caused admins being locked out after upgrade. I don't remember having seen any prior warning about this.

**Describe the solution you'd like**
Describe this breaking change in the Upgrade notes.

**Describe alternatives you've considered**
Allow raw passwords again, but as a deprecated feature. And warn users/admins if raw passwords are used.

**Additional context**
Screenshot of surprise waiting admins who wish to login after upgrade:
![screenshot](https://github.com/user-attachments/assets/a5f7e4b4-91f1-4546-9efe-52f2e40d9650)

jaakristioja:
The documentation is also incorrect about `auth.adminpassword`:

https://github.com/simplesamlphp/simplesamlphp/blob/74c4130659d1574f44aa2606090a8d87ffc95ebb/docs/simplesamlphp-install.md#L223-L231

tvdijen:
@monkeyiq [This](https://github.com/simplesamlphp/simplesamlphp/pull/2113) shouldn't have ended up in a minor release

tvdijen:
99c5458da should cover your remarks. @thijskh do you see any added value in the reference to authcrypt, or the authcrypt module in general?

PR #2222, opened 2024-08-27
v23: allow admin password to be stored in config.php without hashing …

monkeyiq:
Given that this touches the admin login route directly, some more eyes to review are appreciated.

The code uses the result of `password_get_info` and if the `algo` is null then we know the `auth.adminpassword` from config.php is not set to a hashed password. At that stage the new lines do a direct string compare in the same way that `pwValid` does in 2.2 and used to in 2.3. More explicitly:
 https://github.com/simplesamlphp/simplesamlphp/blob/a9dc9579774e4aea4f147d11aee87b8030e3b0d5/src/SimpleSAML/Utils/Crypto.php#L379

This is the intent of the code. Admins can again put plaintext passwords in config.php for auth.adminpassword if desired. Hashing is recommended but not required any more and functionality returns to that of v2.2.

We can see a null for the password for example in the following test
```
php > var_dump(password_get_info("123"));
array(3) {
  ["algo"]=>
  NULL
  ["algoName"]=>
  string(7) "unknown"
  ["options"]=>
  array(0) {
  }
}

```

This was requested here https://github.com/simplesamlphp/simplesamlphp/pull/2216#issuecomment-2311931730

More information about the changes around the admin password hashing and how we got where we are in 2.3 is at https://github.com/simplesamlphp/simplesamlphp/issues/2212#issuecomment-2305550977

### DIFF
--- a/modules/core/src/Auth/Source/AdminPassword.php
+++ b/modules/core/src/Auth/Source/AdminPassword.php
@@ -61,6 +61,12 @@ class AdminPassword extends UserPassBase
 
         $pwinfo = password_get_info($adminPassword);
         if ($pwinfo['algo'] === null) {
+            // @deprecated: remove this in the future.
+            // Continue to allow admin login when the config contains
+            // a password that is not hashed
+            if ($adminPassword === $password) {
+                return ['user' => ['admin']];
+            }
             throw new Error\Error(Error\ErrorCodes::ADMINNOTHASHED);
         }
 


tvdijen (commented):
This looks good @monkeyiq !
I would like to add a deprecation warning to the logs when the password is plain-text, and the upgrade notes need a touch up.

thijskh:
Would it be better to use `hash_equals()` instead of `===`?

monkeyiq:
I was thinking about using something like `hash_equals`. The first PR is using the same thing that 2.2 uses to do the comparison. I think if we want to improve that then we probably want to move to `hash_equals` in 2.2, 2.3+ as a logical unit.

monkeyiq:
The deprecation warning is a great idea. It is much better to continue to work and offer a suggestion to improve security and at some stage remove the old acceptance of plain password from config.php.

monkeyiq:
I think I'll make the deprecation warning and upgrade notes PR separate linking back to this.

monkeyiq:
I was about the cherry pick this forward to master but I am not sure we want it that far out?

monkeyiq:
I picked it into 2.4

tvdijen:
No, master is where we break stuff and force hashed passwords!

monkeyiq:
sounds good!

monkeyiq:
ok, I'll take a look at another PR for deprecation warning and release notes update.

PR #2227, opened 2024-09-01
Disconnect ADFS-module from SimpleSAMLphp

tvdijen:
This particular code was used by the ADFS-module to generate metadata.
The new adfs-module (currently in dev-feature/library) now has it's own metadata-builder and uses the low-level ws-security library to build the SecurityTokenServiceType-element and it's content.

Since this is fully backwards compatible once I tag a new major of the module (people can still use the adfs metadata-endpoint and get the same result), this should be fine to merge in the current release.


src/SimpleSAML/Metadata/SAMLBuilder.php
```diff
@@ -123,28 +122,6 @@ public function getEntityDescriptorText(bool $formatted = true): string
     }
 
 
-    /**
-     * Add a SecurityTokenServiceType for ADFS metadata.
-     *
-     * @param array $metadata The metadata with the information about the SecurityTokenServiceType.
-     */
-    public function addSecurityTokenServiceType(array $metadata): void
```
monkeyiq, line 131: I think there may still be a call to this here
https://github.com/simplesamlphp/simplesamlphp/blob/2509a7abc3e7dc54d1dc02d4d9028732cd15b8fb/modules/admin/src/Controller/Federation.php#L287

tvdijen: Ah, good catch! I think that entire if-block for adfs-entities can also go now.
If we want to display them in the admin module, the adfs-module should generate them and provide them using a hook

monkeyiq: I resurrected my ADFS configuration a little bit. I am still not convinced that the setup I have would work in production. 

The I comment out the `addSecurityTokenServiceType`  in Federation I can see the federation page again. Though there is a reasonably blank "In SAML 2.0 Metadata XML format:" in the web page.

tvdijen:
Hi Ben!  I'm away for a couple of days. Will get back to this ASAP.

monkeyiq:
Hi tim, no worries. Enjoy your away time (hopefully) and get back to it when it happens. No need for ASAP :)

tvdijen:
@monkeyiq https://idp.moo-archive.nl/module.php/adfs/metadata

tvdijen:
Ah wait, you want to test authentication?
You can use https://engine.openconext.moo-archive.nl/authentication/sp/debug , select ADFS, select "ADFS-module test" and login using 'student:studentpass'.. You will end up with an exception, because I messed up my ADFS-install, but you should see a perfectly valid response going from the ADFS-module towards ADFS. I've been successfully using this at work with an Azure-tenant connected to SSP+adfs module.

monkeyiq:
This looks ok to me. It obviously helps that the PR is coming from a core dev :)

The caveat is that I am not really up to speed with ADFS itself and would really need to setup a docker/VM local image to auth against to do rigorous testing on it. The roads I have seen along those lines are not as straightforward as those that lead me to being able to test sql server code. Even then I might run into the issue that my configuration might be bogus and the code doing the right thing.

Since the composer file no longer pulls in the adfs module we should probably mention in the docs/simplesamlphp-changelog that folks using ADFS should pull the adfs module with release $ver or above when upgrading to SSP 2.3.3+?

tvdijen:
Yes, we should mention it. I did add the adfs-module to the full-build, so this change would only affect people who manually checkout.

Setting up ADFS is maybe not necessary at all, because what the adfs-module is mimicking here really is just a WS-federation claims provider. To test this, you would need a dummy WS-federation relying party demo-app.  Maybe this is something you can use?

https://is.docs.wso2.com/en/latest/get-started/try-samples/ws-federation-webapp/

monkeyiq:
Over the weekend I created a setup with SSP 2.3 running as an IdP with the ADFS module. From my understanding, I think this provides the WS-Fed "Security Token Service" side of things.

I mostly got the PassiveSTSSampleApp to authenticate with that SSP-IdP. There is an exception post auth in the tomcat app but that is post auth and I can try to look into that later. The issue was with the DOMImpl and setXMLEncoding so I think it relates to something the tomcat app wants or a version mismatch in jar files somewhere.

I also had to tinker with my `ADFS/sendResponse` to set `$nameid = $randomUtils->generateID();` inside `if (!empty($nameidattribute)) {` which I assume is because I have not gotten my configuration for attribute massaging to the level that it has to be for a clean run.

This should give me some position to do A/B testing of SSP2.3+ADFS and this PR which is the immediate goal.

Thanks for the info and link to the WS-federation relying party demo-app!

tvdijen:
The Adfs-module relies on having the `simplesaml.nameidattribute` setting being set in `adfs-sp-hosted.php`.. I think that's missing and the reason you have to mess with the sendResponse-method.

monkeyiq:
I am using the simplesamlphp-module-sqlauth/PasswordVerify authsource and the uid was not in there. I have `email` and `edupersontargetedid` which I am now mapping one of those with `simplesaml.nameidattribute`. I guess the first phase of my testing was the biggest hammer to fit the peg into the hole. Now with the refinements in place things are looking good. 

I am wondering if we might put a link in the logs to the admin page to show the attributes for that auth source to help folks find and debug quicker. Something like, for example, simplesaml.nameidattribute with value `uid` not found in attributes from auth source. Please visit `https://example.com/your-ssp/admin/test/$authsource` to see available attributes and update your nameidattribute in adfs-sp-hosted.php to select one of those attributes.

Maybe subbing out the host name and your-ssp part and certainly the active authsource so that folks can see what simplesaml.nameidattribute is and what the attributes at hand might be.

monkeyiq:
With some thinkering with composer I have this PR and the matching adfs-module (currently in dev-feature/library)  installed. 

I found one difference initially. If `metadata.sign.enable = false` and keys are not set in the config.php for signing then the attempt to sign in `ADFS::sendResponse()` will result in failure. 

https://github.com/simplesamlphp/simplesamlphp-module-adfs/blob/fae939a543a1a60532dd1a33ffad235436a3cecb/src/IdP/ADFS.php#L451

The logs
```
Caused by: TypeError: SimpleSAML\XMLSecurity\Key\PrivateKey::fromFile(): Argument #2 ($passphrase) must be of type string, null given, called in /opt/sspdev/modules/adfs/src/IdP/ADFS.php on line 197
```

It may be that I am on the edge case here with metadata signing disabled. I am not sure if it is a useful configuration to continue to support but I thought I would mention the difference to 2.3. That was my main goal - to get another test of the code in and try to see if there are other changes.

monkeyiq:
As an aside digression into composer... There is probably a branch-alias type of solution to the testing of modules that relate to a PR.  I thought I would mention how I got my test install going in case my process is adding something that the release might not so we can try to reason about issues now rather than then. On the other hand, if it is working on your site it will likely work OK on release :)

To get this going I added updated my checkout of the PR from the SSP repo editing `composer.json` to add a local repository 
```
+    "repositories": [
+        {
+            "type": "path",
+            "url": "/home/me/src/simplesamlphp/adfs/adfs-feature-library/simplesamlphp-module-adfs"
+        }
+    ],    
```

and temporarily added adfs to the require to have composer pull in the right deps for me.

```
     "require": {
+        "simplesamlphp/simplesamlphp-module-adfs": "dev-feature/library"
```

In my ADFS module checkout I locally modded the SSP dep to v2.3.0. This is likely where the branch-alias type of setup would have helped.

```
-        "simplesamlphp/simplesamlphp": "dev-feature/adfs-upgrade",
+        "simplesamlphp/simplesamlphp": "v2.3.0",
```

IIRC I added the adfs module this way so that its deps are handled by composer. I only really had to tinker this way because the adfs dev-feature/library wanted SSP dev-feature/adfs-upgrade which I couldn't manage to alias to the checked out PR. It is messy and suboptimal but the info might be useful if somebody is wanting to checkout and test a PR that depends on a specific branch of a module as well.

monkeyiq:
I am happy to merge this and maybe we/I can check what needs doing to allow `metadata.sign.enable = false` again in a way that we can merge.

monkeyiq (approved):
We can check out the no metadata signing issue before or after merge.

monkeyiq:
My setup is probably a little strange. 

I have in my config.php:
```
    'metadata.sign.enable' => false,

    'metadata.sign.privatekey' => null,
    'metadata.sign.privatekey_pass' => null,
    'metadata.sign.certificate' => null,
```

I have in my adfs-idp-hosted.php metadata file
```
    'privatekey' => 'host200404.pem',
    'certificate' => 'host200404.crt',
```

With that configuration I have to add a test like this in ADFS.php
```
        if(Configuration::getInstance()->getOptionalBoolean('metadata.sign.enable',true)) {
            if ($privateKeyFile !== null && $certificateFile !== null && $algo !== null) {
                $assertion = ADFS::signAssertion($assertion, $privateKeyFile, $certificateFile, $algo, $passphrase);
                $assertion = Assertion::fromXML($assertion->toXML());
            }
        }

```

After digging into this a bit more what I think we want is for me to set the privatekey/certificate to null in the idp file and the code to load keys changes to something like
```

        $assertion = ADFS::generateAssertion($idpEntityId, $spEntityId, $nameid, $attributes, $assertionLifetime);

        $configUtils = new Utils\Config();
        $privateKeyFile = null;
        $certificateFile = null;
        $path = $idpMetadata->getOptionalString('privatekey',null);
        if( $path ) {
            $privateKeyFile = $configUtils->getCertPath($path);
        }
        $path = $idpMetadata->getOptionalString('certificate',null);
        if( $path ) {
            $certificateFile = $configUtils->getCertPath($path);
        }
        $path = null;
        
        $passphrase = $idpMetadata->getOptionalString('privatekey_pass', null);

```

Otherwise the `getString`  on 'privatekey' fails because it is set to null.  `getString` calls `getValue` which calls `hasValue`  which checks for the key and also `!is_null($this->configuration[$name])`

```
Oct 17 19:47:36 simplesamlphp ERROR [0bb4150be6] 12 /opt/sspdev/vendor/simplesamlphp/assert/src/Assert.php:389 (SimpleSAML\Assert\Assert::__callStatic)
Oct 17 19:47:36 simplesamlphp ERROR [0bb4150be6] 11 /opt/sspdev/src/SimpleSAML/Configuration.php:402 (SimpleSAML\Configuration::getValue)
Oct 17 19:47:36 simplesamlphp ERROR [0bb4150be6] 10 /opt/sspdev/src/SimpleSAML/Configuration.php:693 (SimpleSAML\Configuration::getString)
Oct 17 19:47:36 simplesamlphp ERROR [0bb4150be6] 9 /opt/sspdev/modules/adfs/src/IdP/ADFS.php:445 (SimpleSAML\Module\adfs\IdP\ADFS::sendResponse)
```

monkeyiq:
Thinking a bit more, it is a bit cleaner if the privatekey/certificate calls remain not optional but are moved into a block that is wrapped in a test if signing is desired (true by default).
```
if($idpMetadata->getOptionalBoolean('metadata.sign.enable', true)) {
   // sign it here
   ...
```

tvdijen:
I'm not sure I follow what you're doing?  What does `metadata.sign.enable` have to do with signing assertions?  Metadata-signing uses a separate key-pair from config.php and doesn't affect the authentication process.

monkeyiq:
Please ignore my comment about 'metadata.sign.enable' I was thinking about how the code might be a little simpler if it was using a config key but it was the end of a long day so the key I mentioned was not valid.

tvdijen:
Gotcha, changed two getString with getOptionalString

monkeyiq:
Hmm, `getCertPath` can not take a null value. Perhaps we can wrap the signing block with
```
if( $idpMetadata->getOptionalString('privatekey', null) !== null
  && $idpMetadata->getOptionalString('certificate', null) !== null ) {
...
}

tvdijen:
I think I properly fixed it now

monkeyiq:
Cool. Login works without key/cert here.

monkeyiq:
Should we tag a release of the module before merging?

tvdijen:
I've tagged v3.0.0-rc1 for the module and I will rebase this PR against the 2.4 release branch

monkeyiq:
I took a quick look at the PR after the rebase. It seems LanguageTest.php is being changed in the PR and src/SimpleSAML/Locale/Language.php. The former seems ok but the later seems to be removing the lower casing of the language from 2.4 so maybe something was a little funky in the rebase?

https://github.com/simplesamlphp/simplesamlphp/blob/70a08b97a51e6ac4fe60ebd5b0137f4e8bfa0f9a/src/SimpleSAML/Locale/Language.php#L233

tvdijen:
The strtolower only seems to exist in 2.4 ... Not in 2.3, not in master..  Something wasn't properly cherry-picked along the different branches here

monkeyiq:
Ah I see commit b9e38b9 above seems to be it.

tvdijen:
Right, so it was intentionally removed in the 2.3 branch, but we failed to cherry-pick it to 2.4 ..

monkeyiq:
Ah I see, they were cherry picked into feature/adfs-upgrade and are now being merged into simplesamlphp-2.4.

monkeyiq:
I am thinking the "rebase and merge" option is probably the way to go? This will keep info about the cherry picked stuff separate in the history.

monkeyiq:
I will merge this tomorrow morning.

monkeyiq:
I guess the normal merge is the second best option. The squash seems like the least desirable for this PR.

PR #2229, opened 2024-09-01
Fix deprecation of Twig spaceless-filter

### LINKED ISSUE #2228: Fix deprecation of Twig spaceless-filter
**Describe the issue**
Twig has deprecated the `spaceless` filter, which is used in three places:

templates/base.twig
templates/_table.twig
modules/admin/templates/metadata_converter.twig

The first two are not problematic to remove, but the last one exists because of a bug in clipboard.js.
We need an alternative for spaceless, or figure out the bug in clipboard.js.

tvdijen:
Closes #2228

The whitespace-issue turns out to be not a clipboard.js issue, but appears to be related to the `<pre>`-tag.

### DIFF
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -16,6 +16,11 @@ Released TBD
 `admin`
 
 * Fix metadata-viewer to output a valid PHP array
+* Fix text-overflow of the metadata-converter output-box
+
+Chores:
+
+* Solved deprecation notices in CI regarding the use of the Twig spaceless-filter (#2229)
 
 ## Version 2.3.0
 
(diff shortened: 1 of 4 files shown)


monkeyiq:
For me, the update in the PR as it is now makes the text in the dialog not break out of the code-box as much. yay!

I am not sure where the `code-box-content` is being generated from in stylesheet.css.  The contents of the code-box still runs a little too far to the right here. (In Chrome, others untested as yet). If I override the width to `unset` then the content fits nicely in the code-box.

+++ at end of stylesheet.css
```
.code-box-content {
    width: unset
}
```

monkeyiq (approved):

monkeyiq:
I guess next is working out what needs to change in the docs CI.

thijskh (approved):

tvdijen:
Fixed the docs.

@monkeyiq The slight overflow has been fixed in an update of the simplesamlphp-assets-base package

PR #2234, opened 2024-09-02
Throw user-friendly exception with the proper HTTP statuscode

### LINKED ISSUE #2233: Exceptions on HEAD and OPTIONS requests to the IDP
I have the same bug as already temporally fixed #1400 for the old 1.x version.  When a client (noticed Firefox, Chrome, Edge and Word in the logs) sends an OPTIONS request, SSP throws an unhanded Exception.

There was an temp fix ( https://github.com/simplesamlphp/simplesamlphp/commit/a7600ce477632aca33eb206be7c41056ce808e14 ) for the 1.x version. As I see in the sources for the 2.x there wasn't an check until the commit e895170 but there was added an error report SSOPARAMS so I have now errors in the log files again.

Should it be an solution to add again the first approach with with sending back an header 405 ? Or what is the idea behind the "SSOPARAMS" error ?

tvdijen:
Closes #2233

### DIFF
--- a/modules/core/templates/method_not_allowed.twig
+++ b/modules/core/templates/method_not_allowed.twig
@@ -1,6 +1,6 @@
 <p>{{ 'This is most likely a configuration problem on either the service provider or identity provider.'|trans }}</p>
 
 <ul>
-    <li>{{ 'If you are an user who received this error after following a link on a site, you should report this error to the owner of that site.'|trans }}</li>
+    <li>{{ 'If you are a user who received this error after following a link on a site, you should report this error to the owner of that site.'|trans }}</li>
     <li>{{ 'If you are a developer who is deploying a single sign-on solution, you have are trying to reach an endpoint using the wrong HTTP-method.'|trans }}</li>
 </ul>
--- a/modules/core/templates/method_not_allowed.twig
+++ b/modules/core/templates/method_not_allowed.twig
@@ -1,0 +1,6 @@
+<p>{{ 'This is most likely a configuration problem on either the service provider or identity provider.'|trans }}</p>
+
+<ul>
+    <li>{{ 'If you are an user who received this error after following a link on a site, you should report this error to the owner of that site.'|trans }}</li>
+    <li>{{ 'If you are a developer who is deploying a single sign-on solution, you have are trying to reach an endpoint using the wrong HTTP-method.'|trans }}</li>
+</ul>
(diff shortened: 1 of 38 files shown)


tvdijen:
afbeelding

tvdijen:
Note to self: should fix the debug info box overflow

modules/core/templates/method_not_allowed.twig
```diff
@@ -0,0 +1,6 @@
+<p>{{ 'This is most likely a configuration problem on either the service provider or identity provider.'|trans }}</p>
+
+<ul>
+    <li>{{ 'If you are an user who received this error after following a link on a site, you should report this error to the owner of that site.'|trans }}</li>
```
thijskh, line 4:
```suggestion
    <li>{{ 'If you are a user who received this error after following a link on a site, you should report this error to the owner of that site.'|trans }}</li>
```

thijskh (commented):
It makes it a lot friendlier, but I think the core issue remains and that is that OPTIONS and HEAD requests will still be logging exceptions and I doubt we want that. You get a lot of OPTIONS passing by from browsers etc that should just be silently responded to with a 405 but not generate a lot of logging. I think.

tvdijen:
I'm not sure we can fix that...
We can stop logging this specific error, but that kind of defeats the purpose of the `debug.backtraces = true` setting...
One possible solution could be to add a parameter to `Error::show()` to pass the desired log-level. This would allow us to upscale or downscale exceptions.. In this particular case we could downscale to debug-level to hide the tracebacks (but log-level debug would still show them)

thijskh:
I would only do this in debug mode indeed. Because it's very common for browsers etc to send OPTIONS requests which is spec compliant to do and the server just needs to respond with 405. Generating higher loglevel entries for this common and normal behaviour obscures the actual exceptions.

tvdijen:
So your OK with the changes I made last night?

thijskh:
The changes seem to be about the email form? Or am I misunderstanding it. The browsers sending legitimate OPTIONS requests will not be submitting the email form anyway...

tvdijen:
The email form would still be generated and produce output to the logs.
I've added a boolean to be able to not show the form at all for this type of error (and therefore not produce the log output)

monkeyiq (approved):
I see that in Error::show the old direct call to `logError()` is now moved to `log($logLevel)` which will dispatch the `logBacktrace()` call based on the level that is passed and the system settings.

Just a thinking out loud took a closer look comment :)

thijskh (approved):

tvdijen:
> I see that in Error::show the old direct call to `logError()` is now moved to `log($logLevel)` which will dispatch the `logBacktrace()` call based on the level that is passed and the system settings.

Yes, that's the trick to downgrade exceptions. Any exception we want to downgrade to debug-level can be dealt with this way in `ExceptionHandler`

PR #2269, opened 2024-10-16
Remove non-existing language

RoSk0:
Follow up for #2089 .

At the moment, using default config file would produce `ERROR Language "st" not installed. Check config.`


tvdijen:
Thank you for your contribution @RoSk0 , but this is not the way to fix it.
Similar to what we did with with `pt-br` and `zh-tw`, the `st` code has to be replaced with `en_LS`.

I will try and fix this (hopefully) later today, keeping this open as a reminder to myself.

monkeyiq:
I am also happy to dig into this one

monkeyiq:
I have updated this with the PR above. I also have created a new gh issue, self assigned, to track over other SSP projects looking at the locale directories there https://github.com/simplesamlphp/simplesamlphp/issues/2277

PR #2278, opened 2024-10-24
Destroy session cookies on logout

patrickharriscatalyst1:
During a security review it was found:

_It was possible to retrieve the authentication cookies from the browser and reuse this to
continue on the target website using a different browser and without any authentication.
The cookie and the SAML authentication token were set to HTTPOnly to strengthen the
security of the cookie.
It would be possible to extract the authentication details using CSRF techniques and use
these details to authenticate against the target web application. Similar behaviour would
be observed by users where the application would not request login credentials if the
expire duration has not been reached yet._

Therefore it seems the simplesamlphp cookies should be deleted on logout. I'm not sure if this cookie expiry is perhaps better done in `doLogout()`.

### DIFF
--- a/modules/saml/src/Controller/ServiceProvider.php
+++ b/modules/saml/src/Controller/ServiceProvider.php
@@ -579,7 +579,12 @@ class ServiceProvider
 
             $state = $this->authState::loadState($relayState, 'saml:slosent');
             $state['saml:sp:LogoutStatus'] = $message->getStatus();
-            return $source::completeLogout($state);
+
+            // Destroy session cookies.
+            $session = Session::getSessionFromRequest();
+            $session->updateSessionCookies(['expire' => TRUE]);
+
+	    return $source::completeLogout($state);
         } elseif ($message instanceof LogoutRequest) {
             Logger::debug('module/saml2/sp/logout: Request from ' . $idpEntityId);
             Logger::stats('saml20-idp-SLO idpinit ' . $spEntityId . ' ' . $idpEntityId);
--- a/modules/saml/src/Controller/ServiceProvider.php
+++ b/modules/saml/src/Controller/ServiceProvider.php
@@ -581,8 +581,7 @@ class ServiceProvider
             $state['saml:sp:LogoutStatus'] = $message->getStatus();
 
             // Destroy session cookies.
-            $session = Session::getSessionFromRequest();
-            $session->updateSessionCookies(['expire' => TRUE]);
+            $this->session->updateSessionCookies(['expire' => TRUE]);
 
 	        return $source::completeLogout($state);
         } elseif ($message instanceof LogoutRequest) {
--- a/modules/saml/src/Controller/ServiceProvider.php
+++ b/modules/saml/src/Controller/ServiceProvider.php
@@ -584,7 +584,7 @@ class ServiceProvider
             $session = Session::getSessionFromRequest();
             $session->updateSessionCookies(['expire' => TRUE]);
 
-	    return $source::completeLogout($state);
+	        return $source::completeLogout($state);
         } elseif ($message instanceof LogoutRequest) {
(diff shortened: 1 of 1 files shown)


monkeyiq:
I have been playing around with a single app SP and removing cookies in doLogout().  On thing I see is that `session.authtoken.cookiename` can be deleted if I add some code in there. On the other hand the `sessioncookie session.phpsession.cookiename` will be recreated during logout. So if we want that part to be really clean then we need to also have SSP avoid making a new session cookie as part of returning to the application for this case.

Doing something just in doLogout() would work for really simple apps with a 1-1 to a configured SP in SSP. Not what one wants if they have SSO as the cookies should probably stick around until SLO.

monkeyiq:
Sorry for the delay. I will have to setup an SLO environment for testing. I haven't forgotten the PR :)

monkeyiq:
I now have a basic SingleLogout environment based on `sspsmall` extended with two SPs and two demo apps.

I have verified SingleLogout use with one SP logging out the second one. 

With this patch I see when I do a SingleLogout on sp1 the `SimpleSAMLAuthToken` is deleted and disappears on the second app (sp2) when I go there as well. 

Though without the patch, the `SimpleSAMLAuthToken` cookies remain but if I single logout in sp1 and then try to use the cookies in sp2 I get a full user/pass screen in sp2. So it seems even if the cookies are outliving the logout they are not useful after logout anway?

tvdijen:
I'm not using SLO, so I'm not sure I can help you out here.. Sorry!

monkeyiq:
My main previous exposure to SSP was as in an SP use case. Though I never got around to doing SLO because I was only working on the one app. So it has been very interesting to me seeing how things work at this level with 2 SPs and an IdP.

monkeyiq:
It is a larger concern here if one can extract these cookies "using CSRF techniques" and then use an existing session elsewhere. That part seems to be separate to the cookie lifetime issue as once an SLO has happened it doesn't seem you can use that cookie for anything. Unless I am missing something. 

I do like the patch, having the cookie not floating around after the session SLO is useful for other things such as privacy.

monkeyiq:
I have just reverified this on 2.3 as an A/B test. Without the patch applied the SAMLAuthToken remains after SLO. With the patch that cookie is deleted after logout.

I don't think the token will be accepted after SLO but the PR removes it from the browser as well.

monkeyiq:
I have also picked this back into 2.4.

jaimeperez:
Hi @patrickharriscatalyst1!

Would you be able to share more details on the feedback from the auditors?

By definition, if you grab session cookies and use them in a different browser, you will still be logged in as the original user. However, protecting those cookies is a task for the specific deployment. They should be set to HTTPOnly and marked as secure to avoid their exchange via regular HTTP connections or accessible to scripts.

Furthermore, the session cookies do not need to be expired or deleted upon logout. What determines your session status is the data stored in the backend associated with your session, not your cookie. If your session has been invalidated after logout, having still the cookies doesn't allow you to authenticate again.

We need to know more about what the auditors managed to do and how, in order to figure out if this is indeed an issue in the software or a deployment issue. Therefore, would you mind sharing that with us privately?

Thanks a lot in advance!

patrickharriscatalyst1:
Hi @jaimeperez - what is the best way of contacting you? I tried the blackout email but it was not working for me.

tvdijen:
I think the best way to securely share your info with the developers-team is to go [here](https://github.com/simplesamlphp/simplesamlphp/security/advisories) and press the 'Report vulnerability' button.

jaimeperez:
> Hi @jaimeperez - what is the best way of contacting you? I tried the blackout email but it was not working for me.

In case you have issues reporting it here directly, you can also reach me at _jaimeperez_ at protonmail.

PR #2284, opened 2024-10-28
SSP Metadata Performance Improvements

ioigoume:
improve MetadataStorageHandler getMetadataForEntities:optimize and skip unnecessary loops and fetches.

### DIFF
--- a/src/SimpleSAML/Metadata/MetaDataStorageHandler.php
+++ b/src/SimpleSAML/Metadata/MetaDataStorageHandler.php
@@ -264,7 +264,6 @@ class MetaDataStorageHandler implements ClearableState
      */
     public function getMetaDataForEntities(array $entityIds, string $set): array
     {
-        $startTime = microtime(true);
         $result = [];
         // We are flipping the entityIds array in order to avoid constant iteration over it.
         // Even if it becomes smaller over time.
@@ -295,11 +294,6 @@ class MetaDataStorageHandler implements ClearableState
                 $result[$key] = $le;
             }
         }
-
-        $endTime = microtime(true);
-        $executionTime = $endTime - $startTime;
-        $formattedTime = number_format($executionTime, 6, '.', '');
-        Logger::debug(__METHOD__ . "Execution time: $formattedTime seconds");
 
         return $result;
     }
--- a/src/SimpleSAML/Metadata/MetaDataStorageHandler.php
+++ b/src/SimpleSAML/Metadata/MetaDataStorageHandler.php
@@ -274,11 +274,11 @@ class MetaDataStorageHandler implements ClearableState
 
         foreach ($this->sources as $source) {
             // Break iteration if the entityID List is empty
-            if (empty($entityIdsFlipped)) {
+            if (empty($entityIds)) {
                 break;
             }
 
-            $srcList = $source->getMetaDataForEntities($entityIdsFlipped, $set);
+            $srcList = $source->getMetaDataForEntities($entityIds, $set);
             foreach ($srcList as $key => $le) {
                 if (!empty($le['expire']) && $le['expire'] < time()) {
                     unset($srcList[$key]);
@@ -290,7 +290,7 @@ class MetaDataStorageHandler implements ClearableState
                 }
                 // We found the entity id so remove it from the list that needs resolving
                 /** @psalm-suppress PossiblyInvalidArrayOffset */
-                unset($entityIdsFlipped[$key]);
+                unset($entityIds[$entityIdsFlipped[$key]]);
                 // Add the key to the result set
                 $result[$key] = $le;
             }
@@ -299,7 +299,7 @@ class MetaDataStorageHandler implements ClearableState
         $endTime = microtime(true);
         $executionTime = $endTime - $startTime;
         $formattedTime = number_format($executionTime, 6, '.', '');
-        Logger::info(__METHOD__ . "Execution time: $formattedTime seconds");
+        Logger::debug(__METHOD__ . "Execution time: $formattedTime seconds");
 
         return $result;
     }
--- a/src/SimpleSAML/Metadata/MetaDataStorageHandler.php
+++ b/src/SimpleSAML/Metadata/MetaDataStorageHandler.php
@@ -264,31 +264,45 @@ class MetaDataStorageHandler implements ClearableState
      */
     public function getMetaDataForEntities(array $entityIds, string $set): array
     {
+        $startTime = microtime(true);
         $result = [];
+        // We are flipping the entityIds array in order to avoid constant iteration over it.
+        // Even if it becomes smaller over time.
+        // Still, after flipping all actions will be O(1)
+        $entityIdsFlipped = array_flip($entityIds);
         $timeUtils = new Utils\Time();
+
         foreach ($this->sources as $source) {
-            $srcList = $source->getMetaDataForEntities($entityIds, $set);
+            // Break iteration if the entityID List is empty
+            if (empty($entityIdsFlipped)) {
+                break;
+            }
+
+            $srcList = $source->getMetaDataForEntities($entityIdsFlipped, $set);
             foreach ($srcList as $key => $le) {
-                if (array_key_exists('expire', $le)) {
-                    if ($le['expire'] < time()) {
-                        unset($srcList[$key]);
-                        Logger::warning(
-                            "Dropping metadata entity " . var_export($key, true) . ", expired " .
-                            $timeUtils->generateTimestamp($le['expire']) . ".",
-                        );
-                        continue;
-                    }
+                if (!empty($le['expire']) && $le['expire'] < time()) {
+                    unset($srcList[$key]);
+                    Logger::warning(
+                        'Dropping metadata entity ' . var_export($key, true) . ', expired ' .
+                        $timeUtils->generateTimestamp($le['expire']) . '.',
+                    );
+                    continue;
                 }
                 // We found the entity id so remove it from the list that needs resolving
                 /** @psalm-suppress PossiblyInvalidArrayOffset */
-                unset($entityIds[array_search($key, $entityIds)]);
+                unset($entityIdsFlipped[$key]);
+                // Add the key to the result set
+                $result[$key] = $le;
             }
-            $result = array_merge($srcList, $result);
         }
+
+        $endTime = microtime(true);
+        $executionTime = $endTime - $startTime;
+        $formattedTime = number_format($executionTime, 6, '.', '');
+        Logger::info(__METHOD__ . "Execution time: $formattedTime seconds");
 
         return $result;
     }
-
 
     /**
      * This function looks up the metadata for the given entity id in the given set. It will throw an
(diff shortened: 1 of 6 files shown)


src/SimpleSAML/Metadata/MetaDataStorageHandler.php
```diff
@@ -264,32 +264,46 @@ public function getPreferredEntityIdFromCIDRhint(string $set, string $ip): ?stri
...
+        // We are flipping the entityIds array in order to avoid constant iteration over it.
+        // Even if it becomes smaller over time.
+        // Still, after flipping all actions will be O(1)
+        $entityIdsFlipped = array_flip($entityIds);
         $timeUtils = new Utils\Time();
+
         foreach ($this->sources as $source) {
-            $srcList = $source->getMetaDataForEntities($entityIds, $set);
+            // Break iteration if the entityID List is empty
+            if (empty($entityIdsFlipped)) {
```
tvdijen, line 277: This if-statement could be moved outside the foreach-loop.. If the variable is not empty, run the foreach..

ioigoume: @tvdijen we need to break looping the sources as soon as the `entityIdsFlipped` variable get empty, check line 293, `unset($entityIdsFlipped[$key]);

pradtke: I agree with @ioigoume that the goal was to stop iterating through the sources once all the entityIds have been found.

src/SimpleSAML/Metadata/MetaDataStorageHandler.php
```diff
@@ -264,32 +264,46 @@ public function getPreferredEntityIdFromCIDRhint(string $set, string $ip): ?stri
...
+                // Add the key to the result set
+                $result[$key] = $le;
             }
-            $result = array_merge($srcList, $result);
         }
 
+        $endTime = microtime(true);
+        $executionTime = $endTime - $startTime;
+        $formattedTime = number_format($executionTime, 6, '.', '');
+        Logger::info(__METHOD__ . "Execution time: $formattedTime seconds");
```
pradtke, line 302: I'd prefer not seeing the timing included and logged as part of the PR.     Overall I like the idea of being able to get timing information out of SSP on how long different things takes `authproc`, metadata retrieval,  saving/reading from the store, etc  but I think we should do that in a different conversation and get some more input from community. For example we have thresholds where that log level for timings adjusts based on how long it takes: e.g. fast is debug, slower is info, etc.      So my vote would be for removing timing from this PR and continuing the idea of how to do timing and threshold configuration in slack or github.

src/SimpleSAML/Metadata/MetaDataStorageSource.php
```diff
@@ -260,32 +260,35 @@ public function getMetaData(string $entityId, string $set): ?array
...
-     * @param string[] $entityIds The entity ids to load
-     * @param string $set The set we want to get metadata from.
+     * @param array<string, int> $entityIds        The entity ids to load
+     * @param string             $set              The set we want to get metadata from.
      * @return array An associative array with the metadata for the requested entities, if found.
      */
     public function getMetaDataForEntities(array $entityIds, string $set): array
     {
+        if (empty($entityIds)) {
+            return [];
```
pradtke, line 270: I do like this change.  We had an metadata source with a bug that if you loaded 0 entries it would perform super slowly, and this would have help prevent it from triggering.

src/SimpleSAML/Metadata/MetaDataStorageHandlerSerialize.php
```diff
@@ -271,8 +271,9 @@ public function deleteMetadata(string $entityId, string $set): void
     /**
      * This function loads the metadata for entity IDs in $entityIds. It is returned as an associative array
      * where the key is the entity id. An empty array may be returned if no matching entities were found
-     * @param array $entityIds The entity ids to load
-     * @param string $set The set we want to get metadata from.
+     *
+     * @param  array<string, int> $entityIds The entity ids to load
```
pradtke, line 275: I'm actually quite opposed to changing all these method signatures for any SSP 2.X release.  They are part of the public API contract for people extending SSP by creating their own metadata sources and this is not a backward compatible change.  My vote is to keep `$entityIds` as string[] arrays to make everyone's lives easier when they upgrade to 2.4.

monkeyiq:
Is the primary intention to stop the following loop as soon as possible or to use different data structures?
```
foreach ($this->sources as $source) {
```

If the former then at the end of that loop couldn't the code just fall out like below. At the moment `$entityIds` is used to track how many remain.
```
if (empty($entityIds)) { ... break }
```

This would avoid the further potentially heavy `getMetaDataForEntities` calls.

---

To avoid those `array_search` penalties then we might consider moving things that are using string sets in array values to using array keys instead. Though that might be cleaner to use a StringSet type there instead of having call types like `array<string, int>` just `StringSet`. 

I am wondering what size is your `$entityIds` array might be here? Or what size an SSP install might be throwing around. This would help to see if we might move `master` to using a dedicated type for handling string sets.

src/SimpleSAML/Metadata/MetaDataStorageHandler.php
ioigoume: @pradtke i added the time tracking for testing purposes while validating the fix. I will remove it as soon as we have a consensus on the implementation.

ioigoume:
@monkeyiq,
> Is the primary intention to stop the following loop as soon as possible or to use different data structures?
> 
> ```
> foreach ($this->sources as $source) {
> ```
> 


The goal here is to improve performance. Definitely the critical path is the heavy `getMetaDataForEntities::getMetadataSet` so we need to break when we do not have to call it any more. Then removing any unnecessary loops in favor of an O(1) call will further improve performance. 

> If the former then at the end of that loop couldn't the code just fall out like below. At the moment `$entityIds` is used to track how many remain.
> 
> ```
> if (empty($entityIds)) { ... break }
> ```
> 
> This would avoid the further potentially heavy `getMetaDataForEntities` calls.
> 
> To avoid those `array_search` penalties then we might consider moving things that are using string sets in array values to using array keys instead. Though that might be cleaner to use a StringSet type there instead of having call types like `array<string, int>` just `StringSet`.

Performance wise, using `Ds\Set` structures would make sense.

> 
> I am wondering what size is your `$entityIds` array might be here? Or what size an SSP install might be throwing around. This would help to see if we might move `master` to using a dedicated type for handling string sets.

tvdijen:
> Performance wise, using `Ds\Set` structure would make sense here.
> 

Let's not introduce a dependency on a pecl-extension.. It would be overkill.

ioigoume:
> > Performance wise, using `Ds\Set` structure would make sense here.
> 
> Let's not introduce a dependency on a pecl-extension.. It would be overkill.

I am going to add this here as a reference. The `DS` package can be installed without the use of pecl. I am using the (sury)[https://github.com/oerdnj/deb.sury.org] repository for installing PHP, both for Rocky linux and Debian and using

```bash
apt install php8.1-ds
```
command installs the package.

tvdijen:
That's exactly what we don't want.. Dependencies on third-party repositories..
On CentOS the packages isn't even available.

monkeyiq:
It is a bit annoying that something like a Set hasn't migrated into the core as yet. Or that there isn't a light and large php "variant"...

We could use the few needed methods with the same interface as `Ds\Set`. If things change and it becomes core then migration will be easier and in the meantime we wouldn't have the O(n) overhead of accessing the values in the array as though they are in a set.

This would then also mean we are changing a bunch of API signatures. So would be in a high 2.x if not `master`. 

I think the current version that is using the inverted set inside the single method helps and doesn't change a bunch of APIs.

I would personally move the `if (empty($entityIds))` check to the end of the loop. Also move the unset if possible to just before it. This has the advantage that the reader can see `$entityIds` getting smaller and then a test for extinction. If it is at the top it does lead the reader to think maybe it should be out of the loop. If it stays at the top I would put a comment before the empty test that the array shrinks in the loop. If it moves to the bottom it also has to be tested before the loop in case of an `empty()` entityid set being passed.

```
               // Add the key to the result set
               $result[$key] = $le;
               unset($entityIds[$entityIdsFlipped[$key]]);
               if (empty($entityIds)) {
                     break;
               }
```

ioigoume:
> It is a bit annoying that something like a Set hasn't migrated into the core as yet. Or that there isn't a light and large php "variant"...
> 
> We could use the few needed methods with the same interface as `Ds\Set`. If things change and it becomes core then migration will be easier and in the meantime we wouldn't have the O(n) overhead of accessing the values in the array as though they are in a set.
> 
> This would then also mean we are changing a bunch of API signatures. So would be in a high 2.x if not `master`.
> 
> I think the current version that is using the inverted set inside the single method helps and doesn't change a bunch of APIs.
> 
> I would personally move the `if (empty($entityIds))` check to the end of the loop. Also move the unset if possible to just before it. This has the advantage that the reader can see `$entityIds` getting smaller and then a test for extinction. If it is at the top it does lead the reader to think maybe it should be out of the loop. If it stays at the top I would put a comment before the empty test that the array shrinks in the loop. If it moves to the bottom it also has to be tested before the loop in case of an `empty()` entityid set being passed.
> 
> ```
>                // Add the key to the result set
>                $result[$key] = $le;
>                unset($entityIds[$entityIdsFlipped[$key]]);
>                if (empty($entityIds)) {
>                      break;
>                }
> ```

@monkeyiq the reason why the 
```php
            if (empty($entityIds)) {
                break;
            }
```
is immediately after the `foreach` and not grouped with the rest of the code, as you proposed, is because the `$entityIds` might be empty to begin with. Also, since the `getMetaDataForEntities` might be overridden by a third party module, things might get slow since the whole datasource will be loaded again.

pradtke:
@monkeyiq   You asked "I am wondering what size is your $entityIds array might be here? Or what size an SSP install might be throwing around."

For us the code path that leads to these methods being called is when an SP makes a request with `IDPList`.
In our usage the size is usually 1.   Sometimes we get a request from an SP that lists more, in the 10 - 20 range. 

So we aren't dealing with a large number of entities, and that part wasn't where we ran into performance issues.
The performance issue that we had encountered was, without the short circuits @ioigoume added, that ` getMetaDataForEntities` in MD source would load all the metadata and then discard it if asked to find it for 0 entityIds.  

My general thoughts are that we can forgo incorporating a Set data structure since the size of the `$entityIds` is small.

monkeyiq:
@pradtke Thanks for the info! I had suspected that the array vs set lookup might be very much overshadowed by the call `getMetaDataForEntities` for all sources (even when processing is already complete) performance hit. 

It shouldn't hurt things too much here to use a flipped array inside the one method to track things as the entities are found. It also doesn't need large scale API changes :)  I say it "shouldn't hurt things" because it comes at a slight price in terms of code complexity.

It is not that I wouldn't chase the O(N) -> O(1) lookup gain in general. But that can be a discussion for longer term implementation. Also with wider application across anything that is really a set being stored in an array. This might even be a post master becomes released issue as it may be better to stabilize master and move to that to enjoy all of the benefits it already has.

monkeyiq:
@ioigoume Yeah, if the `if (empty($entityIds)) {` moves to the end of the loop it would also need to be done before the loop is entered with a `return $result;` to short out the case of being called with an empty entity array.

Otherwise... maybe change:
```
// Break iteration if the entityID List is empty
```
to
```
// entityIds may be reduced to being empty in this loop or already empty
```

This would give a strong hint to the reader that the array is reduced during iteration (which happens at the end of the loop so is not in the mind's eye at this line).

tvdijen (approved):

monkeyiq:
also cherry picked to `master` :)

PR #2288, opened 2024-10-28
syntax highlighting for php.dist files

dnmvisser:
This adds syntax highlighting to the various `*.php.dist` files (such as those in `config` and `metadata`.


monkeyiq:
I also cherry picked this back to 2.3 and forward into master.

PR #2292, opened 2024-10-31
docs: add some info about the new entityid metadata setup from pr 2270

monkeyiq:
This relates to https://github.com/simplesamlphp/simplesamlphp/pull/2270


monkeyiq:
This might do with a little rewording etc, but if it is already here there is less chance it gets forgotten for the release notes

PR #2301, opened 2024-11-04
replace assert_options with ini_set, fix commit b20f5714

paulmhh:
due to PHP8.3 deprecation of assert_options
else clause in commit b20f5714 doesnt make sense, reverted to state beforehand


tvdijen:
Hi @paulmhh !  Thank you for your contribution, however v1.19.x is no longer supported and will not receive any updates anymore. I therefore kindly decline this PR. Sorry!

Issue #2305, opened 2024-11-07
Incorrect language code en_LS used for Sesotho


ghalse:
Issues #2269 together with pull requests #2274 and all those mentioned in #2277 erroneously change the language code for Sesotho from `st` to `en_LS`.

This makes no sense whatsoever. Sesotho is a language in its own right, used in Lesotho, South Africa and Zimbabwe. However, `en_LS` suggests it is a Lesotho variant of English. So it's fundamentally the wrong language code for the language.

`st` is the correct ISO639-1 Alpha2 code: https://www.loc.gov/standards/iso639-2/php/langcodes_name.php?iso_639_1=st

In Symfony's locale format, it should be `st_LS`. Or perhaps `st_ZA` since the original translators who produced it [#1260] speak the South African variant of Sesotho (although for these purposes it shouldn't matter). However, this doesn't make sense because the regionalised versions presume a base variant of `st` as a [fallback locale](https://symfony.com/doc/current/translation.html#fallback-translation-locales). Meaning `st` is correct, and should work.

The real problem appears to be that Symfony itself has no idea that Sesotho is a language, irrespective of the regionality of it ([see here](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Intl/Resources/data/locales/en.php#L566-L567)).

Symfony appears to derive that data from the Unicode ICU and claims that [6.4 is up-to-date to ICU 75.1](https://github.com/symfony/symfony/blob/v6.4.14/src/Symfony/Component/Intl/Resources/data/git-info.txt). The ICU includes the st language code (and st_ZA and st_LS) from [release 76.1](https://github.com/unicode-org/icu/blob/release-76-1/icu4c/source/data/locales/st.txt).

So the original st that was there before this change will be correct, when Symfony updates its locale information to match the next release of of the ICU locale data...

tvdijen:
Well, this kind of explains why I had a hard time finding the right code for it..
My suggestion would be to rename it to the appropriate `st_ZA` and then hopefully we can locally fix this until Symfony updates their version of ICU.

ghalse:
I think the original `st` is the right answer, rather than `st_ZA`. I'm opening a separate issue at Symfony for them to update the ICU data, since it affects many official languages of many African countries.

ghalse:
[...]

PR #2312, opened 2024-11-13
Add gettext compatible headers to PO files

ghalse:
Add basic headers to the PO files generated by bin/translations translations:update:translatable to make them more compatible with external translation tools.

The headers aren't [complete](https://www.gnu.org/software/gettext/manual/html_node/Header-Entry.html), but they're now good enough that `msgfmt -c` issues warnings not errors.

There are some assumptions here, although they're not entirely mine ;-).

The big one is that all the PO files are UTF-8 encoded. I can find no mention of what encoding is actually used anywhere in the php-gettext tools, but I think this is fairly safe. Running `files` on the existing .po files suggests they're almost all already UTF-8, and those that aren't are probably ASCII/ISO8859-1 and implicitly compatible with UTF-8. If they're not, they're broken anyway because gettext reading a file without a header appears to assume ISO8859-1.

The other assumption is that the directory name under locales/ matches what should be in the Language header. I think that's fairly safe, particularly given the changes that lead to #2305.

### DIFF
--- a/bin/translations
+++ b/bin/translations
@@ -9,7 +9,11 @@ use Symfony\Component\Console\Application;
 umask(000);
 set_time_limit(0);
 
-require __DIR__.'/../vendor/autoload.php';
+if (str_ends_with(__DIR__, '/vendor/simplesamlphp/simplesamlphp/bin')) {
+    require __DIR__.'/../../../autoload.php';
+} else {
+    require __DIR__.'/../vendor/autoload.php';
+}
 
 $application = new Application();
 $application->add(new UnusedTranslatableStringsCommand());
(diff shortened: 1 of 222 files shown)


ghalse:
The pull request is broken into two commits to make it easer:

The changes in `src/SimpleSAML/Command/UpdateTranslatableStringsCommand.php` are a commit on their own, and should be cherry-pickable into the 2.3 branch if you want to do that. They're the important part of the pull request.

The changes in `bin/translations` aren't strictly needed, but make it easier for me to use this command from vendor/simplesamlphp/simplesamlphp/bin/translations in my module project.

ghalse:
And for completeness, the existing locale files in master updated with headers as their own commit.

monkeyiq:
I think I will cherry pick the first two commits and then run the update script locally and push those changes as a different PR. It will be simpler than trying to eyeball the 200 odd files of changes. I don't think there is anything nasty there but I generally like to eyeball things just in case.

monkeyiq:
The github action also does the rebuild I see. So I did the 2.4 two small cherry picks and then the GHA. This should be in master and 2.4 now.

PR #2315, opened 2024-11-15
Make translations tool theme-aware

ghalse:
With [#2312](https://github.com/simplesamlphp/simplesamlphp/pull/2312/commits/e2f4e0cc53eb1f1b176ce23b94a1000f599e3064), the bin/translations tool becomes useable to module developers as well as SSP itself. However, I realised that there's one bit missing -- theme modules that include additional strings in their overridden templates.

This makes the tool theme-aware, allowing people to (optionally) parse the themes/ directory as well.

In my theme module I can now do:
```sh
dev:~/simplesamlphp-module-safire$ ./vendor/simplesamlphp/simplesamlphp/bin/translations translations:update:translatable --module=safire --themes
```
and end up with updated .po files that include all the translatable strings, including those in my theme overrides. That's a lot easier than the hand extraction I used to do :)

I've made the default for the `translations:update:translatable` command to ignore themes (--no-themes), since I suspect the most common use case is to make minor adjustments to layout and the original translations in the source module would remain usable.

However, I've made the default for `translations:unused` to include themes (--themes), so that we don't mark any existing strings as unused if they are used in themes.

One caveat is that the modules that the theme refers to must exist and be enabled for the tool to work. This is easily achieved by doing a composer require --dev for the appropriate module. The tool will generate an error suggesting this if it encounters an unknown module in a theme.

### DIFF
--- a/src/SimpleSAML/Command/UnusedTranslatableStringsCommand.php
+++ b/src/SimpleSAML/Command/UnusedTranslatableStringsCommand.php
@@ -55,6 +55,13 @@ class UnusedTranslatableStringsCommand extends Command
             InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
             'Which modules to perform this action on',
         );
+        $this->addOption(
+            'themes',
+            null,
+            InputOption::VALUE_NEGATABLE,
+            'Include/exclude templates from themes directories of modules (default --themes)',
+            true, /* safe default, so we don't lose translations */
+        );
     }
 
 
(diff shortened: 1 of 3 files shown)


tvdijen:
I'm happy to merge, if you're done

ghalse:
> I'm happy to merge, if you're done

I'm done. The theme.controller stuff I just added explained the one weird missing bit in my tests.

I think it now has reasonably complete coverage of the use cases. It'll work for simple modules that have templates; it'll work for modules that contain one or more themes; and it will work for modules that require a theme controller. And any combination of those.

The one scenario it won't cover is a module that has multiple theme controllers (presumably applying to multiple themes). Supporting that is hard, and I'm hoping it's a pretty rare case :)

Issue #2324, opened 2024-11-18
PHP 8.4 incompatibility


tvdijen:
From the PHP 8.4 changelog:

Gettext:
bind_textdomain_codeset, textdomain and d(*)gettext functions now throw an exception if the domain argument is empty.
      
We have one case where this affects us:

https://github.com/simplesamlphp/simplesamlphp/blob/76109fbcce35c00d4244ab6cee4c7bf879508091/src/SimpleSAML/Locale/Translate.php#L95-L99

monkeyiq:
I think here the only po files without a domain (`X-Domain` in the po file) are the attributes file in locales/en/LC_MESSAGES.  We could make that domain "attributes" and update this fallback to using that.

tvdijen:
I was hoping it was that simple, yes :D  If you could test this for us, that would be great!

monkeyiq:
Will do. As I am not on the latest Fedora I think I will either update a secondary machine or install F41 in a VM to be able to verify that php 84 is happy with this as well. Having a "Works on 8.3" is weaker than a works on 8.4 goal.

monkeyiq:
more to do on the pr when testing it in anger.

monkeyiq:
After a bit of chasing my tail I see that the `dgettext` call in SSP is actually to a php library:
https://github.com/php-gettext/Translator/blob/a4fa5ed740f304a0ed7b3e169b2b554a195c7570/src/Translator.php#L121

Running current 2.3 (`d31b3b3`) in a Fedora 41 VM with an updated 8.4.1 php version seems to work fine.

monkeyiq:
I will also take a look at composer generating binary mo files in master soon.

PR #2328, opened 2024-11-20
i18n move attributes into an attributes domain

monkeyiq:
A possible fix for https://github.com/simplesamlphp/simplesamlphp/issues/2324

I will be doing some more testing on this in an installed form.

I have been testing this with a bin/test4.php script which is as follows. Adding a new `test4` msgid in the attributes.po for `en` does find the new term when the domain is set in attributes.po and the code is looking for the "attributes" domain.

```
#!/usr/bin/env php
<?php

require __DIR__ . '/../vendor/autoload.php';

use SimpleSAML\Metadata\MetaDataStorageHandler;
use SimpleSAML\Configuration;
use SimpleSAML\Locale\Translate;
use SimpleSAML\Locale\Localization;

use Symfony\Component\Intl\Locales;

echo "---- hello ----- \n";


$configuration = Configuration::getConfig();
$loc = new Localization($configuration);
$tr = new Translate($configuration);
$loc->addAttributeDomains();

echo "locales path " . $configuration->resolvePath('locales') . "\n";


$v = Translate::translateSingularGettext("test4");
echo "$v \n";

```

### DIFF
--- a/bin/translateAttributes.php
+++ b/bin/translateAttributes.php
@@ -121,5 +121,6 @@ foreach (array_keys($languages) as $language) {
 
     $strings->setLanguage($language);
     echo "Saving translations to " . $base . "locales/" . $language . "/LC_MESSAGES/attributes.po\n";
+    $strings->setDomain("attributes");
     Gettext\Generators\Po::toFile($strings, $base . 'locales/' . $language . '/LC_MESSAGES/attributes.po');
 }
--- a/locales/af/LC_MESSAGES/attributes.po
+++ b/locales/af/LC_MESSAGES/attributes.po
@@ -9,6 +9,7 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Language: af\n"
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
+"X-Domain: attributes\n"
 
 msgid "aRecord"
 msgstr "aRecord"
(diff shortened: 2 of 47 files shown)


monkeyiq:
this still has some issues

tvdijen:
Like what?

monkeyiq:
It wasn't working unless I had the `""` line in there too. It didn't make sense.

Having thought about this overnight I thought maybe I had some `mo` files floating around. That triggering the load binary files without domain path which was implicitly working when not explicitly looking for the `attributes` domain. Short story, I did have some of those and removing them allowed translation to work using `"attributes"` as the only fallback. So this PR works on my desktop and on the F41/8.4.1 vm machine. once the errant mo files are removed.

The follow up will be sifting around in the bin scripts etc to see what I ran to make those `mo` files and then to make that stop happening. I think removing the fallback to `mo` files makes sense too.

monkeyiq:
I just reverified this as working in 2.3 using an app pointed at /opt/sspdev. I was previously testing in the ssp/admin test logins.

monkeyiq:
I am still trying to find out what was making those original mo files :/

monkeyiq:
Nothing in ssp/bin seems to be doing it or composer update.

monkeyiq:
It is being done in the `composer update` for master. That is still generating the binary `mo` files. Perhaps because we might have moved back to binary in at the time of master.

monkeyiq:
I am thinking we should remove that binary generation in `master` for now to avoid these possible head scratcher interactions if folks are switching from 2.3, 2.4 to master and back and the `mo` files remain floating around.

tvdijen:
Nuke the mo-files please

monkeyiq:
I have removed the calls that were making the mo files in `master` with https://github.com/simplesamlphp/simplesamlphp/pull/2343

PR #2330, opened 2024-11-21
i18n: this is just the removal of the mo file loading. The patch looks a bit more complex than that

monkeyiq:
One of the scripts I ran made the mo files which were loaded by this path giving unexpected results. I think this path was in place to allow a transition but perhaps we should just say it is `po` files for now.

Perhaps this should be against 2.4 instead.

### DIFF
--- a/src/SimpleSAML/Locale/Localization.php
+++ b/src/SimpleSAML/Locale/Localization.php
@@ -250,29 +250,20 @@ class Localization
             }
         }
 
-        $file = new File($langPath . $domain . '.mo', false);
+        $file = new File($langPath . $domain . '.po', false);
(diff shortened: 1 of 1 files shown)


tvdijen:
The mo files should have been killed a long time ago.
Let's merge this in 2.4 to make sure we don't break stuff for people

monkeyiq:
Changing the base ref gave an output that didn't make much sense. So I just rebased locally onto 2.4 to https://github.com/simplesamlphp/simplesamlphp/pull/2331

monkeyiq:
I have merged that into 2.4 and will cherry pick it over to master.

PR #2332, opened 2024-11-23
Add possibility to provide connection context to XML storage-handler

tvdijen:


monkeyiq:
This looks good. 

One thing that occurs to me is `HTTP::fetch` overrides some context options. At the moment these are proxy related and SNI stuff. Though that might well change over time.

Perhaps the person configuring things might expect that `metadata.sources/context` will override any defaults that are set. At the moment the code flow uses the md.sources `context` as a base and overrides things in `HTTP::fetch`. 

Maybe we should move to storing the $context in fetch, do the initial setting of a new local `$context[]` in fetch and then take what was passed to fetch to copy over to `$localcontext`. Thus making what is passed the final setting? This would protect things for future updates if we have a default_context from the config file and allow the context for a specific metadata stanza to be able to override the ciphers list for example.

monkeyiq (approved):
Looks good. The update I suggest can be done as an addition if we decide to move to that.

tvdijen:
This is a good idea @monkeyiq !

PR #2336, opened 2024-11-28
Fix IDPList setting in saml20-sp-remote not applied when SSP is in proxy mode

pierrebateau:
In #1653 the way scoping is handled by the IDPList setting was changed. Currently this breaks the ability to apply IdP scoping when defined in  saml20-sp-remote.php because the IDPList will be set to the state as `saml:IDPList`.

Following a discussion with @tvdijen on Slack, I re-added the possibility to apply scoping through `$state['saml:IDPList']` add kept the changed version from #1653 (with `$state['IDPList']`) intact to prevent issues when this state parameter is explicitly set. I don't think this is the most beautiful approach, but it fixes this specific use case and is backwards compatible with the change from #1653.

### DIFF
--- a/modules/saml/src/Auth/Source/SP.php
+++ b/modules/saml/src/Auth/Source/SP.php
@@ -582,6 +582,8 @@ class SP extends \SimpleSAML\Auth\Source
         if ($this->disable_scoping !== true && $idpMetadata->getOptionalBoolean('disable_scoping', false) !== true) {
             if (isset($state['IDPList'])) {
                 $ar->setIDPList($state['IDPList']);
+            } elseif (isset($state['saml:IDPList'])) {
+                $ar->setIDPList($state['saml:IDPList']);
             } elseif (!empty($this->metadata->getOptionalArray('IDPList', []))) {
                 $ar->setIDPList($this->metadata->getArray('IDPList'));
             } elseif (!empty($idpMetadata->getOptionalArray('IDPList', []))) {


monkeyiq:
I have been sort of taking a back seat on this PR as a review was requested. To test this I will have to try to setup a proxy mode config to see what is happening there.

monkeyiq:
It seems that from https://github.com/simplesamlphp/simplesamlphp/pull/1653 the two parameters should be exclusive. So there shouldn't be any problem also using `state['saml:IDPList']` to allow things to work on a proxy config?

monkeyiq:
Sorry, I haven't setup an SSP in proxy mode as yet. So I am just looking at the background and change as yet.

monkeyiq:
This appears to be in release v2.3.4+ already. 

Perhaps it was merged but the PR was not updated?

https://github.com/simplesamlphp/simplesamlphp/blob/210a8f677f827b6296f6ae4c056f6297498e7762/modules/saml/src/Auth/Source/SP.php#L583

@tvdijen do you think there anything else to do with this PR other than mark it as fixed in 2.3.4+ ?

tvdijen:
Ouch, I think this was a mistake on my end.. I think it's OK to merge this into 2.4 & master

monkeyiq:
I have brought this forward into 2.4 with b849ced30f2fc30c6f945727531c9c211f540fe2
and master https://github.com/simplesamlphp/simplesamlphp/commit/8f000136541075a15a0cab88c8788b1911f2073f

monkeyiq:
This is now in 2.3, 2.4, and master. It was not directly merged but the update is in all three. Thank you @pierrebateau

PR #2349, opened 2024-12-11
Update the SP and IdP URIs to pass the URI regex.

monkeyiq:
These were all failing because of
  "eugeneIdP" is not a SAML2-compliant URI

It seems that adding in the colon makes these a compliant test.

The regex that was failing things is
  /^([a-z][a-z0-9\+\-\.]+[:])/i

which is not anchored at both ends, so it seems you can have whatever you want after the colon. As long as there is one with a limited prefix.

I discovered this when making the AttributeNameIDTest for pr 2260. It may also cause other tests to fail creating a bunch of smoke around the harder tests to get working.

### DIFF
--- a/tests/modules/saml/src/Auth/Process/NameIDAttributeTest.php
+++ b/tests/modules/saml/src/Auth/Process/NameIDAttributeTest.php
@@ -40,8 +40,8 @@ class NameIDAttributeTest extends TestCase
     public function testMinimalConfig(): void
     {
         $config = [];
-        $spId = 'eugeneSP';
-        $idpId = 'eugeneIdP';
+        $spId = 'eugene:SP';
+        $idpId = 'eugene:IdP';
 
 
         $nameId = new NameID(
@@ -72,8 +72,8 @@ class NameIDAttributeTest extends TestCase
     {
         $attributeName = 'eugeneNameIDAttribute';
         $config = ['attribute' => $attributeName];
-        $spId = 'eugeneSP';
-        $idpId = 'eugeneIdP';
+        $spId = 'eugene:SP';
+        $idpId = 'eugene:IdP';
 
         $nameId = new NameID(
             value: 'eugene@oombaas',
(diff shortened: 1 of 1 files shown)


monkeyiq:
If this is merged I might expand things to other tests to try to lower the overall fail count a bit.

tvdijen (changes requested):
The test should be replaced with \SimpleSAML\SAML2\Assert\Assert::validEntityID and the tests then have to get _real_ entityID's

monkeyiq:
Forcing one of these to fail again the backtrace is as follows.
```
vendor/simplesamlphp/saml2/src/Assert/CustomAssertionTrait.php:70
vendor/simplesamlphp/saml2/src/Assert/CustomAssertionTrait.php:81
vendor/simplesamlphp/saml2/src/Assert/Assert.php:107
vendor/simplesamlphp/saml2/src/Assert/Assert.php:73
vendor/simplesamlphp/saml2/src/XML/saml/NameIDType.php:41
vendor/simplesamlphp/saml2/src/XML/saml/NameID.php:75
tests/modules/saml/src/Auth/Process/NameIDAttributeTest.php:47
```

NameIDType.php:41 / 42 (depending on if it is an invalid entity on sp or idp)
```
SAMLAssert::nullOrValidEntityID($NameQualifier);
SAMLAssert::nullOrValidEntityID($SPNameQualifier);
```

After a bit of digging I see there is an `Assert` and `SimpleSAML\SAML2\Assert\Assert as SAMLAssert;` as a second assert class.

monkeyiq:
I have updated the entity ids here. I guess the regex used here is to try to catch very vague values being used for entities.

tvdijen:
Yes, SAML 2.0 entities are defined as URIs and from the top of my head they have to be absolute URIs, which leaves us with URNs and URLs. The test-values were neither of those.

tvdijen (approved):

PR #2351, opened 2024-12-15
It seems the & should be after the array here.

monkeyiq:
Minor update to the recent commit here https://github.com/simplesamlphp/simplesamlphp/commit/5f4a47f32cd7787e845445595da4b33f28f1415b


modules/core/src/Auth/Source/AbstractSourceSelector.php
```diff
@@ -71,7 +71,7 @@ public function authenticate(array &$state): void
      * @param array $state
      * @return void
      */
-    public static function doAuthentication(Auth\Source $as, &array $state): void
+    public static function doAuthentication(Auth\Source $as, array &$state): void
```
tvdijen, line 74: D'OH! :'(

PR #2355, opened 2024-12-16
Skip disco when single IdP found

thijskh:
When there's only a single IdP to choose from, discovery is useless at best and confusing at worst. Just skip it as we know what the choice will be.

### DIFF
--- a/src/SimpleSAML/XHTML/IdPDisco.php
+++ b/src/SimpleSAML/XHTML/IdPDisco.php
@@ -15,6 +15,7 @@ use function array_fill_keys;
 use function array_intersect;
 use function array_intersect_key;
 use function array_key_exists;
+use function array_key_first;
 use function array_keys;
 use function array_merge;
 use function preg_match;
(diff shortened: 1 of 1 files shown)


tvdijen (approved):

monkeyiq:
I tested this using sspsmall and changing authsources/`sp.idp == null;` 

Without the patch I was offered a disco screen with a choice of 1 IdP with the patch I was directed straight to the IdP.

When I had just one IdP for an SP I would generally state it on the SP page so hadn't noticed this degenerate case. I imagine in real world cases where the IdP list might be defined in a more complex way then this case would also present itself and offer a less than desirable user experience.

monkeyiq:
I have picked this into master https://github.com/simplesamlphp/simplesamlphp/commit/f7fb7bb099c5bb09c1d8717f9cd9d5befbb35afc

monkeyiq:
This seems like a candidate for picking back into the 2.3 series as well?

thijskh:
It has worked in the previous way since day one I guess. So maybe it's more like a feature than an actual bugfix.

PR #2356, opened 2024-12-16
Docs: Clarify that the Simple API needs a saml:SP

monkeyiq:
Perhaps here I could reference sspsmall once that is in better shape to help small deployments get up and runnnig.


monkeyiq:
sspsmall is still in its infancy but is here https://github.com/monkeyiq/sspsmall if folks want to see it. It is probably best to only look at the config and scripts for now until I have tested it a bit more. This should still give an idea of how one might do a small SP/IdP install.

Issue #2327, opened 2024-11-20
Possible auto case folding on language code?


monkeyiq:
I was looking at an issue with language cookies and case folding in the CI of master. The test was setting a custom named cookie with a value `Es` and expecting it to be converted to `es` by SSP. In the past there were a bunch of `strtolower()` for the selected locale/lang code in Language.php which were removed to allow codes like pt_BR to work. 

The question is do we wish to add calls to `\Locale::canonicalize` or something similar to do the same sort of case and implicit conversion that was there in the past? We would want to use `\Locale::canonicalize` via a local static method so we could use our own custom converter instead or in addition to if desired.

tvdijen:
I think we don't need the static method..  The cookie needs to have the right value with lowercase `es` and not `Es`.
Symfony\Component\Intl\Locales::exists() already validates the locale and will fail if the wrong casing is used.

monkeyiq:
Proposed update in https://github.com/simplesamlphp/simplesamlphp/pull/2357

tvdijen:
Sorry, I don't like the added dependency on `ext-intl` for something so trivial... It think it's fine as-is right now.

monkeyiq:
I agree. I hadn't noticed yesterday that Locale was not included in the core. For this it is certainly not worth using it.

We have these issues etc now for searching and if/when something comes up where something is using "Es" for "es" we can take a look at the context there and decide if we want to do something in SSP or not.

PR #2357, opened 2024-12-22
Run the language through Locale::canonicalize to clean it up

monkeyiq:
This relates to
https://github.com/simplesamlphp/simplesamlphp/issues/2327

This may or may not be needed or desired. It should allow more language code to be accepted, and possibly incorrect ones which will be converted into something that can be acted on.


monkeyiq:
As this depends on the intl extension, it is not worth adding this one line change right now. If at some point a source of invalid locale strings is found we can work out how we wish to handle that.

This update was sort of a lingering from a test case that was allowing "Es" to be converted to "es" implicitly by SSP. Unless we know why and where such invalid input is coming from we can leave this out for now.

Issue #2299, opened 2024-11-03
Ability to define hosted SP in saml20-sp-hosted


monkeyiq:
At the moment a hosted SP is defined in `config/authsources.php` with type `saml:SP`. The use of the same file lumps the definition of a hosted SP with the definition(s) of authentication methods. 

It would be nice to be able to define the SP in a `metadata/saml20-sp-hosted.php` file. This would enable the authsources file to be kept to authentication methods that hosted IdP might use to authenticate a user with. For new or casual users of SSP having the SP in `config/authsources.php` might be a little confusing for a number of reasons. 

One confusion being that in the 1.x series one might pass the name of an authentication method directly to the `\SimpleSAML\Auth\Simple` API to authenticate with. This configuration is not desired but be tempting given that the SP and authentication method are in the same file. Moreover the misconfiguration of directly using an AM might have worked in 1.x.

I suggest adding the ability for hosted SP to be placed in `metadata/saml20-sp-hosted.php` and also loaded from `config/authsources.php`. The documentation can be updated to prefer the new way. Over time perhaps defining the SP in the latter can become deprecated and removed.

monkeyiq:
I am working on some code to do this. As it updates a number of methods in SimpleSAML\Auth\Source I will have to add a number of unit tests to try to ensure this is not disruptive to folks. I will initially look at a 2.4 inclusion.

monkeyiq:
I have some initial code for this up at https://github.com/simplesamlphp/simplesamlphp/pull/2360

monkeyiq:
It seems for now that leaving hosted SP in authsource.php is the desired outcome. I have some initial code in https://github.com/simplesamlphp/simplesamlphp/pull/2360 to start on a transition to allowing hosted SP to be defined in `metadata/saml20-sp-hosted.php` but am leaving that aside for now.

PR #2360, opened 2024-12-29
Also allow hosted SP to be defined in metadata/saml20-sp-hosted.php

monkeyiq:
This was raised in issue
https://github.com/simplesamlphp/simplesamlphp/issues/2299

This is a first attempt to make this happen. If you are defining your hosted SP in authsources.php then they should still be picked up. Otherwise you can now define them in metadata/saml20-sp-hosted.php.

This separates the authsources and sp-hosted into two logical locations. The authsources is a good place to define your authentication mechanisms that an IdP might like to use to verify a user. The sp-hosted metadata file is a good place to define a hosted SP. I think the simplesamlphp-sp.md becomes a little verbose with both examples all the time. Perhaps we should mention that there are two formats and the inline definition in authsources is still respected but that folks might like to move to using sp-hosted. New users might like to just start using the later as the location that a hosted SP will be defined.

One slight difference in the format to defining an SP in authsources vs in sp-hosted is that the type is forced to 'saml:SP' authmatically for you in sp-hosted. No initial empty 'saml:SP' key is needed there.

I still have to experiment with the setting of the entityid and entityID for an SP loaded from authsources.php. I suspect that one of those settings needs to be either not there or made optional.

I have expanded SourceTest.php to try to find issues with the code though have only tested that with my local setup and not the one that is employed by phpunit. There are likely to be slight differences to count()s needed with the later environment.

I initially had getAuthSourcesFromConfig toward the top of the file but the diff also conflated that method with getSourcesOfType which was much harder to read. The current location makes the diff much simpler to read.

As the above highlights, there are probably updates desired to this PR before merge but I think it is a good time to share it and gather thoughts and comments.

### DIFF
--- a/docs/simplesamlphp-sp.md
+++ b/docs/simplesamlphp-sp.md
@@ -63,9 +63,10 @@ see the [`saml:SP` reference](./saml:sp).
 If you want multiple Service Providers in the same site and
 installation, you can add more entries in the
 `metadata/saml20-sp-hosted.php` or `authsources.php` configuration. If
-so remember to set the EntityID explicitly. 
+so remember to set the EntityID explicitly.
 
 Here is an example for `metadata/saml20-sp-hosted.php`:
+
 ```php
 $metadata['https://myapp.example.org/'] = [
     'authid' => 'sp1',
--- a/docs/simplesamlphp-sp.md
+++ b/docs/simplesamlphp-sp.md
@@ -18,7 +18,7 @@ metadata loading system as other files in metadata giving you more
 flexibility in how you define your configuration.
 
 The formats for defining a SP in `saml20-sp-hosted.php` and `authsources.php`
-are very similar. The main differences are that hte former uses `$metadata`
+are very similar. The main differences are that the former uses `$metadata`
 and the later a subelement in `$config`.
 
 This is a minimal `metadata/saml20-sp-hosted.php` for a SP:
--- a/docs/simplesamlphp-sp.md
+++ b/docs/simplesamlphp-sp.md
@@ -8,7 +8,29 @@ described in the [SimpleSAMLphp installation instructions](simplesamlphp-install
 
 ## Configuring the SP
 
-The SP is configured by an entry in `config/authsources.php`.
+The SP can be configured in either the `metadata/saml20-sp-hosted.php`
+or by adding an entry in `config/authsources.php`. The
+`saml20-sp-hosted.php` file method was added in SSP 2.4 to simplify
+where a hosted SP is configured. Using `saml20-sp-hosted.php` keeps
+the SP definitions and the authentication methods that are defined in
+`authsources.php` separate. The new system in 2.4 also uses the same
+metadata loading system as other files in metadata giving you more
+flexibility in how you define your configuration.
+
+The formats for defining a SP in `saml20-sp-hosted.php` and `authsources.php`
+are very similar. The main differences are that hte former uses `$metadata`
+and the later a subelement in `$config`.
+
+This is a minimal `metadata/saml20-sp-hosted.php` for a SP:
+
+```php
+<?php 
+
+$metadata['default-sp'] = [
+        // The entity ID of this SP.
+        'entityID' => 'https://myapp.example.org/',
+];
+```
 
 This is a minimal `authsources.php` for a SP:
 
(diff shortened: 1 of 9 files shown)


monkeyiq:
I guess it would be better longer term to setup some elisp to help make the docs lint as expected.

tvdijen:
I don't want to spoil the party here, but I think the FlatFile-metadata loader will break on this because the hosted-sp files do not follow the same format as the other metadata-files do..

monkeyiq:
It looks like it was a good idea to put this out there now. The PR is enough to talk about and refine ideas & code from there. Looking to refine and fully test a solution locally that then wants some changes to how the name/entityID are set is probably the slower way of going overall to the merge goal.

The PR is not complete. I also want to add more test cases to ensure that nothing breaks for people with existing complex configurations for many SPs in `authsources.php`. 

I started out with a test SP in authsources.php and looking to move that configuration into `saml20-sp-hosted.php`. There are some updates I have made to allow that to happen but one thing that the code refined on which now that I am comparing metadata files is perhaps not desired was having the entityID as an attribute. The format of other metadata files seems to have the entityID / URI for the item in the key to `$metadata[]` itself and the settings in an array under that key.

Currently this PR has the "name" for the SP as the `$metadata[]` key and the entityID as a must be present configuration key (or `assert`).

The test metadata file here
https://github.com/simplesamlphp/simplesamlphp/pull/2360/files#diff-945852e553ce2d2b14d3a40995a9ec5107339a69119e0b2b166b9dcc8c408f61

should be updated. The 'saml:SP', line is no longer needed (it is forced to that when loaded from sp-hosted) and I think that the entityID should maybe be the key to the `metadata[]` to be more inline with other metadata files. Considering the `saml20-idp-hosted.php.dist` file as a template.

The new SourceTest.php methods should be loading the SP from `saml20-sp-hosted.php` using the MetaDataStorageHandler. For example in the test for  `testgetSourcesOfTypeID`. I have focused the tests at this stage on making sure the updated `getX` methods are returning an expected result. This is mainly for running the test but also with a bunch of extra logging in the Soruce.php to examine what is happening at a low level.

So with the entityID shuffle an updated `saml20-idp-hosted.php` using the example SP data from `authsources.php.dist` might look like the below... There doesn't seem to be a "simple name" for IdP but I see in admin/federation that my other SPs do have the simple name that was the key used in authsources.php.

```
<?php 

// A hosted SP
$metadata['https://myapp.example.org/'] = [

         // simple name for UI ?
         // this was the key in authsource.php such as
         // 'default-sp' => [ 'saml:SP', ... 
        'ssp-name' => 'default-sp',

        // The entity ID of the IdP this SP should contact.
        // Can be NULL/unset, in which case the user will be shown a list of available IdPs.
        'idp' => null,

...
];
```

monkeyiq:
Testing things more I see I have also made most of my testing on this around saml:SP sources. Some updates are needed to make `getAuthSourcesFromConfig()` work for some other sources in authsources.php.

monkeyiq:
The ssp-name (or ssp.name to be more like other config settings) is the `$authId` in `Source::getById`. So maybe it should be named authid instead of ssp.name.

monkeyiq:
It looks like sp-hosted is harder to add to the metadatahandler xml. That also doesn't seem to allow IDP Hosted.

https://github.com/simplesamlphp/simplesamlphp/blob/355ab8192b4478a71f81ed9da2ba385dc4b401e2/src/SimpleSAML/Metadata/MetaDataStorageHandlerXML.php#L93

monkeyiq:
So `tests/src/SimpleSAML/Auth/test-metadata/source3` now has the same sp metadata (except authid and entityid) in both an `authsources.php` format and `saml20-sp-hosted.php` and loads both and asserts the loaded data.

monkeyiq:
As this touches 3 core methods that access SPs I am planning a wider test suite to try to verify the change doesn't disrupt users. It is a nice to have using an sp-hosted file to lift the SP description to a metadata file and leave the auth mechanisms in authsources.php. Though I need to be more certain that it isn't going to break configurations that are not migrated to that setup. As this is my PR I can pick at the updates as time allows without fear of annoying the submitter :)

tvdijen:
The more I think about this PR, the more I doubt it..
Our 'metadata'-files have a certain format that is recognizable and putting an SP-authsource in it causes errors in my mind, the same way it errors people their mind that SP-config is in authsources.php...

monkeyiq:
We could also shelf this PR and just double down on documentation that hosted SP goes in authsource and there are two types of things in `authsources.php` : authsource (SP) and auth mechanisms.

I don't really feel strongly enough about the change to push it if it makes SSP folks feel at all uneasy :)

Issue #2367, opened 2025-01-06
Infinite recursion due to possible circular references for `var_export`


rharish101:
**Specifics of your environment**
1. Are you acting as SP/IdP/proxy? IdP
2. SimpleSAMLphp: 2.3.5
3. PHP: 8.3.15
4. Platform: Linux
5. Webserver: Apache

**Describe the bug**
When writing a custom auth proc filter by subclassing `SimpleSAML\Auth\ProcessingFilter`, I cannot create private class properties that contain objects, which have circular references. The warning "Warning: var_export does not handle circular references in /path/to/simplesamlphp/src/SimpleSAML/Auth/ProcessingChain.php on line 93" is repeated many times on the error page, which shows the error "Error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 199233536 bytes)".

The problem is solved by commenting out these lines: https://github.com/simplesamlphp/simplesamlphp/blob/e3c148028ba3d0c1cfd941ff885626faecb0fc95/src/SimpleSAML/Auth/ProcessingChain.php#L89-L90
The problem is the `var_export`, whose contents cannot be customized with a `__debugInfo` method in the custom auth proc filter like `print_r`.

**To Reproduce**
1. Create a SimpleSAMLphp IdP.
2. In the IdP's installation directory, create a new directory in `modules/`, say, `modules/test`.
3. Inside this directory, create the file `src/Auth/Process/TestFilter.php`.
4. Subclass `SimpleSAML\Auth\ProcessingFilter` to create a custom auth filter `TestFilter`.
5. Declare a private class property: `private $foo;`
6. Create a constructor and initialize this property with an instance of an object with a circular reference. My code uses an object of class `ProcessWire` (<https://github.com/processwire/processwire>).
7. Set up a local SP (say, using SimpleSAMLphp), configure this SP for the IdP and vice-versa.
8. Specify your custom auth proc filter for this SP (which would be `test:TestFilter`).
9. Test a SAML login request through this SP for this IdP.
10. Observe the recursion error.

**Expected behavior**
There should be no recursion error, and SimpleSAMLphp should finish the login request.

**Screenshots or logs**
![image](https://github.com/user-attachments/assets/071e2bb9-1ec6-4c5f-a9c7-b737cac1230a)

tvdijen:
[...]

PR #2368, opened 2025-01-12
Use print_r to output the filters to handle cycles

monkeyiq:
Using `print_r` will allow `__debugInfo` to customize how things are print/dumped. This should allow for cases with cycles to be handled without running out of memory.

Relates to [issues/2367](https://github.com/simplesamlphp/simplesamlphp/issues/2367)

I also found that although this is a Logger::debug the string being passed to the function is probably being evaluated all the time. Even if debug is not on the export, dump, or print_r would run. To help with this I have the debugActive() method to only run the `debug()` call if debug is active. I have some testing using a getString() function and that will always be called when used as part of the string concat to the Logger::debug() method call. As the dumping could be expensive and is not used at all if not debugging this might be handy to have.

Formatting needs a phpcs or whatnot and the new methods in Logger need some docs. This is mainly an initial drop to discuss.

### DIFF
--- a/src/SimpleSAML/Auth/ProcessingChain.php
+++ b/src/SimpleSAML/Auth/ProcessingChain.php
@@ -89,8 +89,10 @@ class ProcessingChain
             self::addFilters($this->filters, $spFilters);
         }
 
-        Logger::debug('Filter config for ' . $idpMetadata['entityid'] . '->' .
-            $spMetadata['entityid'] . ': ' . str_replace("\n", '', VarExporter::export($this->filters)));
+        if( Logger::debugActive()) {
+            Logger::debug('Filter config for ' . $idpMetadata['entityid'] . '->' .
+                          $spMetadata['entityid'] . ': ' . str_replace("\n", '', print_r($this->filters,true)));
+        }
     }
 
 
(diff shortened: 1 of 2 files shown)


rharish101:
Can confirm that this solves my issue in #2367 with ProcessWire. Thanks a lot!

tvdijen:
@monkeyiq Maybe drop the changes to the logger for now.. I don't think the evaluation is very problematic since it's only a matter of concatenating a few strings.

monkeyiq:
@tvdijen I have removed the `debugActive` update. It is probably best to be in another PR because it is just about efficiency and not cycle links in serializing the auth proc filter list.

The concating of a few strings isn't the issue with debugActive. It is mostly the `var_dump` or `print_r` which I think may take a lot more time. Those variable to string functions are run each time and will also run even if DEBUG is not enabled. The parameter to debug() will be evaluated to be passed to it only to be thrown away inside the function if DEBUG is not active. 

Thinking it over a bit more we could have a `print_r` type function in SSP which only returns a result if the deglevel is sufficient. Maybe a `Logger::print_r()` which would 
```
if(self::$logLevel < self::DEBUG) { 
  return ""; 
} 
```

But I think discussions about timing should be delayed from here to the PR for a possible `debugActive()`. It might be best for me to also create some small ProcessChains to gather some real timing. It will be interesting to see how efficient the dump/print functions are here :)

monkeyiq:
So 11 simple procs in a ProcessChain takes around 15 microseconds to `print_r`. Concat on a string 4-5 times takes around 1 μsec. So a few numbers.

PR #2372, opened 2025-01-20
Configurably allow "Test Authentication Sources" for saml:SP authsources.

jesse-elentra:
I am looking for a way to bring back the "Test Authentication Sources" page for having users test basic SSO login.
While there is the autotest module, it is less refined for a human to interact with.

This PR introduces a configuration option `sp.public.test` which when `true`  allows public access to `module.php/admin/test`. When not logged in as administrator, the listed sources are limited to `saml:SP` sources.

### DIFF
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -294,6 +294,12 @@ $config = [
     ],
      */
 
+    /*
+     * Allow public access to the admin module "Test Authentication Sources"
+     * with only saml:SP authsources listed.
+     */
+    'sp.public.test' => false,
+
 
     /************************
      | ERRORS AND DEBUGGING |
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -294,12 +294,6 @@ $config = [
     ],
      */
 
-    /*
-     * Allow public access to the admin module "Test Authentication Sources"
-     * with only saml:SP authsources listed.
-     */
-    'sp.public.test' => false,
-
 
     /************************
      | ERRORS AND DEBUGGING |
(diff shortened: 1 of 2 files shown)


tvdijen:
I think it's perfectly fine to leave out the config-flag.

jesse-elentra:
> I think it's perfectly fine to leave out the config-flag.

I was trying my best to not change the current functionality unless overridden.
I do agree with you, though.

My understanding is that the test page was locked behind admin authentication to hide IdP links which this still hides, so I personally don't see a security risk.

tvdijen:
Yes, correct! Exposing authsources that are not of the `saml:SP` type _could_ impose a security risk, so that's why we locked it.

The nicest solution would probably be to move this to a separate 'support'-module that people can enable/disable.

monkeyiq:
I have ported this to the 2.3 branch for ease of testing. I see that if I know the SP name I can perform a test login with that selected SP. With just 2.3 installed SSP will ask me to login as admin first.

monkeyiq:
Leaving this as part of the admin module does also mean that the menu is exposed for the admin page. Though you have to authenticate as admin in order to change the page.

monkeyiq:
I am splitting this into a module `splogintest`. This ability has been requested a number of times and if it is in a separate module then it will be available and we can avoid the admin checks because that module can be built to work as a non admin.

monkeyiq:
The module is up at https://github.com/simplesamlphp/simplesamlphp/pull/2381

Maybe we want to split this out into a separate module repo instead of being in the main repo.

monkeyiq:
I have this module fairly much ready now. Once I work out CI I think it should be merged.  

Please feel free to take a look and comment if things could be better https://github.com/simplesamlphp/simplesamlphp/pull/2381

monkeyiq:
I have taken code from this PR and created a new module debugsp which is merged into 2.4 and master. This should bring back the ability for folks to test SP logins if desired without needing admin rights on the machine.

PR #2381, opened 2025-02-05
A new module to allow debugging SP logins

monkeyiq:
The ability to test SP logins by non admin users was available in SSP 1.x. In the 2.x series only the admin can do such testing in the admin module.

This module allows testing SP logins by non admin users. At the moment there is no login needed in order to test an SP.

This is based on the code from
https://github.com/simplesamlphp/simplesamlphp/pull/2372

I have not brought forward the translations that the code was using in the admin module. I will add that as a subsequent update once we know we want this module and where we are wanting to place it.

I have left the list of SP available. Probing for an SP that does not exist is not an error, it just shows the list again. Trying to use a non SP also just shows a list of valid SPs that you might try. This way you can not try to guess the name of an auth mechanism as attempts at both non existing and non SP return the same result.


modules/debugsp/templates/status.twig
```diff
@@ -0,0 +1,100 @@
+{% set pagetitle = 'SimpleSAMLphp installation page'|trans %}
```
tvdijen, line 1: This seems like an exact copy.. Can't we just use the `auth_status.twig` from the base templates-dir?

monkeyiq: Much cleaner using the base template. I had to define a trackid to help the base template. This came from the status.twig in the admin module so it is also very similar to the base `auth_status.twig`.

monkeyiq: This should be resolved in https://github.com/simplesamlphp/simplesamlphp/pull/2381/commits/8aeee60314b0862d6bdfc707793b9240d666101a

monkeyiq:
With the little updates and docs this should be ready to go now? I will revisit it once the Psalm issue is resolved. I would like for the other CI tests to pass as well and the Psalm thing has to be addressed anyway.

monkeyiq:
There is a little issue in the later CI which I might look at to get more green here.

tvdijen:
Needs another rebase and possibly an update of `xml-common` to v1.22.x if that's not already in the lock-file

monkeyiq:
CI is winning! Thanks @tvdijen :)

PR #2386, opened 2025-02-11
Update to supress errors for many tests items for now.

monkeyiq:
Some psalm working is better than failing and preventing other CI from running.


monkeyiq:
I think the validDuration error in subsequent CI items is better to handle in another PR.

This one is only psalm stuff.

Issue #2384, opened 2025-02-11
Overriding translations in a theme only works for EN


tvdijen:
Overriding translations in a custom theme only works for English. Once I switch to another language, the original text is shown instead of the custom translation from my theme.

To reproduce:

- Set `'theme.use' => 'themetim:tim',`
- Set `'language.available' => ['de', 'en', 'nl', 'fr', 'es'],`
- Set `'language.default' => 'en',`
- Deploy attached theme inside `modules`-directory.

[themetim.tar.gz](https://github.com/user-attachments/files/18752596/themetim.tar.gz)

- Go to admin-login and switch languages to see that only the English language show the custom translation.

monkeyiq:
IIRC It was a long time ago when the domains were being ignored. I am happy to take a look at this. I might do a bit of digging at CI/Psalm first.

monkeyiq:
I am also seeing this. With just the following header in `themetim/locales/fr/LC_MESSAGES/themetim.po` the override (`MY CUSTOM TRANSLATION`) for French is used. otherwise it falls back to the default longer real French string.
```
msgid ""
msgstr ""
"X-Domain: themetim\n"

```

monkeyiq:
I am wondering if we should either just override the X-Domain or atleast issue a Logger message if it is not the expected or the empty string. Other values probably indicate an error such as copying the translation file and not updating the X-Domain setting and not realizing that the translation might not be found because of that.

PR #2392, opened 2025-02-19
i18n: set the domain for a theme if it is missing

monkeyiq:
If there is no header in the po file then it was not being found for a theme. This PR allows po files to exist without any header such as the below.

msgid ""
msgstr ""
"X-Domain: worldpeace\n"

This was reported here https://github.com/simplesamlphp/simplesamlphp/issues/2384

I have verified using the theme in that issue and with this PR the "MY CUSTOM TRANSLATION" should appear when one selects French and the po file lacks any explicit header.


tvdijen (approved):
Cool! At least we have restored historical behaviour now!

PR #2401, opened 2025-03-11
Session getData take into consideration the expiration timestamp

ioigoume:


tests/src/SimpleSAML/SessionTest.php
```diff
@@ -47,4 +48,29 @@ public function testSetRememberMeExpireExplicit(): void
 
         $this->assertEquals(time() + 1000, $this->session->getRememberMeExpire());
     }
+
+    public static function getAllowedExpired(): array
+    {
+        return [
+            'Fetch Expired Entries' => [true],
+            'Do not Fetcn Expired Entries' => [false],
```
tvdijen, line 56: typo

src/SimpleSAML/Session.php
```diff
@@ -964,6 +965,10 @@ public function getData(string $type, ?string $id): mixed
             return null;
         }
 
+        if (!$allowExpired && $this->dataStore[$type][$id]['expires'] < time()) {
```
tvdijen, line 968: This should also take into account the case where expiration was set to string `self::DATA_TIMEOUT_SESSION_END`

tests/src/SimpleSAML/SessionTest.php
ioigoume: done

src/SimpleSAML/Session.php
ioigoume: done

ioigoume:
@tvdijen review comments addressed

monkeyiq:
I tried to think of a cleaner way to handle the `self::DATA_TIMEOUT_SESSION_END` case. At the moment it is `assert`ed that if `expires` is a string then it must be that value. I guess how much of an issue it might be depends on how readily we might expect to allow expires to be another string type that means something other than does not expire.

monkeyiq (approved):
This looks good to me.

monkeyiq:
It looks like something has made CI / psalm angry again :/ Unlikely this PR.

ioigoume:
> I tried to think of a cleaner way to handle the `self::DATA_TIMEOUT_SESSION_END` case. At the moment it is `assert`ed that if `expires` is a string then it must be that value. I guess how much of an issue it might be depends on how readily we might expect to allow expires to be another string type that means something other than does not expire.

I thought about that too. This is why i added relevant tests. If SSP allows more strings in the future the tests will raise a flag.

ioigoume:
> It looks like something has made CI / psalm angry again :/ Unlikely this PR.

these are php8.3 related errors.

monkeyiq:
picked into master at https://github.com/simplesamlphp/simplesamlphp/commit/3417b67f04d6824247ee47404484ad64d262933e

Issue #28, opened 2014-02-27
Support metadata DiscoveryResponse for discovery service (was #363)


sspghost:
_From [andreass...@gmail.com](https://code.google.com/u/102889613336421617108/) on November 29, 2010 10:40:01_

We need to support stuffing the idpdisc:DiscoveryResponse in metadata; and more importantly in the discovery service, by an optional configuration, restrict the receivers of responses from the discovrey service to hostnames listed in metadata (of trusted SPs).

As specified in the IdP Discovery profile: http://docs.oasis-open.org/security/saml/Post2.0/sstc-saml-idp-discovery.pdf

_Original issue: http://code.google.com/p/simplesamlphp/issues/detail?id=363_

tvdijen:
Is this still relevant in 2018?

jaimeperez:
I guess the one who could best answer that question is @andreassolberg.

jiscfoo:
> Is this still relevant in 2018?

It's relevant in 2019 :-) At least for people looking to use the UK Federation's Central Discovery Service… the attribute is required in the metadata for the SP in order to use it so currently requires manually adding upon registration.

thijskh:
A pull request is most welcome

alexstuart:
> A pull request is most welcome

Can you provide pointers into which part of the codebase need modifying. Presumably it's only the metadata generation part of the code, because when the UK federation has manually added this endpoint, we've generated it from the URLs in the `AssertionConsumerService` endpoint.

skyforest:
What ever happened with this?

tvdijen:
Support was added to the saml2-library to publish the discovery service in the metadata, but that's about it.

sunetzacharias:
This is now prompted as a warning to the end user in the discovery service SeamlessAccess (RA21 / thiss.io) if the federation metadata of this service provider does not provide this information.

thijskh:
I've observed that and it should not be too hard to fix in SSP. It does need adding DiscoveryResponse support to the 4.x branch of the saml2 lib.

thijskh:
It has been merged for SimpleSAMLphp 2.4

madsi1m:
this is causing me this error from simplesamlphp-module-metarefresh:

XML parser returned: Expected a value identical to "urn:oasis:names:tc:SAML:profiles:SSO:idp-discovery-protocol". Got: "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect

[...]

PR #2402, opened 2025-03-14
Backport support for IDP Discovery protocol

tvdijen:
Closes https://github.com/simplesamlphp/simplesamlphp/issues/28

I know next to nothing about this specification, but I've made the following assumptions:

- Multiple entries of this element are allowed
- These entries only make sense in the Extensions of an SPSSODescriptor

I've verified that:
- Setting this in a `saml:SP`-entry results in an entry in the SP-metadata
- Parsing XML-metadata containing this entry results in a proper SSP metadata-array.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "simplesamlphp/assert": "^1.1",
         "simplesamlphp/composer-module-installer": "^1.3",
         "simplesamlphp/saml2": "^5@dev",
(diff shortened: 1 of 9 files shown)


thijskh (approved):
Great work.

One I think crucial thing was missing and that is that we by default add our own discoresponse URL to the metadata we generate. So that all SPs that use default endpoints will publish it so it can be used by the discovery services. I've added that now.

PR #2403, opened 2025-03-16
Throw exception also when ReturnTo is specified but empty

thijskh:
Some bots generate requests that have an empty ReturnTo. This causes a lot of load (e.g. generating a discovery screen) while it's obviously wrong. Therefore check for this explicitly early on and stop processing.

### DIFF
--- a/modules/saml/src/Controller/ServiceProvider.php
+++ b/modules/saml/src/Controller/ServiceProvider.php
@@ -165,6 +165,13 @@ class ServiceProvider
         ) {
             throw new Error\BadRequest('Missing ReturnTo parameter.');
         }
+        if (
+            $request->query->has('ReturnTo') &&
+            $request->query->getString('ReturnTo') === ""
+        ) {
+            throw new Error\BadRequest('Empty ReturnTo parameter specified.');
+        }
+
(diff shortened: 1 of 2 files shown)


modules/saml/src/Controller/ServiceProvider.php
```diff
@@ -165,6 +165,14 @@ protected function loginHandler(
         ) {
             throw new Error\BadRequest('Missing ReturnTo parameter.');
         }
+        if (
+            $request->query->has('ReturnTo') &&
```
tvdijen, line 169: I think this entire block can be replaced with `$request->query->getString('ReturnTo') === ''`.
If the parameter doesn't exist, the function will return `''`

thijskh: Indeed, much simpler

thijskh: Yeah, but it does need to exist, but be empty for this exception to trigger. The rest is handled above and is more complex (it's ok if it's completely missing)

thijskh: I've simplified the code a bit with getString()

tvdijen (approved):

PR #2405, opened 2025-03-18
skip test if xdebug.mode does not contain develop

verfriemelt-dot-org:
usually if have the extension xdebug loaded but configured to off. so test like this:

```php
    #[RequiresPhpExtension('xdebug')]
    #[RunInSeparateProcess]
    public function testSetCookieSameSite(): void
```

will run but fail due to `xdebug_get_headers()` not being defined. so i added checks to mark them as skipped, if the extension is not configured as expected

### DIFF
--- a/tests/src/SimpleSAML/SessionHandlerPHPTest.php
+++ b/tests/src/SimpleSAML/SessionHandlerPHPTest.php
@@ -72,6 +72,10 @@ class SessionHandlerPHPTest extends ClearStateTestCase
     #[RunInSeparateProcess]
     public function testSetCookie(): void
     {
+        if (!in_array('develop', xdebug_info( 'mode' ))) {
+            $this->markTestSkipped('xdebug.mode != develop');
+        }
+
         Configuration::loadFromArray($this->sessionConfig, '[ARRAY]', 'simplesaml');
         $sh = SessionHandlerPHP::getSessionHandler();
         $sh->setCookie('SimpleSAMLSessionID', '1');
@@ -95,6 +99,10 @@ class SessionHandlerPHPTest extends ClearStateTestCase
     #[RunInSeparateProcess]
     public function testSetCookieSameSiteNone(): void
     {
+        if (!in_array('develop', xdebug_info( 'mode' ))) {
+            $this->markTestSkipped('xdebug.mode != develop');
+        }
+
         Configuration::loadFromArray(
             array_merge($this->sessionConfig, ['session.cookie.samesite' => 'None']),
             '[ARRAY]',
@@ -115,6 +123,10 @@ class SessionHandlerPHPTest extends ClearStateTestCase
     #[RunInSeparateProcess]
     public function testSetCookieSameSiteLax(): void
     {
+        if (!in_array('develop', xdebug_info( 'mode' ))) {
+            $this->markTestSkipped('xdebug.mode != develop');
+        }
+
         Configuration::loadFromArray(
             array_merge($this->sessionConfig, ['session.cookie.samesite' => 'Lax']),
             '[ARRAY]',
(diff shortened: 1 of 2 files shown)


tvdijen:
Would it work if we add another PHP attribute `#[RequiresFunction('xdebug_get_headers')]` ?
I don't like the idea of fixing this in code, because it kind of defeats the purpose of the attributes..

verfriemelt-dot-org:
did not tried that, but tried the RequiresConfiguration :thinking: which did not work. but that could work, let me test.

verfriemelt-dot-org:
does not work as expected :thinking: 

```shell
$ XDEBUG_MODE=off php --rf xdebug_get_headers
Function [ <internal:xdebug> function xdebug_get_headers ] {

  - Parameters [0] {
  }
  - Return [ array ]
}
```

the method is defined - even with mode=off - but returns just an empty result:

```shell
$ XDEBUG_MODE=off php -r 'header("foo: bar"); var_dump(xdebug_get_headers());'
array(0) {
}
$ XDEBUG_MODE=develop php -r 'header("foo: bar"); var_dump(xdebug_get_headers());'
array(1) {
  [0] =>
  string(8) "foo: bar"
}
```

tvdijen:
How did you try `RequiresSetting`?  I think it takes an ini-setting, so you would have to do `#[RequiresSetting('xdebug.mode', 'develop')]`

verfriemelt-dot-org:
> tried the RequiresSetting 🤔 which did not work.

yeah, first thing i tried, did not work either

monkeyiq:
This is not a 100% ideal workaround, but... In bootstrap.php I have
```
if (in_array('develop', xdebug_info( 'mode' ))) {
    function ssp_debug_mode() {
    }    
}
```

Then in one of my local phpunit tests I have
```
#[RequiresFunction('ssp_debug_mode')]
public function testEmpty(): array
{
   $this->assertEmpty(array(123444));
    ...
}
```

Then on the command line the following will show me 1 skipped test
```
XDEBUG_MODE=off  phpunit10 
```

and the following tries to execute the test which fails.
```
XDEBUG_MODE=develop  phpunit10 
```

This was a bit of me trying a few things until something stuck by way of trying to keep the dirty updates in bootstrap.php and just use the normal pragmas in the test code.

tvdijen:
The putenv solution doesn't work.. The env is set too late in the process.
I have an alternate solution here which I think is a little cleaner than the one in this PR: https://github.com/simplesamlphp/simplesamlphp/commit/dc407482d9e23a777a2b987e8fa8c29aef995d21

@verfriemelt-dot-org Could you verify that the commit above solves your issue?

monkeyiq:
The `Depends('testXdebugMode')` seems to work here. I'll do more testing on it tomorrow. Much cleaner than my hack using a global function!

tvdijen:
> The `Depends('testXdebugMode')` seems to work here. I'll do more testing on it tomorrow. Much cleaner than my hack using a global function!

I agree it's much less hacky, that's why I came up with it! If we have more cases like this, we might even offload the 'configuration tests' to a separate class and use 'dependsOnExternal'.. This would eliminate the need to do the xdebug-mode test twice in different test-classes.

tvdijen:
I've merged my version of the patch. Thanks all!

verfriemelt-dot-org:
works :)
thank you!

```shell
$ phpunit --filter testSetCookieSameSite
PHPUnit 10.5.45 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.32
Configuration: /home/easteregg/src/simplesamlphp/phpunit.xml

SSSS                                                                4 / 4 (100%)

Time: 00:00.016, Memory: 10.00 MB

There was 1 PHPUnit test runner warning:

1) XDEBUG_MODE=coverage (environment variable) or xdebug.mode=coverage (PHP configuration setting) has to be set

WARNINGS!
Tests: 4, Assertions: 0, Warnings: 1, Skipped: 4.
```

PR #2408, opened 2025-03-19
i18n warn if a module has an unexpected domain in its po file.

monkeyiq:
This is a follow up to the theme domain handling update from https://github.com/simplesamlphp/simplesamlphp/issues/2384

If you copy the translations for an existing theme or module and forget to change the domain setting in the po file then it might not be found by SSP. With this warning message you will at least have some indication as to why this might have happened.


tvdijen (approved):

tvdijen (approved):

monkeyiq:
I picked this forward to master and back to 2.3. Since it is a warning it should be fine as a bug fix and may help folks who are not getting the translations they are expecting.

PR #2410, opened 2025-03-20
Clear caches script and hook

ioigoume:


tvdijen:
This should be converted to a symfony console command

ioigoume:
> This should be converted to a symfony console command

@tvdijen if i understand this correctly the symfony `cache:clear` will handle the `../cache/core` directory. What about the modules cache directories?

ioigoume:
The updated version now clears the cache using the symfony way. It clears the cache for all loaded modules.

ioigoume:
Screenshot from 2025-03-20 17-19-37

tvdijen:
Already much better! But I meant to say it should be converted to a class extending `\Symfony\Component\Console\Command\Command` (just like for example `src/SimpleSAML/Command/UnusedTranslatableStringsCommand.php`)

monkeyiq:
I am in the process of digging into this. It will be very handy as I have in the past gotten some of the cache stuff lodged and had to do a manual clearing out.

monkeyiq:
This may be a catch 22. For the command to show up I had to do the following

```
php bin/console cache:clear --no-warmup
```

Then the command would be found with
```
php bin/console
```

Not the end of the world, but it might break things that expect the command to be found and it is not because of the old cache.

ioigoume:
> This may be a catch 22. For the command to show up I had to do the following
> 
> ```
> php bin/console cache:clear --no-warmup
> ```
> 
> Then the command would be found with
> 
> ```
> php bin/console
> ```
> 
> Not the end of the world, but it might break things that expect the command to be found and it is not because of the old cache.

@monkeyiq you tried to list all the commands:
```php
./bin/console list
```
and it was not there? Auto-wiring should have taken care of this without needing a clear cache or anything else.

monkeyiq:
Maybe I was doing something wrong but it took a moment for me to get it to turn up in the commands list. I will try to replicate that situation later today.

Another issue I found which might only apply to me was that I made the cache directory owned by and writable by apache. It wasn't 777 protected as I don't see that as necessary. This meant that, for me, I couldn't clear the cache as my normal user but had to run that command as apache. I am not sure how many would share my configuration but if they did it might also pose an issue with running the cache clearing from composer.

monkeyiq (approved):
This is looking good. I think we might need to add some cache:clear stuff to the composer file so that we can switch to older versions like 2.3 and back and not have symfony command not found messages.

monkeyiq:
I guess the developer permissions on `/var/cache/ssp` is another issue. Perhaps we should check for that and either do no cache clear in the composer or fail and provide a link to an explanation as to why we failed. 

 It might be better to not clear the cache if permissions on the top level directory do not allow it so that we can use composer to setup a development env where that cache is not writable by the developer?

monkeyiq (changes requested):
I guess the `cache` issue and permissions on `/var/cache/ssp` issues are better addressed before a merge.

ioigoume:
> Maybe I was doing something wrong but it took a moment for me to get it to turn up in the commands list. I will try to replicate that situation later today.
> 
> Another issue I found which might only apply to me was that I made the cache directory owned by and writable by apache. It wasn't 777 protected as I don't see that as necessary. This meant that, for me, I couldn't clear the cache as my normal user but had to run that command as apache. I am not sure how many would share my configuration but if they did it might also pose an issue with running the cache clearing from composer.

@monkeyiq i noticed that when i ran the `cache:clear` command the permissions of the cache directory changed. I will look into that.

monkeyiq:
It is interesting. Perhaps the cache:clear should be run as the web server uid? If I gave myself permission to that directory then I notice that the webserver was having some trouble after I run cache:clear as my personal uid. Visiting the `admin` module in the browser...
```
Caused by: RuntimeException: Unable to create the "cache" directory (/var/cache/sspdev/admin
```
I will have to do more reading about these commands.

monkeyiq:
ok, so after running the two setfacl commands based on the ones in  https://symfony.com/doc/current/setup/file_permissions.html I can cache:clear and ssp-cache clear and continue using that ssp install in the browser ok. I have some memory of these commands but probably ran them on `/var/cache/sspdev` as I did now to limit the scope and maybe deleted that directory at some stage along the way wiping away the acl stuff in the process.

```
 (pr-2410)]$ php bin/console cache:clear --no-warmup
 (pr-2410)]$ php bin/console ssp-cache:clear --no-warmup
```

monkeyiq:
Hopefully with https://github.com/simplesamlphp/simplesamlphp/pull/2410/commits/9dbb14d63f3923b6a23b349b691c6a4b2ec5fe54 composer should be able to find the ssp-cache command. It may come down to documenting the setfacl requirement to provide normal user and web server dual access and these console commands to work together.

monkeyiq:
It might also be an idea for me to pick that docs update back to 2.3 even if the rest of the update is only in 2.4+.

phpcs.xml
```diff
@@ -16,6 +16,7 @@
 
     <exclude-pattern>modules/adfs/*</exclude-pattern>
     <exclude-pattern>public/assets/*</exclude-pattern>
+    <exclude-pattern>config/*</exclude-pattern>
```
tvdijen, line 19: Why do we have to completely ignore the config-directory? Can't we suppress a specific issue if there is something we cannot fix?

ioigoume: @tvdijen  We do not. I did this for consistency. Previously we were including the config directory but no file was taken into account since all files were template ones. Do you want me to include the config directory and retest?

ioigoume:
@monkeyiq i am not sure how to proceed. Should i make any changes?

monkeyiq:
I will take another look at this today. 

I added the composer cache clear stuff in there to do that before trying the ssp-cache clear runs so that if a dev is bouncing from 2.2 or 2.3 up to 2.4 then it should be able to find the command with no issue. For my noted ssp cache permissions I have updated the docs (which already metioned the ACL commands from the Symfony website) to be a little stronger and explain the possible reach out to the cache directory from composer commands.

monkeyiq (approved):
I am happy with this.

thijskh (approved):

PR #2415, opened 2025-04-08
Move to direct execution of these to allow env to find php

monkeyiq:
Sometimes folks might have php executable as php84 for example and the env command might be needed to find or resolve the php in use.


tvdijen (approved):

monkeyiq:
merged and picked to master

Issue #2425, opened 2025-04-24
Unrecognized key error


tvdijen:
> The `'expire'` key provided here results in a `Warning:  session_set_cookie_params(): Argument #1 ($lifetime_or_options) contains an unrecognized key "expire" in ...\src\SimpleSAML\SessionHandlerPHP.php on line 345` when logging out of `debugsp/test`. 

 _Originally posted by @goozak in [b02d7b3](https://github.com/simplesamlphp/simplesamlphp/commit/b02d7b3739aa5c8efbadf476e270f27ff710e132#commitcomment-155840689)_

goozak:
I wasn't sure if it warranted a full bug report...

Acting as SP
SimpleSAMLphp: 2.4.0 Light (with `'module.enable' => [ 'debugsp' => true, ],` in `config.php`)
PHP: 8.4.5 (with `error_reporting = E_ALL` and `display_errors = On` in `php.ini`)
Platform: Windows
Webserver: IIS

After login in as an SP from `debugsp/test` and clicking the "Logout" button, we get the warning.

I was just wondering if author expected the `'expire'` key to actually do something specific.

If I turn off error reporting, the "Logout" completes as expected, with the `SimpleSAMLAuthToken` cookie removed. and a new `SimpleSAML` cookie.

tvdijen:
You mention `debugsp`..  Is that the new 2.4 module you're using?

goozak:
Yes, but the code that generate this warning is not related to this module.

monkeyiq:
If I dump the `$cookieParams` just before `session_set_cookie_params` in 2.3 and 2.4 I see that in 2.3 the 'expire' key is missing. The other cookie params seem to be the same. Digging...

monkeyiq:
That makes sense the call isn't there is 2.3.

goozak:
Maybe it's a confusion with the `expires` (with an `s`)  parameter of the `setcookie` function?

monkeyiq:
It looks also like `session_set_cookie_params()` wants to use a `lifetime` option instead of an `expires` that `setcookie()` wants.
https://www.php.net/manual/en/function.session-set-cookie-params.php
https://www.php.net/manual/en/function.setcookie.php

monkeyiq:
[...]

PR #2426, opened 2025-04-25
Scrub the expire key for session_set_cookie_params

monkeyiq:
Scrub the `expire` key for `session_set_cookie_params`

If that is set then the session_set_cookie_params() will complain. It
is probably best to convert it to working on lifetime but only if
lifetime is also not explicitly set.

Util\HTTP::setCookie() handles the expire and passes that on to the php API.
https://github.com/simplesamlphp/simplesamlphp/blob/93ca2d0043ae1175192ab4aded579e328a3ee4fa/src/SimpleSAML/Utils/HTTP.php#L1117

This relates to https://github.com/simplesamlphp/simplesamlphp/issues/2425

### DIFF
--- a/src/SimpleSAML/SessionHandlerPHP.php
+++ b/src/SimpleSAML/SessionHandlerPHP.php
@@ -342,7 +342,7 @@ class SessionHandlerPHP extends SessionHandler
         }
 
         if (array_key_exists('expire', $cookieParams)) {
-            // Simiar to the Util\HTTP::setCookie()
+            // Similar to the Utils\HTTP::setCookie()
             if (isset($cookieParams['expire'])) {
                 $expire = intval($cookieParams['expire']);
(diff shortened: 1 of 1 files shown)


src/SimpleSAML/SessionHandlerPHP.php
```diff
@@ -341,6 +341,15 @@ public function setCookie(string $sessionName, ?string $sessionID, ?array $cooki
             session_write_close();
         }
 
+        if (array_key_exists('expire', $cookieParams)) {
+            // Simiar to the Util\HTTP::setCookie()
```
tvdijen, line 345: typo

monkeyiq: updated. I think the comment was a last minute thing to mention where some the `isset` part came from. Since we know it is already set, since we just checked array_key_exists.

PR #2427, opened 2025-05-01
update the list of encryption algorithms supported

briskt:
See this thread: https://groups.google.com/g/simplesamlphp/c/qL1VSnMyj7k/m/uWaSvCamAwAJ

### DIFF
--- a/docs/simplesamlphp-reference-sp-remote.md
+++ b/docs/simplesamlphp-reference-sp-remote.md
@@ -279,7 +279,14 @@ The following options can be set:
 ## Encrypting assertions
 
 It is possible to encrypt the assertions sent to a SP. Currently the
-only algorithm supported is `AES128_CBC` or `RIJNDAEL_128`.
+following algorithms are supported:
+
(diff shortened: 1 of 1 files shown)


tvdijen:
Those are already listed under 'sharedkey_algorithms' a few lines down?

briskt:
> Those are already listed under 'sharedkey_algorithms' a few lines down?

That's where I got this list, but was unsure about deleting the incorrect text vs. correcting it. Based on your comment, I assume you want to delete it.

monkeyiq:
It seems from digging around in the code a bit that the constant `RIJNDAEL_128` might not be supported any more. In particular code like in `xml-security/src/Alg/Encryption/AES.php` wants to use AES constants and `xml-security/src/Constants.php` doesn't list rijndael constants. So making the docs just refer to the AES constants is probably a good move.

A similar statement is in `simplesamlphp-reference-idp-remote.md` so should be simplified to use AES. I might add that to a subsequent PR.

PR #2430, opened 2025-05-04
Update the IdP docs to reference the sharedkey_algorithm list

monkeyiq:
Since that list includes both CBC and GCM modes and is explicit with possible values that are full identifiers.


monkeyiq:
This is a bit of a follow up to https://github.com/simplesamlphp/simplesamlphp/pull/2427

PR #2434, opened 2025-05-12
Add exception annotations to PhpDoc docblocks

HenkPoley:
Used PhpStorm and Psalm to thread the thrown exceptions throughout the code. The addition of extra `@throws` should be pretty minimal.

Added checkForThrowsDocblock and checkForThrowsInGlobalScope to psalm.xml

Psalm will now emit:

* **MissingThrowsDocblock**, for exceptions that can be added as `@throws`
* **UncaughtThrowInGlobalScope**, for exceptions that happen at the non-class level. Meaning you can probably get to see the uncaught results of them as an end-user. These probably need some try-catch handling right there or somewhere deeper in the call chain.

### DIFF
--- a/modules/core/src/Auth/Source/AbstractSourceSelector.php
+++ b/modules/core/src/Auth/Source/AbstractSourceSelector.php
@@ -29,6 +29,7 @@ abstract class AbstractSourceSelector extends Auth\Source
      *
      * @param array $info Information about this authentication source.
      * @param array $config Configuration.
+     * @throws Exception
      */
     public function __construct(array $info, array $config)
     {
@@ -54,6 +55,9 @@ abstract class AbstractSourceSelector extends Auth\Source
      *
      * @param \Symfony\Component\HttpFoundation\Request $request The current request
      * @param array &$state Information about the current authentication.
+     * @throws Exception
+     * @throws Error\Exception
+     * @throws \Throwable
      */
     public function authenticate(Request $request, array &$state): ?Response
     {
@@ -72,6 +76,9 @@ abstract class AbstractSourceSelector extends Auth\Source
      * @param \Symfony\Component\HttpFoundation\Request $request
      * @param \SimpleSAML\Auth\Source $as
      * @param array $state
+     * @throws Error\Exception
+     * @throws Exception
+     * @throws \Throwable
      */
     public static function doAuthentication(Request $request, Auth\Source $as, array &$state): ?Response
     {
--- a/modules/saml/src/IdP/SAML2.php
+++ b/modules/saml/src/IdP/SAML2.php
@@ -11,7 +11,7 @@ use DOMNodeList;
 use Exception;
 use Nyholm\Psr7\Factory\Psr17Factory;
 use RobRichards\XMLSecLibs\XMLSecurityKey;
-use SimpleSAML\{Auth, Configuration, Error, IdP, Logger, Module, Stats, Utils};
+use SimpleSAML\{Auth, Configuration, Error, Error\MetadataNotFound, IdP, Logger, Module, Stats, Utils};
 use SimpleSAML\Assert\{Assert, AssertionFailedException};
 use SimpleSAML\Metadata\MetaDataStorageHandler;
 use SimpleSAML\Module\saml\Message;
@@ -62,6 +62,9 @@ class SAML2
      * Send a response to the SP.
      *
      * @param array $state The authentication state.
+     * @throws Error\Exception
+     * @throws Exception
+     * @throws \Throwable
      */
     public static function sendResponse(array $state): Response
     {
@@ -138,6 +141,8 @@ class SAML2
      * \SimpleSAML\Error\Exception $exception  The exception.
      *
      * @param array $state The error state.
+     * @throws Exception
+     * @throws \Throwable
      */
     public static function handleAuthError(Error\Exception $exception, array $state): Response
     {
@@ -218,6 +223,7 @@ class SAML2
      * @param bool                      $authnRequestSigned Whether or not the authn request was signed.
      *
      * @return array|null  Array with the Location and Binding we should use for the response.
+     * @throws Exception
      */
     private static function getAssertionConsumerService(
         array $supportedBindings,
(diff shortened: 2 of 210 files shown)


tvdijen:
I'm worried that this will throw a shitload of errors in the quality-assurance, because you are adding many use-statements for classes that are not _actually_ used inside a class (just as @throws)

HenkPoley:
I see.

So instead of:

```php
<?php

use Foo\Bar\Baz;

/**
 * @throws Baz;
 */
function fizz() {}
```

Do:

```php
<?php

/**
 * @throws \Foo\Bar\Baz;
 */
function fizz() {}
```

I had not seen you have test time checks for these use statements.

Have to say that I largely did this in an automated way, with lots of manual checking because these AST manipulation tools can't leave other things alone. So that way of writing is clearly not what the tooling does. I'll have to see if I can whip something together. I do like Fully Qualified _Class_ Names (FQCNs) in places like tests. Often have to combat PhpStorm 'cleaning' that up.

Would make the difference between `@throws \Exception` (PHP's built-in Exception class) and the one-character-different `@throws Exception` (often some Exception that appears directly in the file's namespace such as `\SimpleSAML\Error\Exception`).

Sometimes the current simplesamlphp code already uses the short `@throws Exception`  syntax. Would you mind having those replaced with with the FQCN?

HenkPoley:
And for example in `\SimpleSAML\Configuration::getConfig()` I would throw `\SimpleSAML\Error\CriticalConfigurationError` instead of \Exception.

tvdijen:
> So instead of [..]:
Yes, that's exactly what I mean!

> I had not seen you have test time checks for these use statements.
Yes, it's a custom sniffer in PHPCS (part of the Quality-step) that sniffs for unused or missing use-statements. We even enforce ordering them alphabetically!

> [..] would you mind replacing them with [..]
Actually, I would love that!

> And for example in \SimpleSAML\Configuration::getConfig() I would throw \SimpleSAML\Error\CriticalConfigurationError instead of \Exception.
Yes, I agree this would be better!

monkeyiq:
I assume there is another git push coming before a final review of this...?

modules/core/src/Auth/Source/AbstractSourceSelector.php
```diff
@@ -5,7 +5,7 @@
 namespace SimpleSAML\Module\core\Auth\Source;
 
 use Exception;
-use SimpleSAML\{Auth, Configuration, Error, Session};
+use SimpleSAML\{Auth, Auth\Source, Configuration, Error};
```
tvdijen, line 8: `Auth\Source` is wrong and should be removed

modules/saml/src/IdP/SAML2.php
```diff
@@ -11,8 +11,8 @@
 use Exception;
 use Nyholm\Psr7\Factory\Psr17Factory;
 use RobRichards\XMLSecLibs\XMLSecurityKey;
-use SimpleSAML\{Auth, Configuration, Error, Error\MetadataNotFound, IdP, Logger, Module, Stats, Utils};
-use SimpleSAML\Assert\{Assert, AssertionFailedException};
+use SimpleSAML\{Auth, Configuration, Error, IdP, Logger, Module, Stats, Utils};
+use SimpleSAML\Assert\{Assert};
```
tvdijen, line 15: same

modules/saml/src/Message.php
```diff
@@ -14,7 +14,7 @@
 use SimpleSAML\SAML2\XML\saml\Issuer;
 use SimpleSAML\SAML2\XML\saml\AuthnContextClassRef;
 use SimpleSAML\SAML2\XML\samlp\RequestedAuthnContext;
-use SimpleSAML\SAML2\XML\samlp\{StatusCode, StatusMessage}; // Status
+use SimpleSAML\SAML2\XML\samlp\{StatusCode}; // Status
```
tvdijen, line 17: same

src/SimpleSAML/Metadata/SAMLBuilder.php
```diff
@@ -5,12 +5,11 @@
 namespace SimpleSAML\Metadata;
 
 use DOMElement;
-use SimpleSAML\{Configuration, Error\Exception, Module, Logger, Utils};
-use SimpleSAML\Assert\{Assert, AssertionFailedException};
+use SimpleSAML\{Configuration, Module, Logger, Utils};
+use SimpleSAML\Assert\{Assert};
```
tvdijen, line 9: same

src/SimpleSAML/Utils/Config/Metadata.php
```diff
@@ -4,7 +4,7 @@
 
 namespace SimpleSAML\Utils\Config;
 
-use SimpleSAML\{Configuration, Logger};
+use SimpleSAML\{Configuration};
```
tvdijen, line 7: same

src/SimpleSAML/Utils/EMail.php
```diff
@@ -7,7 +7,7 @@
 use Exception;
 use InvalidArgumentException;
 use PHPMailer\PHPMailer\PHPMailer;
-use SimpleSAML\{Configuration, Error\ConfigurationError, Logger};
+use SimpleSAML\{Configuration};
```
tvdijen, line 10: same

tvdijen (approved):
Excellent work! If you could fix the few minor comments I will merge this.

HenkPoley:
I'm certainly not finished yet.

HenkPoley:
Since you have your own assertion library, did you mean to use Webmozart\Assert\Assert here: https://github.com/simplesamlphp/simplesamlphp/blob/4994422cba4decd71cf14e4c249b394aa98ecf56/modules/saml/hooks/hook_sanitycheck.php#L5

tvdijen:
Very well, but maybe it's a good idea to spread across multiple PR's, because the bigger they are, the harder to review.

HenkPoley:
I had some better fixes ready. Messed up some git rebase. Made a new checkout. And I copied them the wrong way, so they're gone. Maybe another week.

So.. how's your day?

HenkPoley:
Please **don't** merge this. I think it erases some Exception annotations with help text.

Maybe I'll clean it up later. Probably not this week.

tvdijen:
No worries, I'll leave this open until you sign off!

HenkPoley:
☺️ Made something to help fix people's exception mess in PHP: https://github.com/HenkPoley/PhpDocBlockDoctor

HenkPoley:
@tvdijen meanwhile I've made https://github.com/HenkPoley/PhpDocBlockDoctor

Would it be possible for you to run it, and see if you're happy with the proposed `@throws` annotation fixes?

```
vendor/bin/doc-block-doctor --read-dirs=attributemap,bin,modules,public,src,tests,vendor --write-dirs=attributemap,bin,modules,public,src,tests
```

I could also make a new PR, since this one seems to have some issues now.

Issue #2435, opened 2025-05-13
Release v2.4.0 breaks SLO when using SimpleSAMLphp as a bridge


NoseSnatcher:
**Specifics of your environment**
1. Are you acting as SP/IdP/proxy? **Proxy/Bridge**
2. SimpleSAMLphp: What version are you using? **v2.4.0**
3. PHP: What version are you using? **8.3**
4. Platform: unix or Windows? **unix**
5. Webserver: Apache/Nginx/ISS? **Apache**

**Describe the bug**
When running SimpleSAMLphp 2.4.0 as a bridge to another IdP ([see docs](https://simplesamlphp.org/docs/stable/simplesamlphp-advancedfeatures)) triggering SLO using an SP that is connected to SimpleSAMLphp, raises an `SimpleSAML\Error\NoState: NOSTATE` Error

**To Reproduce**
The overall setup of SPs and IdP is illustrated in the following picture

1. Setup SimpleSAMLphp 2.4.0 as an IdP
2. Add a SAML-SP authsource using another (upstream) IdP
3. Configure the SimpleSAMLphp IdP to use the SAML-SP authsource
4. Add a SAML service provider to the SimpleSAMLphp IdP. At this point LoginRequests from this service provider should be triggering a login at the upstream IdP
5. Have the service provider do a logout with SimpleSAMLphp (i.e. send a SAML LogoutRequest)
6. SimpleSAMLphp raises an exception and shows an error page

**Screenshots or logs**
Stack trace of the error
```
May 13 10:51:58.170401 ERROR [4285395ef7] SimpleSAML\Error\NoState: NOSTATE
May 13 10:51:58.170442 ERROR [f7e0b280f4] Backtrace:
May 13 10:51:58.171680 ERROR [f7e0b280f4] 6 /var/www/simplesamlphp/src/SimpleSAML/Auth/State.php:295 (SimpleSAML\Auth\State::loadState)
May 13 10:51:58.172997 ERROR [f7e0b280f4] 5 /var/www/simplesamlphp/modules/core/src/Controller/Logout.php:398 (SimpleSAML\Module\core\Controller\Logout::resumeLogout)
May 13 10:51:58.174379 ERROR [f7e0b280f4] 4 /var/www/simplesamlphp/vendor/symfony/http-kernel/HttpKernel.php:181 (Symfony\Component\HttpKernel\HttpKernel::handleRaw)
[...]

PR #2438, opened 2025-05-19
Update for cookie deletion on logout.

monkeyiq:
Because a bridge setup runs the IdP and SP (that you are bridging too) on the same installation instance then cookie handling is a little more complicated. The SP should not delete the cookie on logout here as it is shared with the IdP that is bridging to the SP. For that there is a new saml:SPBridge type that can be used which will be the same as the normal SP for now but does not delete the session cookie on logout. For a bridge the IdP and SP are on the same instance so the IdP can be left to delete the session cookie for that SSP install instance.

See simplesamlphp-advancedfeatures.md for information about the saml:SPBridge.

This relates to https://github.com/simplesamlphp/simplesamlphp/issues/2435

This PR also has the IdP deleting its cookie on logout. This can be turned off with the logout_delete_session_cookie (`true` by default) setting for the IdP. This setting allows degenerate installations with the IdP and SP on the same instance to continue to function. It is also a good backup mechanism allowing the new IdP cookie deletion to be turned off if it is causing issues on a bridge or other setup instance where further investigation may be needed.

Somehow there were different git hashes there and instead of merging I rolled back my 2.4 branch and moved it forward. So that should be the same as my v2 branch but with the current remote history.

### DIFF
--- a/docs/simplesamlphp-advancedfeatures.md
+++ b/docs/simplesamlphp-advancedfeatures.md
@@ -18,7 +18,22 @@ A bridge between two protocols is built using both an IdP and an SP, connected t
 To let a SAML 2.0 SP talk to a SAML 1.1 IdP, you build a SimpleSAMLphp bridge from a SAML 2.0 IdP and a SAML 1.1 SP.
 The SAML 2.0 SP talks to the SAML 2.0 IdP, which hands the request over to the SAML 1.1 SP, which forwards it to the SAML 1.1 IdP.
 
-If you have followed the instructions for setting up an SP, and have configured an authentication source, all you need to do is to add that authentication source to the IdP.
+If you have followed the instructions for setting up an SP, and have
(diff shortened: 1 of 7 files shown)


monkeyiq:
I know this needs some php*lint --autofix foo.

This will by default delete the session cookie for the SP and IdP when these instances are different (default configuration).

If you are wanting to make a degenerate install you can have the IdP not delete its cookie so both will run on a single instance. This is also a useful fallback if somebody wishes to disable the IdP cookie deletion for another reason (for example, they are doing it in a hook or other code).

If you are running a bridge the new spbridge class will leave the cookie floating around for the IdP to delete. Since the IdP is running on the same host as the SP in this configuration.

jaimeperez (changes requested):
Please don't merge this. As discussed previously, session cookies don't need to be deleted and doing so might result on unforeseen side effects.

PR #2439, opened 2025-05-19
Fix metadata generation for indexed endpoints with isDefault set

tvdijen:


monkeyiq (approved):

monkeyiq:
It might be an idea to update https://simplesamlphp.org/docs/stable/simplesamlphp-metadata-endpoints.html to mention that `isDefault` is only applicable to indexed endpoint configurations?

PR #2449, opened 2025-05-28
bugfix: log a warning instead on throwing an exception for invalid entityIDs

### LINKED ISSUE #2448: Regression in SP EntityID format on upgrade to 2.4
While upgrading from 2.2.7 to 2.4.1 I encountered the following error:
```
SimpleSAML\SAML2\Exception\ProtocolViolationException: "DOWJONES" is not a SAML2-compliant URI
```

I was able to trace this down to the upgrade in the `saml2` library from 4.x to 5.x.

I did perform an upgrade to 2.3.7, and that worked fine with my existing SPs (saml2 is on 4.x there),  so this is definitely a BC break from 2.3 to 2.4. This also wasn't mentioned in the [Upgrade notes for version 2.4](https://simplesamlphp.org/docs/stable/simplesamlphp-upgrade-notes-2.4.html), so that should at least be updated.

Note: This is different from #1891 where it was dealing with the IdP EntityID, and was much easier to perform the upgrade from 1.x to 2.x.

If this is something we can revert easily for SP EntityID's without interfering with other security validations, that'd be great.

I have multiple external SPs that I don't control, and that aren't URI/URN format. These are largish companies with a track record of ignoring, or taking their time (months to years) to make code changes. While I realize these SP's are likely out of spec, they've been around for a long time in this condition.

Otherwise I'll plan on figuring out how to patch a workaround for this in the short term while I'm performing the upgrade.

tvdijen:
Closes #2448

While throwing an exception is the right thing to do, it may also be a little harsh.
Definitely not something we should do in a minor release.

### DIFF
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -10,6 +10,8 @@ See the [upgrade notes](https://simplesamlphp.org/docs/stable/simplesamlphp-upgr
 Released TBD
 
 * Fixed a bug where metadata-endpoints with isDefault set would not yield the expected metadata (#2439)
+* Fixed a backwards incompatibility that would throw an exception on an invalid entityID.
+  The exception was downgraded to a warning in the log (#2448)
 
 `memcacheMonitor`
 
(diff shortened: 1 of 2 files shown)


monkeyiq:
I have cherry picked this to `master` as well. If we are loading these from a site's config then it is probably best to warn rather than refuse for these. If the SP metadata is coming from external sites which use poor URNs for entityId then we either log and allow or we force the sysadmin to script an update to change things and maybe try to change the entityids mid flight. 

At any rate, I think working here is best for now and we can remove that try/catch if we want to deny such configs in the future.

PR #2454, opened 2025-05-29
Fix unauthorized use of cron with default key

### LINKED ISSUE #2453: Cron CLI can be executed by everyone through the URL
SimpleSAMLphp v2.4.1

https://github.com/simplesamlphp/simplesamlphp/blob/5c3fd07137369a28814f6c7f72e09a9c3e391058/modules/cron/src/Controller/Cron.php#L119

Please see mentioned code above.

If somebody send a request with the key secret or RANDOM_KEY, the $key !== $configKey never executes.

elseifs should be replaced by standalone ifs.

tvdijen:
Closes #2453


modules/cron/routing/routes/routes.yml
```diff
@@ -14,5 +14,4 @@ cron-run:
     output: 'silent'
   }
   methods: [GET]
-#  TODO: Use this instead of controller-code when Symfony 6.1 is available
-#  condition:  "params['key'] not in ['secret', 'RANDOM_KEY']"
+  condition: "params['key'] not in ['secret', 'RANDOM_KEY']"
```
thijskh, line 17: The drawback of this is that it's rather implicit and in a different part of the code than the actual controler, so likely to get lost at some point. Maybe more clear to keep everything around the key together in one place?

modules/cron/src/Controller/Cron.php
```diff
@@ -115,13 +115,7 @@ public function info(): Template
     public function run(string $tag, string $key, string $output = 'xhtml'): Response
     {
         $configKey = $this->cronconfig->getOptionalString('key', 'secret');
```
thijskh, line 117: This is rather implicit. If you forget or misset the key, it will be "secret". Maybe just make it a nonoptional string which would give a much more concrete result?

monkeyiq (approved):

thijskh (approved):

PR #2456, opened 2025-05-30
Update base.twig

DanielBuerckner:
_head.twig does not exist. The header is included in line 17.

Also, this line somehow causes a segmentation fault with PHP 8.2 and PHP 8.4. We are investigating this issue.

### DIFF
--- a/templates/base.twig
+++ b/templates/base.twig
@@ -9,7 +9,6 @@
     <link rel="stylesheet" href="{{ asset("css/stylesheet.css") }}">
     <link rel="icon" href="{{ asset("icons/favicon.ico") }}">
     <meta name="robots" content="noindex, nofollow">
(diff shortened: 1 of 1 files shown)


DanielBuerckner:
Of course, the segfault is caused by something else. It's only recently causing problems with our development computers. Production works without a problem for years now.

We did not realize there is a reason for a missing file.

monkeyiq:
If the segv is caused by something else, perhaps we can close this PR and look at backtraces or other hints for the segv in a github issue.

tvdijen:
segfaults always originate from PHP or on of its extensions, so yeah, not much we would be able to do about it anyway

tvdijen:
Yeah, I've encountered the same issue today, although my stack-trace is different..  Downgrading or removing xdebug solves the issue.

PR #2466, opened 2025-06-12
Warn the user if they do not also set sign.enabled in metadata file

monkeyiq:
This block needs to be enabled in the metadata file to have the effect.


monkeyiq:
I am thinking the final CI failure is not directly related to this PR

tvdijen:
Idk what the issue was, but a `composer update` solved it

Issue #2447, opened 2025-05-23
IdP metadata signing option undocumented and problematic fallback implementation


mrvanes:
## Missing documentation
In `modules/saml/src/Controller/Metadata.php` in function metadata() metadata can be signed by `SSPMetadata\Signer::sign()` depending on boolean `metadata.sign.enable` which is undocumented in `simplesamlphp-reference-idp-hosted.html`

## Failing fallback feature
`SSPMetadata\Signer::sign()` tries to collect the signing key in `findKeyCert()` but fails to _fall back_ to `privatekey` and `certificate` because in `SAML2_IdP::getHostedMetadata()` the `privatekey` and `certificate` keys are not exposed in the internal representation of the metadata, they are stored in the `keys` key after processing, so fallback will always fail and raise the last
```
'Could not find what key & certificate should be used to sign the metadata' .
            ' for the ' . $type . ' "' . $entityMetadata['entityid'] . '".'
```
error.

monkeyiq:
I think I see the problematic fallback now. I have also discovered some other interesting items along with the deep dive. PR(s) to come.

Thank you for reporting this issue.

monkeyiq:
For the fallback I have two PRs, one to allow the fallback and one to explicitly remove it from the docs (https://github.com/simplesamlphp/simplesamlphp/pull/2470).  

At this stage I think the team are leaning toward the later, if a fallback is desired it shouldn't be that hard to use a script to modify the metadata file to copy the relevant fields explicitly to md.sign.privatekey for example. This way there is less chance that the code does something unexpected; if you want to sign metadata you should use the md.sign.privatekey and not plan on an implicit fallback. That is at this stage, discussion may end up in another place but as it seems right now that is the leaning.

monkeyiq:
Thank you for the detailed report. The more we can make things explicit and unsurprising the safer configurations will be for everyone.

monkeyiq:
It was decided that the fallback docs should be removed. This is a good time to do it as the fallback is not working currently.
This was done in https://github.com/simplesamlphp/simplesamlphp/pull/2470

IMHO it is a bit easier for the configuration to explicitly capture which keys to use to sign what. If a fallback is desired it can still be done by using the same key file for an IdP in `metadata.sign.privatekey` for example. Or using an explicit path and just making that symlink to the main `privatekey` if that is what you really want. It seems a bit more likely that you would want an IdP to have its own key/cert files though. Eggs and baskets as they say.

[...]

PR #2469, opened 2025-06-16
Allow the fallback to the pk/cert as mentioned in the docs

monkeyiq:
This was raised in https://github.com/simplesamlphp/simplesamlphp/issues/2447

When you have a hosted IdP with
```
    'metadata.sign.enable' => true,
```

This PR will allow the privatekey, certificate to be used for signing the metadata.

One case that is not handled at the moment is if
metadata.sign.enable is set in the config.php and you are wanting the privatekey and certificate to be used from the hosted IdP file. If metadata.sign.enable is set in config.php then unless you set metadata.sign.enable for that hosted file then things will fail to work (the IdP doesn't show up in the admin/federation page).

This is sort of inline with the restrictions made explicit in https://github.com/simplesamlphp/simplesamlphp/pull/2466

### DIFF
--- a/src/SimpleSAML/Metadata/Signer.php
+++ b/src/SimpleSAML/Metadata/Signer.php
@@ -117,7 +117,8 @@ class Signer
                 $ret['privatekey_pass'] = $entityMetadata['privatekey_pass'];
             }
 
-            Logger::info("Falling back to using privatekey and certificate for the IdP for signing for " . $entityMetadata['entityid']);
+            Logger::info("Falling back to using privatekey and certificate " .
+                         " for the IdP for signing for " . $entityMetadata['entityid']);
 
             return $ret;
         }


monkeyiq:
I will also cast my vote for updating the docs to remove the implicit fallback. As the original github issue mentioned the fallback is not working at the moment then we have a good opportunity to change the documentation to document that we do not intend for any fallback to be in effect. 

If folks are wanting a fallback then it makes sense for that limited user base to update the metadata file to copy the keys for `privatekey` and `certificate` to the `metadata.sign.` prefixed config settings as well. It is better for that group to script that update if needed. Less moving parts in SSP and less fallbacks maybe doing something unintended is a better outcome. Especially if it is near the crypto code.

I should have taken a step back and thought of that as the better way to go rather than digging into the code. But I now have some more familiarity for that code path so it may be useful at some stage.

monkeyiq:
I will make a docs PR as the counter to this one. As this is not time critical I will give time for @thijskh to chime in with his thoughts.

At the moment the expected outcome is to close this PR and merge the new docs update that the fallback to `privatekey`  when `metadata.sign.privatekey` is not provided is not to be expected.

tvdijen:
I'd say instead of 'unexpected' we should make it 'deprecated' and effectively kill it in a next major.

PR #2475, opened 2025-06-18
Explicitly reject HEAD requests for SSO service

procky:
Fix for HEAD requests reaching the controller. I commented in issue https://github.com/simplesamlphp/simplesamlphp/issues/2233. 
I tried to match the error normally logged in Symfony RouterListener.  
Test appear to test core SAML functionality rather than routing edge cases like this but it might be worth it.

### DIFF
--- a/modules/saml/src/Controller/WebBrowserSingleSignOn.php
+++ b/modules/saml/src/Controller/WebBrowserSingleSignOn.php
@@ -116,6 +116,19 @@ class WebBrowserSingleSignOn
      */
     public function singleSignOnService(Request $request): Response
     {
+        if ($request->isMethod('HEAD')) {
+            // These are the allowed methods from routes.yml
+            $allowedMethods = ['GET', 'POST'];
+            $message = sprintf(
+                'No route found for "%s %s": Method Not Allowed (Allow: %s)', 
+                $request->getMethod(), 
+                $request->getUriForPath($request->getPathInfo()), 
+                implode(', ', $allowedMethods)
+            );
+
+            throw new MethodNotAllowedHttpException($allowedMethods, $message);
+        }
+
         Logger::info('SAML2.0 - IdP.SSOService: Accessing SAML 2.0 IdP endpoint SSOService');
 
         if ($this->config->getBoolean('enable.saml20-idp') === false || !Module::isModuleEnabled('saml')) {


tvdijen:
I'm a little confused why this isn't captured by this code:
https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-2.4/src/SimpleSAML/Error/ExceptionHandler.php#L30-L32

procky:
> I'm a little confused why this isn't captured by this code: https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-2.4/src/SimpleSAML/Error/ExceptionHandler.php#L30-L32

https://github.com/simplesamlphp/simplesamlphp/issues/2233#issuecomment-2985884080

In short, Symfony treats HEAD requests the same as GET when matching routes, but in this case, this behaviour causes an issue.

monkeyiq:
Using something like this in ssp-2.4 branch head
```
curl --insecure -I -X HEAD 'https:.../singleSignOnService?SAMLRequest=...'
```
Gives me a 400 bad request. This is due to `Binding::getCurrentBinding()` failing in SAML2.php:388.

```
... simplesamlphp ERROR [450aa91d3c] SimpleSAML\Error\Error: SSOPARAMS

Caused by: SAML2\Exception\Protocol\UnsupportedBindingException: Unable to find the SAML 2 binding used for this request.
```

With this patch I am seeing
```
SimpleSAML\Error\MethodNotAllowed: METHODNOTALLOWED('%MESSAGE%' => 'No route found for "HEAD https://sam/sspdev/module.php/saml/idp/singleSignOnService": Method Not Allowed (Allow: GET, POST)')
and some backtraces...
```

This is using a full URL harvested from SAMLTracer using a login started in Chrome.

I have modified the PR slightly for my 2.4 branch testing. I am happy to do more digging, this was just an initial investigation so far.

monkeyiq:
@tvdijen I think the `MethodNotAllowedHttpException` code you link above is not being used as the HEAD gets through the routing. The Binding.php code will throw a `UnsupportedBindingException` in `getCurrentBinding()` IIRC. If I put a `case 'HEAD':` above the `case 'GET':` in that method then it will run as normal and result in a HTTP 303 see other just as the `GET` issued request would.

tvdijen:
Ah, I understand now. Sorry I haven't been able to take a closer look earlier.

Now, for the solution, I would like to make this a little bit more explicit by defining two router-methods, one matching the GET and process as usual and one matching the HEAD and throwing a `MethodNotAllowedException`.  This new method could be reused to match any other SAML-endpoint as well to consistently deal with HEAD-requests sent to them.

Does that make sense @monkeyiq ?

monkeyiq:
I was concerned that the HEAD might be only matchable as a GET. It seems you can explicitly match the HEAD request though.

monkeyiq:
The pronouns are making me want to clarify. By "It does a check" I am thinking we talking about `It`==Symfony. If that is the case then I from what I have seen and read Symfony will match a HEAD to a GET implicitly. It seems from https://github.com/simplesamlphp/simplesamlphp/pull/2481 that you can explicitly match a HEAD as well. The routes in 2481 look a bit funky at first inspection. I started out there with a match on HEAD

https://github.com/simplesamlphp/simplesamlphp/pull/2481/commits/501443e095067f808fa7c4f2d2897937e681ecee#diff-4d2082df56f6a3e4968cd6b3c08daac50509d0a7acceafb11576d6c951ee9984

From testing for 2481 it seems if you have a HEAD match before the GET match then the HEAD route is used. The current 2481 is a generalization from there.

tvdijen:
No no, by 'It' Thijs means 'OpenConext Engineblock' that I specifically asked about to see how other SAML-products deal with this.

tvdijen:
@procky We took a slightly different approach, so I'm closing this PR!
Thanks for your time and patience! The solution will be in the next release.

PR #2481, opened 2025-07-02
Draft: Another way to return a 405 if calling SSO with HEAD req.

monkeyiq:
This relates to https://github.com/simplesamlphp/simplesamlphp/pull/2475

To play with this, goto admin/test and login with an authentication source. Harvest the SignOnService full link from the web console.

Test on the console with the following...
```
$  curl --insecure -I -X HEAD  'https://fqdn/sspdev/module.php/saml/idp/singleSignOnService?SAMLRequest=...'
HTTP/1.1 405 Method Not Allowed

$  curl --insecure -I -X GET   'https://fqdn/sspdev/module.php/saml/idp/singleSignOnService?SAMLRequest=...'
HTTP/1.1 303 See Other
```

I was also experimenting with rejecting HEAD requests for all single idp lookups with the following. This seems to work but you also loose the ability to get a 404 for links which will never actually work. eg `singleSignOnServiceFOO?SAMLRequest=` will just give a 405 if you wildcard with the below.

```
websso-idp-head:
  path: /idp/{methodname}
  controller: 'SimpleSAML\Module\saml\Controller\WebBrowserSingleSignOn::headRequestNotAllowed'
  defaults:
    methodname: 1  
  methods: [HEAD]
```

### DIFF
--- a/modules/saml/routing/routes/routes.yml
+++ b/modules/saml/routing/routes/routes.yml
@@ -64,6 +64,13 @@ saml-legacy-sp-metadata:
   }
   methods: [GET]
 
+websso-single-sign-on-head:
+  path: /idp/singleSignOnService
+  defaults: {
+    _controller: 'SimpleSAML\Module\saml\Controller\WebBrowserSingleSignOn::headRequestNotAllowed'
+  }
+  methods: [HEAD]
+  
 websso-single-sign-on:
   path: /idp/singleSignOnService
   defaults: {
--- a/modules/saml/src/Controller/WebBrowserSingleSignOn.php
+++ b/modules/saml/src/Controller/WebBrowserSingleSignOn.php
@@ -20,6 +20,9 @@ use SimpleSAML\Logger;
 use SimpleSAML\Metadata;
 use SimpleSAML\Module;
 use SimpleSAML\Store\StoreFactory;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException;
+
 
 /**
  * Controller class for the Web Browser Single Sign On profile.
@@ -107,6 +110,23 @@ class WebBrowserSingleSignOn
         return new RunnableResponse([$binding, 'send'], [$artifactResponse]);
     }
 
+    public function headRequestNotAllowed(): RunnableResponse
+    {
+        Logger::info('Handling a HEAD request by returning method not allowed...');
+        
+        $request = Request::createFromGlobals();
+
+        // These are the allowed methods from routes.yml
+        $allowedMethods = ['GET', 'POST'];
+        $message = sprintf(
+            'No route found for "%s %s": Method Not Allowed (Allow: %s)',
+            $request->getMethod(),
+            $request->getUriForPath($request->getPathInfo()),
+            implode(', ', $allowedMethods)
+        );
+
+        throw new MethodNotAllowedHttpException($allowedMethods, $message);
+    }
 
     /**
      * The SSOService is part of the SAML 2.0 IdP code, and it receives incoming Authentication Requests


modules/saml/src/Controller/WebBrowserSingleSignOn.php
```diff
@@ -107,6 +110,23 @@ public function artifactResolutionService(): RunnableResponse
         return new RunnableResponse([$binding, 'send'], [$artifactResponse]);
     }
 
+    public function headRequestNotAllowed(): RunnableResponse
+    {
+        Logger::info('Handling a HEAD request by returning method not allowed...');
```
tvdijen, line 115: I wouldn't want to see all this in the logs.. Downgrade to 'debug'

modules/saml/routing/routes/routes.yml
```diff
@@ -64,6 +64,13 @@ saml-legacy-sp-metadata:
   }
   methods: [GET]
 
+websso-single-sign-on-head:
+  path: /idp/singleSignOnService
+  defaults: {
+    _controller: 'SimpleSAML\Module\saml\Controller\WebBrowserSingleSignOn::headRequestNotAllowed'
+  }
+  methods: [HEAD]
```
tvdijen, line 72: Instead, use a condition `context.getMethod() not in ["GET", "POST"]`

modules/saml/src/Controller/WebBrowserSingleSignOn.php
```diff
@@ -107,6 +110,23 @@ public function artifactResolutionService(): RunnableResponse
...
+        
+        $request = Request::createFromGlobals();
+
+        // These are the allowed methods from routes.yml
+        $allowedMethods = ['GET', 'POST'];
+        $message = sprintf(
+            'No route found for "%s %s": Method Not Allowed (Allow: %s)',
```
tvdijen, line 122: Since we want to use this controller generically for all SAML-endpoints, we cannot know the 'allowed methods' and have to make this error message a little more generic

tvdijen:
Wildcards should work if you specify all possible 'methodName' in a condition.

modules/saml/routing/routes/routes.yml
monkeyiq: That update is at https://github.com/simplesamlphp/simplesamlphp/pull/2481/commits/dc29a4b7276c7029bf330c73ed23d566e8268df6

modules/saml/src/Controller/WebBrowserSingleSignOn.php
monkeyiq: If there are only a handful of GET,POST combinations we could pass in the allowedMethods to the method from the routes yml so it can pass that info on. https://github.com/simplesamlphp/simplesamlphp/pull/2481/commits/b423796fd418382d97df4abc314c0182a57bf76f

tvdijen: For another PR:  we should probably ditch the routes.yml files and use PHP attribute-annotations instead

tvdijen:
@monkeyiq I hope you don't mind I hijacked your pull request..  It is now the way I had in mind and ready for review

monkeyiq:
@tvdijen It wasn't really a hijacking, more of a collaboration on the solution ;)

monkeyiq:
I am hapoy with the move from WebBrowserSingleSignOn::methodNotAllowed to Exception::methodNotAllowed. It needed to move somewhere more globally accessible for wider scope work.

monkeyiq:
It seems here that the `$config` that is passed to the constructor of SimpleSAML\Module\saml\Controller\Exception is not used. If I leave that in then I get this when I try to execute a HEAD request:

```
The controller for URI "/saml/idp/singleSignOnService" is not callable: 
Controller "SimpleSAML\Module\saml\Controller\Exception" 
has required constructor arguments and does not exist in the container. 
Did you forget to define the controller as a service?
```

Removing that `$config` allows things to work as expected.

monkeyiq:
I also played around with making the Exception classs `extend AbstractController` to no success.

monkeyiq:
I am thinking this might have to do with route caching for me

PR #2490, opened 2025-07-22
Enhance_assert_tag

ioigoume:
This Pull Request tries to address the following issues:
- the cache buster tag can change/renew when a module is updated to a newer/different version
- the cache buster tag doesn’t leak the ssp version

### DIFF
--- a/src/SimpleSAML/XHTML/Template.php
+++ b/src/SimpleSAML/XHTML/Template.php
@@ -37,7 +37,7 @@ use function class_exists;
 use function count;
 use function date;
 use function explode;
-use function hash;
+use function hash_hmac;
 use function in_array;
 use function is_null;
 use function key;
@@ -208,10 +208,20 @@ class Template extends Response
         $file = new File($file);
 
         $tag = $this->configuration->getVersion();
+        if ($module !== null) {
+            // Modules can be updated more frequently than the core. Especially the custom ones.
+            // As a result, we will use a different tagging method
+            $composerLock = new File($baseDir . 'composer.lock');
+            $tag = md5($composerLock->getContent());
+        }
+
         if ($tag === 'master') {
             $tag = strval($file->getMtime());
         }
-        $tag = substr(hash('md5', $tag), 0, 5);
+        // Use the `secretsalt` to enhance security.
+        // Do not make it easy to guess the underlying SSP version.
+        $secretSalt = $this->configuration->getString('secretsalt');
+        $tag = substr(hash_hmac('sha256', $tag, $secretSalt), 0, 5);
 
         return $path . '?tag=' . $tag;
     }
--- a/src/SimpleSAML/XHTML/Template.php
+++ b/src/SimpleSAML/XHTML/Template.php
@@ -182,6 +182,8 @@ class Template extends Response
      */
     public function asset(string $asset, ?string $module = null, bool $tag = true): string
     {
+        $tagLength = 6;
+
         $baseDir = $this->configuration->getBaseDir();
         $basePath = $this->configuration->getBasePath();
         if (is_null($module)) {
@@ -212,20 +214,42 @@ class Template extends Response
             // Modules can be updated more frequently than the core. Especially the custom ones.
             // As a result, we will use a different tagging method
             $composerLock = new File($baseDir . 'composer.lock');
-            $tag = md5($composerLock->getContent());
+            $tag .= $this->getAssetTagForFile($composerLock);
         }
 
         if ($tag === 'master') {
             $tag = strval($file->getMtime());
+        } else {
+            $configfile = new File($this->configuration->getConfigPath());
+            $tag .= $this->getAssetTagForFile($configfile);
         }
         // Use the `secretsalt` to enhance security.
         // Do not make it easy to guess the underlying SSP version.
-        $secretSalt = $this->configuration->getString('secretsalt');
-        $tag = substr(hash_hmac('sha256', $tag, $secretSalt), 0, 5);
+        $salt = $this->configuration->getOptionalString('tagsalt', 'defaulttagsalt');
+        $mac = hash_hmac('sha256', $tag, $salt, true);
+        $tag = substr(base64_encode($mac), 0, $tagLength);
 
         return $path . '?tag=' . $tag;
     }
 
+    /**
+     * Get a string to use in tag generation for a file. This can be done from the contents
+     * of the file (expensive) or the mtime of the file (cheap). If a store is using many SSP
+     * instances and knows that the mtime is synced for the files in the installation then
+     * the cheap method is preferred.
+     */
+    public function getAssetTagForFile(File $f): string
+    {
+        $useContents = false;
+
+        if ($useContents) {
+            $r = md5($f->getContent());
+        } else {
+            $r = ":" . strval($f->getMtime());
+        }
+
+        return $r;
+    }
 
     /**
      * Get the normalized template name.
--- a/src/SimpleSAML/XHTML/Template.php
+++ b/src/SimpleSAML/XHTML/Template.php
@@ -220,12 +220,11 @@ class Template extends Response
         if ($tag === 'master') {
             $tag = strval($file->getMtime());
         } else {
-            $configfile = new File($this->configuration->getConfigPath());
-            $tag .= $this->getAssetTagForFile($configfile);
+            $tag .= $this->getAssetTagForFile($file);
         }
         // Use the `secretsalt` to enhance security.
         // Do not make it easy to guess the underlying SSP version.
-        $salt = $this->configuration->getOptionalString('tagsalt', 'defaulttagsalt');
+        $salt = $this->configuration->getOptionalString('assets.secretsalt', 'default.assets.secretsalt');
         $mac = hash_hmac('sha256', $tag, $salt, true);
         $tag = substr(base64_encode($mac), 0, $tagLength);
 
--- a/src/SimpleSAML/XHTML/Template.php
+++ b/src/SimpleSAML/XHTML/Template.php
@@ -38,7 +38,7 @@ use function class_exists;
 use function count;
 use function date;
 use function explode;
-use function hash_hmac;
+use function hash_hmac_file;
 use function in_array;
 use function is_null;
 use function key;
--- a/src/SimpleSAML/XHTML/Template.php
+++ b/src/SimpleSAML/XHTML/Template.php
@@ -212,8 +212,10 @@ class Template extends Response
 
         // Use the `assets.secretsalt` to enhance security.
         // Do not make it easy to guess the underlying SSP version.
-        $salt = $this->configuration->getOptionalStringNested( 'assets.secretsalt',
-                                                               'assets.secretsalt.default');
+        $salt = $this->configuration->getOptionalStringNested(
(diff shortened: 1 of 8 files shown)


monkeyiq:
I like this update. Sharing the ssp version as part of the tag seems like an unnecessary thing to do.

monkeyiq:
We may also want to ask `hash_hmac` for a binary result and encode that to base64 or ascii85 (or a safe variant for cgi params) before taking the truncation for the tag. 

For a 5 digit lowercase hexit we get a one in a million chance of arbitrary collision, if we ask for binary data from hmac and base64 it we end up with 1 in a billion chance of collision.

If we then extend this to a 10 character prefix of the base64 we get into somewhat large numbers.

monkeyiq:
The main reservation I am thinking about with the use of `hash_hmac` here is that the `tag` passed into it is going to be relatively known to a user. If it is the md5 of the composer.lock file or the release version or the mtime of the composer.lock file etc then a bad actor will either know it from a relatively small set of known values or know a somewhat limited range of values it can be (mtime). 

Then they could if they wanted to do an offline attack on the salt. This is likely just being paranoid but it can help to have that mindset for security related things. Especially if the `secretsalt` is used in other code (I see it is passed do Symfony).

This could be mitigated to some degree by having a second salt just used for this purpose with a slightly added admin/install burden. The potential successful attacker then would be able to work out (release version) + guessed salt = captured value. So they would know the version then. 

I have been looking at some code using `Symfony\Component\Cache` to cache a salt used just for the tag generation. As this is a throw away value the cache seems like a good use. We could tell Symfony to maybe leave a large window for the validity of the value to avoid tags changing to quickly leading to cache misses. 

```
   $salt = $this->cache->get('asset_tag_salt', function (ItemInterface $item): string {
            $item->expiresAfter(3600);
            $salt = random_bytes(32);
            $salt = base64_encode($salt);
            return $salt;
   });
...
   // todo: change to return binary and then use base64
   // such as generateHash() in 
   // Symfony/Component/Security/Core/Signature/SignatureHasher.php
   $tag = substr(hash_hmac('sha256', $tag, $salt), 0, 5);
```

Unfortunately the code I have is still very prototype level. It is just using `Adapter\FilesystemAdapter` which is not the best choice for production.

tvdijen:
I don't think using a store is a solution either @ioigoume .. The data in a store can have a very short lifetime (for example memcached will destroy it oldest record when storage is full)..

The problem we're trying to solve is to update the cache-buster when a module is updated, right?
Why not generate something completely random during `composer install/require/update` and store it in the cachedir?

ioigoume:
> I don't think using a store is a solution either @ioigoume .. The data in a store can have a very short lifetime (for example memcached will destroy it oldest record when storage is full)..
> 
> The problem we're trying to solve is to update the cache-buster when a module is updated, right? Why not generate something completely random during `composer install/require/update` and store it in the cachedir?

I will do that. Thank you for the feedback @tvdijen

tvdijen:
> [..] but I believe @monkeyiq thought was the initial mtime for the main SSP installations `assets` folder is the same and would indicate the ssp version.

Ah right, for people who are using the tarball that might indeed be the case.
My suggestion unfortunately does not cover the second requirement. It's impossible to install on multiple nodes and end up with the _exact_ same timestamp

tvdijen:
OK, after considering all the options, I agree with adding the second secret.. It's the only way to get a consistent result over multiple nodes.

monkeyiq:
The other solution I had was using a bin/update-tag-salt command. But backed off that because it would require some additional knowledge or commands for admins. Even if the tag-salt was generated on one machine you could just be able to specify the same tag-salt on another node in a cluster to remake the same tag for busting purposes. But you then have a general purpose "make the tag-salt" for single nodes and a "copy the tag-salt" for replications.

If you are doing that then maybe just having two salts in config.php is simpler as it is an extension to an already used admin path. Copying the salt is as simple as copying the config.php over nodes. The salt doesn't need updating ever as the SSP version is updated. For production we could use salt+`information`. The information might be the sspversion, or perhaps based in mtimes. At least trying to crack the salt for tags only gains the salt for that purpose. Whatever `secretsalt` is used for is not exposed.

The round robin connection to pool of SSP makes the `cache` use scenario I was looking at much worse than adding a new `tagsalt` to the config.php file.

monkeyiq:
I have an updated version which I can push back to this PR if we are happy with it's direction.

This uses a new 'tagsalt' config with a default value of 'defaulttagsalt'. I have made it optional so that folks do not have to change it and 2.4.x compatibility is maintained. Obviously it is best to actually change the new salt. I also have `getAssetTagForFile` which allows either md5 from contents or mtime. If you have synced mtimes then the mtime will be much faster.

https://github.com/monkeyiq/simplesamlphp/blob/c490b1d75ac3c06bf08f468775a99ebefaa436b4/src/SimpleSAML/XHTML/Template.php#L228

This also puts the mtime of the config.php into the base tag for a released version. With that and the specific salt just used for these tags it should be fairly hard to guess what version is used from the tag. At that point it probably becomes quicker to try to scrape what assets you can get from the site and compare those to released assets to figure it out. And there is no ability to attack the secretsalt used for other things if you successfully attack the tag salt.

monkeyiq:
I'm happy with the config key rename to `assets.secretsalt`. I'll update.

monkeyiq:
easy to remove the mtime of config.php from the code path

ioigoume:
> I'm happy with the config key rename to `assets.secretsalt`. I'll update.

@monkeyiq how do we proceed? I restored the initial commit that uses the secret salt. So this needs to be updated. Also the tests need some refactoring.  Should i copy the changes over the PR? Will you do the changes directly to this PR and then i will update the  tests? What do you think?

monkeyiq:
So the tag is now based on the mtime of the asset regardless of if it is `master` or in a module. This has one side effect that each asset will almost certainly have its own individual hashed tag. In the code in 2.4.x there would be a single tag for all assets for a release. I am not sure this is an issue but it is something we should consider.

The code also relies on the mtime of the asset for making the tag. So properly synced times on a cluster is needed to have the same tag. The fallback is to hash the file instead of using mtime, so the timestamp can change to whatever as long as the file content is identical. That is why I still have the `getAssetTagForFile` method. Maybe for hashing we could just use up to the first 4k of the file as we dont care about the md5 itself so a partial will likely be ok to tell if an asset has changed in content.

src/SimpleSAML/XHTML/Template.php
```diff
@@ -206,16 +208,38 @@ public function asset(string $asset, ?string $module = null, bool $tag = true):
...
-        }
-        $tag = substr(hash('md5', $tag), 0, 5);
+        // Use the `assets.secretsalt` to enhance security.
+        // Do not make it easy to guess the underlying SSP version.
+        $salt = $this->configuration->getOptionalStringNested(
+            'assets.secretsalt',
+            'assets.secretsalt.default',
+        );
+        $mac = hash_hmac('sha256', $tag, $salt, true);
+        $tag = substr(base64_encode($mac), 0, $tagLength);
```
pradtke, line 220: From some oidc work I learned  that base64 encodings are not url safe since they can produce `+`, `/` and `=`.  Some languages have a b64 url safe version but I did not see one for php

ioigoume: this is easily fixed with this:

tvdijen:
> So the tag is now based on the mtime [..]

I don't understand why we're still using the mtime here. A salted hash of the file's checksum is all we need. Now we have introduced a problem for clustered nodes because there is no way to keep the timestamps over multiple nodes.
I'm also not a fan of the new config-getter.. It all adding too much complexity

monkeyiq:
re complexity we can strip out the base64 stuff and move back to lower case hexits to reduce that part of it.

monkeyiq:
I have removed the nested config getter method now

monkeyiq:
I have moved to reduce the size of the patch again. I can also drop out the base64 stuff and just go back to lower case hexits for the tag. That part of this patch is less interesting to me. It was intended as a gain in entropy per limited number of chars by increasing the base from 16 to 64. Higher retained entropy = lower chance of collision.

I have left in `getAssetTagForFile` but can inline the `md5($f->getContent())` to it's call site and remove it as well. If we have the method then we can just use mtime on a non clustered install and use hashing otherwise. I'm not super attached to any solution. Though hashing every asset for every access for a production release will likely add CPU load.

monkeyiq:
I have removed `getAssetTagForFile` and made the code as it was before but with the initial update from @ioigoume for considering the composer.lock file when in a module.

src/SimpleSAML/XHTML/Template.php
```diff
@@ -217,7 +218,7 @@ public function asset(string $asset, ?string $module = null, bool $tag = true):
             'assets.secretsalt.default',
         );
         $mac = hash_hmac('sha256', $tag, $salt, true);
```
tvdijen, line 232: If we set the fourth parameter to `false`, we are given a url-safe hash and we don't need the added `base64_url_encode` utility method.

src/SimpleSAML/XHTML/Template.php
```diff
@@ -208,10 +211,26 @@ public function asset(string $asset, ?string $module = null, bool $tag = true):
         $file = new File($file);
 
         $tag = $this->configuration->getVersion();
```
tvdijen, line 213: I'm not sure what using the version and the md5 of the composer.lock is adding besides complexity.
An md5 hash of the file should work for all cases and CPU-load is _not_ an issue, because md5 is lightning-fast.. I'd reduce the two if-blocks to `$tag = md5_file($file);`

tvdijen: Or even better, use `hash_hmac_file` and use the salt as the key-parameter.

tvdijen:
> This way if you update from say 2.4.0 to 2.4.1 at the same disk location and URL then the tag will be different [..]

Yes, but why is that necessary if the actual asset hasn't changed between versions?

> One of the things this PR originally wanted to do was base the tag on the composer.lock [..]

Again, why?! This will create new tags even if `composer update` didn't update any of the assets

My point is that all three cases can be dealt with in the same way by using a salted file checksum and it will remain consistent over time _and_ over multiple nodes as long as the asset hasn't changed.

tvdijen:
Yes, now we're talking!! Only thing left is the default salt...

monkeyiq:
Since this is doing a) : new and optional at the moment I think it is ready for a merge.

monkeyiq:
There are some things I would have liked to integrate but which I think we can talk about after the merge. This PR has become fairly long in the tooth already so moving incrementally is probably a good things from here.

Things we might like to consider include:

1) Moving from lower case hexits to url safe base64 to go from 16 to 64 as the base for each digit. For a 5 digit tag this would go from the current 1 million values to 1 billion values so the chance of  an arbitrary "same tag" collision would be much less. A 1 in a million chance of collision seems like something that might, in very rare cases, appear in production and be hard to tell why assets are not being updated.

2) Maybe as well go to 7-8 digits long. 64^8 = 281,474,976,710,656 which would make for a significant improvement over the 1 in a million that the current code offers.

monkeyiq:
@pradtke thank you for the information about salts in the wild. It is extremely valuable to have such feedback :)

PR #2498, opened 2025-08-14
Consider it an error if session.check_function can not be called

monkeyiq:
If the user configuration includes a definition of the check_function we should not silently ignore it if we can not call it. Consider for example if the admin is using that to reject or drop a session for a group of nefarious users. They would rather know that the function is not callable than assume everything is ok. This could happen after initial testing for example if the function is in a module that was not enabled during an upgrade etc.

My main doubt here is that it could cause error messages to appear during a 2.4 series update. But then it will only happen if there is an existing error in config.php.


tvdijen:
I think it should actually _throw_ an error instead of just logging one and then continue execution.
How about replacing this with `Assert::isCallable` instead?

PR #2499, opened 2025-08-15
Add ProfileAuth authsource

mrvanes:
A new authsource that lets user choose preconfigured profiles from a list.

authsources.php:
```
    'profileauth' => [
        'exampleauth:UserClick',
        'users' => [
            [
                'uid' => ['student'],
                'displayName' => ['Student'],
                'eduPersonAffiliation' => ['student', 'member'],
            ],
            [
                'uid' => ['employee'],
                'displayName' => ['Employee'],
                'eduPersonAffiliation' => ['employee', 'member'],
            ],
        ],
    ],
```

If only one user is configured, the profile chooser will be skipped and login is automatic.

### DIFF
--- a/modules/exampleauth/src/Auth/Source/UserClick.php
+++ b/modules/exampleauth/src/Auth/Source/UserClick.php
@@ -147,8 +147,7 @@ class UserClick extends Auth\Source
      *
      * Note that both the username and the password are UTF-8 encoded.
      *
-     * @param int $id  The username the user wrote.
-     * @param string $password  The password the user wrote.
+     * @param int $id  The userprofile the user clicked on.
      * @return array  Associative array with the users attributes.
      */
     protected function login(int $id): array
--- a/modules/exampleauth/src/Auth/Source/UserClick.php
+++ b/modules/exampleauth/src/Auth/Source/UserClick.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace SimpleSAML\Module\exampleauth\Auth\Source;
 
 use Exception;
-use SAML2\Constants;
 use SimpleSAML\Auth;
 use SimpleSAML\Assert\Assert;
 use SimpleSAML\Error;
@@ -13,6 +12,9 @@ use SimpleSAML\Logger;
 use SimpleSAML\Module;
 use SimpleSAML\Utils;
 use Symfony\Component\HttpFoundation\{Request, Response};
+
+use function array_key_exists;
+use function sprintf;
 
 /**
  * Profileauth authentication source.
@@ -43,23 +45,35 @@ class UserClick extends Auth\Source
      */
     public array $users = [];
 
-    protected function getUsers($config) {
+
+    /**
+     * @param array<mixed> $config
+     * @return array<mixed>
+     */
+    protected function getUsers(array $config): array
+    {
         $users = [];
+        $attrUtils = new Utils\Attributes();
 
         // Validate and parse our configuration
         foreach ($config as $id => $attributes) {
-            $attrUtils = new Utils\Attributes();
-
             try {
                 $attributes = $attrUtils->normalizeAttributesArray($attributes);
             } catch (Exception $e) {
-                throw new Exception('Invalid attributes for user ' . $id .
-                    ' in authentication source ' . $this->authId . ': ' . $e->getMessage());
+                throw new Exception(sprintf(
+                    'Invalid attributes for user %s in authentication source %s: %s',
+                    $id,
+                    $this->authId,
+                    $e->getMessage(),
+                ));
             }
+
             $users[$id] = $attributes;
         }
+
         return $users;
     }
+
 
     /**
      * Constructor for this authentication source.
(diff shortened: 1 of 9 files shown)


tvdijen:
Thanks a bunch @mrvanes ! Could you please confirm that the code I have removed was indeed a development leftover?

modules/exampleauth/src/Auth/Source/UserClick.php
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * Profileauth authentication source.
+ *
+ * This class is an authentication source which stores all users in an array,
+ * and authenticates users by clicking on their name.
+ *
+ * @package SimpleSAMLphp
+ */
+
+class UserClick extends Auth\Source
```
thijskh, line 28: I'm confused as to the difference between `Profileauth` (l 20) and `UserClick`. What do we call this auth source?

modules/exampleauth/src/Auth/Source/UserClick.php
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * Attempt to log in using the given user id.
+     *
+     * On a successful login, this function should return the users attributes. On failure,
+     * it should throw an exception. If the error was caused by the user entering the wrong
+     * username or password, a \SimpleSAML\Error\Error(\SimpleSAML\Error\ErrorCodes::WRONGUSERPASS) should be thrown.
+     *
+     * Note that both the username and the password are UTF-8 encoded.
+     *
+     * @param int $id  The username the user wrote.
```
thijskh, line 150: This is not correct. I understand that we want to use an int as the key. It does mean a user cannot reuse the config from exampleauth:UserPass which uses strings as the keys.

modules/exampleauth/templates/userclick.twig
```diff
@@ -0,0 +1,63 @@
+{% set pagetitle = 'Continue as persona' %}
+
+{% extends "@core/base.twig" %}
+
+{% block preload %}
+    <link rel="stylesheet" href="{{ asset('css/default.css', 'exampleauth') }}">
+{% endblock %}
+
+{% block content %}
+    {%- if not isProduction %}
```
thijskh, line 10: This seems a bit misplaced. This auth source is never for production?

thijskh (commented):
This seems for sure useful to have. I'm missing any documentation for users to know that this functionality even exists other than by reading the code. Since the config is non-trivial (e.g. you need to know about numeric indexes, that DisplayName is mandatory to have for a user...) so some docs guidance is needed I think.

modules/exampleauth/src/Auth/Source/UserClick.php
mrvanes: There are lots of leftovers from UserPass, thx for the extra attention, will clean up

modules/exampleauth/templates/userclick.twig
tvdijen: afbeelding

It will display this warning if `'production' => false` in config.php

Nothing more.. I think Thijs overlooked something, but I this this can be marked as resolved.

monkeyiq:
It may make sense to group information about this and some other information into a new "Testing" page under the "Further topics" page on the top docs page. The install page makes some mention of the admin page but that is before the reader has setup as an SP or IdP etc. 

Perhaps a page with tips on the admin interface (admin/test for checking logins) and this module for clicking profiles.

monkeyiq:
I have also made a follow up PR at https://github.com/simplesamlphp/simplesamlphp/pull/2504 to add some docs for this functionality and start a core "testing an install" page.

tvdijen (approved):

modules/exampleauth/templates/userclick.twig
```diff
@@ -0,0 +1,63 @@
...
+
+            </div>
+        </div>
+    </div>
+    {%- endif %}
+
+    <p><b>{{ 'Continue as persona' }}</b></p>
+
+    {% for id, attributes in users %}
+    <a href="{{formURL}}&id={{ id }}">
```
tvdijen, line 39: I'm a little worried about proper url encoding when we compose the urls like this in Twig.
Perhaps we can generate the entire url in the controller

modules/exampleauth/src/Controller/ProfileAuth.php
```diff
@@ -0,0 +1,165 @@
...
+     * @param \Symfony\Component\HttpFoundation\Request $request
+     * @param \SimpleSAML\Module\exampleauth\Auth\UserClick $source
+     * @param array $state
+     */
+    private function handleLogin(Request $request, UserClick $source, array $state): Response
+    {
+        $authStateId = $request->query->get('AuthState');
+        $this->authState::validateStateId($authStateId);
+
+        $organizations = $organization = null;
```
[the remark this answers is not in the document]
monkeyiq: While I had the repo cloned... https://github.com/simplesamlphp/simplesamlphp/pull/2499/commits/ea50934ee3d03a74acea1153358c3708cbf409be

modules/exampleauth/templates/userclick.twig
tvdijen: Yes, exactly what I was after! Thanks @monkeyiq

monkeyiq:
I picked that forward to master. It seems the `setAuthSource` had a new `authenticate` method which might need picking into the `$this->authSource` as well.

PR #2509, opened 2025-09-02
Update FontAwesome icon names

ghalse:
A few of the Twig templates still have old FontAwesome 4 icon names. While these are mostly backwards compatible still, at least one used in the logout-iframe.twig has gone completely (circle-o-notch) resulting in a broken display.

This updates the names to the version 6+ naming per https://docs.fontawesome.com/v6/web/setup/upgrade/whats-changed#icons-renamed-in-version-6


monkeyiq:
For SSP 2.4 I see Font Awesome Free 6.7.1 in the base assets of my install. I am happy to move to the version 6+ names for 2.4+

monkeyiq:
As you say, `fa-circle-o-notch` doesn't show anything in ssp 2.4. Just forcing it by adding to `modules/admin/templates/logout.twig` for local testing.

monkeyiq (approved):

PR #2513, opened 2025-09-10
disable showerrors by default

mcdruid:
SimpleSAMLphp has `showerrors` enabled by default.

```
     * When 'showerrors' is enabled, all error messages and stack traces will be output
     * to the browser.
```

Although this is no doubt convenient at times, it's risky from a security point of view. See, for example:

https://owasp.org/www-community/Improper_Error_Handling

> The most common problem is when detailed internal error messages such as stack traces, database dumps, and error codes are displayed to the user (hacker).

A specific example of this in SimpleSAMLphp was [CVE-2024-52596](https://github.com/advisories/GHSA-2x65-fpch-2fcm) whereby an attacker could induce a specific error condition via [XXE](https://owasp.org/www-community/vulnerabilities/XML_External_Entity_(XXE)_Processing) that causes the application to reveal potentially sensitive information within an error message.

`showerrors` should be disabled by default to avoid such problems.

Filing this as a public PR as all of the details are public already; there is no novel information about an exploitable vulnerability here.

### DIFF
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -343,7 +343,7 @@ $config = [
      * When 'errorreporting' is enabled, a form will be presented for the user to report
      * the error to 'technicalcontact_email'.
      */
-    'showerrors' => true,
+    'showerrors' => false,
     'errorreporting' => true,
 
     /*
--- a/src/SimpleSAML/Error/Error.php
+++ b/src/SimpleSAML/Error/Error.php
@@ -253,7 +253,7 @@ class Error extends Exception
         $config = Configuration::getInstance();
 
         $data = [];
-        $data['showerrors'] = $config->getOptionalBoolean('showerrors', true);
+        $data['showerrors'] = $config->getOptionalBoolean('showerrors', false);
         $data['error'] = $errorData;
         $data['errorCode'] = $this->errorCode;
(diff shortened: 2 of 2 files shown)


tvdijen:
> However in the case of the XXE vulnerability the display of verbose error messages can definitely be abused - happy to share more details if that would help.

Yes, error messages should never reveal file paths or anything like that. However, this is not controlled with the `showerrors` setting

drupal-spider:
Given that, we have already had the ['backtraces'](https://github.com/simplesamlphp/simplesamlphp/blob/master/config/config.php.dist#L323) option for the debug setting. We can respect that setting with the error setting to surface the detailed debug information in the error page.

The proposed changes are:
- 	The 'showerrors' is disabled by default. 
- 	If no debug flags are set, no extra information will be shown in the error page.
- 	If 'backtraces' is enabled in the debug option, the stack traces information will display along with the error messages.

In a user case where the user does want to have the stack trace details in the logs, but does not want those information expose to the client side (a browser). The configuration should be set as:
```
'debug' => [
        'backtraces' => true,
    ],
'showerrors' => false,
```

which aligns with the current behaviour.

Proposed PR: https://github.com/simplesamlphp/simplesamlphp/pull/2540

PR #2516, opened 2025-09-18
slow post ux improvement

ioigoume:
Improves user experience during cross-site POST when the SP is slow or unresponsive, which previously left users on a blank SSP page and suggested a failure.
Introduces a loader and explanatory messaging to indicate the POST is in progress and set expectations.
Accessibility considerations were incorporated and validated using the WAVE browser tool.

### DIFF
--- a/src/SimpleSAML/Utils/HTTP.php
+++ b/src/SimpleSAML/Utils/HTTP.php
@@ -1192,6 +1192,14 @@ class HTTP
         $p = new Template($config, 'post.twig');
         $p->data['destination'] = $destination;
         $p->data['post'] = $data;
+
+        // Read optional config override; default to 30s, ensure non-negative integer
+        $delay = (int) ($config->getOptionalValue('slow_post_delay_ms', 30000));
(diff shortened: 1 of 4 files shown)


src/SimpleSAML/Utils/HTTP.php
```diff
@@ -1187,12 +1189,23 @@ public function submitPOSTData(string $destination, array $data): void
...
             $this->redirect($this->getSecurePOSTRedirectURL($destination, $data));
+            return;
         }
 
         $p = new Template($config, 'post.twig');
         $p->data['destination'] = $destination;
         $p->data['post'] = $data;
+
+        // Read optional config override; default to 30s, ensure non-negative integer
+        $delay = (int) ($config->getOptionalValue('slow_post_delay_ms', 30000));
```
tvdijen, line 1200: Use getOptionalInteger or getOptionalIntegerRange instead and drop the cast.

ioigoume: @tvdijen done.

PR #2521, opened 2025-10-06
Allow a whitelist of errors to be shown when showerrors is true

monkeyiq:
If this whitelist is not used then all errors are shown if showerrors is true. If you use this new option then you have to list every error you would like to be shown to the user with a description and backtrace.

This was raised in
https://github.com/simplesamlphp/simplesamlphp/pull/2513

### DIFF
--- a/src/SimpleSAML/Error/Error.php
+++ b/src/SimpleSAML/Error/Error.php
@@ -206,6 +206,25 @@ class Error extends Exception
         } else {
             $referer = 'unknown';
         }
+
+        $showerrors = $config->getOptionalBoolean('showerrors', true);
+
+        $whitelist = Configuration::getInstance()->getOptionalArray('showerrors.whitelist', ['*' => true]);
+        if( count($whitelist)==1 && array_key_exists('*', $whitelist)) {
+            // no filtering
+        } else {
+            $showRealError = false;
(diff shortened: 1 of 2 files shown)


monkeyiq:
In this PR I have attempted to allow the admin to select which of the exceptions will have the message and backtrace shown to the user. This is to provide a grey scale between turning `showerrors=false` (nothing) and `showerrors=true` (everything).

If there are other ways to allow this to be configured I am very happy to discuss them

src/SimpleSAML/Error/Error.php
```diff
@@ -206,6 +206,25 @@ protected function saveError(): array
         } else {
             $referer = 'unknown';
         }
+
+        $showerrors = $config->getOptionalBoolean('showerrors', true);
+
+        $whitelist = Configuration::getInstance()->getOptionalArray('showerrors.whitelist', ['*' => true]);
```
drupal-spider: In terms of the backward compatibility, if the new configuration of 'showerrors.whitelist' is missing from the exisiting configuration file, it will not filter out any error as default?

monkeyiq: I have made the initial version of the PR not change behaviour by default. If this PR seems like a good thing to pursue  we can create a curated list of errors to hide and add that as a breaking change at some version.

PR #2540, opened 2025-10-15
Hide the debug information in the error page by default

mingsong-hu:
The proposed changes are:
- 	The 'showerrors' is disabled by default. 
- 	If no debug flags are configured, no additional information will appear on the error page. Currently, stack traces are always displayed whenever showerrors is enabled, regardless of the debug settings.
- 	If 'backtraces' is enabled in the debug option, the stack traces information will display along with the error messages.

In a user case where the user does want to have the stack trace details in the logs, but does not want those information expose to the client side (a browser). The configuration should be set as:
```
'debug' => [
        'backtraces' => true,
    ],
'showerrors' => false,
```

which aligns with the current behaviour.

### DIFF
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -341,13 +341,13 @@ $config = [
     ],
 
     /*
-     * When 'showerrors' is enabled, all error messages and stack traces will be output
-     * to the browser.
-     *
-     * When 'errorreporting' is enabled, a form will be presented for the user to report
-     * the error to 'technicalcontact_email'.
+     * Error display behaviour:
+     * - Enable 'showerrors' to allow detailed messages in the browser; leave it disabled for production.
+     * - Add at least one entry to the 'debug' option to surface those messages. Without any debug flags, no extra details appear.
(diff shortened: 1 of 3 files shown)


drupal-spider:
> > Agree. Having a dedicated configuration for ui report would be more flexible.
> 
> That's effectively what the `showerrors` flag is doing.. What am I missing here?

My point is that, if we want to provide an option for an administrator to signify that only output the  error exception message in the UI (browser) without exposing the stack traces information which are considered as sensitive. Currently, this option is not in place.

Take the following screenshot as an example, the exception message is "SimpleSAML\Error\Error: UNHANDLEDEXCEPTION". 
And the back traces information highlighted in blue. Those two part of error message currently can't be switched off separately.  It has to show both or neither.

Error messages

drupal-spider:
Quite frankly, I don't know how popular this request is. 

No matter, this merge request is out of date as the https://github.com/simplesamlphp/simplesamlphp/pull/2513 has been merged.  So I am closing this merge request for now. If we do need a new configuration after discussing, I can create a new issue for that.

PR #2543, opened 2025-10-17
Equalize some docs with master

cicnavi:
@thijskh Note that this overrides your recent modifications in branch `simplesamlphp-2.5`, since they were not in line with the docs in `master`.


monkeyiq:
Perhaps in a follow up to this PR ... since I was looking at the various branches and this PR ...

It seems that the link used in the  `IdP-initiated logout`

https://idp.example.org/simplesaml/saml2/idp/SingleLogoutService.php

Might be able to change to 

simplesaml/saml2/idp/initSingleLogout

assuming that we are preferring the routes to the raw php files. That could change in all version 2.4, 2.5, and master.

monkeyiq:
Since this has some "overrides your recent modifications" made by @thijskh I am leaving merging this for a little bit to allow time for feedback.

PR #2546, opened 2025-10-20
Allow a whitelist of errors to be shown when showerrors is true

monkeyiq:
If this whitelist is not used then all errors are shown if showerrors is true. You can use this new option to explicitly allow backtraces and descriptions to be shown to the user for only select error events.

If you provide a list of errors to show then anything not on that list will not be shown to the user. The error will be logged etc as normal.

This was raised in
https://github.com/simplesamlphp/simplesamlphp/pull/2513

This is a rebase of https://github.com/simplesamlphp/simplesamlphp/pull/2521 to `master`

### DIFF
--- a/config/config.php.dist
+++ b/config/config.php.dist
@@ -348,6 +348,66 @@ $config = [
     'errorreporting' => true,
 
     /*
+     * When showerrors is true, this is an array of which errors
+     * should still be shown to the user. By default an error will
+     * always be shown if showerrors==true and this setting is at the default value to allow all.
+     *
+     * If you list anything in this option you have to explicitly list each error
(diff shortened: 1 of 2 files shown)


monkeyiq:
I have moved https://github.com/simplesamlphp/simplesamlphp/pull/2521 over to `master`. I think it is useful to have as it explicitly allows the list of which errors will be filtered through to the user. The admin might wish to allow some errors to permeate while others can be kept silent from the user. It also allows for specific errors to be hidden if they are seen to be problematic while leaving the remainder of the system as it as rather than shutting off all errors in a single option.

If people don't want the complexity then they can safely ignore the option and set `showerrors=true` or `showerrors=false` and either swing the big axe or not as in 2.4.

PR #2559, opened 2025-11-18
Add conditional options to AttributeAdd authproc

nathanjrobertson:
I have a use case where we have a number of dynamic attributes that follow a particular pattern, and if any of them exist, I need to add an attribute.

As a concrete example, if any of `customerTypeA`, `customerTypeB`, `customerTypeC`, ... exists, then set `isCustomer` = ['true'].

With this PR, this can easily be done with:

```php
    'authproc' => [
        50 => [
            'class' => 'core:AttributeAdd',
            '%if_attr_regex_matches' => '/^customerType/',
            'isCustomer' => ['true'],
        ],
    ],
```

I added both the string and regular expression options, as not every user will understand regular expressions, yet it's the only sensible way to support a dynamic list of attributes (such as my use case).

Fully backward compatible as usual. If neither options are specified then it will add the attribute, as per existing functionality. PHPUnit tests and documentation updated too.

### DIFF
--- a/modules/core/docs/authproc_attributeadd.md
+++ b/modules/core/docs/authproc_attributeadd.md
@@ -6,6 +6,10 @@ Filter that adds attributes to the user.
 If the attribute already exists, the values added will be merged into a multi-valued attribute.
 If you instead want to replace the existing attribute, you may add the `%replace` option.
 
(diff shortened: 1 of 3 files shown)


nathanjrobertson:
PR #2579 was merged, and this functionality overlaps that PR. I'll close this PR.

Issue #2555, opened 2025-11-12
ADFS iframe logout handling broken


m0ark:
During logout via the iframe logout handler the ADFS associations are not terminated.

Module `adfs` [prefixes its entityIDs](https://github.com/simplesamlphp/simplesamlphp-module-adfs/blob/c1b753d3558d6c6f599b2b8883fc97044ebcde6f/src/IdP/ADFS.php#L576) with `adfs:`, but the [LogoutController](https://github.com/m0ark/simplesamlphp/blob/bc3aed2a6bd10842c7b5c791f983211b014f468a/modules/core/src/Controller/Logout.php#L386) only looks for keys prefixed with `saml:`.

Possible fixes:
1. dynamically resolve entityID using the same mechanism to [determine the metadata set](https://github.com/m0ark/simplesamlphp/blob/bc3aed2a6bd10842c7b5c791f983211b014f468a/modules/core/src/Controller/Logout.php#L362)
2. change the entityID prefix in module `adfs` to 'saml'
3. use the hook infrastructure to define a protocol agnostic interface for logout scenarios

Any ideas or preferences on how to solve this?

tvdijen:
Yes, replacing the hooks with PSR-14 dispatchers has been the idea for quite some time, but I never got to it.

monkeyiq:
Doing a bit of a pre-PR look over... 

* in `modules/admin/src/Controller/Menu.php` I wonder if `$t` should be passed instead of `$template` to `Module::callHooks('adminmenu', $template);` in `insert`.

I like the move from hooks to PSR-14: Event Dispatcher. I look forward to combing over things again once this becomes a PR. Thanks for the update in your branch and the chance to look it over.

PR #2560, opened 2025-11-18
Use PSR-14 events to replace hook infrastructure

m0ark:
As suggested by @cicnavi in https://github.com/simplesamlphp/simplesamlphp/issues/2555#issuecomment-3521114391 I took an approach to replace the existing hook infrastructure with PSR-14 events.

The current implementation listens for both hooks and events (in the modules present in this repository) and merges their results to support a smooth migration across modules.

I do not claim this to be perfect, so feel free to roast it :smile:

### DIFF
--- a/src/SimpleSAML/Event/Dispatcher/ModuleEventDispatcherFactory.php
+++ b/src/SimpleSAML/Event/Dispatcher/ModuleEventDispatcherFactory.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace SimpleSAML\Event\Dispatcher;
 
-use SimpleSAML\Configuration;
 use SimpleSAML\Event\Provider\ModuleListenerProvider;
 
 class ModuleEventDispatcherFactory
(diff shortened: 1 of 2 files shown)


monkeyiq:
Since this is looking more on the lines of a merge and cherrypick into 2.5, perhaps we should also update docs/simplesamlphp-modules.html to mention the new PSR-14 hook interface in 2.5+ and retain the docs renamed to the old hooks interface (pre 2.5).

tvdijen:
Agreed and possibly provide an example in the upgrade notes on how to migrate your hook to PSR-14

monkeyiq:
I will soon merge this and backport into the 2.5 branch.  
I will look at the docs update after that as I have done some hook &rarr; event listener in my testing

monkeyiq:
Yep. I see some starters
https://github.com/simplesamlphp/simplesamlphp/actions/runs/19722356052/job/56507208229

PR #2568, opened 2025-11-27
tests: update for psr14 with cron looking at config page

monkeyiq:
This is not pretty but is a step in the right direction. That asserteq could be made more robust to change.


monkeyiq:
This is a start to cleanup some of the tests I broke backporting https://github.com/simplesamlphp/simplesamlphp/pull/2560 to branch 2.5.

monkeyiq:
adding cron to the essential module.enable makes the 2 failing tests work in CI. Now it is just a matter of making the tests do that adding and moving from from essential to optional agian

Issue #2572, opened 2025-12-03
Attributes translation doesn't work?


Kojiro-Goto:
When display consentform.twig, attributes translated name isn't displayed.
But if you add "X-Domain messages\n" into attributes.po on locales/ folder, translated name is displayed.

Please tell me appropriate solution to display translated attribute name.

monkeyiq:
I will take care of this one. Thanks @tvdijen

PR #2576, opened 2025-12-11
fix for attributes.po translations

monkeyiq:
As this now passes I will merge this and cherry pick forwards from here.

Translations from attributes.po are implicitly put into the attributes domain. The code looking for them was looking in the empty domain still. Also addAttributeDomains is included now to explicitly make sure they are loaded.

This was raised in
https://github.com/simplesamlphp/simplesamlphp/issues/2572

https://github.com/simplesamlphp/simplesamlphp/pull/2575


monkeyiq:
That seems to come down to the new `addAttributeDomains` in the `Localization` ctor.

Issue #2586, opened 2026-01-22
\SimpleSAML\Utils\System::resolvePath fails for UNC Paths


raubv0gel:
`\SimpleSAML\Utils\System::resolvePath` does not properly return UNC paths on Windows. E. g., for param `$path` being `\\some-pc\some-share\some.key` it returns `/some-pc/some-share/some.key` not being a valid path. Therefore `$data = @file_get_contents($location);` at `\SimpleSAML\Utils\Crypto::retrieveCertOrKey` fails to load the key file.

raubv0gel:
`var_dump(file_get_contents('//some-pc/some-share/some.key'));` and `var_dump(file_get_contents('\\\\some-pc\\some-share\\some.key'));` do work (PHP 8.2).

PR #2591, opened 2026-02-05
Update UNC path handling

monkeyiq:
This

```
\\some-pc\some-share\some.key
```

Should result in

```
var_dump($ret);
string(29) "//some-pc/some-share/some.key"
```

Which according to
https://github.com/simplesamlphp/simplesamlphp/issues/2586 should be able to be loaded.

I only have a partial php win env so it needs verification against a real SSP install.

### DIFF
--- a/tests/src/SimpleSAML/Utils/SystemTest.php
+++ b/tests/src/SimpleSAML/Utils/SystemTest.php
@@ -279,4 +279,21 @@ class SystemTest extends TestCase
         $reflectedInstance->setValue($service, []);
         $reflectedInstance->setAccessible(false);
     }
+
+
+    /**
+     */
+    public function testUNCpaths(): void
+    {
+        if (DIRECTORY_SEPARATOR == '\\') {
+
+            $base = null;
+            $path = "\\some-pc\some-share\some.key";
+
+            $res = $this->sysUtils->resolvePath($path, $base);
+            $expected = "//some-pc/some-share/some.key";
+
+            $this->assertEquals($expected, $res);
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


monkeyiq:
I now have a simple test case for this on windows. I have used a hack based on DIRECTORY_SEPARATOR to know if we are running on Win, perhaps there is a DEFINE or something that is more direct to use here instead.

tests/src/SimpleSAML/Utils/SystemTest.php
```diff
@@ -279,4 +279,21 @@ protected function clearInstance(Configuration $service, string $className): voi
         $reflectedInstance->setValue($service, []);
         $reflectedInstance->setAccessible(false);
     }
+
+
+    /**
+     */
+    public function testUNCpaths(): void
```
tvdijen: This is not right. If the separator is `\\` then the UNC-path should also use backslashes?

monkeyiq: I think the UNC path should, I was looking at first at the comment from the original issue which mentions that php 8.2 seems to allow `//server/path` as well and work https://github.com/simplesamlphp/simplesamlphp/issues/2586#issuecomment-3783674998

If we are looking to make the returned path use '\\' then we could change the rebuild code at
https://github.com/simplesamlphp/simplesamlphp/blob/5cae09a277a37729f17654384fdd455387eaafa8/src/SimpleSAML/Utils/System.php#L163
and see what else falls over with that and try to get those ripple effects under control again.

monkeyiq:
I am moving this to a draft until I have a win install to test this on and/or feedback from the initial bug report is received.

PR #2595, opened 2026-02-15
Updates for CI in 2.5 series

monkeyiq:


monkeyiq:
This should maybe ok if the `saml2-legacy` dep is updated to use ->value on this enum. Perhaps we create an s2l v4.19.2 with something like that and switch to it.

PR #2629, opened 2026-04-15
Duplicate requirement in composer.json

cicnavi:
`gettext/php-scanner` and `symfony/translation` were both in `require` and in `require-dev`. I've removed them from `require-dev` and ran `composer update`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -106,11 +106,9 @@
         "ext-curl": "*",
         "ext-pdo_sqlite": "*",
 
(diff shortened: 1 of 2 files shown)


cicnavi:
I now see that after update, `phpstan` now has some new reports. Will check this out tomorrow.

tvdijen:
A rebase should fix things.

cicnavi:
I see that you made fixes in master, so I've merged with it, actions pass now...

PR #2630, opened 2026-04-16
Move to global Kernel

cicnavi:
This is a proposal only PR for v3, and should be carefully tested before merge.

The point is to route all requests through `public/index.php`, where global `Kernel` is instantiated, which loads all available services and routes, including from all enabled modules.

Module endpoints would now live under `/module/<module>/...`. ~~Legacy `module.php` entry point is no longer supported. Also, any other direct public PHP entry point is no longer supported.~~  Legacy URLs containing `/module.php/` are supported in v3.0 to allow a transition period, so all peers can update the newly published URLs in a reasonable timeframe. The module assets now have to be "published" to SSPs public dir to be available (dedicated asset publishing command is introduced). 

My initial thought was to "only" remove "module.php", but this ended up being whole request process refactoring.... But this should enable easier maintenance and testing, and adding more Symfony features in the future, I think....

I did some manual testing locally where I had an IdP with this implementation and an SP using SSP v2.4. I tried doing basic SAML authentication and it worked. However, I didn't do any other testing with the overall SSP ecosystem...


public/saml2/idp/initSLO.php
```diff
@@ -1,30 +1,11 @@
...
-foreach ($headers as $header => $value) {
-    // Some pages may have specific requirements that we must follow. Don't touch them.
-    if (!$response->headers->has($header)) {
-        $response->headers->set($header, $value);
-    }
-}
-$response->send();
+http_response_code(410);
+header('Content-Type: text/plain; charset=UTF-8');
+echo "Legacy public/saml2/idp scripts are no longer supported. Use the routed /module/saml endpoints instead.\n";
```
tvdijen: I think it makes sense to create a controller for this that can be re-used for every legacy-endpoint. We can then remove these old php-files.

To give you an idea of what I mean, see this controller:
https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/saml/src/Controller/Exception.php#L35-L54

And the (multiple) routes that go with it:
https://github.com/simplesamlphp/simplesamlphp/blob/master/modules/saml/routing/routes/routes.yml

cicnavi: Ok, will check a bit later.

public/index.php
```diff
@@ -6,12 +6,13 @@
 
 require_once('_include.php');
 
-$config = Configuration::getInstance();
-$httpUtils = new Utils\HTTP();
+$kernel = new Kernel();
+$request = \Symfony\Component\HttpFoundation\Request::createFromGlobals();
+$response = $kernel->handle($request);
+$kernel->terminate($request, $response);
```
monkeyiq: Are we sure we can call `$response->headers` after kernel `terminate` on the response?

monkeyiq:
Maybe we should document somewhere that the included modules also have an explicit line in the .gitignore to include the deployed files such as 
```
!/public/assets/admin/
```

Since the included modules is not likely to change I am not sure if it is worth adding a test in the deployment script `ModuleAssetPublisher.php` checking if there are assests from modules that are directly in `modules` and that have no override in the .gitignore file for their content. Or we just add some note in docs/simplesamlphp-developer-information.md if more primary modules (those in the repo itself) have assets to make sure the .gitignore file is updated for them if they have public assests.

public/index.php
cicnavi: I'll double check.

public/saml2/idp/initSLO.php
cicnavi: I've added a new `ObsoleteRouteController` which can be used to mark routes as "gone" (HTTP 410), so you can take a look. I've used it for several entry points from `public`. I've also used `RedirectController` from Symfony to for `admin-shortcut` and legacy `admin/index.php`.

public/index.php
cicnavi: In https://github.com/symfony/symfony-docs/blob/7.4/components/http_kernel.rst I see that the `terminate` should be called after the response was sent, so I've adjusted accordingly... Thanks for the hint!

cicnavi:
> Maybe we should document somewhere that the included modules also have an explicit line in the .gitignore to include the deployed files such as
> 
> ```
> !/public/assets/admin/
> ```
> 
> Since the included modules is not likely to change I am not sure if it is worth adding a test in the deployment script `ModuleAssetPublisher.php` checking if there are assests from modules that are directly in `modules` and that have no override in the .gitignore file for their content. Or we just add some note in docs/simplesamlphp-developer-information.md if more primary modules (those in the repo itself) have assets to make sure the .gitignore file is updated for them if they have public assests.


I've updated the docs and also added a small unit tests that will check modules in the repo that have assets if they are noted in .gitignore, this should do it...

cicnavi:
`composer-unused` now reports `file=composer.json,line=52,col=0::ext-fileinfo is unused` but I'm not sure if it is really ok to remove `ext-fileinfo` from `require`.

tvdijen:
> `composer-unused` now reports `file=composer.json,line=52,col=0::ext-fileinfo is unused` but I'm not sure if it is really ok to remove `ext-fileinfo` from `require`.

Yes, it's only use was the `mime_content_type()` call in `Module.php` and that is now gone, so it's safe to remove this from `composer.json`

cicnavi:
I removed all of the Symfony services definition in SSP, as those seemed redundant to me. However, when I first did that, custom SSP error handling didn't work anymore (Symfony was used under the hood). So, that's why now there is a `ExceptionSubscriber` class which is hooked on Symfony kernel exception event and this calls the SSP exception handler. So, this works for showing SSP error pages... HOWEVER.... , there is one problem with custom SSP error handler - it uses `exit()` and this totally breaks the Symfony request - response flow... for example, the kernel `terminate()` is never called, nor are the security headers are set in `public/index.php`.... I'll try to refactor this to return proper Symfony response in this case.

cicnavi:
So there is a way to hook into exception events, and this is now used to show custom SSP error pages (https://symfony.com/doc/7.4/event_dispatcher.html). I think I can also use the similar logic to set the headers on responses: https://symfony.com/doc/7.4/reference/events.html#kernel-response. This would replace the code in public/index.php.

cicnavi:
I think this is now rounded enough as a PR, so please review...

monkeyiq:
Thank you for the info about ready for review. I have reviewed many parts but left some things to review in my local notes until we got to this stage.

monkeyiq:
I have read over the bulk of this now. Thank you for the substantial change request! :)

One caveat is that I have skipped over the apache config fragments as my knowledge of that is not enough to analyse it as I was going. I will now go through my notes and try to make them more coherent in comments here.

monkeyiq:
One thought for migration...

If modules all need to update their paths on the release maybe we should tell module authors to update paths to using these helpers prior to 3.0? See the part of "docs/simplesamlphp-install.md"

Does it make sense for a late 2.x series release to rewrite /module/ requests to /module.php/ so that modules might already be using the new URL format prior to 3.x?

Maybe we should try to have 
```
host/simplesaml/module.php/cron/...
host/simplesaml/module/cron/...
```

monkeyiq:
Another broad question without a specific answer...

As the  `AssertionConsumerService` updates seem fairly mechanical should we maybe allow /module.php/ to be in there and update that to /module/ for 3.0 so that the same metadata files will work. This will allow limited migration between 2.9 and 3.0 with the same metadata files and if those files are being synced from other sources will allow older ones to continue to work for a limited time.

monkeyiq:
It might be nice for onKernelException() to have an example that also uses TWIG for the returned page. See the example in docs/simplesamlphp-ecp-idp.md which returns some inline html.

Category: Can easily delay until after merge.

monkeyiq:
In "upgrade-notes-3.0.md" we might like to expand on
```
For module URLs, use the URL helpers provided by SimpleSAMLphp
```
to have one or more examples of the code update for this. Maybe just mentioning to use `Module::getModuleURL()`.

Likewise "For assets, use the published asset URLs." having a URL there to allow
folks to avoid having to work it out. Perhaps mention the new `Module::getModuleAssetUrl()`.

monkeyiq:
For the error handling...

Somewhere in the docs we might like to add
```
php bin/console debug:event-dispatcher kernel.exception 
```
to see if the error handler is registered.

As an aside I plan to but have not run a local test on the error handling stuff. Likely an easy exercise but something I plan to do but have not done as yet.

Easy to delay to after merge: Maybe throw an error if `error.show_function` remains in place in the config? The downside is that the same config.php couldn't then be used between 2.9 and 3.0. The upside is that the user will not be left thinking they have a custom error handler in 3.0 and not have one because they forgot to migrate that part of their setup. A soft option here is we could have a `pedantic_config_check` type of option that is recommended but can be set to `false` to allow a more migration friendly config with the caveat that some options might be ignored and not reported if you force pedantic to false.

monkeyiq:
I notice the path change in docs/simplesamlphp-sp-api.md which I assume is ok but thought I'd mention it to double check...

a link containing
```
   module.php/core
```
going to
```
   module/saml
```

monkeyiq:
Other than those thoughts I need to pull this and test it locally and I guess then the "should be carefully tested before merge" part you mention in the initial comment phase begins :)

tvdijen:
> The downside is that the same config.php couldn't then be used between 2.9 and 3.0

I really don't mind.. This is our opportunity to break things.

Ideally I want to split the current huge `config.php` into several files with settings that can be grouped (i.e. sessions.php, logging.php) and introduce config-classes that extensively validate the values (like this:  https://github.com/simplesamlphp/simplesamlphp-module-logpeek/pull/3)

cicnavi:
> Ideally I want to split the current huge `config.php` into several files with settings that can be grouped (i.e. sessions.php, logging.php) and introduce config-classes that extensively validate the values (like this: [simplesamlphp/simplesamlphp-module-logpeek#3](https://github.com/simplesamlphp/simplesamlphp-module-logpeek/pull/3))

SSP is currently using Symfony "only" for HTTP request / response handling. So, we could also widen the SSP - Symfony integration with using their configuration components... The similar thing is also for sessions, database abstractions, etc.

tvdijen:
Yes, I would really like to use Symfony for session-handling too!
Also, our controllers do not yet extend from Symfony's AbstractController.. I've tried to do this a while back, but failed.

Also, hidden somewhere in the backup-branch of the old `master` (I think summer 2024 or even 2023), I did a _lot_ of work to make all the authsources/authproc-filters/utilities return a Symfony response. That was a huge improvement, because currently a lot of methods (especially in the http-utilities) never return. This work should also find it's way back into a 3.0 release

Perhaps we should create a Github project and create some issues for the things we would like to add to the 3.0 release. While writing this, a PSR-3 compliant logger also comes to mind

cicnavi:
@tvdijen Ok, we can check those in other PRs so we don't put more stuff in this one. Fine by me for additional issues for v3 milestone or project...

@monkeyiq I'll go through every comment from you a bit later.

tvdijen:
I went over them and it's fair to say we need to get the documentation right immediately, because well.. We all know what happens if we don't :')  Other than that I would merge this.

monkeyiq:
Some of the comments I had were more "nice to have" and I have tried to add that to them when converting from brain-dump in emacs to this PR. 

I am also happy to move more to symfony and think that having it in subsequent PRs is a good way to break it up a little for reasonable sized review chunks.

monkeyiq:
I think the logging change was mentioned in a few places. Symfony comes with some PSR-3 loggers including monolog support. 

https://symfony.com/doc/current/logging.html

If there was a common base class or some form of mix in we might be able to have the existing Logger class forward to the injected Symfony logger and then avoid needing to think about that change when logging. I haven't explored that but it might be a handy way to move to that so that as a dev we don't have to think is it `Logger::error()` or `$logger->error()` for this release.

cicnavi:
> One thought for migration...
> 
> If modules all need to update their paths on the release maybe we should tell module authors to update paths to using these helpers prior to 3.0? See the part of "docs/simplesamlphp-install.md"
> 
> Does it make sense for a late 2.x series release to rewrite /module/ requests to /module.php/ so that modules might already be using the new URL format prior to 3.x?
> 
> Maybe we should try to have
> 
> ```
> host/simplesaml/module.php/cron/...
> host/simplesaml/module/cron/...
> ```

Huh, good point... since exchanged metadata will not be updated between entities at the exact time of upgrade to v3, bunch of things will break when they try to upgrade, until they refresh their metadata. We definitely need a transition... How about in v3 I simply do a second pass when loading routes and give those a `module.php` prefix. So we would have both `module` and `module.php` of all routes. I could also introduce a config option that would enable that...

tvdijen:
> So we would have both `module` and `module.php` of all routes.

This is only necessary for the metadata-endpoints and other SAML endpoints where we receive SAML-messages.

cicnavi:
> > So we would have both `module` and `module.php` of all routes.
> 
> This is only necessary for the metadata-endpoints and other SAML endpoints where we receive SAML-messages.

It is also relevant for oidc module which exposes its metadada and relevant endponts in it, I guss casserver also... any endpoint that other services had to register, any API endpoint that is memorized... this will all stop working for them until they update to new URLs

tvdijen:
> It is also relevant for oidc module which exposes its metadada and relevant endponts in it, I guss casserver also...

Absolutely! Add the adfs-module to that list too!
I got the impression that you wanted to keep the old module.php alive even for stuff like the admin-module where an endpoint-change shouldn't be an issue. We should fix this for the relevant endpoints only, not some generic solution for _every_ route.

cicnavi:
> > It is also relevant for oidc module which exposes its metadada and relevant endponts in it, I guss casserver also...
> 
> Absolutely! Add the adfs-module to that list too! I got the impression that you wanted to keep the old module.php alive even for stuff like the admin-module where an endpoint-change shouldn't be an issue. We should fix this for the relevant endpoints only, not some generic solution for _every_ route.

I'm afraid that we don't know all of the relevant endpoints... Those are not just protocol related endpoints. For example, cron module - in our federation we call its endpoints in crons. WIth metadata aggregator we have automations... The problem is, when upgrading to v3 of SimpleSAMLphp I'll have to upgrade all related external stuff that calls SSP endpoints at the same time. I can't do that in advance because new URLs won't work until I upgrade SSP to v3, and old URLs won't work after I upgrade to v3. 

I'm in favor of "transition period", where in v3 we will support all routes - new and legacy (with new ones being published in metadata). And in a next major version, or even minor - so in v3.1 - we remove support for legacy routes. This way people will have a chance to announce the change, publish any new metadata, for which ever protocol, change any external calls to SSP endpoints to cron, metadata, whatever - while old routes work. With this, in theory, there shouldn't be downtime regarding the new URL form for any module, even custom module ones...

cicnavi:
> > > It is also relevant for oidc module which exposes its metadada and relevant endponts in it, I guss casserver also...
> > 
> > 
> > Absolutely! Add the adfs-module to that list too! I got the impression that you wanted to keep the old module.php alive even for stuff like the admin-module where an endpoint-change shouldn't be an issue. We should fix this for the relevant endpoints only, not some generic solution for _every_ route.
> 
> I'm afraid that we don't know all of the relevant endpoints... Those are not just protocol related endpoints. For example, cron module - in our federation we call its endpoints in crons. WIth metadata aggregator we have automations... The problem is, when upgrading to v3 of SimpleSAMLphp I'll have to upgrade all related external stuff that calls SSP endpoints at the same time. I can't do that in advance because new URLs won't work until I upgrade SSP to v3, and old URLs won't work after I upgrade to v3.
> 
> I'm in favor of "transition period", where in v3 we will support all routes - new and legacy (with new ones being published in metadata). And in a next major version, or even minor - so in v3.1 - we remove support for legacy routes. This way people will have a chance to announce the change, publish any new metadata, for which ever protocol, change any external calls to SSP endpoints to cron, metadata, whatever - while old routes work. With this, in theory, there shouldn't be downtime regarding the new URL form for any module, even custom module ones...

I suggest the following as a solution in transition period: https://github.com/simplesamlphp/simplesamlphp/pull/2630/changes/b17496b079fbc917fb9d1a4de34dd61b70e26ebb

cicnavi:
> Another broad question without a specific answer...
> 
> As the `AssertionConsumerService` updates seem fairly mechanical should we maybe allow /module.php/ to be in there and update that to /module/ for 3.0 so that the same metadata files will work. This will allow limited migration between 2.9 and 3.0 with the same metadata files and if those files are being synced from other sources will allow older ones to continue to work for a limited time.

I suggest the following solution in v3.0 in tranistion period: https://github.com/simplesamlphp/simplesamlphp/pull/2630/changes/b17496b079fbc917fb9d1a4de34dd61b70e26ebb

cicnavi:
> It might be nice for onKernelException() to have an example that also uses TWIG for the returned page. See the example in docs/simplesamlphp-ecp-idp.md which returns some inline html.
> 
> Category: Can easily delay until after merge.

Yes, indeed... I've added twig example in `docs/simplesamlphp-errorhandling.md` (I think you meant that file): https://github.com/simplesamlphp/simplesamlphp/pull/2630/changes/b9cf65583d3413d4c3f3797c990d7f07319cb416

cicnavi:
> In "upgrade-notes-3.0.md" we might like to expand on
> 
> ```
> For module URLs, use the URL helpers provided by SimpleSAMLphp
> ```
> 
> to have one or more examples of the code update for this. Maybe just mentioning to use `Module::getModuleURL()`.
> 
> Likewise "For assets, use the published asset URLs." having a URL there to allow folks to avoid having to work it out. Perhaps mention the new `Module::getModuleAssetUrl()`.

Indeed... here is the added note: https://github.com/simplesamlphp/simplesamlphp/pull/2630/commits/66a9cc71d305f43ed5b1239e00a4523cbccf8b6b

cicnavi:
> For the error handling...
> 
> Somewhere in the docs we might like to add
> 
> ```
> php bin/console debug:event-dispatcher kernel.exception 
> ```
> 
> to see if the error handler is registered.
> 
> As an aside I plan to but have not run a local test on the error handling stuff. Likely an easy exercise but something I plan to do but have not done as yet.

Ok, added: https://github.com/simplesamlphp/simplesamlphp/pull/2630/commits/595d9a4be6d5b92f18213d05f8b0a722b4eb311a

> Easy to delay to after merge: Maybe throw an error if `error.show_function` remains in place in the config? The downside is that the same config.php couldn't then be used between 2.9 and 3.0. The upside is that the user will not be left thinking they have a custom error handler in 3.0 and not have one because they forgot to migrate that part of their setup. A soft option here is we could have a `pedantic_config_check` type of option that is recommended but can be set to `false` to allow a more migration friendly config with the caveat that some options might be ignored and not reported if you force pedantic to false.

Huh... I see what you mean, this could even be implemented in a way to track many config options, and present any deprecated ones in the UI where users will clearly get the message... but I'm now sure about it... but I would certainly like to keep it out of scope of this PR...

tvdijen:
> I'm in favor of "transition period", where in v3 we will support all routes

I can live with that :) One question though:  what would be a reasonable time for the transition period? What is a "normal" interval to refresh metadata? I do this once every three year, because that's the lifetime of our signing keys, however SAML2INT for instance preaches to use long-lived self-signed certificates.. Could easily be 10 yrs?

Other than that, I think this PR is ready to merge.

cicnavi:
> I notice the path change in docs/simplesamlphp-sp-api.md which I assume is ok but thought I'd mention it to double check...
> 
> a link containing
> 
> ```
>    module.php/core
> ```
> 
> going to
> 
> ```
>    module/saml
> ```

The related method is below, it points to `'saml/sp/login/...'`, so the mistake was before..

```php
    public function getLoginURL(?string $returnTo = null): string
    {
        if ($returnTo === null) {
            $httpUtils = new Utils\HTTP();
            $returnTo = $httpUtils->getSelfURL();
        }

        $login = Module::getModuleURL('saml/sp/login/' . urlencode($this->authSource), [
            'ReturnTo' => $returnTo,
        ]);

        return $login;
    }
 ```

tvdijen:
We should check the docs for any mention of `.php` if that didn't already happen as part of this PR

cicnavi:
> > I'm in favor of "transition period", where in v3 we will support all routes
> 
> I can live with that :) One question though: what would be a reasonable time for the transition period? What is a "normal" interval to refresh metadata? I do this once every three year, because that's the lifetime of our signing keys, however SAML2INT for instance preaches to use long-lived self-signed certificates.. Could easily be 10 yrs?
> 
> Other than that, I think this PR is ready to merge.

I don't know what is reasonable timeframe would be.... I didn't introduce a config option to enable or disable legacy URLs in the end, to not pollute config file with something that will be removed anyways in the future. So everyone will have to wait for a new SSP version which will have legacy URLs removed. For which version should this be, you guys can decide this at any time - just remove the one-line in Kernel to stop supporting legacy URLs in some future version release...

cicnavi:
> We should check the docs for any mention of `.php` if that didn't already happen as part of this PR

This should already be fixed.

cicnavi:
> Other than those thoughts I need to pull this and test it locally and I guess then the "should be carefully tested before merge" part you mention in the initial comment phase begins :)

I think I went through all your suggestions. I did try things out while working on this, but it would be great if you guys do that also...

tvdijen (approved):

monkeyiq:
I knew this would have some interaction with https://github.com/simplesamlphp/simplesamlphp/pull/2645

I can take a look at updating the 4 breaks in 2645

Issue #2484, opened 2025-07-09
SimpleSAMLphp Session Management Issues with Nginx Unit


vaillen:
## Overview

We've encountered significant challenges when deploying SimpleSAMLphp behind Nginx Unit, particularly with session persistence and SAML authentication flows. While SimpleSAMLphp functions correctly with Apache, we're consistently seeing session data loss when using Nginx Unit, primarily manifesting as "The POST data we should restore was lost" errors during SAML authentication.

## Symptoms

1. **Session and POST Data Loss**
   - SAML responses from IdP to SP through the ACS endpoint fail to maintain POST data
   - Session data doesn't persist between authentication requests
   - SAML authentication flow breaks with "The POST data we should restore was lost"

2. **Nginx Unit-Specific Challenges**
   - Inability to modify incoming request headers in Nginx Unit
   - No straightforward method to inject `X-Forwarded-Proto: https` header
   - Routing configuration in Nginx Unit significantly impacts SimpleSAMLphp's behavior

## Root Cause Analysis

1. **`isHTTPS()` Implementation Limitations**
   The current implementation in SimpleSAMLphp:
   ```php
   public function isHTTPS(): bool
   {
       return strpos($this->getSelfURL(), 'https://') === 0;
   }
   ```
   This approach is too simplistic as it only checks the URL scheme without considering proxy headers or the specific requirements of SAML ACS endpoints.

2. **Session Cookie Handling**
   - Mismatch between `session.cookie.secure` settings and actual protocol
   - Cross-domain POST request issues with `SameSite` cookie attribute
   - Improper `session.cookie.path` configurations leading to session data loss

3. **Lack of POST Data Recovery**
   No robust mechanism exists to recover POST data when session failures occur

## Implemented Solutions

1. **Enhanced `isHTTPS()` Method**
   We've improved the HTTPS detection to handle various proxy configurations and edge cases:
   ```php
   public function isHTTPS(): bool
   {
       // Special handling for SAML ACS endpoint
       $currentUrl = $_SERVER['REQUEST_URI'] ?? '';
       if (strpos($currentUrl, '/module.php/saml/sp/saml2-acs.php/') !== false) {
           // Force HTTPS for SAML ACS to maintain session integrity
           Logger::debug('isHTTPS: Treating SAML ACS endpoint as HTTPS for session handling');
           return true;
       }
       
       // Support for common proxy headers
[...]

Issue #879, opened 2018-06-27
Allow setting secure cookie when server connection is http


jplahti:
Checking of secure server connection with setting "session.cookie.secure = true" leads to unwanted behaviour on load balancer TLS-terminated environments, where the application server running SimpleSAMLphp sees internal http connections only and prevents setting cookie secure flag.

Please remove this check, because interpreting the secure flag on cookie should be left for the browser to deside. Or maybe even better add extra setting e.g. "behindsecureproxy", which can be used to fine tune the behaviour.

jaimeperez:
Hi @jplahti,

The test serves a purpose: letting the admin know that there is a misconfiguration. If you are accessing SimpleSAMLphp through plain HTTP and `session.secure.cookie` is `true`, there is an incongruency there. Cookies marked secure (as you are requesting in the configuration) cannot be sent through plain HTTP, and that leads to lost sessions.

If that's causing issues in your particular deployment, then you haven't configured SimpleSAMLphp properly. Since you have a load balancer in front of your backend, and you want to set `session.cookie.secure` in the configuration, I assume all access to your app/SimpleSAMLphp must be done via HTTPS. In that case, you should configure the base URL accordingly. Look at the `baseurlpath` configuration option. We even have a comment explaining how to configure it in your specific deployment:

> The full url format is useful if your SimpleSAMLphp setup is hosted behind a reverse proxy. In that case you can specify the external url here.

jplahti:
"baseurlpath" was configured correctly, but while melding configs from older version I had missed the new "application->baseURL" setting. Setting that fixed the issue.

Issue #808, opened 2018-03-06
CriticalConfigurationError secure cookie on plain http after upgrade to 1.15.4


Don-Raf:
After upgrading to 1.15.4 I started getting "Uncaught SimpleSAML\Error\CriticalConfigurationError: The configuration is invalid: Setting secure cookie on plain HTTP is not allowed. /lib/SimpleSAML/Session.php:296"

In the config.php the following was set and was working fine behind a reverse nginx proxy which terminates the TLS connection:

'baseurlpath' => 'https://FQDN/simplesaml/',
'session.cookie.secure' => true,

After the upgrade I had to change 'session.cookie.secure' to false to prevent the error.

Something that changed between 1.13.2 and 1.15.4 that I should be aware of?
I compared the config.php files and couldn't find anything obvious.

jaimeperez:
Can you provide a full stack trace of the error? Do you get it when you call SimpleSAMLphp for authentication from your application?

If so, you need to set the following configuration option:

```php
    'application' => array(
        'baseURL' => 'https://FQDN'
    ),
```

Don-Raf:
Sorry for the late reply. 

Set up a test installation and verified the 'application' setting worked.
Didn't find that option since the old 1.13.2 installation config.php didn't have it. Copied in from new config and using secure cookie worked again. 

Thanks for the help!

modxcustomize:
> session.cookie.secure

Hi, sorry but what is the FQDN standing for? Is it the URL of the app like http://appdomain.tld ? What exactly it should stand for?

thijskh:
Fully Qualified Domain Name, indeed as you say.

modxcustomize:
> Fully Qualified Domain Name, indeed as you say.

Thanks. So, it is set as 
`'application' => array(
        //'baseURL' => 'https://FQDN'
        'baseURL' => 'http://sso.sergey.....com'
    ),`
did it needs any slash or else, please? The problem I am trying to resolve is that the Admin login keeps redirecting and finally dies with browser error like ERR TOO MANY REDIRECTS

modxcustomize:
> The admin login is something from SimpleSAMLphp itself. The config you reference is in your own aplication, so unrelated.
> 
> I would investigate with the browser debug bar what exactly happens on authentication, where it redirects to and whether there's anything like cookies getting lost or other things you would not expect in the login flow. This requires some debugging from your side, not something we can remotely help with just by describing the symptom.

[...]

PR #2638, opened 2026-05-26
HTTP: have getServerHTTPS() honor https baseurlpath when $_SERVER['HTTPS'] is absent

### LINKED ISSUE #2637: Utils\HTTP::getServerHTTPS() ignores explicit https baseurlpath behind TLS-terminating proxy, breaking secure-cookie SP path
# `Utils\HTTP::getSelfURL()` external-script branch ignores `baseurlpath`, throws `CriticalConfigurationError` behind TLS-terminating proxy

## Specifics of your environment

1. **Role:** SP (using `\SimpleSAML\Auth\Simple` programmatically from a host application — the canonical pattern shown in [`docs/simplesamlphp-sp.md`](https://github.com/simplesamlphp/simplesamlphp/blob/master/docs/simplesamlphp-sp.md))
2. **SimpleSAMLphp:** v2.5.0 (the relevant code in `Utils/HTTP.php` is identical on `master` as of 2026-05-25; verified by diff)
3. **PHP:** verified to reproduce identically on PHP 8.3 and PHP 8.5 (`php:8.3-fpm-alpine` and `php:8.5-fpm-alpine` base images)
4. **Platform:** verified on Alpine Linux (the reproducer's base image)
5. **Webserver:** nginx 1.27 in front of php-fpm in the reproducer. The bug is rooted in `$_SERVER['HTTPS']` being empty when PHP is reached, so any topology with the same property (TLS terminated upstream of nginx and stock nginx `fastcgi_params`, or any other web server that does not set `$_SERVER['HTTPS']='on'` for proxied HTTPS requests) should exhibit it.

## Describe the bug

When `\SimpleSAML\Auth\Simple` is invoked from a script outside SimpleSAMLphp's own `public/` directory (the canonical SP usage pattern shown in `docs/simplesamlphp-sp.md`, used when a host PHP application embeds SimpleSAMLphp as a library and triggers authentication from its own controller code), `Session::getSessionFromRequest()` throws:

```
SimpleSAML\Error\CriticalConfigurationError:
  Setting secure cookie on plain HTTP (except on localhost) is not allowed.
```

…**even when `baseurlpath` is set to a full `https://` URL**, which the maintainers have repeatedly stated ([#795][1], [#376][2], [#324][3]) is the correct way to declare the deployment scheme behind a TLS-terminating reverse proxy.

The root cause is that `Utils\HTTP::getSelfURL()` has two branches:

- **"Normal" branch** (script inside `public/`): builds the URL from `getBaseURL()`, which correctly honors `baseurlpath`'s scheme. Works as advertised.
- **"External script" branch** (script outside `public/`): consults `application.baseURL` if set, otherwise falls back to `getServerHTTPS()` which reads `$_SERVER['HTTPS']` directly. **`baseurlpath` is never consulted.**

Behind a typical nginx → php-fpm topology where TLS is terminated *upstream* of nginx (e.g., by a cloud load balancer), `$_SERVER['HTTPS']` is empty in PHP. The stock `nginx:1.27-alpine` `/etc/nginx/fastcgi_params` does include the line `fastcgi_param HTTPS $https if_not_empty;`, but `$https` is empty when nginx itself receives a plain HTTP connection, so `if_not_empty` skips it and the HTTPS param is never passed to PHP-FPM. Result: `getServerHTTPS()` returns `false`, `isHTTPS()` returns `false`, the secure-cookie check in `setCookie()` throws `CannotSetCookie::SECURE_COOKIE`, and `Session::getSessionFromRequest()` promotes it to `CriticalConfigurationError`.

The workaround that does work is to set `application.baseURL`. Maintainer guidance in [#2484][i2484] (2024-07-10) makes this explicit:

> "Like I said, the `baseurlpath` is important, and if you are the SP (which I understand you are) **you should also configure the `application.baseURL` with the same value as the `baseurlpath`**."  
> — @tvdijen

This is correct, and it works. The discoverability problem is that this guidance lives only in the issue-tracker history:

- The config option is **commented out** in the shipped `config.php` template with a description ("useful when the environment observed in the server differs from the one observed by end users") that does not flag it as **required** for the external-script SP pattern.
- The error message at the throw site does not point to it.
- The main SP documentation ([`docs/simplesamlphp-sp.md`][docs-sp]) shows `new \SimpleSAML\Auth\Simple('default-sp')` as the canonical pattern, but does not mention `application.baseURL` at all.

Users have hit this same root cause repeatedly over a decade — verified by direct review of each:

- [#77][i77] (2013) — AWS load balancer
- [#343][i343] (2015) — AWS ELB, exact same symptom
- [#654][i654] (2017) — direct getServerHTTPS() bug report
- [#808][i808] (2017) — post-upgrade regression, same error string verbatim
- [#879][i879] (2017) — load balancer TLS termination, requesting the same workaround
- [#2484][i2484] (2024) — same root cause, Nginx Unit deployment

## To Reproduce

A minimal Docker Compose reproducer is attached separately. The essence:

1. nginx receives **plain HTTP** from a TLS-terminating upstream proxy (a common production topology where a cloud load balancer terminates TLS and forwards over HTTP: client → load balancer → nginx → php-fpm). The upstream proxy is simulated by the curl invocation.
2. The php-fpm container runs SimpleSAMLphp v2.5.0 with:
   - `baseurlpath = 'https://app.example.test/simplesaml/'`  (full URL — the documented workaround)
   - `session.cookie.secure = true`
   - `application.baseURL` unset (matching the shipped default)
3. A small `external.php` script outside SimpleSAMLphp's `public/` does:
   ```php
   require_once __DIR__ . '/../vendor/autoload.php';
   $as = new \SimpleSAML\Auth\Simple('demo-sp');
   ```
4. `curl -H "X-Forwarded-Proto: https" -H "Host: app.example.test" http://localhost:8080/external.php`

Verified output (captured against `simplesamlphp/simplesamlphp:2.5.0`, `php:8.3-fpm-alpine`, `nginx:1.27-alpine`):

```
=== $_SERVER snapshot (relevant keys) ===
  $_SERVER[HTTPS] = NULL
  $_SERVER[HTTP_X_FORWARDED_PROTO] = 'https'
  $_SERVER[HTTP_HOST] = 'app.example.test'
  $_SERVER[SERVER_PORT] = '80'
  $_SERVER[REQUEST_URI] = '/external.php'
  $_SERVER[SCRIPT_FILENAME] = '/var/www/app/external.php'

=== SimpleSAMLphp Utils\HTTP probes ===
  getServerHTTPS()        = false
  isHTTPS()               = false
  isSecureCookieAllowed() = false
  getSelfURL()            = 'http://app.example.test/external.php'
  getBaseURL()            = 'https://app.example.test/simplesaml/'

=== Now trying Auth\Simple ===
REPRODUCED: SimpleSAML\Error\CriticalConfigurationError:
  The configuration is invalid: Setting secure cookie on plain HTTP (except on localhost) is not allowed.
```

Notice that `getBaseURL()` correctly reports `https://...` (it consults `baseurlpath`), but `getSelfURL()` reports `http://...` for the same request — its external-script branch ignores `baseurlpath`. The two methods of the same `Utils\HTTP` instance disagree about the scheme of the same request. That is the bug in a single line of output.

### ⚠️ Reproducibility trap

This bug **does not reproduce** if nginx itself terminates TLS (`listen 443 ssl`), because nginx's stock `/etc/nginx/fastcgi_params` includes the line `fastcgi_param HTTPS $https if_not_empty;`. When nginx has the TLS connection itself, `$https` evaluates to `'on'` and the param is forwarded; PHP sees `$_SERVER['HTTPS']='on'` and the bug never fires. The reproducer here is set up specifically with nginx receiving a plain HTTP connection from a simulated upstream LB so that `$https` is empty and `if_not_empty` skips the HTTPS param.

If you try to reproduce by terminating TLS at nginx directly, you will see the bug fail to reproduce and may incorrectly close as not-a-bug.

## Expected behavior

Either:

(a) Setting `baseurlpath` to a full `https://` URL declares the deployment scheme consistently across both branches of `getSelfURL()`, OR

(b) The library makes it unambiguous that `application.baseURL` is **required** for the external-script SP pattern behind a TLS-terminating proxy — via the error message, the config template comment, and the documentation.

The current behavior — `baseurlpath` works for `/simplesaml/*` URLs but is silently ignored for external SP scripts — is a discoverability trap. Users who follow the maintainer-stated workaround (full-URL `baseurlpath`) and use the documented `Auth\Simple` API hit a generic-sounding error with no hint of the actual fix.

## A note on production behaviour

The reproducer above is **deterministic** — the bug fires on every request because every request hits the cookie-setting path. In real production traffic the same configuration can present intermittently (some external-script requests in a burst succeed, others throw this error within seconds of each other on the same client). The mechanism is the cookie short-circuit in `Session::getSessionFromRequest()`:

```php
if ($session instanceof Session) {
    return self::load($session);   // returns existing session, NEVER reaches cookie-setting
}
...
self::load(new Session());        // ← throws CannotSetCookie::SECURE_COOKIE here
```

Requests that arrive with a valid `SimpleSAMLSessionID` cookie from a prior session load the existing session and bypass the cookie-setting path entirely, succeeding. Requests in the same burst that don't include the cookie (for example, parallel-fetch races in a WebView client, or `credentials: 'omit'` on some calls) hit `new Session()` and throw. Same client, same URL, alternating outcome within milliseconds.

## Proposed fix

A separate PR proposes a narrow, backward-compatible change: have `getServerHTTPS()` consider `baseurlpath`'s declared scheme **when `$_SERVER['HTTPS']` is absent AND baseurlpath explicitly declares `https://` for the current request's host**. This:

- Never reads any client-controlled header (consistent with the maintainer position taken in [#795][1], [#376][2], [#324][3], and [#491][p491]).
- Preserves admin's explicit `$_SERVER['HTTPS'] = 'off'` (IIS pattern) — fallback only activates when the key is missing, not when it's explicitly off.
- Preserves all 880 existing tests, verified by running the full PHPUnit suite on the upstream master branch.

Because `getServerHTTPS()` is also called by `getServerPort()`, `getBaseURL()`'s relative-path branch, `getSelfURL()`'s external-script branch (the bug site), and `Auth\Simple::getProcessedURL()`, the fix benefits multiple call sites — most notably `Auth\Simple::getProcessedURL()`, which currently builds `http://` return URLs behind reverse proxies.

[1]: https://github.com/simplesamlphp/simplesamlphp/pull/795
[2]: https://github.com/simplesamlphp/simplesamlphp/pull/376
[3]: https://github.com/simplesamlphp/simplesamlphp/pull/324
[p491]: https://github.com/simplesamlphp/simplesamlphp/pull/491
[i77]: https://github.com/simplesamlphp/simplesamlphp/issues/77
[i343]: https://github.com/simplesamlphp/simplesamlphp/issues/343
[i654]: https://github.com/simplesamlphp/simplesamlphp/issues/654
[i808]: https://github.com/simplesamlphp/simplesamlphp/issues/808
[i879]: https://github.com/simplesamlphp/simplesamlphp/issues/879
[i2484]: https://github.com/simplesamlphp/simplesamlphp/issues/2484
[docs-sp]: https://github.com/simplesamlphp/simplesamlphp/blob/master/docs/simplesamlphp-sp.md

esetnik:
# Make `baseurlpath`'s HTTPS declaration apply consistently in `getServerHTTPS()`

Fixes #2637.

## Problem

`Utils\HTTP::getServerHTTPS()` consults only `$_SERVER['HTTPS']`. In nginx → php-fpm deployments behind a TLS-terminating reverse proxy, `$_SERVER['HTTPS']` is absent — nginx's stock `fastcgi_params` ships `fastcgi_param HTTPS $https if_not_empty;`, and `$https` is empty when nginx itself receives plain HTTP from the upstream proxy. So `getServerHTTPS()` returns `false` for what is actually an HTTPS request, breaking five downstream call sites:

| Caller | Symptom of the bug |
|---|---|
| `Utils\HTTP::getSelfURL()` external-script branch | `$as = new \SimpleSAML\Auth\Simple(…)` from a host application's own controller throws `CriticalConfigurationError: Setting secure cookie on plain HTTP` (the most visible failure mode) |
| `Auth\Simple::getProcessedURL()` | Returns `http://...` return-to URLs even when the original request was HTTPS |
| `Utils\HTTP::getServerPort()` | Defaults to port 80 instead of 443 when `SERVER_PORT` is missing |
| `Utils\HTTP::getBaseURL()` relative-path branch | Builds the auto-detected base URL with `http://` instead of `https://` when baseurlpath is a relative path |

The maintainer-prescribed workaround for reverse-proxy deployments — set `baseurlpath` to a full `https://` URL, restated multiple times in [#795](https://github.com/simplesamlphp/simplesamlphp/pull/795), [#376](https://github.com/simplesamlphp/simplesamlphp/pull/376), [#324](https://github.com/simplesamlphp/simplesamlphp/pull/324), [#491](https://github.com/simplesamlphp/simplesamlphp/pull/491) — works for `Utils\HTTP::getBaseURL()`'s full-URL branch (used by the "normal" branch of `getSelfURL()` for SimpleSAMLphp's own `/saml/*` URLs) but is silently ignored by `getServerHTTPS()` and therefore by the four other callers above.

For SPs that use the canonical `Auth\Simple` pattern from a host application's own controller, the maintainer's verbal guidance in [#2484](https://github.com/simplesamlphp/simplesamlphp/issues/2484#issuecomment) is "also set `application.baseURL`." That works, but it's not documented in the config template, the SP documentation, or the error message — users repeatedly burn hours discovering it (a decade of duplicate reports: [#77](https://github.com/simplesamlphp/simplesamlphp/issues/77), [#343](https://github.com/simplesamlphp/simplesamlphp/issues/343), [#654](https://github.com/simplesamlphp/simplesamlphp/issues/654), [#808](https://github.com/simplesamlphp/simplesamlphp/issues/808), [#879](https://github.com/simplesamlphp/simplesamlphp/issues/879), [#2484](https://github.com/simplesamlphp/simplesamlphp/issues/2484)).

## Approach

Threads the maintainer-stated philosophy explicitly:

1. **Do not trust client-supplied `X-Forwarded-*` headers.** Consistent with the position taken in [#795](https://github.com/simplesamlphp/simplesamlphp/pull/795), [#376](https://github.com/simplesamlphp/simplesamlphp/pull/376), [#324](https://github.com/simplesamlphp/simplesamlphp/pull/324), and [#491](https://github.com/simplesamlphp/simplesamlphp/pull/491).
2. **Trust admin-set configuration.** `baseurlpath` is set in `config.php` by the deployment administrator — this is the same source of truth the maintainers have repeatedly endorsed.
3. **Preserve every existing semantic.** When `$_SERVER['HTTPS']` is present (set or `'off'`), behavior is byte-for-byte unchanged. The new fallback only activates when the key is **absent**.
4. **Conservative host match.** Only trust `baseurlpath`'s declared scheme when its host exactly matches the current request's host (case-insensitive). Prevents a multi-host installation from over-promoting unrelated requests to HTTPS just because one configured vhost happens to declare HTTPS.

### `src/SimpleSAML/Utils/HTTP.php` — `getServerHTTPS()`

Refactored so the original three-case `$_SERVER['HTTPS']` interpretation runs only when `array_key_exists('HTTPS', $_SERVER)`. When the key is absent, falls through to a new admin-config-based check:

```php
public function getServerHTTPS(): bool
{
    if (array_key_exists('HTTPS', $_SERVER)) {
        if ($_SERVER['HTTPS'] === 'off') {
            return false;
        }
        return !empty($_SERVER['HTTPS']);
    }

    try {
        $cfg = Configuration::getInstance();
    } catch (\Exception $e) {
        return false;
    }

    $baseURL = $cfg->getOptionalString('baseurlpath', null);
    if (
        $baseURL !== null
        && preg_match('#^https://([^/:]+)(?::([0-9]+))?#', $baseURL, $matches)
    ) {
        $configuredHost = strtolower($matches[1]);
        $currentHost = strtolower($this->getServerHost());
        if ($configuredHost === $currentHost) {
            return true;
        }
    }

    return false;
}
```

### `tests/src/SimpleSAML/Utils/HTTPTest.php` — 8 new test methods

| Test | Asserts |
|---|---|
| `testGetServerHTTPSHonorsHttpsBaseurlpathBehindProxy` | The bug — fails on master without the fix, passes with it |
| `testGetServerHTTPSBaseurlpathHostMismatchDoesNotForceHTTPS` | Host-match guard rejects cross-host baseurlpath |
| `testGetServerHTTPSReturnsTrueWhenHttpsEnvIsSet` | Branch A short-circuit preserved (HTTPS=on with conflicting baseurlpath still returns true) |
| `testGetServerHTTPSRespectsExplicitHttpsOff` | IIS-style explicit `HTTPS=off` is not overridden by fallback |
| `testGetServerHTTPSCaseInsensitiveHostMatch` | Host comparison is case-insensitive (DNS rule) |
| `testGetServerHTTPSHostWithPortInRequest` | `HTTP_HOST` with port matches baseurlpath without port (port stripped by `getServerHost()`) |
| `testGetServerHTTPSIgnoresRelativeBaseurlpath` | Default relative-path baseurlpath doesn't trigger the fallback |
| `testGetServerHTTPSIgnoresHttpBaseurlpath` | `http://` baseurlpath is not coerced into HTTPS |

## Verification

```
vendor/bin/phpunit --no-coverage
...
Tests: 888, Assertions: 2633, Warnings: 1, PHPUnit Notices: 24, Skipped: 11.
```

- 880 → **888 tests, all pass** (8 added, 0 regressions).
- The new bug-demonstration test fails on unmodified master:
  > `Failed asserting that false is true.` at `testGetServerHTTPSHonorsHttpsBaseurlpathBehindProxy`
- The same test passes after the patch is applied.
- All 4 other `getServerHTTPS()` callers in the library benefit: `getServerPort()`, `getBaseURL()`'s relative-path branch, `getSelfURL()`'s external-script branch (the bug site), and `Auth\Simple::getProcessedURL()`.

## Backward compatibility

| Pre-existing scenario | Behavior on master | Behavior with this PR |
|---|---|---|
| `$_SERVER['HTTPS']='on'` | true | **true** (unchanged) |
| `$_SERVER['HTTPS']='off'` | false | **false** (unchanged — explicit off still wins) |
| `$_SERVER['HTTPS']=''` | false (empty string check) | **false** (unchanged) |
| `$_SERVER['HTTPS']` missing, no baseurlpath / relative baseurlpath | false | **false** (unchanged) |
| `$_SERVER['HTTPS']` missing, http:// baseurlpath | false | **false** (unchanged — http baseurlpath rejected) |
| `$_SERVER['HTTPS']` missing, https:// baseurlpath, different host | false | **false** (unchanged — host-match guard rejects) |
| `$_SERVER['HTTPS']` missing, https:// baseurlpath, **same host** | false (THE BUG) | **true** (FIXED) |

Production deployments that already set `application.baseURL` as a workaround are unaffected — `getSelfURL()`'s external branch checks that first.

## Security considerations

The new fallback never reads any client-controlled header. The two inputs are:
- `baseurlpath` from `Configuration` (admin-set in `config.php`).
- `$_SERVER['HTTP_HOST']` via `getServerHost()` (already used by the library; `getServerHost()` strips any port and falls back to `SERVER_NAME` / `'localhost'`).

There is no widening of the trust surface; the change is strictly within the library's existing trust model.

## Reproducer

A self-contained Docker Compose reproducer is attached to the linked issue. Tear-down command included.

### DIFF
--- a/src/SimpleSAML/Utils/HTTP.php
+++ b/src/SimpleSAML/Utils/HTTP.php
@@ -209,13 +209,7 @@ class HTTP
         // virtual host the library knows about is configured for HTTPS.
         //
         // No client-controlled header is read at any point.
-        try {
-            $cfg = Configuration::getInstance();
-        } catch (\Exception $e) {
-            // Configuration not yet loaded — nothing more to check.
-            return false;
-        }
-
+        $cfg = Configuration::getInstance();
         $baseURL = $cfg->getOptionalString('baseurlpath', null);
         if (
             $baseURL !== null
--- a/src/SimpleSAML/Utils/HTTP.php
+++ b/src/SimpleSAML/Utils/HTTP.php
@@ -178,18 +178,57 @@ class HTTP
      */
     public function getServerHTTPS(): bool
     {
-        if (!array_key_exists('HTTPS', $_SERVER)) {
-            // not an https-request
+        // When $_SERVER['HTTPS'] is set — by a web server that terminates
+        // TLS itself — it is the authoritative signal. Preserve the
+        // existing three-case interpretation exactly, so an admin's
+        // explicit 'off' (IIS convention) still wins.
+        if (array_key_exists('HTTPS', $_SERVER)) {
+            if ($_SERVER['HTTPS'] === 'off') {
+                return false;
+            }
+            return !empty($_SERVER['HTTPS']);
+        }
+
+        // $_SERVER['HTTPS'] is COMPLETELY ABSENT. This is the typical
+        // php-fpm case behind a TLS-terminating reverse proxy: nginx's
+        // stock fastcgi_params line `fastcgi_param HTTPS $https
+        // if_not_empty;` only sends the param when $https itself is
+        // non-empty, which only happens when nginx terminates TLS.
+        //
+        // Fall back to the admin-set baseurlpath. We trust it only when:
+        //   (a) it is a full URL starting with https://, AND
+        //   (b) its host exactly matches the current request's host.
+        //
+        // (a) restricts to deployments where the admin has explicitly
+        // declared an HTTPS deployment scheme (the maintainer-prescribed
+        // form for reverse-proxy setups, e.g. as in
+        // github.com/simplesamlphp/simplesamlphp/pull/795).
+        //
+        // (b) prevents a multi-host SimpleSAMLphp installation from
+        // over-promoting unrelated requests to HTTPS just because some
+        // virtual host the library knows about is configured for HTTPS.
+        //
+        // No client-controlled header is read at any point.
+        try {
+            $cfg = Configuration::getInstance();
+        } catch (\Exception $e) {
+            // Configuration not yet loaded — nothing more to check.
             return false;
         }
 
-        if ($_SERVER['HTTPS'] === 'off') {
-            // IIS with HTTPS off
-            return false;
+        $baseURL = $cfg->getOptionalString('baseurlpath', null);
+        if (
+            $baseURL !== null
+            && preg_match('#^https://([^/:]+)(?::([0-9]+))?#', $baseURL, $matches)
+        ) {
+            $configuredHost = strtolower($matches[1]);
+            $currentHost = strtolower($this->getServerHost());
+            if ($configuredHost === $currentHost) {
+                return true;
+            }
         }
 
-        // otherwise, HTTPS will be non-empty
-        return !empty($_SERVER['HTTPS']);
+        return false;
     }
 
 
--- a/src/SimpleSAML/Utils/HTTP.php
+++ b/src/SimpleSAML/Utils/HTTP.php
@@ -218,9 +218,29 @@ class HTTP
             $configuredHost = strtolower($matches[1]);
             $currentHost = strtolower($this->getServerHost());
             if ($configuredHost === $currentHost) {
+                Logger::debug(
+                    "getServerHTTPS(): no \$_SERVER['HTTPS']; treating the request as HTTPS "
+                    . "because the 'baseurlpath' host '" . $configuredHost
+                    . "' matches the current host.",
+                );
+
                 return true;
             }
+
+            Logger::debug(
+                "getServerHTTPS(): no \$_SERVER['HTTPS']; not treating the request as HTTPS "
+                . "because the 'baseurlpath' host '" . $configuredHost
+                . "' does not match the current host '" . $currentHost . "'.",
+            );
+
+            return false;
         }
+
+        Logger::debug(
+            "getServerHTTPS(): no \$_SERVER['HTTPS'] and 'baseurlpath' is not a full https:// "
+            . "URL, so the request is not treated as HTTPS. If TLS is terminated at an upstream "
+            . "proxy, set 'baseurlpath' (or 'application.baseURL') to your full https:// URL.",
+        );
 
         return false;
     }
--- a/tests/src/SimpleSAML/Utils/HTTPTest.php
+++ b/tests/src/SimpleSAML/Utils/HTTPTest.php
@@ -808,4 +808,267 @@ class HTTPTest extends ClearStateTestCase
 
         $_SERVER = $originalServer;
     }
+
+    /**
+     * Reproducer guard: when SimpleSAMLphp is invoked from an external
+     * script (script outside its own public/ directory) behind a
+     * TLS-terminating reverse proxy that does not set $_SERVER['HTTPS'],
+     * getServerHTTPS() must honor an https:// baseurlpath whose host
+     * matches the current request's host. Without this, isHTTPS()
+     * returns false and setting a secure cookie raises
+     * CriticalConfigurationError ("Setting secure cookie on plain HTTP").
+     */
+    public function testGetServerHTTPSHonorsHttpsBaseurlpathBehindProxy(): void
+    {
+        $original = $_SERVER;
+        $httpUtils = new Utils\HTTP();
+
+        Configuration::loadFromArray([
+            'baseurlpath' => 'https://app.example.test/simplesaml/',
+        ], '[ARRAY]', 'simplesaml');
+
+        $_SERVER = [
+            'HTTP_HOST'              => 'app.example.test',
+            'HTTP_X_FORWARDED_PROTO' => 'https',
+            'REQUEST_URI'            => '/app/sp.php?foo=bar',
+            'SCRIPT_FILENAME'        => '/var/www/app/sp.php',
+            'SERVER_PORT'            => '80',
+        ];
+
+        $this->assertTrue(
+            $httpUtils->getServerHTTPS(),
+            'getServerHTTPS() must return true when baseurlpath declares https for the current host, '
+            . 'even when $_SERVER[HTTPS] is empty (TLS-terminating-proxy topology).',
+        );
+        $this->assertTrue(
+            $httpUtils->isHTTPS(),
+            'isHTTPS() must return true in the same topology, otherwise secure-cookie setup fails.',
+        );
+        $this->assertStringStartsWith(
+            'https://',
+            $httpUtils->getSelfURL(),
+            'getSelfURL() must return an https:// URL when baseurlpath declares https for the current host.',
+        );
+
+        $_SERVER = $original;
+    }
+
+    /**
+     * Companion guard: when baseurlpath declares https for a DIFFERENT
+     * host than the current request, the host-match guard must prevent
+     * the new fallback from triggering — getServerHTTPS() must fall
+     * through to the original $_SERVER-based detection (which here
+     * yields false because $_SERVER['HTTPS'] is unset).
+     */
+    public function testGetServerHTTPSBaseurlpathHostMismatchDoesNotForceHTTPS(): void
+    {
+        $original = $_SERVER;
+        $httpUtils = new Utils\HTTP();
+
+        Configuration::loadFromArray([
+            'baseurlpath' => 'https://different-host.example.test/simplesaml/',
+        ], '[ARRAY]', 'simplesaml');
+
+        $_SERVER = [
+            'HTTP_HOST'       => 'app.example.test',
+            'REQUEST_URI'     => '/app/sp.php',
+            'SCRIPT_FILENAME' => '/var/www/app/sp.php',
+            'SERVER_PORT'     => '80',
+        ];
+
+        $this->assertFalse(
+            $httpUtils->getServerHTTPS(),
+            'getServerHTTPS() must NOT be coerced to true when baseurlpath declares https for a different host.',
+        );
+
+        $_SERVER = $original;
+    }
+
+
+    /**
+     * Branch A regression guard: when $_SERVER['HTTPS'] is set to 'on',
+     * the existing short-circuit must continue to return true regardless
+     * of any baseurlpath configuration. This guards against a future
+     * refactor that accidentally reorders the checks.
+     */
+    public function testGetServerHTTPSReturnsTrueWhenHttpsEnvIsSet(): void
+    {
+        $original = $_SERVER;
+        $httpUtils = new Utils\HTTP();
+
+        // No baseurlpath at all; HTTPS=on must still be honoured.
+        Configuration::loadFromArray([], '[ARRAY]', 'simplesaml');
+
+        $_SERVER = [
+            'HTTPS'           => 'on',
+            'HTTP_HOST'       => 'app.example.test',
+            'REQUEST_URI'     => '/',
+            'SCRIPT_FILENAME' => '/var/www/app/sp.php',
+            'SERVER_PORT'     => '443',
+        ];
+        $this->assertTrue($httpUtils->getServerHTTPS());
+
+        // HTTPS=on with a totally unrelated baseurlpath also stays true.
+        Configuration::loadFromArray([
+            'baseurlpath' => 'http://different-host.example.test/simplesaml/',
+        ], '[ARRAY]', 'simplesaml');
+        $this->assertTrue(
+            $httpUtils->getServerHTTPS(),
+            'HTTPS=on env must win over a conflicting (http) baseurlpath.',
+        );
+
+        $_SERVER = $original;
+    }
+
+    /**
+     * Critical preservation guard: an admin's explicit $_SERVER['HTTPS']
+     * = 'off' (IIS convention for "no HTTPS for this request") must
+     * still return false, even when baseurlpath declares https for the
+     * matching host. The fallback only applies when HTTPS is absent
+     * from $_SERVER, not when it is explicitly off.
+     */
+    public function testGetServerHTTPSRespectsExplicitHttpsOff(): void
+    {
+        $original = $_SERVER;
+        $httpUtils = new Utils\HTTP();
+
+        Configuration::loadFromArray([
+            'baseurlpath' => 'https://app.example.test/simplesaml/',
+        ], '[ARRAY]', 'simplesaml');
+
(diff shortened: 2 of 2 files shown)


esetnik:
Re: codecov 92.85% patch coverage — the 1 uncovered line was a defensive `try/catch` around `Configuration::getInstance()` that didn't match the rest of `HTTP.php` (e.g., `getBaseURL()` calls `Configuration::getInstance()` without a try/catch). Removed it in [`171f6924a`](https://github.com/simplesamlphp/simplesamlphp/pull/2638/commits/171f6924a) — patch coverage should now be 100%.

tvdijen:
LGTM!  @thijskh would you mind taking a look before I merge this?
We've had tons of issues in the past with changes like this, so a second pair of eyes never hurts.

thijskh:
It looks like a valid improvement. Given the various decisions based on different variables, I wonder if adding some debug-level logging would help. So if deployers do not get the URL they expect, they can know what triggered SSP to conclude or not that it is HTTPS.

tvdijen:
Makes sense! @esetnik is this something you can add for us, please?

esetnik:
Good call — added in [`8c11988`](https://github.com/simplesamlphp/simplesamlphp/pull/2638/commits/8c1198868248213686daada2c453b6d20b632ada). When `$_SERVER['HTTPS']` is absent and `getServerHTTPS()` falls back to the `baseurlpath` check, it now emits a debug-level line for each of the three outcomes:

- baseurlpath host matches the current host → treated as HTTPS (names the host)
- baseurlpath host does **not** match → not HTTPS (names both hosts — this is the "deployer expected a URL but didn't get it" case)
- baseurlpath isn't a full `https://` URL → not HTTPS, with a hint to set `baseurlpath`/`application.baseURL` for upstream-TLS-terminating deployments

I kept the logging to the new fallback branch only — the pre-existing `$_SERVER['HTTPS']` paths are unchanged and unlogged, since `getServerHTTPS()` is called several times per request and logging the common path would just be noise. All three lines are covered by the existing tests.

esetnik:
@tvdijen already added it in [`8c11988`](https://github.com/simplesamlphp/simplesamlphp/pull/2638/commits/8c1198868248213686daada2c453b6d20b632ada) (our comments crossed) — debug logging on each of the three baseurlpath-fallback outcomes, scoped to the new branch only.

One thing I can't fix from a fork: the **PHP-Linter (8.4)** job failed on a transient GitHub API rate limit fetching the `overtrue/phplint` PHAR, not on anything in the diff:

```
[WARNING] GitHub API Rate Limit exceeded - cannot resolve "overtrue/phplint"
[ERROR]   Could not resolve requested PHAR overtrue/phplint
```

PHP-Linter on **8.3 and 8.5 passed on the same commit** (run 26628316432). Could one of you re-run the failed job when convenient? `gh run rerun` needs repo admin rights, which I don't have on the fork.

esetnik:
The **8.5-ubuntu unit-test** failure is also infra, not the diff: `IndexTest::testRedirection` / `testRedirectionFrontpageRedirectOption` died at `BuiltInServer.php:229` with `explode(): Argument #2 must be of type string, false given` — i.e. `curl_exec()` returned `false` because the test's built-in PHP server wasn't reachable (the log also shows `sh: 1: kill: No such process`).

Evidence it's unrelated to this change:
- The same two tests pass locally on this branch, and pass on **8.5-windows + 8.3 + 8.4** in this very run.
- `IndexTest` also exercises a `baseurlpath => 'https://example.org/'` case (line 85) — the one most sensitive to this PR's HTTPS-detection change — and it **passed**. Only the two `curl_exec`-returned-false cases errored.

Both this and the earlier PHP-Linter (8.4) PHAR rate-limit are transient. A re-run should green the pipeline; I can't trigger it from the fork. Thanks for your patience with the flakes!

tvdijen:
It's passing now.. The linters are always failing, so annoying :(

PR #2639, opened 2026-05-28
syntax fix for Accept header

guillomovitch:
An invalid option syntax create two empty headers:
Accept:\r\n
application/samlmetadata+xml:\r\n

Instead of the expected:
Accept: application/samlmetadata+xml:\r\n

This makes strict MDQ servers reject the request.


tvdijen:
Ohh, good catch!  Thanks @guillomovitch

tvdijen (approved):

Issue #2643, opened 2026-06-03
remember_organization cookie is never sent on successful LdapMulti/UserPassOrg login


szabogyula:
## Describe the bug

When using an authentication source based on `UserPassOrgBase` (e.g. `ldap:LdapMulti`) with `remember.organization.enabled => true`, the "Remember my organization" checkbox is shown and checked, but the `<authsource>-organization` cookie is never sent to the browser on a successful login. The organization choice is therefore not remembered on the next visit.

## To Reproduce

1. Configure an `ldap:LdapMulti` authsource with `remember.organization.enabled => true`, `remember.organization.checked => true`, `username_organization_method => none` and a `mapping` of two orgs.
2. Open the login page, select an organization, tick "Remember my organization", log in.
3. Inspect the `Set-Cookie` headers on the `core/loginuserpassorg` POST response.

**Expected:** a `Set-Cookie: <authsource>-organization=<org>; expires=<1 year>; ...` header.
**Actual:** only the standard `SimpleSAML` and `SimpleSAMLAuthToken` session cookies are sent; no `-organization` cookie.

## Root cause

In `modules/core/src/Controller/Login.php`, `handleLogin()` collects the organization cookie into a local `$cookies[]` array, calls `$source::handleLogin(...)`, and only attaches the cookies to the response at the very end of the method (`foreach ($cookies ...) { $t->headers->setCookie(...) } return $t;`).

On a successful authentication, `UserPassOrgBase::handleLogin()` (and the `LdapMulti` override) calls `Auth\Source::completeAuth($state)`, which redirects and does not return. The `foreach` block and `return $t` that attach the cookies are therefore never executed, so the `Set-Cookie` header is never generated. The cookie is only ever attached on the failure path, where the template is re-rendered — which is why the choice is never remembered after a successful login.

## Environment

- SimpleSAMLphp 2.x (master / 2.5.x-dev; the code path is identical in shipped 2.x releases)
- simplesamlphp-module-ldap 2.0.x (`ldap:LdapMulti`)
- PHP 8.2

## Suggested fix

The cookies must be emitted before `$source::handleLogin()` is invoked, since that call redirects and never returns on success.

**Option A:** emit the collected cookies onto the outgoing response right before the `try { $source::handleLogin(...) }` call instead of at the end of the method, keeping the Symfony `Cookie` abstraction.

**Option B:** set the organization cookie directly after it is rendered, before the redirect:

```php
setcookie(
[...]

PR #2645, opened 2026-06-13
Propagate Symfony responses throughout the codebase (src & modules)

tvdijen:
This is a re-do of an older PR: https://github.com/simplesamlphp/simplesamlphp/commit/a6458f5b21297eac76df0565f9f395575871c1b6#diff-cd2cc539ac38ea3aaa66c905a882503ca5dfa6714b7b4ef51ed1f3537a9135cf  (https://github.com/simplesamlphp/simplesamlphp/pull/1760)

I've been picking the changes in src/ and modules/, but I haven't come to the tests just yet.
It's hard, because the old master-branch was half-what migrated to the new saml2-lib. With this PR I'm hoping to peel Symfony responses off.


tvdijen:
@cicnavi @monkeyiq First look?
I think it's already a huge improvement and we could for example fix #2643 with this.

tvdijen:
Idea:

I have left out the migration to PSR-17 to interact with the saml2-lib (see old PR)..
We can easily re-introduce this by backporting the saml2-bindings from saml2v6 into samlv4

monkeyiq:
```
Edit: Never mind, I see now that Template here is XHTML/Template and so is a response.
```
Looking at this, modules/admin/src/Controller/Config.php still falls through to the code that does
```
        return $this->menu->insert($t);
````
Which by looking at Menu should be then trying to return a `Template` object from Config `diagnostics` and `main`.

Looking at  https://github.com/simplesamlphp/simplesamlphp/pull/1760 it doesnt seem there was a Menu.php update there to be picked still.

**Comment by monkeyiq** [comment]
I will go back to reading and using emacs to make cross check comments before posting to avoid noise

File: modules/core/src/Auth/Process/WarnShortSSOInterval.php [hunk 1]
```diff
@@ -52,7 +51,7 @@ public function process(array &$state): void
         // Save state and redirect
         $id = Auth\State::saveState($state, 'core:short_sso_interval');
         $url = Module::getModuleURL('core/short_sso_interval');
-        $httpUtils = new Utils\HTTP();
-        $httpUtils->redirectTrustedURL($url, ['StateId' => $id]);
+        $response = $httpUtils->redirectTrustedURL($url, ['StateId' => $id]);
```
monkeyiq, line 56: I think the httpUtils should still be declared here?
```
$httpUtils = new Utils\HTTP();
```

modules/core/src/Controller/Logout.php
```diff
@@ -72,21 +73,16 @@ public function setAuthState(Auth\State $authState): void
...
-    public function logout(Request $request, string $as): RunnableResponse
+    public function logout(Request $request, string $as): Response
     {
         $auth = new Auth\Simple($as);
-        $returnTo = $this->getReturnPath($request);
-        return new RunnableResponse(
-            [$auth, 'logout'],
-            [$returnTo],
-        );
+        return $auth->login($this->getReturnPath($request));
```
monkeyiq: Shouldn't this be a call to `$auth->logout` instead?

monkeyiq:
I have read over most of this now. I have not checked it out and run it yet.

monkeyiq:
NB: This is more of a comment for a future update if we like the idea. 

One thought I had was to do with getting query parameters...
```
$consumerURL = null;
if ($request->query->has('ConsumerURL')) {
    $consumerURL = $request->query->get('ConsumerURL');
}
```

Looking at the symfony source code it seems that this could be a single line. Though the temptation there is also to use the FILTER to validate things at that time as well. (https://github.com/symfony/symfony/blob/e9a41cc7483070563d0e8264842e064716319e11/src/Symfony/Component/HttpFoundation/InputBag.php#L143)
```
$consumerURL = $request->query->filter( 'ConsumerURL', null );
```

monkeyiq (changes requested):
Only a few little things that were perhaps discovered by another set of eyes.

tvdijen:
> Looking at the symfony source code it seems that this could be a single line.

I'm trying to not change anything in this already huge PR that is not related to passing requests/responses back and forth. Also, I would never use `->filter(.., null)` because it gives a false sense that we're filtering something when we don't.. Reducing the if/has/get to a oneliner makes sense here.

modules/core/src/Auth/Source/AbstractSourceSelector.php
```diff
@@ -62,27 +65,31 @@ public function authenticate(array &$state): void
         }
 
         $state['sourceSelector:selected'] = $source;
-        static::doAuthentication($as, $state);
+        return parent::doAuthentication($request, $as, $state);
```
cicnavi: There is no `parent::doAuthentication` method. Method `doAuthentication` is defined in this file.

modules/multiauth/src/Auth/Source/MultiAuth.php
```diff
@@ -177,26 +181,30 @@ public static function delegateAuthentication(string $authId, array $state): Run
...
+     * @return \Symfony\Component\HttpFoundation\Response
      */
-    public static function doAuthentication(Auth\Source $as, array $state): void
+    public static function doAuthentication(Request $request, Auth\Source $as, array $state): Response
     {
         try {
-            $as->authenticate($state);
+            $response = $as->authenticate($request, $state);
+            if ($response instanceof Response) {
+                return $reponse;
```
cicnavi: Typo causing undefined variable...

modules/saml/src/IdP/SAML2.php
```diff
@@ -299,34 +299,35 @@ private static function getAssertionConsumerService(
...
         $username = null;
 
-        if (isset($_REQUEST['spentityid']) || isset($_REQUEST['providerId'])) {
+        if ($request->query->has('spentityid') || $request->query->has('providerId')) {
             /* IdP initiated authentication. */
 
-            if (isset($_REQUEST['cookieTime'])) {
-                $cookieTime = (int) $_REQUEST['cookieTime'];
+            if ($request->query->has('cookieTime')) {
+                $cookieTime = intval($requqest->query->get('cookieTime'));
```
cicnavi: Typo `requqest`, again unknown variable.

modules/core/src/Auth/Process/Cardinality.php
```diff
@@ -195,8 +195,8 @@ public function process(array &$state): void
         if (array_key_exists('core:cardinality:errorAttributes', $state)) {
             $id = Auth\State::saveState($state, 'core:cardinality');
             $url = Module::getModuleURL('core/error/cardinality');
-            $this->httpUtils->redirectTrustedURL($url, ['StateId' => $id]);
-            return;
+            $response = $httpUtils->redirectTrustedURL($url, ['StateId' => $id]);
```
cicnavi: Undefined `httpUtils`, I see that there is one passed in constructor and defined as property, so that can be used.

modules/core/src/Auth/Process/CardinalitySingle.php
```diff
@@ -122,8 +122,8 @@ public function process(array &$state): void
         if (array_key_exists('core:cardinality:errorAttributes', $state)) {
             $id = Auth\State::saveState($state, 'core:cardinality');
             $url = Module::getModuleURL('core/error/cardinality');
-            $this->httpUtils->redirectTrustedURL($url, ['StateId' => $id]);
-            return;
+            $response = $httpUtils->redirectTrustedURL($url, ['StateId' => $id]);
```
cicnavi: Undefined `httpUtils` again, there is one defined as property which can be used.

modules/saml/src/Auth/Source/SP.php
```diff
@@ -1058,19 +1074,21 @@ public static function reauthPostLogout(IdP $idp, array $state): never
         }
 
         /** @var \SimpleSAML\Module\saml\Auth\Source\SP $sp */
-        $sp = Auth\Source::getById($state['saml:sp:AuthId'], Module\saml\Auth\Source\SP::class);
+        $sp = Auth\Source::getById($state['saml:sp:AuthId'], selfSP::class);
```
cicnavi: I get `undefined class` for `selfSP`, I guess a typo of `self`.

modules/saml/src/Auth/Source/SP.php
```diff
@@ -1138,10 +1157,10 @@ public function logout(array &$state): void
 
         // State variable saml:logout:Type is set to saml1 by us if we cannot properly logout the user
         if ($logoutType === 'saml1') {
-            return;
+            return null;
         }
 
-        $this->startSLO2($state);
+        return new RunnableResponse([$this, 'startSLO2'], [$this->config, $state]);
```
cicnavi: I get `undefined class` for `RunnableResponse`. I guess you wanted to move away from that class anyways, and return Symfony response from `startSLO2`.

modules/cron/src/Controller/Cron.php
```diff
@@ -66,12 +66,15 @@ public function setAuthUtils(Utils\Auth $authUtils): void
     /**
      * Show cron info.
      *
-     * @return \SimpleSAML\XHTML\Template
+     * @return \Symfony\Component\HttpFoundation\Response
      *   An HTML template or a redirection if we are not authenticated.
      */
     public function info(): Template
```
cicnavi: The return type should be `Response`

src/SimpleSAML/Metadata/MetaDataStorageHandler.php
```diff
@@ -58,12 +63,14 @@ public static function getMetadataHandler(): MetaDataStorageHandler
     /**
      * This constructor initializes this metadata storage handler. It will load and
      * parse the configuration, and initialize the metadata source list.
+     *
+     * @param \SimpleSAML\Configuration $globalConfig
      */
-    protected function __construct()
+    protected function __construct(Configuration $globalConfig)
```
cicnavi, line 70: I see that `MetadataStorageHandler` now accepts a global config instance, and you had to pass that instance in a bunch of places. Me personally, would define that parameter as nullable, and then still instantiate it inside if I get null. I mean, the global config instance is singleton anyways... This example is also a perfect one to be defined as a Symfony service, but there are also many other such cases, so not really important for now... Just making comment, and not something that I "require fixing", for both things...

src/SimpleSAML/XHTML/IdPDisco.php
```diff
@@ -405,7 +418,7 @@ protected function saveIdP(): bool
             return false;
         }
 
-        if (array_key_exists('remember', $_GET)) {
+        if ($this->request->request->has('remember')) {
```
cicnavi: Previously this was using GET parameters and now you are using POST. You probably wanted to do `->query`...

modules/admin/src/Controller/Config.php
```diff
@@ -113,12 +116,14 @@ public function diagnostics(Request $request): Template
      * Display the main admin page.
      *
      * @param \Symfony\Component\HttpFoundation\Request $request The current request.
-     *
-     * @return \SimpleSAML\XHTML\Template
+     * @return \Symfony\Components\HttpFoundation\Response
```
cicnavi: I get `undefined namespace` here.

modules/core/src/Controller/Login.php
```diff
@@ -101,6 +101,7 @@ public function welcome(): Template
      * username/password authentication.
      *
      * @param \Symfony\Component\HttpFoundation\Request $request
+     * @param \Symfony\Component\HttpFoudnation\Response
```
cicnavi: Undefined namespace again.

cicnavi: Oh, this should be return, not param

modules/multiauth/src/Auth/Source/MultiAuth.php
```diff
@@ -89,12 +90,13 @@ public function __construct(array $info, array $config)
      * and redirects to a page where the user must select one of these
      * authentication sources.
      *
-     * This method never return. The authentication process is finished
-     * in the delegateAuthentication method.
+     * The authentication process is finished in the delegateAuthentication method.
      *
+     * @oaram \Symfony\Component\HttpFoundation\Request $request
```
cicnavi: Typo oaram

src/SimpleSAML/XHTML/IdPDisco.php
```diff
@@ -504,16 +517,19 @@ protected function filterList(array $list): array
 
     /**
      * Check if an IdP is set or if the request is passive, and redirect accordingly.
+     *
+     * @preturn \Symfony\Component\HttpFoundation\Response|null
```
cicnavi: Typo preturn

modules/exampleauth/src/Auth/Source/External.php
```diff
@@ -185,25 +181,27 @@ public function authenticate(array &$state): void
...
-    public static function resume(Request $request): void
+    public static function resume(Request $request, Auth\State $authState): Response
     {
         /*
          * First we need to restore the $state-array. We should have the identifier for
          * it in the 'State' request parameter.
          */
-        if (!$request->query->has('State')) {
+        if (!$request->query->has('AuthState')) {
             throw new Error\BadRequest('Missing "State" parameter.');
```
cicnavi: The check was changed from checking State to checking AuthState, but the comment and error message stayed with mentioning State.

modules/saml/src/Auth/Source/SP.php
tvdijen: Yes, we want to move away, but I deliberately left this out in a few places because this requires changes to the saml2-lib as well.  I need to double-check, but I suspect this is one of those places.

tvdijen: For now solving the unresolved class by adding a use-statement.

tvdijen:
This PR wouldn't be complete if I left out the saml2-endpoints, but they need the Binding-classes from the saml2v6-lib to properly work with Symfony responses.. Therefore I converted them and then convert the response captured by the v6 bindings back to a v4 message-object. It's not pretty, but we need small steps and this PR is already huge.

Where possible I used the Constants and the exceptions from the new lib.

monkeyiq:
It looks like IndexTest is the only code in the test suite using simplesamlphp-test-framework/src/BuiltInServer.php

There may be a symfony way to do that instead.

tvdijen:
> There may be a symfony way to do that instead.

Probably, yes.. Although  I think what the test is doing can also be tested in the frontpage controller tests.. It's only testing redirects to specific destinations.

tvdijen:
Something strange is going on.. ServiceProviderTest::testACSExpectedIssuerMismatchHardFails is passing locally, but failing on GH Actions

monkeyiq:
The issue with tests/www/IndexTest.php seemed to happen when I merged the ssp3 branch into this here [42a7252](https://github.com/simplesamlphp/simplesamlphp/pull/2645/commits/42a72528b11445a0ff7553ed57ab07bff65b106d) to try to resolve those 4 conflicts that were created.

The IndexTest test runs ok on the ssp3 timeline commits https://github.com/simplesamlphp/simplesamlphp/commits/simplesamlphp-3.0/ ... cfe871c466be4ccb8914dacf1818283fd31bf71e through 4f5d8452ed7fe0373dd68eef873a3d16020198e3

I will dig more into IndexTest tomorrow to try to resolve it. (unless it has been discovered in the meantime).

monkeyiq:
This also completes ok here...
```
$ ./vendor/bin/phpunit --display-warnings --no-coverage ./tests/modules/saml/src/Controller/ServiceProviderTest.php
```

monkeyiq:
I will be trying to get IndexTest happening again as it is currently rather than chasing other changes for the simple error/redirection testing.

tvdijen:
Oh crap, that wasn't supposed to happen :/

PR #2649, opened 2026-07-02
Expose available languages publicly

cicnavi:
I'm implementing support for `ui_locales` paremeter for `oidc` module https://github.com/simplesamlphp/simplesamlphp-module-oidc/pull/350

I need to know langauages supported by SSP, and my initial thought was to rely on `language.available` confg option, but now I see there is additional handling in SSP method `\SimpleSAML\Locale\Language::getInstalledLanguages` which is used to populate `availableLanguages` property. So I propose to make a public getter for that property so this can be fetched by other consumers...

Also, in the constructor there is a step which checks for language request parameter in $_GET and, if set, uses it to set the language. I would propose a switch which can be used to turn this off, so when I call it from `oidc` I don't have this side effect. 

This all should be fully backward compatible.

I see there is a public method `\SimpleSAML\Locale\Language::getLanguageList`, but I would rather propose also adding getter for `availableLanguages`, if you don't mind.

### DIFF
--- a/src/SimpleSAML/Locale/Language.php
+++ b/src/SimpleSAML/Locale/Language.php
@@ -144,16 +144,20 @@ class Language
      * Constructor
      *
      * @param \SimpleSAML\Configuration $configuration Configuration object
+     * @param bool $handleLanguageRequestParameter Whether to handle the language request parameter, if present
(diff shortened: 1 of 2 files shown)


tvdijen:
Makes sense @cicnavi ! Code looks good.

monkeyiq (approved):

PR #2650, opened 2026-07-02
Add some validations when resolving language

cicnavi:
Add some validations against `availableLanguages` when setting a language.


cicnavi:
I'll fix tests...

cicnavi:
This is now updated so it takes into account the scenario when the default language and the custom language function use a value not listed in language.available....

tvdijen (approved):

monkeyiq:
This stood on the toes of the test case of https://github.com/simplesamlphp/simplesamlphp/pull/2649 a little bit. As they didn't really conflict I did a simple keep both merge and a tiny, tiny frame update for the patch to get it going. That seems to have run into issues with whitespace in the web editor but I should be able to update for that now.

monkeyiq (approved):

PR #2654, opened 2026-07-16
fix: header.twig: fix typo in LI HTML tag

vladimir-mencl-eresearch:
Use a *closing* `</li>` tag.

Avoid rendering an extra empty item on the list (really visible only when stylesheets are not loaded)


tvdijen:
Thank you Vladimir!

PR #2658, opened 2026-07-24
Fix discovery with preselect

### LINKED ISSUE #2657: Discovery tries to delegate to multiauth:preselect source even if it is not a valid one for this request
**Specifics of your environment**
I'm on 2.5.2 but i suspect bug to be reproducible in any version >=2.0
This did not occur in <1.18

**To Reproduce**
1. Have a config with multiauth with a preselect authsource
2. Send a SAML request to the SSOService with a RequestedAuthnContext that cannot be fulfilled by preselected authsource
4. Kaboom, Exception 'Invalid authentication source: preselect-example'

**Cause**
This bug is not trivial to reproduce but very easy to understand the cause.
If you reach discovery with a list of sources in state (under MultiAuth::SOURCESID) but have configured multiauth preselect source that isn't in the list, the discovery process will eagerly and wrongly delegate to the preselect despite it not being a valid source at the moment.

The root of the problem is here:
https://github.com/simplesamlphp/simplesamlphp/blob/7665841cbc1d7b48148ca20648f304327da3d5a4/modules/multiauth/src/Controller/DiscoController.php#L123-L126

Which will then cause a failure immediately afterwards:
https://github.com/simplesamlphp/simplesamlphp/blob/7665841cbc1d7b48148ca20648f304327da3d5a4/modules/multiauth/src/Auth/Source/MultiAuth.php#L167-L168

**Expected behavior**
It should not try to delegate to preselect when it is not on the list

SrGesus:
Fixes #2657

It's a pretty straight forward fix, simply don't delegate to preselect if it is not on the list.

### DIFF
--- a/modules/multiauth/src/Controller/DiscoController.php
+++ b/modules/multiauth/src/Controller/DiscoController.php
@@ -120,7 +120,7 @@ class DiscoController
             return MultiAuth::delegateAuthentication($selectedSource, $state);
         }
 
-        if (array_key_exists('multiauth:preselect', $state)) {
+        if (array_key_exists('multiauth:preselect', $state) && array_key_exists($state['multiauth:preselect'], $state[MultiAuth::SOURCESID])) {
             $source = $state['multiauth:preselect'];
             return MultiAuth::delegateAuthentication($source, $state);
         }


SrGesus:
Now that i think about it this could also be done the other way

Here in multiauth unset preselect if is not in new_sources

https://github.com/simplesamlphp/simplesamlphp/blob/e5ce356c42789cdd4921e8248d7d2a11b7c159a3/modules/multiauth/src/Auth/Source/MultiAuth.php#L104-L130

tvdijen:
Good catch! Thanks @SrGesus

tvdijen (approved):

PR #2666, opened 2026-08-07
Fix variable name typo in `MetaDataStorageHandlerXML`

cicnavi:
In `src/SimpleSAML/Metadata/MetaDataStorageHandlerXML.php` there was `$srcXML` which should be `$srcXml`. This caused `$entities = SAMLParser::parseDescriptorsString($srcXml);` to throw even when XML was successfully fetched from URI.

There are also other commits which are not mine, and which I'm not sure why they are here. I guess because again there was cherry-picking, or rebasing...

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -12,7 +12,14 @@ The 1.19 branch will receive security and critical bug fixes until the end of 20
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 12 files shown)


tvdijen (approved):

tvdijen:
I wasn't aware that cherry-picking would cause this. I use it a lot to distribute fixes across multiple release-branches.. Not sure what the appropriate way would be to distribute this fix to the other branches now

PR #2672, opened 2026-08-11
Fix admin federation page calling a removed SAMLBuilder method for `adfs`

cicnavi:
This delegates metadata generation to the adfs module, which already builds and signs this document for its own endpoint. Also, widen the catch to Throwable so an \Error can be caught too.

### DIFF
--- a/docs/simplesamlphp-changelog.md
+++ b/docs/simplesamlphp-changelog.md
@@ -7,6 +7,8 @@ See the [upgrade notes](https://simplesamlphp.org/docs/stable/simplesamlphp-upgr
 
 ## Version 2.5.4
 
+Released TBD
+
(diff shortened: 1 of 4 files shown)


tvdijen:
Thanks @cicnavi ! While this is already a great improvement over a broken admin-page, it re-introduces a direct dependency on the admin-module.
I believe the idea @monkeyiq and I had was to leverage the event dispatcher (formerly called 'hooks') to delegate metadata-creation to other modules.

cicnavi:
Hm, ok, but since this is broken for SSP v2.5, how about we fix this for this minor version, and move to using events in v2.6?

tvdijen:
Very well!

tvdijen (approved):

PR #2675, opened 2026-08-15
Feature/requesting different authncontexts from a proxy

ioigoume:
**Use Case:**
When our system acts as a proxy, customers often want to request strong authentication methods (like MFA) without making them strictly mandatory. Currently, if we ask for a specific method, some Identity Providers (IdPs) will simply throw an error if they don't support it. On the other hand, if we ask for multiple methods at once, IdPs often default to weaker options like passwords. To follow best practices, we need a way to try one authentication method at a time. If an IdP rejects our first choice (like phishing-resistant MFA), the system should automatically step down and try the next option, eventually falling back to a standard login if needed.

**Implementation:**
~~To make this possible, we updated the code to support a sequential fallback mechanism for authentication requests. When the system receives an error from an IdP, it will now automatically retry the login using the next context in our fallback list. We also fixed a few important edge cases for when the system is operating as a proxy. Specifically, we made sure that the system doesn't get stuck in an endless loop by prioritizing the active fallback attempt over standard configurations. Additionally, we ensured that on the final fallback attempt (when no specific method is requested), the original service provider's request doesn't accidentally interfere, allowing the login to proceed smoothly.~~

This PR introduces a pre-flight mapping mechanism (`AuthnContextClassRefMapping`). 

Context selection is evaluated *before* the authentication request is ever sent to the IdP. By defining an `AuthnContextClassRefMapping` table in the global SP configuration or directly in a specific IdP's remote metadata, the proxy can intercept the requested contexts and translate them into contexts that are actually supported by the upstream IdP. 

This allows the system to seamlessly step down (e.g., mapping phishing-resistant MFA to standard MFA) or drop the requirement entirely (falling back to standard password login) cleanly and efficiently. The mapping happens dynamically per IdP, preventing invalid requests, mitigating downgrade attack risks, and avoiding broken login flows. This PR also includes comprehensive unit tests and documentation detailing how to configure these **mappings.**


tvdijen:
> Currently, if we ask for a specific method, some Identity Providers (IdPs) will simply throw an error if they don't support it.

Yes, this is required by the specification. They should respond with a SAML NoAuthnContext exception.

> If an IdP rejects our first choice (like phishing-resistant MFA), the system should automatically step down and try the next option, eventually falling back to a standard login if needed.

No, that would be bad practise and would open up for an entire new attack vector > downgrade attacks.

> To make this possible, we updated the code to support a sequential fallback mechanism for authentication requests.

That's a fair approach, but the selection should be done before a request is sent, based on a mapping table. Imagine what it would do to the IDPs logfiles if some SP is throwing a bunch of requests with unsupported AuthnContexts at them..  Every IDP of federation should have a publicly available list of supported AuthnContexts (LIke i.e. https://refeds.org/profile/mfa)

So while I understand your problem, I don't think this is the right approach. A mapping table in the metadata would make a lot more sense.

pradtke:
I'm kind of lost in what this is trying to accomplish now.  To me the original PR/goal was to automatically handle the "MFA with Retry" from Refeds Wiki ( https://wiki.refeds.org/pages/viewpage.action?pageId=77465130), where MFA context is desired, but not required, and you start with requesting the desired context, and fall through to other acceptable contexts if the IdP does not support, or does not support for that user, the initially requested contexts.

monkeyiq:
If my understanding is correct from what I have read elsewhere by those explaining it the core of this is:

SAML Core §3.3.2.2.1 explicitly states that `<RequestedAuthnContext>` elements "**MUST** be evaluated as an ordered set, where the first element is the most preferred",

However: In real-world deployments, IdPs prioritize preserving Single Sign-On and mitigating user prompt fatigue ("prompting people constantly is a pretty hated behavior"). If an IdP receives a request listing both strong MFA and a weaker context (or standard login), and the user already has an active session, the IdP defaults to satisfying the request immediately using the existing SSO session rather than interrupting the user with a step-up prompt. That part happens because the lists were initially designed with disjoint, non-overlapping primary credentials in mind.

In current SimpleSAMLphp proxy flows, expressing a preference for REFEDS MFA without making it mandatory creates a dilemma: requesting only the MFA AuthnContextClassRef causes IdPs to reject users without MFA with a Responder/NoAuthnContext error, while bundling MFA and password contexts into a single request causes many IdPs to default to password authentication even when MFA was available for that user.

It is also noted that trying to get metadata about this from an IdP has issues including time based issues. A user might have added MFA support and can do that at the IdP at a particular time. This might also be revoked or suspended.

I think the gist of the PR is then to be able to ask for MFA for a user if the IdP supports it and allow fallback if not. But not allow implicit automatic fallback if the user is already authenticated without MFA. If you get implicit fallback you would then have to turn around and ask for MFA as an exact match if your logic requires it to cover if the IdP could do it but chose not to.

An installation is looking to test this to see if it meets these needs. I will then review with an eye toward ensuring that the changes are only taking effect when `AuthnContextClassRefMapping` is in use. ie, SSP remains the same unless you are looking to enable this feature.

PR #2682, opened 2026-09-10
Fix file config for `MetaDataStorageHandlerXML`

lscharmer:
With commit https://github.com/simplesamlphp/simplesamlphp/commit/ec3fb13a63c9518930a53c3548a3a86d1b091723 `metadata.sources` crashes with `[['type' => 'xml', 'file' => '/foo/bar.xml']]` (the code previously checked for `$srcXml` _and_ `$src`).

In >= `v2.5` it works correctly. I'm guessing this is due to some refactoring + cherry-picking issues.

I added the same line as it is implemented in `master`: https://github.com/simplesamlphp/simplesamlphp/blob/master/src/SimpleSAML/Metadata/MetaDataStorageHandlerXML.php#L49

I also added the missing namespace for the Exception case (also not applicable for >= `2.5`).


monkeyiq (approved):
This looks good. The code after the initial if() block in `__construct` certainly wants to see `$srcXml` being set and the 2.5 code contains this get_contents call as shown.

